PR #7, opened 2018-11-07
Add SAMD support and simplify OS_NONE

tannewt:
I will rebase this tomorrow on the latest develop branch with a better PR description as well.

### DIFF
--- a/src/class/cdc/cdc.h
+++ b/src/class/cdc/cdc.h
@@ -50,6 +50,10 @@
  extern "C" {
 #endif
 
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wpacked"
+#pragma GCC diagnostic ignored "-Wattributes"
+
 /** \defgroup ClassDriver_CDC_Common Common Definitions
  *  @{ */
 
@@ -400,6 +404,8 @@ typedef struct ATTR_PACKED
 } cdc_line_control_state_t;
 
 TU_VERIFY_STATIC(sizeof(cdc_line_control_state_t) == 2, "size is not correct");
+
+#pragma GCC diagnostic pop
 
 /** @} */
 
--- a/src/class/cdc/cdc.h
+++ b/src/class/cdc/cdc.h
@@ -90,7 +90,8 @@ typedef enum
 /// Communication Interface Protocol Codes
 typedef enum
 {
-  CDC_COMM_PROTOCOL_ATCOMMAND              = 0x01 , ///< AT Commands: V.250 etc
+  CDC_COMM_PROTOCOL_NONE                   = 0x00 , ///< No specific protocol
+  CDC_COMM_PROTOCOL_ATCOMMAND                     , ///< AT Commands: V.250 etc
   CDC_COMM_PROTOCOL_ATCOMMAND_PCCA_101            , ///< AT Commands defined by PCCA-101
   CDC_COMM_PROTOCOL_ATCOMMAND_PCCA_101_AND_ANNEXO , ///< AT Commands defined by PCCA-101 & Annex O
   CDC_COMM_PROTOCOL_ATCOMMAND_GSM_707             , ///< AT Commands defined by GSM 07.07
--- a/src/class/msc/msc_device.h
+++ b/src/class/msc/msc_device.h
@@ -199,6 +199,7 @@ ATTR_WEAK bool tud_lun_capacity_cb(uint8_t lun, uint32_t* last_valid_sector, uin
 void mscd_init(void);
 tusb_error_t mscd_open(uint8_t rhport, tusb_desc_interface_t const * p_interface_desc, uint16_t *p_length);
 tusb_error_t mscd_control_request(uint8_t rhport, tusb_control_request_t const * p_request, uint16_t bytes_already_sent);
+void mscd_control_request_complete (uint8_t rhport, tusb_control_request_t const * p_request);
 tusb_error_t mscd_xfer_cb(uint8_t rhport, uint8_t edpt_addr, tusb_event_t event, uint32_t xferred_bytes);
 void mscd_reset(uint8_t rhport);
 
--- a/src/class/msc/msc_device.h
+++ b/src/class/msc/msc_device.h
@@ -198,7 +198,7 @@ ATTR_WEAK bool tud_lun_capacity_cb(uint8_t lun, uint32_t* last_valid_sector, uin
 
 void mscd_init(void);
 tusb_error_t mscd_open(uint8_t rhport, tusb_desc_interface_t const * p_interface_desc, uint16_t *p_length);
-tusb_error_t mscd_control_request_st(uint8_t rhport, tusb_control_request_t const * p_request);
+tusb_error_t mscd_control_request(uint8_t rhport, tusb_control_request_t const * p_request, uint16_t bytes_already_sent);
 tusb_error_t mscd_xfer_cb(uint8_t rhport, uint8_t edpt_addr, tusb_event_t event, uint32_t xferred_bytes);
 void mscd_reset(uint8_t rhport);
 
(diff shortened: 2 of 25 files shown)


hathach:
Hi @tannewt , give me several more hours. I am swimming in the nrf52840 arduino. I will check and test this out a bit later. Meanwhile you should sleep :)

tannewt:
I moved CircuitPython to this change in https://github.com/adafruit/circuitpython/pull/1321. Let me know what changes you'd like to see on it. Thanks!

hathach:
Thanks @tannewt sorry for the delay, I have been trying to release the Arduino before switching task. I wilk definitely check this out Monday :)

src/class/cdc/cdc.h
```diff
@@ -50,6 +50,10 @@
  extern "C" {
 #endif
 
+#pragma GCC diagnostic push
```
hathach: did gcc issue any warnings for packed attribute. I didn't see it, maybe we should solve the warning instead of suppressing it

src/class/msc/msc_device.h
```diff
@@ -164,6 +182,12 @@ ATTR_WEAK void tud_msc_read10_complete_cb(uint8_t lun);
 ATTR_WEAK void tud_msc_write10_complete_cb(uint8_t lun);
 ATTR_WEAK void tud_msc_scsi_complete_cb(uint8_t lun, uint8_t const scsi_cmd[16]);
 
+// Hook to make a mass storage device read-only.
+ATTR_WEAK bool tud_msc_is_writable_cb(uint8_t lun);
+
+// Override for dynamic LUN sizes.
+ATTR_WEAK bool tud_lun_capacity_cb(uint8_t lun, uint32_t* last_valid_sector, uint16_t* block_size);
```
hathach: yeah, capacity callback is better than using CFG_TUD_MSC_BLOCK_NUM/CFG_TUD_MSC_BLOCK_SZ for multiple LUN support. I will remove those macro in the config file and use this instead. Though the function name should be tud_msc_read_capacity_cb()

src/class/msc/msc_device.h
```diff
@@ -164,6 +182,12 @@ ATTR_WEAK void tud_msc_read10_complete_cb(uint8_t lun);
 ATTR_WEAK void tud_msc_write10_complete_cb(uint8_t lun);
 ATTR_WEAK void tud_msc_scsi_complete_cb(uint8_t lun, uint8_t const scsi_cmd[16]);
 
+// Hook to make a mass storage device read-only.
+ATTR_WEAK bool tud_msc_is_writable_cb(uint8_t lun);
```
hathach: Yeah, stack lack this feature, but it shouldn't be implemented as callback I think. It should be set by application using an class API by setting mode sense/sensekey response similar to tud_msc_set_sense() . It is OK to merge this now, but I will likely to change it later

src/osal/osal_none.h
```diff
@@ -145,22 +98,21 @@ static inline void osal_semaphore_reset(osal_semaphore_t sem_hdl)
...
-      *(_err) = TUSB_ERROR_NONE;                                                                 \
-    }                                                                                            \
-  }while(0)
+static inline tusb_error_t osal_semaphore_wait(osal_semaphore_t sem_hdl, uint32_t msec) {
+  (void) msec;
+  while (true) {
+      while (sem_hdl->count == 0) {
+      }
+      // tusb_hal_int_disable_all();
+      if (sem_hdl->count == 0) {
```
hathach: should be count > 0, I think

src/class/msc/msc_device.h
```diff
@@ -89,6 +81,32 @@ TU_VERIFY_STATIC(CFG_TUD_MSC_BUFSIZE < UINT16_MAX, "Size is not correct");
...
+  uint32_t total_len;
+  uint32_t xferred_len; // numbered of bytes transferred so far in the Data Stage
+
+  // Sense Response Data
+  uint8_t sense_key;
+  uint8_t add_sense_code;
+  uint8_t add_sense_qualifier;
+}mscd_interface_t;
+
+extern mscd_interface_t _mscd_itf;
```
hathach: msc interface driver is meant to be private, there is no need to expose its structure as well, I think.

src/osal/osal_none.h
```diff
@@ -218,22 +170,13 @@ static inline void osal_queue_reset(osal_queue_t const queue_hdl)
...
-      else                                                                                        \
-        return TUSB_ERROR_OSAL_WAITING;                                                           \
-    } else{                                                                                       \
-      /* Enter critical ? */                                                                      \
-      tu_fifo_read(_q_hdl, p_data);                                                               \
-      /* Exit critical ? */                                                                       \
-      *(_err) = TUSB_ERROR_NONE;                                                                  \
-    }                                                                                             \
-  }while(0)
+static inline tusb_error_t osal_queue_receive(osal_queue_t const queue_hdl, void* data) {
```
hathach: the API must be the same, since other RTOS share the same API. I woudn't change API, just in case I need to go back to the complicated OSAL NONE. When we got confident enough that we don't need it, we will change the API for all RTOSes.

src/device/dcd.h
```diff
@@ -154,22 +135,6 @@ void dcd_edpt_stall       (uint8_t rhport, uint8_t ep_addr);
 void dcd_edpt_clear_stall (uint8_t rhport, uint8_t ep_addr);
 bool dcd_edpt_stalled     (uint8_t rhport, uint8_t ep_addr);
 
-//------------- Control Endpoint -------------//
-bool dcd_control_xfer     (uint8_t rhport, uint8_t dir, uint8_t * buffer, uint16_t length);
```
hathach: why would you want to merge the control xfer into the normal transfer. Although it is similar, control pipe is bidirectional, the direction is needed without requiring dcd to guess/cached which direction stack trying to carry

src/portable/nordic/nrf5x/dcd_nrf5x.c
```diff
@@ -82,17 +82,8 @@ typedef struct
...
-    uint16_t total_len;
-    volatile uint16_t actual_len;
-
-    uint8_t  dir;
-  }control;
-
-  // Non control: 7 endpoints IN & OUT (offset 1)
-  nom_xfer_t xfer[7][2];
+  // All 8 endpoints including control IN & OUT (offset 1)
+  nom_xfer_t xfer[8][2];
```
hathach: Yeah :+1: , I also want to merge them together, but always be stuck to something else :(

src/class/msc/msc_device.h
tannewt: Why would you change it? This approach means tinyusb doesn't need to store extra state on its own. The compiler can optimize the call away if it never changes.

hathach: since this is rather specific, not as generic as an usb stack should be. An sense/disk readiness state should be actively set by application to inform the stack.

src/device/dcd.h
hathach: the endpoint 0 is bidirectional, other endpoin can only be either IN or OUT. But indeed merging them is good way. It will require usbd to pass 0x80 as control IN address, 0x00 for control OUT. Note: 0x80 for control IN is not part of usb specs, but we could document for dcd to take that.

src/class/msc/msc_device.h
hathach: I am not sure either, most user is not familiar with MSC sense/sensekey etc.. Maybe leave it as you did for now, I check the scsi docs and find a better way

src/device/control.h
```diff
@@ -0,0 +1,98 @@
...
+} control_stage_t;
+
+typedef struct {
+    control_stage_t current_stage;
+    tusb_control_request_t current_request;
+    uint16_t total_transferred;
+    uint8_t config;
+} control_t;
+
+CFG_TUSB_ATTR_USBRAM CFG_TUSB_MEM_ALIGN uint8_t _shared_control_buffer[64];
```
hathach: @tannewt this could be a potential bug, every one has its own def of this buffer. It should be extern here, and defined in control.c

src/class/cdc/cdc.h
hathach: @tannewt I still didn't see any gcc warning regarding packed locally, can you explain it a bit more

PR #10, opened 2018-11-16
merge for local develop

hathach:
- fix a bug with control transfer direction for nrf5x dcd
- make control transfer as part of usbd. Class driver must use usbd_control_ API() instead of dcd_ api.
- change the signature of class driver's control_request
- allow control request complete to stall in status stage
- move control request parser & handling to usbd.


src/device/usbd.c
```diff
@@ -316,66 +318,142 @@ void usbd_task( void* param)
...
 
-void tud_control_interface_control_complete_cb(uint8_t rhport, uint8_t interface, tusb_control_request_t const * const p_request) {
-    if (_usbd_dev.itf2drv[ interface ] < USBD_CLASS_DRIVER_COUNT)
+//--------------------------------------------------------------------+
+// Control Request Parser & Handling
+//--------------------------------------------------------------------+
+
+// This handles the actual request and its response.
+// return false will cause its caller to stall control endpoint
+static bool process_control_request(uint8_t rhport, tusb_control_request_t const * p_request)
```
tannewt: Why not have this in usbd_control.c?

hathach: usbd_control.c is only used for carrying out control transfer: data, status and invoke complete callback. Parsing and making sense of request is part of usbd.

hathach:
my idea is usbd_control only manages control pipe/transfer (from start, continue 64byte packet until complete, then callback and status). While making sense (or parse) the request meaning is usbd (usb device driver) 's job.  

Control pipe is managed by usbd (it is kind of shared-resource of class drivers), thus I rename it to usbd_control. All class driver must access control endpoint through usbd: i.e must use usbd_control_xfer/status/stall() instead of dcd_edpt_(0, )

tannewt:
Ok, that makes sense. Thanks for the explanation!

hathach:
NP, I am glad that make sense :)

PR #11, opened 2018-11-20
fix samd51 dcd setup packet prepare

hathach:
samd51 should work more reliably in general. Segger embedded studio for samd51 is also added

@tannewt


tannewt:
Ok, it looks good to me.

hathach:
hmm, could see interrupt when host send setup packet to samd21. The dcd is almost indentical to samd51. There is isr when bus reset though.

hathach:
Haven't test with MSC demo (since I am not familliar with flash api of samd yet). but CDC work much better with samd port now. Merge now and can improve later if needed

tannewt:
Awesome! I'll update CircuitPython to this later today and PR any changes that I make. I was thinking the MSC demo could use RAM as a very small drive.

PR #14, opened 2018-11-23
A few more updates

tannewt:
This does a few things:
* Fixes one warning about dropping a cast
* Enables interrupt protection of the queue. If nrf doesn't want this we should change its interrupt enable/disable. We probably want usb at a lesser nvic priority than the SD.
* Changes samd USB start up to wait for sync.
* Ensures that the samd control endpoint always has a valid buffer to write to.

CircuitPython now uses this branch as of https://github.com/adafruit/circuitpython/pull/1353

### DIFF
--- a/src/class/hid/hid_device.c
+++ b/src/class/hid/hid_device.c
@@ -418,7 +418,8 @@ bool hidd_control_request(uint8_t rhport, tusb_control_request_t const * p_reque
 
     if (p_request->bRequest == TUSB_REQ_GET_DESCRIPTOR && desc_type == HID_DESC_TYPE_REPORT)
     {
-      usbd_control_xfer(rhport, p_request, p_hid->desc_report, p_hid->desc_len);
+      // Cast away the const on p_hid->desc_report because we know it won't be modified.
+      usbd_control_xfer(rhport, p_request, (void *)p_hid->desc_report, p_hid->desc_len);
     }else
     {
       return false; // stall unsupported request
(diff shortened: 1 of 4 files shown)


src/class/hid/hid_device.c
```diff
@@ -418,7 +418,8 @@ bool hidd_control_request(uint8_t rhport, tusb_control_request_t const * p_reque
 
     if (p_request->bRequest == TUSB_REQ_GET_DESCRIPTOR && desc_type == HID_DESC_TYPE_REPORT)
     {
-      usbd_control_xfer(rhport, p_request, p_hid->desc_report, p_hid->desc_len);
+      // Cast away the const on p_hid->desc_report because we know it won't be modified.
```
hathach: ah right, forgot about this, hmm, I need to raise my ses warning level to catch this.

hathach:
@tannewt ah thanks for the warning, regarding the irq queue protection. I intended to add it as well. Need a bit of testing to make sure, nrf also need this, and yes the current irq prio of usb is lower than SD ( 2 is safe enough, 0,1 is reserved by SD). 

Though the interrupt disable all protection is only temporarily, since host stack will also need that, both stack can be running concurrently, it should only disable the device/host accordingly. Anyway, I merge now and do more refractor later. Make sure you tested with samd, some controller may not be happy with disable/enable dynamically (e.g pending interrupt not trigger etc ..). I will do tests with nrf locally.

tannewt:
Thanks for the merge! I tested on nrf and samd and it worked ok. We commonly disable interrupts on SAMD so I'm not worry about it.

PR #15, opened 2018-11-25
Add Makefile support for the cdc_msc_hid example

tannewt:
And use it in Travis to validate a build. Removed the failing test which was old.


hathach:
superb !!! Thanks for the work :D

PR #30, opened 2019-01-05
Add MIDI support

tannewt:
This is nearly finished so please take a look. It adds USB MIDI support and a TX buffer that doesn't incur a copy on USB packet send. I only stubbed out the descriptor side because we have our own USB descriptor for CircuitPython.

Please let me know what you think.


hathach:
@tannewt  this is great, there is lots of works. You even add unit testing for tx_buf as well  :+1:

PR #31, opened 2019-01-08
Remove txbuf in favor of fifo.

tannewt:
Fifo incurs a second copy into a USB buffer but this ensures the
correct USB buffer alignment. Some LPC chips require 64 byte
alignment that txbuf couldn't do.


hathach:
That is not fair for samd, being generic stack, it has to be conform popular mcu which is a pain :(

PR #36, opened 2019-02-20
Fix write protected MSC. The bits were flipped.

tannewt:
I'm not exactly sure the CSW response is correct. It doesn't happen on Mac when the right Mode 6 bit is set.

### DIFF
--- a/src/class/msc/msc.h
+++ b/src/class/msc/msc.h
@@ -284,8 +284,8 @@ typedef struct ATTR_PACKED
 {
   uint8_t data_len;
   uint8_t medium_type;
-  bool write_protected : 1;
   uint8_t reserved : 7;
+  bool write_protected : 1;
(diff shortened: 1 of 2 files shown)


src/class/msc/msc.h
```diff
@@ -284,8 +284,8 @@ typedef struct ATTR_PACKED
 {
   uint8_t data_len;
   uint8_t medium_type;
-  bool write_protected : 1;
   uint8_t reserved : 7;
+  bool write_protected : 1;
```
hathach: this is not important, but it would be nice if you change the write_protected to uint8_t instead of bool, make it clearer that WP is part of 8-bit bitfiled with above reserved.

hathach:
right, bitfield counts from bit 0 first, this is correct place of write protected (bit 7).

PR #38, opened 2019-02-22
STM32F4 Family Support

cr1901:
This PR adds support for the STM32F4 family of microcontrollers, and was developed for Adafruit industries. As a demo, I added support for the STM32F4-Discovery board, and ported the CDC portion of the `cdc_hid_msc` demo to said board. This PR has been tested on a Windows 7 machine, using Tera Term and [a pyserial-based terminal](https://github.com/enjoy-digital/litex/blob/master/litex/utils/litex_term.py).

There are still a few issues that need resolving (all issues refer to the `cdc_hid_msc` demo firmware), but I want to go ahead and file this PR before they are resolved; @tannewt and I can hash out the details. I am looking into their causes in parallel with this PR:

* I am able to break the demo by sending an ASCII file (> 64 bytes) to the serial port from my terminal program, such that a flood of OUT-NAKs occurs in response (and the file contents are never echo'd back).

* Disconnect/reconnect issues
  * Occasionally, the core sends a malformed IN response to a SETUP packet (truncated packet which 
  leads to bad CRC). Enumeration never finishes in this case and the Windows claims the device malfunctions.
  * For this next issue, I'm going to explain the steps to duplicate in detail. The order is in fact
  important; just resetting the firmware and unplugging/plugging the core back in will have the 
  correct behavior.
    * Start the firmware (via reset or gdb) 
    * Opening a terminal program (Tera Term) and connect to the STM32F4 COM port.
    * Type some characters into the terminal.
    * Close the connection to the COM port.
    * Unplug and plug the Discovery board's USB port back in.
    * Reopen your terminal and reconnect to the STM32F4 COM port.
    * Typing some more chars.
  
    If these steps are followed in order, when you type some characters as in the last step, some initial 
    characters will not echo back to your terminal, and the “TinyUSB CDC MSC HID device example” 
    banner will not appear. Perhaps these two disconnect/reconnect issues are related?

* Not an essential change, but I want to tweak IN/OUT/SETUP packet buffer sizes for performance.

* Not a functional change, but I want to slightly refactor the interrupt handler into smaller static functions.

The large "Files changed" number is from bringing in STM32 headers; the majority of changes are in `src\portable\stm\stm32f4\dcd_stm32f4.c` Please let me know in comments if you have any questions.

cc: @tannewt @ladyada @ptorrone


hathach:
Superb !!! Thank you very much @cr1901 , I have always wanted to add stm32 ported. I did have and stm32f407 discovery board here so I could help troubleshoot those issue that you listed as well since I did worked with stm32 usb a little bit previously.

Do you want me to merge this now to master, or wait a little bit more, either way is fine with me. I will try to get it running here on my F4 discovery board and help with the issue.

cr1901:
@hathach Please wait a little bit more, I'd like Scott to also review my work :).

Thanks for the offer to help :)!

ladyada:
thanks @tannewt that sounds good - @hathach please take a first pass review of this PR when you have time this week on your stm32 board :)

hathach:
@ladyada @tannewt yeah sure, I am happy to help out with this PR, will look at this closer tomorrow. Since this PR mostly add mcu and hw files without any core file changes. It can be safely merge into master.

hw/mcu/stm/stm32f4/Device/ST/STM32F4xx/Include/stm32f4xx.h
```diff
@@ -0,0 +1,269 @@
+/**
```
hathach: It will be better if `hw/mcu/st` use this repo as submodule
https://github.com/micropython/stm32lib

It is maintained and updated by microphython, with all of the stm32 mcu family. Making porting tinyusb to circuitpython much easier later on.

Another minor thing is we should rename the vendor from stm to **st** which is more accurate.

src/portable/stm/stm32f4/hal_stm32f4.c
```diff
@@ -0,0 +1,69 @@
+/**************************************************************************/
```
hathach: Here, we should also rename the vendor from stm folder to **st** which is more accurate.

examples/device/cdc_msc_hid/src/tusb_config.h
```diff
@@ -108,7 +108,7 @@
 
 //------------- CLASS -------------//
 #define CFG_TUD_CDC                 1
-#define CFG_TUD_MSC                 1
+#define CFG_TUD_MSC                 0
```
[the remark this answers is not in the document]
hathach: yeah, please make this change local.

hw/mcu/stm/stm32f4/Device/ST/STM32F4xx/Include/stm32f4xx.h
cr1901: I agree w/ both changes, but I don't believe the repo you linked (https://github.com/micropython/stm32lib) corresponds to the header files I actually have.

I don't quite remember where I found the headers- @tannewt did the work for me there. Although I did _rediscover_ where he got them from when I started working on this.

Scott, could you point me to the repo or zip file where you found the source files under `hw/mcu/stm/stm32f4`?

hathach: @tannewt do you have any suggestion on which repo is used for stm32lib, I am OK with anything that works :)

cr1901: @hathach I found the relevant zip file. The STM32F4-specific files that are _currently_ checked into tinyusb are _a subdirectory_ of the [Standard Peripheral Library](https://www.st.com/content/st_com/en/products/embedded-software/mcu-mpu-embedded-software/stm32-embedded-software/stm32-standard-peripheral-libraries/stsw-stm32065.html). Relative to the root of the zip file, the relevant files are under `STM32F4xx_DSP_StdPeriph_Lib_V1.8.0\Libraries\CMSIS`. The `DSP_Lib` directory from the zip file was omitted (it wasn't needed I imagine).

@ladyada, would it be okay if the Standard Peripheral Library was kept as a repository under the @adafruit account, as is [currently done](https://github.com/adafruit/asf4) with the Atmel headers?

If so, I think the entire repository should be kept and not just the subdirectory that made it into the tinyusb repo (@tannewt, feel free to disagree :P). Then I will convert `hw/st/smt32f4` into a submodule and update Makefile paths accordingly.

hathach: seem like the std peripheral is legacy lib now, it is replaced by stm32cube lib, ST recommends that in @cr1901 's link. It is probably better to stick with latest lib for better support with new devices later on, unless it has a major drawback or restriction !!

cr1901: I have stm32cube installed; in general, stm32cube will ask you to _download_ all dependencies/headers from a GUI Wizard, and then place the all the dependencies for your current project inside your project directory. So projects are self-contained, but you need to generate the headers for each project using ST's IDE beforehand<sup>1</sup>.

The equivalent directory to the Standard Peripheral Library in an stm32cube project should be `Drivers\CMSIS`. Or at least it is on my machine :P.

1. Requiring the vendor IDE to get started seems par for the course for ARM vendors. NXP does it with lpcxpresso and ST seems to do it with stm32cube.

cr1901:
@hathach I have completed my desired refactoring; I'm now satisfied :D. There's no functional change, but you may wish to re-review the OTG interrupt handler and functions called by it :)!

hw/mcu/stm/stm32f4/Device/ST/STM32F4xx/Include/stm32f4xx.h
hathach: @cr1901 It is troubleshome, sound similar to samd asf4 (must be a trend in silicon vendor :D), though it isresolvable. Micropython already download and put [stm32cube](https://github.com/micropython/stm32lib) library in the above link. We can either use it or folk it if needed.

hathach:
thanks @cr1901 , I know that satisfied feeling :D since tinyusb went through lots of rework lately :D. As long as it works as proof of concept, it is good enough for the PR. We could always refactor/enhance it later on :)

hw/mcu/stm/stm32f4/Device/ST/STM32F4xx/Include/stm32f4xx.h
cr1901: The repo you linked is missing the `Source` directory under `CMSIS/STM32F4xx`, which is needed to provide both the `.s` startup code and the `SystemInit()`/`SystemCoreClockUpdate()` functions in a separate `.c` file.

Additionally, as `board.mk` is currently coded, I've elected to use the CMSIS headers provided by the peripheral library (under `hw/mcu/stm/stm32f4/Include`) instead of the CMSIS headers you provided under `hw/cmsis`. Chances are, both copies are quite similar. But I should check that my code works with your copy (and see what the other MCU families do).

tannewt: I don't really remember where I got the files. I'd have do some sleuthing and don't have time for it at the moment.

hathach: @cr1901 for the missing startup and linker script, we can add it in `hw/bsp`, I am not sure why micropython didn't include the system_stm32.c in the repo, maybe they want to customize it ( or it is how the stm32cube is designed to), but we can also store it in the repo in another folder under `hw/mcu/st`

Note; [system_stm32.c ](https://github.com/micropython/micropython/blob/master/ports/stm32/system_stm32.c) seems to be generic for the cube

@tannewt  no problems at all.

hathach: for `cmsis` folder, yeah, we should use the one that provided by st in their library. Feel free to change is. `hw/cmsis` is included for mcu which the prph lib doesn't include cmsis such as Nordic's nrfx :)

cr1901: @hathach Okay, let's go with your idea, with one suggestion- can we store the startup code (`.s`) along with `system_stm32f4xx.c` files under `hw/mcu/st/startup/$family` and `hw/mcu/st/system-init` instead of under `hw/bsp`? The startup code is board agnostic, and none of the other TinyUSB ports seem to place startup code w/ the boards.

Just for completeness' sake, I'm gonna sum up what I've found (and maybe stick it into a README.md file):

The CMSIS by ARM is an attempt to unify how system initialization and headers for ARM microcontrollers are handled across vendors. A vendor is expected to provide a number of files, including vendor-specific device headers, vendor-specific system init code (in C), vendor-specific startup code (coded in assembly _and_ meant to make writing C interrupt handlers easy), and ARM architecture-specific headers. In practice, it's actually difficult to find all these files in one place.

* stm32cube generates a self-contained application. All the relevant headers, system init, and startup code are under `Drivers/CMSIS`. The linker script is stored at the root of your project, _along_ with a copy of the startup code. Relative to `Drivers/CMSIS`:
  * Device header files are stored under `Device/ST/$family/Include`.
  * CMSIS headers are stored under `Include`.
  * The portable C portion of the startup code, called `system_$family.c`, is stored under `Device/ST/$family/Source/Templates`. This "system init" code is required by CMSIS.
  * The assembly portion of the startup code, called `startup_$device.c`, is stored under `Device\ST\STM32F4xx\Source\Templates\gcc`.

* Micropython maintains a repo with _only_ the device header files. We use that repo and provide the remaining files ourselves.
  * Micropython keeps the linker script and assembly portion of the startup code under the [boards](https://github.com/micropython/micropython/tree/master/ports/stm32/boards) in the stm32 port (but the code is board agnostic).
  * Micropython also embeds its own [copy](https://github.com/micropython/micropython/tree/master/lib/cmsis/inc) of CMSIS headers for all ARM targets into the repo.
  * Micropython indeed _does_ use their own custom C startup code (CMSIS system init) for [stm32](https://github.com/micropython/micropython/blob/master/ports/stm32/system_stm32.c) targets. It is [missing](https://github.com/micropython/micropython/commit/fdbc22e4d27b9d1c686647c9c67a32206f7bc83a) the `SystemCoreClockUpdate` function.

* TinyUSB-specific info over where files are goes here- to be written :).

hathach: @cr1901 thank you very much, yeah sure, you can put them in any folder under `hw/mcu/st`, the number of files and folder are rather small, it should be no issue at all. You could also move it around later on as well if needed, not a problem at all. I am looking forward to your next push and try it on my f4 discovery board here :D

Later on when integrating this into CPY, Adafruit may even wants to maintain Adafruit's own stm32lib like [Nordic nrfx](https://github.com/adafruit/nrfx), and [Samd asf4](https://github.com/adafruit/asf4) . It is only a matter of changing the submodule link reference :D

cr1901: @hathach I have updated the repo structure as requested. We use submodules when possible (stm32lib) and I have updated the remaining headers we keep around to be more up-to-date (from a dummy stm32cube project I used for testing).

I had to add a dummy `_init()` function to `hw/bsp/stm32f407g_disc1/board_stm32f407g_disc1.c` to satisfy the linker (this seems to be a known issue- CTRL+F "_init" on [this page](https://github.com/bjornfor/stm32-test)). And of course I had to update some paths. But other than that, the demo still works fine. I even collected my thoughts into `hw/mcu/st/README.md` so I don't forget the layout :P.

@tannewt, @ladyada, @ptorrone: If @hathach approves my changes now, is it okay if we merge and deal with further issues/fixes in subsequent PRs and commits to keep momentum going?

hathach:
@cr1901 thanks for latest push, everything looks OK, should I merge it now or you still want to work on it a bit more.

cr1901:
@hathach Let's wait for @tannewt's approval.

hathach:
@cr1901 Scott is busy, no need to wait for him. There will be more follow up, no problem at all

cr1901:
Alright, then feel free to merge :)!

hathach:
merged thanks a lot for the PR. I will spend a bit of time this week testing out this port on my local st f4 board :D

hathach:
installing toolchain and programmer for stm32, just realize that my board is the old version stm32f4discovery (not newer disc1), hopefully that won't affect the building/testing.

hathach:
hmm loaded application on my old revision of stm32f4 discovery board, doesn't blink an led at all. I guess it is clock set up or something is difference between revision. I will try to buy latest version to test with.

cr1901:
@hathach Definitely possible... do you have the user's manual for the old version that you could upload? It should come w/ a schematic and I could check the pinouts/clocks compared to my version.

hathach:
@cr1901 thanks a lot, it is false alarm. I manage to get led blink witb segger studio (need to convert onboard stlink to jlink) in effort to debug the issue. I will try to nail it down next week. ST chip is very easy to work with and have decent price, no wonder why it is very popular :D

hathach:
making some progress with my stm32f4 board. Device response with get device descriptor and set address. Afterwards it didn't response to request with new address. It is already a good start here :), not sure what is missing, I am troubleshooting it :D

hathach:
Ok according to STM reference manual, when receive set address command, we change the address first then send the status IN afterwards. 
image

Currently, tinyusb did the other way around, response with IN first and set address afterwards. Swap it, and stm32 work well afterwards.

However, there is another mcu (samd or lpc) where the set address will take affect immediately making responding with IN later is not possible. Hmmm. look like we have to give the Status response back to dcd_set_address() ( I moved it up to usbd.c not long ago) ~.~

hathach:
phew, change dcd_set_address behavior and have to run all the tests with all the board :D. Finally stm32f4 work with cdc, there is still issue with MSC WRITE10, somehow endpoint doesn't response, it response fine with READ10. Doing more trouleshooting

image

cr1901:
>Currently, tinyusb did the other way around, response with IN first and set address afterwards. Swap it, and stm32 work well afterwards.
Hmmm, I don't think I encountered a "change address" command during testing. Is there a way to force the OS to assign a different address?

hathach:
> > Currently, tinyusb did the other way around, response with IN first and set address afterwards. Swap it, and stm32 work well afterwards.
> 
> Hmmm, I don't think I encountered a "change address" command during testing. Is there a way to force the OS to assign a different address?

I have fixed that issue, don't worry about it, dcd_set_address() will need to handle the status IN now (previously it is handled by usbd before call this function) 
https://github.com/hathach/tinyusb/blob/develop/src/portable/st/stm32f4/dcd_stm32f4.c#L187

FYI, Set address is sent after the very first get device descriptor, it is a must, and there is no way to force usb host to change bus address.

PR #39, opened 2019-03-04
Fixes #33: Remove invalid use of strncpy()

### LINKED ISSUE #33: gcc version 8 compiler warning in msc_device.c
When building circuitpython **with arm-none-eabi-gcc 8.2.1** (arm toolchain 8-2018-q4), I get this fatal warning:
```
../../lib/tinyusb/src/class/msc/msc_device.c: In function 'proc_builtin_scsi':
../../lib/tinyusb/src/class/msc/msc_device.c:273:7: error: 'strncpy' output truncated before terminating nul copying 8 bytes from a string of the same length [-Werror=stringop-truncation]
       strncpy((char*) inquiry_rsp.vendor_id  , CFG_TUD_MSC_VENDOR     , sizeof(inquiry_rsp.vendor_id));
       ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
../../lib/tinyusb/src/class/msc/msc_device.c:274:7: error: 'strncpy' output truncated before terminating nul copying 16 bytes from a string of the same length [-Werror=stringop-truncation]
       strncpy((char*) inquiry_rsp.product_id , CFG_TUD_MSC_PRODUCT    , sizeof(inquiry_rsp.product_id));
       ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```
I tried a few ways to fix it, first using `memcpy()` and I think I remember that `CFG_TUD_MSC_PRODUCT` and `VENDOR` are not the same length as `inquiry_rsp.product_id` and `.vendor_id` correspondingly, which causes various issues.

Only gcc version 8 is catching this issue. We haven't moved up to gcc 8 yet, but we will sooner or later.

gpshead:
This was causing a stringop-truncation compiler warning in gcc 8 when
the #defined values being copied from were string literals.
 `error: 'strncpy' output truncated before terminating nul copying 8 bytes from a string of the same length [-Werror=stringop-truncation]`

These fields aren't NUL terminated C strings, they are a fixed width buffer
that is supposed to be space (0x20) padded.

See #33 .

I initialized the values to raw space filled constants.  that could instead be done as a post memcpy step based on the difference between len and the sizeof if you want to preserve your initial adafruit/feather/1.0 values in there (though i don't see why you would - everything should be defining the `CFG_TUD_` macros and using the copies)

It compiles on gcc-8.  i haven't run the code yet.  you'll know much better how to test this than I will.

### DIFF
--- a/src/class/msc/msc_device.c
+++ b/src/class/msc/msc_device.c
@@ -265,14 +265,20 @@ int32_t proc_builtin_scsi(msc_cbw_t const * p_cbw, uint8_t* buffer, uint32_t buf
           .is_removable         = 1,
           .version              = 2,
           .response_data_format = 2,
-          .vendor_id            = "Adafruit",
-          .product_id           = "Feather52840",
-          .product_rev          = "1.0"
+          .vendor_id            = "        ",
+          .product_id           = "                ",
+          .product_rev          = "    ",
       };
+      size_t len;
 
-      strncpy((char*) inquiry_rsp.vendor_id  , CFG_TUD_MSC_VENDOR     , sizeof(inquiry_rsp.vendor_id));
-      strncpy((char*) inquiry_rsp.product_id , CFG_TUD_MSC_PRODUCT    , sizeof(inquiry_rsp.product_id));
-      strncpy((char*) inquiry_rsp.product_rev, CFG_TUD_MSC_PRODUCT_REV, sizeof(inquiry_rsp.product_rev));
+      #define _min(a,b) ((a) < (b) ? (a) : (b))
+      len = strlen(CFG_TUD_MSC_VENDOR);
+      memcpy(inquiry_rsp.vendor_id  , CFG_TUD_MSC_VENDOR     , _min(len, sizeof(inquiry_rsp.vendor_id)));
+      len = strlen(CFG_TUD_MSC_PRODUCT);
+      memcpy(inquiry_rsp.product_id , CFG_TUD_MSC_PRODUCT    , _min(len, sizeof(inquiry_rsp.product_id)));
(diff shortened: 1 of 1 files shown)


src/class/msc/msc_device.c
```diff
@@ -265,14 +265,20 @@ int32_t proc_builtin_scsi(msc_cbw_t const * p_cbw, uint8_t* buffer, uint32_t buf
...
+          .product_id           = "                ",
+          .product_rev          = "    ",
       };
-
-      strncpy((char*) inquiry_rsp.vendor_id  , CFG_TUD_MSC_VENDOR     , sizeof(inquiry_rsp.vendor_id));
-      strncpy((char*) inquiry_rsp.product_id , CFG_TUD_MSC_PRODUCT    , sizeof(inquiry_rsp.product_id));
-      strncpy((char*) inquiry_rsp.product_rev, CFG_TUD_MSC_PRODUCT_REV, sizeof(inquiry_rsp.product_rev));
+      size_t len;
+
+      #define _min(a,b) ((a) < (b) ? (a) : (b))
```
hathach: there is `tu_min32()` that can be used :D

hathach:
Thank you very much for the PR, indeed the initialized value is left-over when switching back and forth between repo :D. It should use the `CFG_TUD_` only :)

PR #40, opened 2019-03-15
stm32f4 port (Cont)

hathach:
This is continue work from PR #38 thanks to @cr1901. 
After testing on stm32f4discovery, I made some modification
- Update TX FIFO allocation scheme for IN Endpoint
- Call dcd_event_xfer_complete() when all byte in total_len is received.
- Added segger studio project for debugging (require to [convert on-board stlink to jlink](https://www.segger.com/products/debug-probes/j-link/models/other-j-links/st-link-on-board/) ) 
- Move zero length status response into dcd_set_address()

The current example has both CDC + MSC working pretty well. Issue described in previous PR seems to be resolved as well. @cr1901 Please take a look and let's me know what do you think about this modification.

@tannewt @ladyada @ptorrone


hathach:
PS: There is a small issue though I  have added make's `flash-stlink` target to flash board with [STM32_Programmer_CLI](https://www.st.com/en/development-tools/stm32cubeprog.html) but somehow the produced binary from make target couldn't run with my old f4 discovery board, the segger studio project does run fine. Maybe I haven't flashed it correctly, @cr1901 would you mind telling me how do you flash it on your pc, I will try to follow that.

tannewt:
@ladyada want to review too? Looks good to me.

cr1901:
@hathach I personally use openocd on Windows. Since STM32F4-DISC attaches as a WinUSB device, it works "right out of the box" if openocd has libusb support! :)

ladyada:
im unable to do an effective review - i should order a discovery board to test tho - i'll trust @hathach and @cr1901 's reviews :)

hathach:
@cr1901 thank for the detail explanation, I will try to get the simple flash target work so that everyone can test this quickly.  `ses` is segger embedded studio which is cross platform and free for non-commercial
https://www.segger.com/products/development-tools/embedded-studio/ 

Since it supports most of popular MCU, and I hate to switch IDE to vendor's specific for each MCU and couldn't in love with GDB. SES is ideal solution for debugging. Installed the tool and stm32 package, then you are good to go. Though it only works with jlink, so you need to convert on-board stlink to jlink first here. Yeah, it has its own startup etc.. It can be configured to used with existing startup (not sure), though `ses` is for debugging purpose, so I just leave it as it is when created project by IDE, not affecting our main codebase much :). May need to clean it up a bit later, but not now.

image

src/portable/st/stm32f4/dcd_stm32f4.c
```diff
@@ -80,20 +80,39 @@ static void bus_reset(void) {
   dev->DOEPMSK |= USB_OTG_DOEPMSK_STUPM | USB_OTG_DOEPMSK_XFRCM;
   dev->DIEPMSK |= USB_OTG_DIEPMSK_TOM | USB_OTG_DIEPMSK_XFRCM;
 
+  // Peripheral FIFO architecture (Rev18 RM 29.11)
```
hathach: @cr1901 please let me know if I understand the Reference Manual's FIFO allocation correctly. All RM seem to be written by hw engineer who expect everyone is an expect of their design, so it is easy to mis-interprete :D.

src/portable/st/stm32f4/dcd_stm32f4.c
```diff
@@ -80,20 +80,39 @@ static void bus_reset(void) {
...
+  // --------------- 0
+  //
   // FIFO sizes are set up by the following rules (each word 32-bits):
-  // OUT FIFO uses (based on page 1354 of Rev 17 of reference manual):
+  // All EP OUT shared a unique OUT FIFO which uses (based on page 1354 of Rev 17 of reference manual):
   // * 10 locations in hardware for setup packets + setup control words
   // (up to 3 setup packets).
   // * 2 locations for OUT endpoint control words.
-  // * 64 bytes for maximum control packet size.
+  // * 16 + 1 (data + info) for largest packet size of 64 bytes.
```
cr1901: No need for the `+1` for `info`. I already account for the `info`  when I said "2 locations for OUT endpoint control words."

src/portable/st/stm32f4/dcd_stm32f4.c
```diff
@@ -80,20 +80,39 @@ static void bus_reset(void) {
...
-  // IN FIFO uses 64 bytes for maximum control packet size.
   //
   // However, for OUT FIFO, 10 + 2 + 16 = 28 doesn't seem to work (TODO: why?).
   // Minimum that works in practice is 35, so allocate 40 32-bit locations
   // as a buffer.
-  USB_OTG_FS->GRXFSIZ = 40;
-  USB_OTG_FS->DIEPTXF0_HNPTXFSIZ |= (16 << USB_OTG_TX0FD_Pos); // 16 32-bit words = 64 bytes
+  //
+  // It is recommended to allocate 2 times the largest packet size, therefore
+  // Recommended value = 10 + 1 + 2 x (16+1) = 45 --> Let's make it 50
```
cr1901: Because each OUT packet will have two control words, I think this comment should _more accurately_ be the following: `10 + 1 + 2 x (16+2) = 47`. Doesn't change the `GRXFSIZ` you chose; I just want the comment to be consistent :).

src/portable/st/stm32f4/dcd_stm32f4.c
```diff
@@ -80,20 +80,39 @@ static void bus_reset(void) {
   dev->DOEPMSK |= USB_OTG_DOEPMSK_STUPM | USB_OTG_DOEPMSK_XFRCM;
   dev->DIEPMSK |= USB_OTG_DIEPMSK_TOM | USB_OTG_DIEPMSK_XFRCM;
 
+  // Peripheral FIFO architecture (Rev18 RM 29.11)
+  //
+  // --------------- 312.5 ( 1250 bytes )
```
cr1901: Should be "`320 (1280 bytes)`".

src/portable/st/stm32f4/dcd_stm32f4.c
```diff
@@ -80,20 +80,39 @@ static void bus_reset(void) {
...
+  // All EP OUT shared a unique OUT FIFO which uses (based on page 1354 of Rev 17 of reference manual):
   // * 10 locations in hardware for setup packets + setup control words
   // (up to 3 setup packets).
   // * 2 locations for OUT endpoint control words.
-  // * 64 bytes for maximum control packet size.
+  // * 16 + 1 (data + info) for largest packet size of 64 bytes.
   // * 1 location for global NAK (not required/used here).
-  // IN FIFO uses 64 bytes for maximum control packet size.
   //
   // However, for OUT FIFO, 10 + 2 + 16 = 28 doesn't seem to work (TODO: why?).
```
cr1901: I should probably have explicitly commented that I deliberately excluded space for global NAK here when testing.

src/portable/st/stm32f4/dcd_stm32f4.c
```diff
@@ -80,20 +80,39 @@ static void bus_reset(void) {
...
 
+  // Peripheral FIFO architecture (Rev18 RM 29.11)
+  //
+  // --------------- 312.5 ( 1250 bytes )
+  // | IN FIFO 3  |
+  // --------------- y + x + 16 + GRXFSIZ
+  // | IN FIFO 2  |
+  // --------------- x + 16 + GRXFSIZ
+  // | IN FIFO 1  |
+  // --------------- 16 + GRXFSIZ
```
cr1901: Just a side note I think worth mentioning: `IN FIFO 0` does _not_ have to map to endpoint 0 (which is where you got `16` entries from I assume). It's simply the easiest mapping :).

src/portable/st/stm32f4/dcd_stm32f4.c
cr1901: @hathach I added my own comments to my interpretation of the FIFO allocation described in the RM. Enjoy! :)



>`// * 2 locations for OUT endpoint control words.`

One control word is sent when the OUT packet is received. The other control word is when the OUT packet has been fully read from the FIFO.

src/portable/st/stm32f4/dcd_stm32f4.c
hathach: Yeah, right, any IN FIFO can be assigned to any IN endpoint, it doesn't need to have fifo index = epnum IN. Currently, we fixed 16 words for control IN which is enough since control doesn't carry much payload, and 80 for each enpoint IN. Later on we can enhance it with better FIFO allocation scheme. e.g Interrupt endpoint only has up to 16 words since Interrupt Endpoint does transfer much, Bulk Endpoint can have as much as possible better is  128 word (512 since it is a sweet spot for example msc class). But we can always improve this later :)

src/portable/st/stm32f4/dcd_stm32f4.c
hathach: ah thanks, it isn't so clear to me from the RM :D

hathach:
@cr1901 I manually flashed the produced elf from make target but it doesn't run 

`$ make BOARD=stm32f407g_disc1 all flash-jlink
Top directory is /home/hathach/Dropbox/tinyusb/workspace/tinyusb
Path from top is examples/device/cdc_msc_hid

  11496    8324   21804   41624    a298 build-stm32f407g_disc1/stm32f407g_disc1-firmware.elf
`
Could you mind verifying that the produced elf (of this develop branch) is running with your setup there. Maybe I mess up something ~.~. Thanks in advance :).

hathach:
> @hathach I also approve these changes
> 
> > Call dcd_event_xfer_complete() when all byte in total_len is received.
> 
> I have a feeling that me not checking `total_len` was a cause of nasty bugs :D!
> 
Yeah, that causes issue with MSC WRITE10 command :D 

> Re: the stall logic, did you add a check for endpoint 0 because the loop waiting for endpoint 0 to disable would never terminate?

Yes, according to RM, control endpoint cannot be disabled, in my testing, the loop never end when ubsd try to stall control endpoint.

> I'll test your changes with my setup tonight or tomorrow and see if I can duplicate your issues. In particular, a text of `11496` bytes seems to low... mine was ~32kB (but _maybe_ that was padding?)...

Ah right, my segger project produce some what 28KB, something is missing here in the makefile. Didn't notice that, thanks for your willingness. I will check why the size is too small as well here.

cr1901:
@hathach I've tested your branch with the following command line: `make -C examples/device/cdc_msc_hid BOARD=stm32f407g_disc1 DEBUG=1`.

The demo still works fine for me. Here is my command line output of `size` for comparison:

```
Top directory is /home/William/src/tinyusb
Path from top is examples/device/cdc_msc_hid

  22536    8356   21844   52736    ce00 build-stm32f407g_disc1/stm32f407g_disc1-firmware.elf
```


Here is an image of the MSC part working correctly!
image

hathach:
> @hathach I've tested your branch with the following command line: `make -C examples/device/cdc_msc_hid BOARD=stm32f407g_disc1 DEBUG=1`.
> 
> The demo still works fine for me. Here is my command line output of `size` for comparison:
> 
> ```
> Top directory is /home/William/src/tinyusb
> Path from top is examples/device/cdc_msc_hid
> 
>    text    data     bss     dec     hex filename
>   22536    8356   21844   52736    ce00 build-stm32f407g_disc1/stm32f407g_disc1-firmware.elf
> ```
> Here is an image of the MSC part working correctly!
> image
> 
> _Through repeated plugging/unplugging on Windows, it is still possible to get the device to fail to enumerate._ But perhaps we can fix that in the next iteration.

Right, when I try to build with `DEBUG=1`, it works, though there is a bit of warning. Hmm, seem like it is optimization related thing.

```
/usr/bin/../lib/gcc/arm-none-eabi/7.3.1/../../../../arm-none-eabi/bin/ld: warning: /usr/bin/../lib/gcc/arm-none-eabi/7.3.1/../../../../arm-none-eabi/lib/thumb/v7e-m/fpv4-sp/hard/libnosys.a(write.o) uses variable-size enums yet the output is to use 32-bit enums; use of enum values across objects may fail
CREATE build-stm32f407g_disc1/stm32f407g_disc1-firmware.bin

  22316    8356   21828   52500    cd14 build-stm32f407g_disc1/stm32f407g_disc1-firmware.elf
```

hathach:
@cr1901 this is crazy enough, seem like it is the bug with gcc version when enable flto
https://bugs.launchpad.net/gcc-arm-embedded/+bug/1747966

The issue author experience the issue with stm32f as well, the walkaround is placing the startup.o (containing weak symbol) first before other *.o with strong symbol. What a issue !!! OK, I think we could merge this PR now, there is still issue with repeatedly plug and unplug, but we will consider it as bug and will fix in later PR.

@tannewt  @cr1901 @ladyada the port working fine with simple example of msc and cdc. I think we could start to add stm32f port to circuitpython. There is also uf2-stm32f bootloader ready as well here https://github.com/mmoskal/uf2-stm32f

PR #46, opened 2019-03-21
Fix slow CDC OUT by NAKing

tannewt:
This NAKs CDC OUT packets when the ring buffer doesn't have
enough space for it. This makes CDC OUT reliable rather than
allowing overwriting into the ring buffer.

### DIFF
--- a/src/class/cdc/cdc_device.c
+++ b/src/class/cdc/cdc_device.c
@@ -87,6 +87,24 @@ typedef struct
 //--------------------------------------------------------------------+
 CFG_TUSB_MEM_SECTION static cdcd_interface_t _cdcd_itf[CFG_TUD_CDC] = { { 0 } };
 
+bool pending_read_from_host;
+static void _prep_next_transaction(void) {
+    uint8_t const itf = 0;
+    cdcd_interface_t* p_cdc = &_cdcd_itf[itf];
+
+    // skip if previous transfer not complete
+    if (pending_read_from_host) {
+        return;
+    }
+
+    // Prepare for incoming data but only allow what we can store in the ring buffer.
+    uint16_t max_read = tu_fifo_remaining(&p_cdc->rx_ff);
+    if (max_read >= CFG_TUD_CDC_EPSIZE) {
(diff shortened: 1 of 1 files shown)


src/class/cdc/cdc_device.c
```diff
@@ -87,6 +87,24 @@ typedef struct
 //--------------------------------------------------------------------+
 CFG_TUSB_MEM_SECTION static cdcd_interface_t _cdcd_itf[CFG_TUD_CDC] = { { 0 } };
 
+bool pending_read_from_host;
+static void _prep_next_transaction(void) {
```
hathach: we should have `itf` passed to this prepare function. Although it is assumed to zero for now in some functions, those are top-level (require a bit of code to detect the interface number).

src/class/cdc/cdc_device.c
```diff
@@ -87,6 +87,24 @@ typedef struct
 //--------------------------------------------------------------------+
 CFG_TUSB_MEM_SECTION static cdcd_interface_t _cdcd_itf[CFG_TUD_CDC] = { { 0 } };
 
+bool pending_read_from_host;
+static void _prep_next_transaction(void) {
+    uint8_t const itf = 0;
+    cdcd_interface_t* p_cdc = &_cdcd_itf[itf];
+
+    // skip if previous transfer not complete
+    if (pending_read_from_host) {
```
hathach: we can use dcd_edpt_busy() function for checking if we already queued an IN transfer

tannewt:
Thanks for the quick review! I'll address the comments early next week when I'm back from travelling. Thanks!

hathach:
> Thanks for the quick review! I'll address the comments early next week when I'm back from travelling. Thanks!

If you are busy and like it to merge asap, I could merge and update it as well. Just give me your thought on my comment above

dhalbert:
@tannewt I can get CircuitPython updated if @hathach's propsosals are fine with you, and you wish to proceed.

tannewt:
Yup I’m fine with both. Just didn’t have time to change and test before travel. Will do it Tuesday if you haven’t beaten me to it. Thanks!

hathach:
@tannewt I merged it now and make modification as above comment and will carry out the test. Working on the stack anyway for another issue https://github.com/adafruit/circuitpython/issues/1617

PR #48, opened 2019-03-25
follow up to pr #46

hathach:
follow to #46  the code work great with nrf5x but struggle to simply echo demo with samd port. There is probably other issue, I am reviewing the samd dcd now. Hopefully could find one or two problem that causes multiple bugs recently ~.~

@dhalbert


hathach:
samd port has issue replacing `pending_read_from_host` by `dcd_edpt_busy()` probably the function is called between the physical transfer complete and DMA moving data and/or handled properly by the stack (with queuing). Revert back to the original variable check, will improve stack with edpt_busy()  later on.

PR #56, opened 2019-04-18
refactor hid device, remove auto descriptor, update example

hathach:


dhalbert:
@hathach Is this all just cleanup, or might it have something to do with https://github.com/adafruit/circuitpython/issues/1782 (Windows USB enumeration)?

dhalbert:
Should we try merging this into CircuitPython now, or can it wait until 4.1?

PR #68, opened 2019-05-18
Fix missing section in binary output

zardam:
The binary outputs of examples are not working, the global variables initialization section was missing.


hathach:
Ah, right. It may miss a section or so. Which mcu/board you are compiling with. I need to check with its linker

zardam:
I'm using the board "metro_m0_express".

hathach:
thanks @zardam , it is indeed an issue as pointed in #69 by Damien, the linker for m0/m4 also include `.relocate` section.

PR #77, opened 2019-05-24
make hid_generic_inout tester have reportLength-sized buffer

todbot:
... since some OSes need that (particularly MacOS)


hathach:
ah thanks, I didn't know that :)

PR #78, opened 2019-05-26
Made node.js hid example more friendly for new users

PTS93:
Hi, tried to make the example a bit more user friendly for someone who may have never used node.js before.
It now send a string that the user can define and change.
I also annotated anything that might be ambiguous to someone new to node.js or node-hid, to make it clear what is happening and why.

I also started a convention to define supported boards (as there will be A LOT when SAMD21 and SAMD51 support comes along) to make it easy to switch between them. Instead of manually changing the vid and pid you can just pass f.e. 'Feather_nRF52840' into the find() function.

### DIFF
--- a/examples/device/hid_generic_inout/hid_test.js
+++ b/examples/device/hid_generic_inout/hid_test.js
@@ -1,22 +1,67 @@
+// IMPORTANT: install dependency via 'npm i node-hid' in the same location as the script
+// If the install fails on windows you may need to run 'npm i -g windows-build-tools' first to be able to compile native code needed for this library
+
 var HID = require('node-hid');
+var os = require('os')
+
+var isWin = (os.platform() === 'win32');
 var devices = HID.devices();
-var deviceInfo = devices.find( function(d) {
-    var isNRF = d.vendorId===0Xcafe && d.productId===0X4004;
-    return isNRF;
-});
+
+// choose either of the following supported devices:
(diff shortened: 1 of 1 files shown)


PTS93:
I kept the change that @todbot made for OSX to keep the buffer size the same. Its just done a little different as the array is now created very differently.

hathach:
superb!! Thank you very much, it looks a lot better and easier to understand now even to me :)

hathach:
@PTS93 one question, instead of choosing the device metro/feather can we loop through an list containing known vid/pid and or by vendor only (since adafruit has a tons of samd21/samd15 board with different PID). User can of course manually add more to the list ( and submit  PR), if they use a custom board.

PTS93:
I was about to do it like that and we could certainly still do it that way.
My idea was that you might not want it to randomly select the first board it finds if you got multiple attached but can define the specific board type.
But I could actually have it both, simply by having a list of the objects instead of pids and vids.
I will test something this evening and if it woks nicely code wise I would add a "any_supported_board" option.

hathach:
@PTS93 multiple board connected with HID device is not very common for most users. Though, manually changing nodejs is very headache. It is true for me, since I typically testing multiple boards making sure all ports are still working after some major changes :). Please make a list of vid/pid if possible, thank you in advance :)

PTS93:
Yea, you're right. The more I think about it the less it makes sense for most users.

PR #79, opened 2019-05-29
Modified example to use list of supported boards

PTS93:
There is now a seperate file to maintain all supported boards.
Maybe Adafruit can generate this for their line of boards?

### DIFF
--- a/examples/device/hid_generic_inout/boards.js
+++ b/examples/device/hid_generic_inout/boards.js
@@ -1,0 +1,4 @@
+module.exports = {
+	"Feather_nRF52840":[0X239A,0X8029],
+	"Metro_nRF52840":[0X239A,0X803F],
+}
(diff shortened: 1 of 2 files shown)


hathach:
superb! thanks, I could fill out the board later. Don't worry too much about it, btw can we take the vid/pid from command line e.g `node hid_test.js -d 239a:1234`  just in case other vendor want to try it out as well, and we cannot keep the list up to date in time.

PS: I still prefer to have only one hid_test.js, but it is ok to have 2 :)

PTS93:
I thought it might make sense to keep that seperated to not end up with a 600 line example at some point :)
If someone want to test their own vid pid pair it would make sense they just add that to the list I would say, they don't have to push the change if they just want to test something.

PTS93:
we could make it a json file instead, then other applications could also read it, like python

hathach:
Make all the sense, thanks for the pr :)

PR #95, opened 2019-08-11
cdc_msc_hid: Limit number of endpoints used by STM32F4 for demo.

cr1901:
Since I last checked, the default options to the `cdc_msc_hid` demo have changed. The default options allocate more endpoints than exist on the STM32 used on the STM32F4DISC. I [guard](https://github.com/hathach/tinyusb/blob/master/src/portable/st/stm32f4/dcd_stm32f4.c#L221-L223) against this, but ultimately tinyusb will end up triggering an assert because the endpoint open failed.

As a compromise, I add an `#if` guard to avoid allocating too many endpoints on STM32F4DISC in `tusb_config.h`.

### DIFF
--- a/examples/device/cdc_msc_hid/src/tusb_config.h
+++ b/examples/device/cdc_msc_hid/src/tusb_config.h
@@ -74,7 +74,14 @@
 //------------- CLASS -------------//
 #define CFG_TUD_CDC              1
 #define CFG_TUD_MSC              1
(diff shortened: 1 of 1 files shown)


hathach:
Yeah, stm32f4 doesn't have enough endpoint for this example. Actually the example doesn't change, we did modify it locally previously. Some higher end f4 (highspeed core) have 5 endpoints, but let's merge it for now. We will modify it later on.

Issue #98, opened 2019-08-15
[Bug] Midi "program change" messages not understood on Windows, Mac


jepler:
**Describe the bug**
@CedarGroveStudios reported that in Circuitpython, Windows was having trouble receiving "program change" messages.  In particular, "[midi tools](https://mountainutilities.eu/miditools)" sees a spurious Program Change 0 before a correct Program Change 33 using a test Python program.  (different OSes seem to treat this differently, I didn't see the problem on Linux but did reproduce it on Windows 10)

**Set up (please complete the following information):**
 - OS: Windows 10 64-bit
 - Board: Particle Xenon
 - Firmware Code: Circuitpython master branch with tinyusb submodule at legacy-1076-g1ee9ef4f (I quickly inspected tinyusb master branch for relevant changes but didn't find any)

**To Reproduce**
 1. Using a supported board with CircuitPython master branch, run https://github.com/CedarGroveStudios/Classic_MIDI_FeatherWing/blob/master/examples/ProgramChange_test.py (some additional support libraries are required, let me know if this poses a problem for your testing procedure and we can minimize it).  The CircuitPython console will show the sent messages as it understood them.
 1. On Windows, start "MidiTls", open the "MIDI input messages" window, and "Record"
 1. Note that with the same or nearly same timestamp, there are two "Program Change" messages reported

**Expected behavior**
Only a single Program Change message should be recorded in each repetition.

**Screenshots**

![](https://user-images.githubusercontent.com/29906257/62904404-065c7500-bd1b-11e9-9145-9d5d2af6ffae.png)

![](https://user-images.githubusercontent.com/29906257/62904473-5d624a00-bd1b-11e9-8388-18db313ae53a.png)

**Additional context**
https://github.com/adafruit/circuitpython/issues/2057

I have a fix for the problem, pull request incoming.

PR #99, opened 2019-08-15
midi_device: Send message classes C and D as 2 bytes

jepler:
.. sending these as two 1-byte messages with CIN of 0xF is misinterpreted by OSes including Windows 10 and (reportedly) Mac.

Testing performed: Ran the reproducer script, which sends Program Change (0xCx) messages.  I did not check with Change Pressure (0xDx) messages, but expect the same thing.

Closes: #98


hathach:
Thank you for your PR

Issue #97, opened 2019-08-13
stm32f4 doesn't work when compile with gcc 8.3.1 with -Os  


hathach:
**Describe the bug**
- example cdc_msc_hid doesn't work when compiled with gcc 8.3.1 with -Os. It works with gcc v7 though  

**Set up (please complete the following information):**
 - OS:  Ubuntu 18.04
 - Board: stm32f407disco
 - Firmware Code: examples/device/cdc_msc_hid

**To Reproduce**
compile with gcc 8.3.1

**Expected behavior**
usb device is not recognized

cr1901:
Commit 8f8ca77 will fixes this for Nucleo, but the fix applies to all STM32 targets.

AFAIK, there isn't a nice `typedef` in the STM32 headers for accessing the TX/RX FIFOs, so I used a `uint32_t *`. Because I forgot the `volatile` qualifier for the FIFO pointers, `gcc` 8.3.1 decided during optimizations to turn the setup packet-handling code from this:

```
_setup_packet[4 - 2*setup_left] = (* rx_fifo);
_setup_packet[5 - 2*setup_left] = (* rx_fifo);
```

to this:

```
_setup_packet[4 - 2*setup_left] = (* rx_fifo);
_setup_packet[5 - 2*setup_left] = _setup_packet[4 - 2*setup_left];
```

Because a read from RX FIFO was optimized away, this meant that the RX FIFO wasn't in a proper state (empty _or_ aligned to a status word) when `OTG_FS_IRQHandler` finished. Then the `RXLEVEL` interrupt triggered again, and the `read_rx_fifo` function read the RX FIFO entry that _should've_ been placed in `_setup_packet[5 - 2*setup_left]`.

Since the data that goes into `_setup_packet[5 - 2*setup_left]` is _not_ a status word, `read_rx_fifo` sees garbage and triggers a `TU_BREAKPOINT`. Marking the FIFO pointers as `volatile` forces all RX/TX FIFO reads and writes to actually occur, fixing the bug.

This was fun to debug :D!... I think?

hathach:
Superb !! thanks for the fix. Will close this issue once that PR is merged. `volatile` keyword indeed is source of many optimization issue, I made lots of mistake with it in the past as well :)

hathach:
close by PR #114

PR #114, opened 2019-09-03
STM32H7 Support

cr1901:
This PR implements the tinyUSB `cdc_msc_hid` demo for the `stm32h743nucleo` board.

The STM32H7 microcontroller uses the same Synopsys core as the STM32F4, with some minor programming differences. Consequently, I was able to reuse a large amount of the STM32F4 driver to implement support for H7. _It probably makes sense at some point to combine F3, F4, and H7 into a single driver at some point._

For debugging purposes, and to keep changes manageable, I didn't just copy and paste the STM32F4 driver into `dcd_stm32h7.c`. There are a few general differences:
* H7 has 8 endpoints, so all endpoint limits were updated.
* Peripheral naming differs between F4 and H7:  `USB_OTG_FS` vs `USB2_OTG_FS`.
* TX FIFO sizes are updated accordingly to accommodate 8 endpoints and more FIFO RAM (1.25kB vs 4kB on H7).

Additionally, there are a few bugfixes and behaviors from the `TODOs` that I implemented while writing this H7 driver, compared to F4:

* I managed to fix the setup packet handling code in the H7 port :D! In F4 currently, due to a bug, I programmed the USB core to [only accept one](https://github.com/hathach/tinyusb/blob/master/src/portable/st/stm32f4/dcd_stm32f4.c#L534-L539) SETUP packet at a time instead of 3. This is technically out of spec, but it worked at the time. This is now [fixed](https://github.com/hathach/tinyusb/commit/d4302dacc584c4488cf1905caa822f96bd64930d) in the STM32H7 driver.
* In the F4 driver, I had an open [TODO](https://github.com/hathach/tinyusb/blob/master/src/portable/st/stm32f4/dcd_stm32f4.c#L544) to `TU_BREAKPOINT()` if an invalid RX FIFO word was found. That is [implemented](https://github.com/hathach/tinyusb/compare/master...cr1901:stm32h7?expand=1#diff-ed90207b1ce4da43a089e757e02b93edR544) in the H7 driver.
* Additionally in the F4 driver, I made a [comment](https://github.com/hathach/tinyusb/blob/master/src/portable/st/stm32f4/dcd_stm32f4.c#L374-L380) to future-me to experiment with STALL behavior due to ambiguous language about the behavior of `STALL` and `SNAK` bits in the F4 reference manual. There's no such ambiguity in the H7 manual, so I removed that comment and "set both `STALL` and `SNAK`" behavior.<sup>1</sup>

I believe all the above can safely be implemented in F4 and F3 drivers without issue. _There are still numerous TODOs left, but progress has been made!_

~As with the `stm32f407disco` demo, this demo currently only works in debug mode; the demo fails to enumerate when compiled in release mode with recent gcc.~

Fixed as of 8f8ca77a. That was [fun to debug](https://github.com/hathach/tinyusb/issues/97#issuecomment-527744295)!

cc: @adafruit @tannewt @ptorrone

1. In fact, I can't find the ambiguous language in my copy of the F4 reference manual anymore! So maybe I misunderstood the manual when doing the F4 port?

### DIFF
--- a/hw/bsp/stm32h743nucleo/STM32H743ZITx_FLASH.ld
+++ b/hw/bsp/stm32h743nucleo/STM32H743ZITx_FLASH.ld
@@ -1,0 +1,173 @@
+/*
+*****************************************************************************
+**
+
+**  File        : LinkerScript.ld
+**
+**  Abstract    : Linker script for STM32H743ZITx Device with
+**                2048KByte FLASH, 1056KByte RAM
+**
+**                Set heap size, stack size and stack location according
+**                to application requirements.
+**
+**                Set memory bank area and size if external memory is used.
+**
+**  Target      : STMicroelectronics STM32
+**
+**
+**  Distribution: The file is distributed as is, without any warranty
+**                of any kind.
+**
+**  (c)Copyright Ac6.
+**  You may use this file as-is or modify it according to the needs of your
+**  project. Distribution of this file (unmodified or modified) is not
+**  permitted. Ac6 permit registered System Workbench for MCU users the
+**  rights to distribute the assembled, compiled & linked contents of this
+**  file as part of an application binary file, provided that it is built
+**  using the System Workbench for MCU toolchain.
+**
+*****************************************************************************
+*/
+
+/* Entry Point */
+ENTRY(Reset_Handler)
+
+/* Highest address of the user mode stack */
+_estack = 0x20020000;    /* end of RAM */
+/* Generate a link error if heap and stack don't fit into RAM */
+_Min_Heap_Size = 0x200;      /* required amount of heap  */
+_Min_Stack_Size = 0x400; /* required amount of stack */
+
+/* Specify the memory areas */
+MEMORY
+{
+DTCMRAM (xrw)      : ORIGIN = 0x20000000, LENGTH = 128K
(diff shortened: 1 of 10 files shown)


hathach:
superb !!! Thank you very much again @cr1901 . Please give me a couple of days, I am ordering the h7 board https://www.proe.vn/nucleo-h743zi2 for testing :)

ladyada:
thanks for the update! @hathach will review this PR
@tannewt feel free to test as well

cr1901:
> typedef usb_fifo_t is a bit overkill, we can just use `volatile uint32_t *` here, but I am fine with either way :)

Oh right, the reason I did that was to reduce the change of me forgetting `volatile` in future code :). It still may be overkill tho.

hathach:
> > typedef usb_fifo_t is a bit overkill, we can just use `volatile uint32_t *` here, but I am fine with either way :)
> 
> Oh right, the reason I did that was to reduce the change of me forgetting `volatile` in future code :). It still may be overkill tho.

It is merged, don't worry, we will change it later on if needed :) . This PR helps me realize that the stm32lib folder structure is not ideal. I make a PR to switch it to easy-to-upgrade st_driver submodule 
https://github.com/hathach/tinyusb/pull/118 . Be sure to merge and re-run 

`git submodule update --init --recursive`

when you pull from master :)

PR #120, opened 2019-09-08
Ported STM32F412 Discovery board, close #102

### LINKED ISSUE #102: Port NXP LPC55S6x
**Porting layer**
 - [x] Device Controller Driver (DCD)
 - [ ] Host Controller Driver (HCD)
 - [x] Board Supported Package (BSP)
 - [ ] OS Abstraction Layer (OSAL)

**Description**
This new mcu share same usb controller with already ported lpc, it should be quick to get supported.

hathach:
- stm32f412 has more endpoints, 6 instead of 4 from f407/f411). dcd_stm32f4 is updated to support either 4 or 6 max endpoints. 
- Also pca10059 and lpcxpresso11u37 board are supported as well since they are easy enough

@cr1901 @hierophect


ladyada:
@hierophect

PR #122, opened 2019-09-09
Synopsys Refactor

cr1901:
This PR refactors the STM32F4 and STM32H7 drivers to use a single `dcd_synopsys.c` driver file, since a number of STM32s opt to use [DesignWare Hi-Speed USB 2.0 On-the-Go Controller](https://www.synopsys.com/dw/ipdir.php?ds=dwc_usb_2_0_hs_otg) to provide USB.

_In addition, some non-ARM microcontrollers also use this same IP; I've heard of RISCV and ESP32 cores using it. I think it may be worthwhile to use the same source file for non-ARM microcontrollers as well within reason._

I discovered that `USB_OTG_FS` is a working alias for `USB2_OTG_FS` in the H7 headers, so I've elected to use `USB_OTG_FS` for everything for now.

I have tested the following boards with this PR, and they work fine:
* `hw\bsp\stm32f407disco`
* `hw\bsp\stm32h743nucleo`

@hathach: Could you please test the remaining F4 boards? I already modified the `board.mk` files, I just can't test them. Additionally _I have not ported over F3 to the new `dcd_synopsys.c` file yet._ Could you perhaps test that out as well?

cc: @ladyada @tannewt @ptorrone


ladyada:
awesome, this is great to see
FYI also to @spritetm

pigrew:
Versus my FSUSB branch, this has a different style of setting "DCFG_TUSB_MCU". I set mine to "OPT_MCU_STM32_FSDEV" (based on the generic driver name), but you set yours to "OPT_MCU_STM32F4" (based on the MCU name). I see advantages both ways, and all I know is that we should be consistent between the branches.

In my code, I configure the driver based on, e.g. the "STM32F407xx" define and CMSIS headers, as I have to use those variables anyway since different models have different RAM or other tweaks.

I see two other related issues:

* Some F1 MCU use the synopsis driver whereas some F1 use the ST_FSDEV driver.
* The STM32F429 (for example) has two cores FS and HS. I've not looked, but are these similar enough that they should share a single driver? Or do they need two separate drivers (syn_hs and syn_fs)? If separate, how will the build system manage using two separate drivers for the same MCU?

Thoughts? I'm a little bit lost, but I don't think I'm doing things the intended way.

cr1901:
@pigrew 
> Versus my FSUSB branch, this has a different style of setting "DCFG_TUSB_MCU". I set mine to "OPT_MCU_STM32_FSDEV" (based on the generic driver name), but you set yours to "OPT_MCU_STM32F4" (based on the MCU name). I see advantages both ways, and all I know is that we should be consistent between the branches.

[Prior art](https://github.com/hathach/tinyusb/blob/master/src/portable/nxp/lpc_ip3511/dcd_lpc_ip3511.c#L39-L43) uses the MCU name. Similar to the linked source, I also use the `DCFG_TUSB_MCU` define (as opposed to the CMSIS header define) to gate MCU-specific limits (endpoints, RAM, etc).

>Some F1 MCU use the synopsis driver whereas some F1 use the ST_FSDEV driver.

If this PR is accepted as-is, F1 MCUs using the `ST_FSDEV` driver will need to set the `CHIP_FAMILY` Makefile var to `stm32_fsdev`. Likewise, F1 MCUs using the Synopsys IP will need to set `CHIP_FAMILY` to `synopsys`. I don't see an issue as long as each family member is documented as to which "`CHIP_FAMILY`" (kinda a misnomer now :P?) is required.

>The STM32F429 (for example) has two cores FS and HS. I've not looked, but are these similar enough that they should share a single driver? Or do they need two separate drivers (syn_hs and syn_fs)? If separate, how will the build system manage using two separate drivers for the same MCU?

AFAIK, the same driver should work for both, or the HS functionality can be gated. But I'm not prepared to debug HS issues.

Spritetm:
> FYI also to @Spritetm

Thanks for the headsup, I'll have the person who's doing tinyusb integration look at this.

hathach:
> Versus my FSUSB branch, this has a different style of setting "DCFG_TUSB_MCU". I set mine to "OPT_MCU_STM32_FSDEV" (based on the generic driver name), but you set yours to "OPT_MCU_STM32F4" (based on the MCU name). I see advantages both ways, and all I know is that we should be consistent between the branches.
> 
> In my code, I configure the driver based on, e.g. the "STM32F407xx" define and CMSIS headers, as I have to use those variables anyway since different models have different RAM or other tweaks.

IMO, OPT_MCU should really is the exact mcu family rather than the controller using by the mcu. The controller can be inferred  from specific mcu family. There is probably minor difference between mcu we need to know as well (chip header, peripheral address or naming etc...) It will also be more consistent with other vendor mcu such as NXP, Microchip. 

> I see two other related issues:
> 
> * Some F1 MCU use the synopsis driver whereas some F1 use the ST_FSDEV driver.

No worry, the dcd_.c file can be selectively picked up for appropriate bsp/project from the makefile.We can have both dcd e.g `dcd_dcrotg` and `dcd_stm32_fsdev` supporting STM32F1, since only one dcd_*.c is selcted for compiling.

> * The STM32F429 (for example) has two cores FS and HS. I've not looked, but are these similar enough that they should share a single driver? Or do they need two separate drivers (syn_hs and syn_fs)? If separate, how will the build system manage using two separate drivers for the same MCU?
> 
> Thoughts? I'm a little bit lost, but I don't think I'm doing things the intended way.

FS and HS has several minor difference e.g naming, irq number, irq handler name etc ... however in general it is pretty much the same (haven't checked it thoroughly) . But the are merged able. To support OTG_HS there is probably more works, but it can wait for next PR, not a problem at all.

hw/bsp/stm32f407disco/board.mk
```diff
@@ -30,10 +30,10 @@ INC += \
 	$(TOP)/$(ST_CMSIS)/Include \
 	$(TOP)/$(ST_HAL_DRIVER)/Inc \
 	$(TOP)/hw/bsp/$(BOARD)
-	
+
 # For TinyUSB port source
 VENDOR = st
-CHIP_FAMILY = stm32f4
+CHIP_FAMILY = synopsys
```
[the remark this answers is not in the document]
cr1901: >hmm, this naming is not ideal, If we aim to make this dcd generic across vendors. We should have it under synopsys/dcd_dwcotg. 

What I did was just the _easiest_ solution, not the best :).

>I would suggest to change the VENDOR to synopsys, CHIP_FAMILY to dwcotg (design ware otg or something similar that you could comeup), or even better to figure out the exact product code/revision for the IP. 

This works fine for me, I'll fix it later today or tomorrow.

>Synopsys seems to have several of them. I tried to downoad the databook & programming guide but no luck in the registration.

I couldn't even find the programming guide, tbh :P.

src/portable/st/synopsys/dcd_synopsys.c
```diff
@@ -27,10 +27,24 @@
 
 #include "tusb_option.h"
 
-#if TUSB_OPT_DEVICE_ENABLED && CFG_TUSB_MCU == OPT_MCU_STM32F4
+#if TUSB_OPT_DEVICE_ENABLED && ( CFG_TUSB_MCU == OPT_MCU_STM32F4 || \
+                                 CFG_TUSB_MCU == OPT_MCU_STM32H7 )
```
pigrew: Can we add the supported F1 variants to this ifdef? We can say ((CFG_TUSB_MCU == OPT_MCU_STM32F1) && (defined(stm32f105xc) || defined(stm32f107xc))).

src/portable/st/synopsys/dcd_synopsys.c
```diff
@@ -27,10 +27,24 @@
...
+  #include "stm32f4xx.h"
+  // TODO Merge with OTG_HS
+  // Max endpoints for each direction
+  #define EP_MAX        USB_OTG_FS_MAX_IN_ENDPOINTS
+  #define EP_FIFO_SIZE  USB_OTG_FS_TOTAL_FIFO_SIZE
+#elif CFG_TUSB_MCU == OPT_MCU_STM32H7
+  #include "stm32h7xx.h"
+  // TODO There is no equivalent macro for max endpoints in H7 header.
+  #define EP_MAX        8
+  #define EP_FIFO_SIZE  4096
```
pigrew: For F1, use USB_OTG_FIFO_SIZE, but not sure about the endpoint count.

src/portable/st/synopsys/dcd_synopsys.c
```diff
@@ -27,10 +27,24 @@
...
+
+#if CFG_TUSB_MCU == OPT_MCU_STM32F4
+  #include "stm32f4xx.h"
+  // TODO Merge with OTG_HS
+  // Max endpoints for each direction
+  #define EP_MAX        USB_OTG_FS_MAX_IN_ENDPOINTS
+  #define EP_FIFO_SIZE  USB_OTG_FS_TOTAL_FIFO_SIZE
+#elif CFG_TUSB_MCU == OPT_MCU_STM32H7
+  #include "stm32h7xx.h"
+  // TODO There is no equivalent macro for max endpoints in H7 header.
```
pigrew: Maybe use `defined(USB_OTG_EPNUM_3) && !defined(USB_OTG_EPNUM_4)` (to know that there are 4 OUT endpoints), and then add the OUT and the IN counts? This same logic applies for the F1.

src/portable/st/synopsys/dcd_synopsys.c
cr1901: `USB_OTG_FIFO_SIZE` has the correct value, but AFAICT it's not meant for actually querying how much FIFO space is present; it's used to get bounds on the amount of address space used for registers and I/O (look at the other variables it's grouped with in the header). So I'm hesitant to use that define for querying FIFO size.

src/portable/st/synopsys/dcd_synopsys.c
cr1901: This sounds like it'll work if you can test it beforehand. I am unable to test F1 boards because I don't have any :).

pigrew: Unfortunately I don't have a Synopsys F1...

(I only have the ST-USB F1). We can skip it you think it would add confusion.

hw/bsp/stm32f407disco/board.mk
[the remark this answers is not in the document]
hathach: > > hmm, this naming is not ideal, If we aim to make this dcd generic across vendors. We should have it under synopsys/dcd_dwcotg.
> 
> What I did was just the _easiest_ solution, not the best :).
> 
> > I would suggest to change the VENDOR to synopsys, CHIP_FAMILY to dwcotg (design ware otg or something similar that you could comeup), or even better to figure out the exact product code/revision for the IP.
> 
> This works fine for me, I'll fix it later today or tomorrow.
> 
> > Synopsys seems to have several of them. I tried to downoad the databook & programming guide but no luck in the registration.
> 
> I couldn't even find the programming guide, tbh :P.

I couldn't get around registration to download databook and programming guide as well. What the heck is site-id , but I guess we will just adding codes when porting new mcu with same synopsys ip :) . 

If you wish not to rename the vendor/family, we can just merge it now and rename it later on when we work with other mcu. I am totally fine with either way, I will let you decide on this matter.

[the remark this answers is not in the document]
pigrew: My Synopsys account expired. Site ID is a number describing which institution you work for. I think that they are assigned in numerical ascending order. I requested a "reactivation", but I'm not sure if it'll go through without bugging real people with authority....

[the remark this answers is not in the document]
cr1901: >I am totally fine with either way, I will let you decide on this matter.

I would prefer to rename it later then, thanks :). IME, incremental changes, such as renaming later, are still working fine in general for tinyusb.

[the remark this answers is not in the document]
hathach: @cr1901 I will merge it then, we will do the renaming later

src/portable/st/synopsys/dcd_synopsys.c
hathach: oops, sorry, I followed the email link and merged the PR without knowing you guys are discussing . @pigrew I like your #ifdef EP checks. Probably will do that when doing more test/code with this dcd. Or if you could, submit your PR :wink:

hathach: @pigrew I just checked, seem like all the header of st, only define up to USB_OTG_EPNUM_3, there is no USB_OTG_EPNUM_4 even on F7 which should support up to 6 !!! It seems to be only an bit definition only for the registers not the number of supported endpoints. I guess we have to it by hands :)

Issue #104, opened 2019-08-24
STM32F0x2 Support


cr1901:
**Porting layer**
 - [x] Device Controller Driver (DCD)
 - [ ] Host Controller Driver (HCD)
 - [x] Board Supported Package (BSP)
 - [ ] OS Abstraction Layer (OSAL)

**Description**
I do not know when I'll get around to this, but I have an interest in using the crystal less USB feature of the [STM32F0x2](https://www.st.com/en/microcontrollers-microprocessors/stm32f0x2.html#overview) family for future projects as, say, an cheap FPGA bootloader interface.

This is a Cortex-M0, and more importantly, the _PHY provided is different from the Synopsys IP for other STM32 families_. So no code duplication will be possible from STM32F4/F3/H7/etc. Thankfully, the PHY is also simpler and less featureful, so I think a port wouldn't take as long as it took me with the Synopsys PHYs.

pigrew:
It looks like the STM32F070x6/B, STM32F0708, and STM32F048 are very similar (but some don't have the automatic clock adjustment). It'd be nice if the port would support the entire family.

(I've been doing development with the 070 nucleo board with the ST stack, but am evaluating if I want to switch to TinyUSB.)

cr1901:
@pigrew Well, the STM32 families using the Synopsys PHY are about to be abstracted into a single driver. I don't see why the same can't be done with the homegrown (?) PHY. But as of right now, I am unsure when I will be working on this as my "for fun" projects are all on the backburner.

pigrew:
Thanks for your reply! If I make any progress towards porting, I'll post a note here (I have to wrap my head around the library architecture first). My medium-term goal is writing a USBTMC class driver, but I would need the HAL port first.

As a side note, the F070 board doesn't have a USB device connector. I cut a USB cable in half and crimped on a 0.100" pitch header to connect to the board's pins (GND, D+, D-). I couldn't get the DFU bootloader working, but running their CDC example is successful.

pigrew:
I've started working on the code a little bit. I'll post my fork on github once I get EP0 up and running, but probably not make a pull request until I get some transfers going. I'm having trouble with the initialization methods for the endpoint states at the moment.... Also, the initialization done by HAL is a little bit different than what is described in the reference manual (e.g. adding time delays to let the hardware settle).

pigrew:
[...]

PR #123, opened 2019-09-09
STM32 FS Device Driver

pigrew:
The attached driver has been tested on the F070RB, but should apply to other models (though I've not yet tested on any model). It has a bunch of limitations (like not supporting isochronous transfers), but is a good starting point.

For development, I've been using STM32CubeIDE, in a project created by STM32CubeMX.

This is related to issue #104.


pigrew:
It looks like the warnings I added to the source are causing the build to fail. They are saying that the driver is untested.... Should I remove them?

Also, the linker is running out of RAM for the "msc_dual_lun" example on the F070. The MCU has 16k but says it needs an additional 4k.

Should we create a "#define CFG_EXAMPLE_MSC_DISK_DUAL_READONLY" which we add into the BSP makefile?

src/portable/st/stm32_fsdev/dcd_stm32_fsdev_pvt_st.h
```diff
@@ -0,0 +1,288 @@
...
+// On 1024-byte devices, access with a stride of one word (use every 16-bit address)
+
+#ifndef PORTABLE_ST_STM32F0_DCD_STM32F0_FSDEV_PVT_ST_H_
+#define PORTABLE_ST_STM32F0_DCD_STM32F0_FSDEV_PVT_ST_H_
+
+#if defined(STM32F042x6) | \
+    defined(STM32F070x6) | defined(STM32F070xB) | \
+    defined(STM32F072xB) | \
+    defined(STM32F078xx)
+#include "stm32f0xx.h"
```
hathach: it would be simpler to use OPT_MCU_STM32F0 to just include `#include "stm32f0xx.h"` here right :)

hw/bsp/stm32f070rbnucleo/stm32f070rbnucleo.c
```diff
@@ -0,0 +1,158 @@
...
+  GPIO_InitStruct.Pull = GPIO_PULLDOWN;
+  GPIO_InitStruct.Speed = GPIO_SPEED_FREQ_HIGH;
+  HAL_GPIO_Init(BUTTON_PORT, &GPIO_InitStruct);
+
+  // Start USB clock
+  __HAL_RCC_USB_CLK_ENABLE();
+}
+
+void dcd_fs_irqHandler(void);
+void USB_IRQHandler(void)
```
hathach: I like this idea, to give a bit more flexibility instead of binding the IRQ handler by default name. However, to be consistent with other port implementation. Would you mind just move this into the dcd. We can #if def to have the exactly IRQ name depending on the OPT_MCU_ if needed. We could revisit this `dcd_irq_handler()` later on if needed. It would need the whole codebase changes since we adding an API for dcd layer.

PS: this also seems to cause travis build failed on `board_test` example

hathach:
> The attached driver has been tested on the F070RB, but should apply to other models (though I've not yet tested on any model). It has a bunch of limitations (like not supporting isochronous transfers), but is a good starting point.
> 
> For development, I've been using STM32CubeIDE, in a project created by STM32CubeMX.
> 
> This is related to issue #104.

No problem, as long as it works, I will verify with other board/port at my side with gcc. Don't worry about Isochronous, the stack actually doesn't handle isochronous for now. We can always add features later :)

> Should we create a "#define CFG_EXAMPLE_MSC_DISK_DUAL_READONLY" which we add into the BSP makefile?

Yeah, I think we could do this, let's called it `CFG_EXAMPLE_MSC_READONLY`, even with single disk, there is mcu that doesn't have enough 8KB of SRAM.  If you don't have time, I will do it myself when the PR is merged.

hw/bsp/stm32f070rbnucleo/board.mk
```diff
@@ -0,0 +1,50 @@
...
+
+# For TinyUSB port source
+VENDOR = st
+CHIP_FAMILY = stm32_fsdev
+
+# For freeRTOS port source
+FREERTOS_PORT = ARM_CM4F
+
+# For flash-jlink target
+JLINK_DEVICE = stm32f303vc
```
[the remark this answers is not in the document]
pigrew: Yes. I've changed it (and FREERTOS_PORT).

src/common/tusb_verify.h
```diff
@@ -105,7 +105,7 @@
 #define TU_VERIFY_1ARGS(_cond)                         TU_VERIFY_DEFINE(_cond, , false)
 #define TU_VERIFY_2ARGS(_cond, _ret)                   TU_VERIFY_DEFINE(_cond, , _ret)
 
-#define TU_VERIFY(...)                   GET_3RD_ARG(__VA_ARGS__, TU_VERIFY_2ARGS, TU_VERIFY_1ARGS)(__VA_ARGS__)
+#define TU_VERIFY(...)                   GET_3RD_ARG(__VA_ARGS__, TU_VERIFY_2ARGS, TU_VERIFY_1ARGS, UNUSED)(__VA_ARGS__)
```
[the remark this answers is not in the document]
pigrew: GCC gave warnings (in pedantic mode) about C99 compliance stating that a variadic define must have at least one argument for the "...". When TU_VERIFY() was called here with one argument, it gave exactly three to GET_3RD_ARG(), leading to the warning.

hw/bsp/stm32f070rbnucleo/board.mk
```diff
@@ -0,0 +1,50 @@
+CFLAGS += \
+	-DHSE_VALUE=8000000 \
+	-DSTM32F070xB \
+	-mthumb \
+	-mabi=aapcs-linux \
+	-mcpu=cortex-m0 \
+	-mfloat-abi=soft \
+	-nostdlib -nostartfiles \
+	-DCFG_TUSB_MCU=OPT_MCU_STM32_FSDEV
```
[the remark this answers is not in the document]
pigrew: The fact that the F102 uses FSDEV while the F105 uses synopsis adds complexity. I'll use the full number in the define (for all STM32, or at least the ones supported by FSDEV).

hw/bsp/stm32f070rbnucleo/stm32f070rbnucleo.c
pigrew: Done

pigrew:
@hathach I believe I've resolved the issues you mentioned.

src/tusb_option.h
```diff
@@ -51,9 +51,29 @@
 #define OPT_MCU_SAMD21        200 ///< MicroChip SAMD21
 #define OPT_MCU_SAMD51        201 ///< MicroChip SAMD51
 
+// ST Synopsis OTG devices
 #define OPT_MCU_STM32F4       300 ///< ST STM32F4
-#define OPT_MCU_STM32F3       301 ///< ST STM32F3
 #define OPT_MCU_STM32H7       302 ///< ST STM32H7
+// ST FSDEV Devices
+#define OPT_MCU_STM32F0x0     330 ///< ST STM32F0x0
```
hathach: why would you need this many of OPT_MCU. I would only prefer a family line such as OPT_MCU_STM32F0 / F1/F3 . They are 90% like each other within the line, these OPT_MCU_* is meant for USBD and higher layer to know which family high layer working with. For specific/detail mcu, the dcd_*.c could check the mcu specific macro for usage.

pigrew:
Will remove in a few minutes. I had added them since they all use the same "ST_FSDEV" IP (so the driver should work on them all).

hw/bsp/stm32f070rbnucleo/board.mk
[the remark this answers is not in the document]
hathach: > The fact that the F102 uses FSDEV while the F105 uses synopsis adds complexity. I'll use the something like STM32F1x2 or STM32F3x3 for all of the ST parts. It seems like this is enough to decide which driver to use.

OK, I didn't know this until you mention, let's just keep them as you listed here

pigrew:
Defines are now only for the F[013] series.

As I've only tested on the F070, I expect other bugs will appear.... My plan is to handle issues on other MCU with more of a "bug fixing" attitude. Once this PR merged, I'll do some testing on a STM32F042.

Note that the driver should work on many STM32 L/G/MP series MCUs. Once this PR is merged, I can add some comments on those feature requests mentioning that it's more of a testing issue than writing an entire driver from scratch.

hathach:
@pigrew OK, I only read your comment https://github.com/hathach/tinyusb/pull/123#discussion_r322715935 in this issue.  Now I understand why you separate them. For B,C,D usb core, they are all sysnopsis, I think we can merge them together. Only the A stm32_fsdev is different. F1 is probably the only line has 102,103 for A and 105/107 for B. Maybe we can have 2 OPT_MCU for that `OPT_MCU_STM32F1x2_1x3` and `OPT_MCU_STM32F1x5_1x7`. Though since application should pick the correct dcd_*.c, maybe it we can get away if #ifdef in the dcd_*.c file

```
Each device with USB support embeds at least one of the following interfaces:
A: USB 2.0 FS device interface
B: USB 2.0 OTG FS, that is, USB 2.0 FS device/host/OTG controller with on-chip FS PHY
C: USB 2.0 OTG HS, that is, USB 2.0 FS/HS device/host/OTG controller, integrating the transceivers for full-
speed operation, and featuring an ULPI for high-speed operation: an external PHY device connected to the
ULPI is required.
D: USB 2.0 OTG HS controller with embedded on-chip HS PHYs
```

pigrew:
> @pigrew OK, I only read your comment [#123 (comment)](https://github.com/hathach/tinyusb/pull/123#discussion_r322715935) in this issue. Now I understand why you separate them. For B,C,D usb core, they are all sysnopsis, I think we can merge them together. Only the A stm32_fsdev is different. F1 is probably the only line has 102,103 for A and 105/107 for B. Maybe we can have 2 OPT_MCU for that `OPT_MCU_STM32F1x2_1x3` and `OPT_MCU_STM32F1x5_1x7`. Though since application should pick the correct dcd__.c, maybe it we can get away if #ifdef in the dcd__.c file

My preference would be to leave all STM32 as a 3-digit code for three reasons:

- Consistancy. If they are all [0-9]x[0-9], then I don't need to think about which define to use on STM32 since they all are using the same format.
- Future-proof: It's possible that ST will extend a line (like the newer L or G) to have both the synopsys and ST IP. I expect defines based on the first and third digits to work for the future.
- I can include all of the _dcd drivers in my project at the moment. For my IDE, I include the entire "tusb/src" directory without needing to exclude certain drivers. If we need the user to only include the particular driver, then makefiles become more complicated.

pigrew:
Or do you propose to have the drivers decide if they should be enabled based on the CMSIS defines, instead of only the OPT_MCU define?

src/tusb_option.h
hathach: > > @pigrew OK, I only read your comment [#123 (comment)](https://github.com/hathach/tinyusb/pull/123#discussion_r322715935) in this issue. Now I understand why you separate them. For B,C,D usb core, they are all sysnopsis, I think we can merge them together. Only the A stm32_fsdev is different. F1 is probably the only line has 102,103 for A and 105/107 for B. Maybe we can have 2 OPT_MCU for that `OPT_MCU_STM32F1x2_1x3` and `OPT_MCU_STM32F1x5_1x7`. Though since application should pick the correct dcd__.c, maybe it we can get away if #ifdef in the dcd__.c file
> 
> My preference would be to leave all STM32 as a 3-digit code for three reasons:
> 
> * Consistancy. If they are all [0-9]x[0-9], then I don't need to think about which define to use on STM32 since they all are using the same format.
> * Future-proof: It's possible that ST will extend a line (like the newer L or G) to have both the Synopsys and ST IP. I expect defines based on the first and third digits to work for the future.
> * I can include all of the _dcd drivers in my project at the moment. For my IDE, I include the entire "tusb/src" directory without needing to exclude certain drivers. If we need the user to only include the particular driver, then makefiles become more complicated.

Yeah, that make sense, though I still think it is too many, Though I think we should use the ST defined macros in both dcd_*.c to include the correct driver F102,103 vs F105,107.   

```
#if !defined (STM32F301x8) && !defined (STM32F302x8) && !defined (STM32F318xx) && \
    !defined (STM32F302xC) && !defined (STM32F303xC) && !defined (STM32F358xx) && \
    !defined (STM32F303x8) && !defined (STM32F334x8) && !defined (STM32F328xx) && \
    !defined (STM32F302xE) && !defined (STM32F303xE) && !defined (STM32F398xx) && \
    !defined (STM32F373xC) && !defined (STM32F378xx)
```

hathach:
> > @pigrew OK, I only read your comment [#123 (comment)](https://github.com/hathach/tinyusb/pull/123#discussion_r322715935) in this issue. Now I understand why you separate them. For B,C,D usb core, they are all sysnopsis, I think we can merge them together. Only the A stm32_fsdev is different. F1 is probably the only line has 102,103 for A and 105/107 for B. Maybe we can have 2 OPT_MCU for that `OPT_MCU_STM32F1x2_1x3` and `OPT_MCU_STM32F1x5_1x7`. Though since application should pick the correct dcd__.c, maybe it we can get away if #ifdef in the dcd__.c file
> 
> My preference would be to leave all STM32 as a 3-digit code for three reasons:
> 
> * Consistancy. If they are all [0-9]x[0-9], then I don't need to think about which define to use on STM32 since they all are using the same format.
> * Future-proof: It's possible that ST will extend a line (like the newer L or G) to have both the Synopsys and ST IP. I expect defines based on the first and third digits to work for the future.
> * I can include all of the _dcd drivers in my project at the moment. For my IDE, I include the entire "tusb/src" directory without needing to exclude certain drivers. If we need the user to only include the particular driver, then makefiles become more complicated.

Yeah, that make sense, though I still think it is too many, Though I think we should use the ST defined macros in both dcd_*.c to include the correct driver F102,103 vs F105,107.   

```
#if !defined (STM32F301x8) && !defined (STM32F302x8) && !defined (STM32F318xx) && \
    !defined (STM32F302xC) && !defined (STM32F303xC) && !defined (STM32F358xx) && \
    !defined (STM32F303x8) && !defined (STM32F334x8) && !defined (STM32F328xx) && \
    !defined (STM32F302xE) && !defined (STM32F303xE) && !defined (STM32F398xx) && \
    !defined (STM32F373xC) && !defined (STM32F378xx)
```

hathach:
> Or do you propose to have the drivers decide if they should be enabled based on the CMSIS defines, instead of only the OPT_MCU define?

Yeah, in the dcd_*.c we are free to use any mcu specific macro, sorry, i think I cause you an misunderstanding in the previous comment

pigrew:
Ok. Looks like the L4 series is another edge case. L4x2 uses ST-USB while L4x5 uses Synopsys.

I'll change to the MCU being OPT_MCU_STM32[FLG][0-9].

The driver will be included based on a proper (OPT_MCU_STM32F0 ) || (OPT_MCU_STM32F1 && any (supported CMSIS defines)) || (OPT_MCU_STM32F3).

I'll only specify F series.

hathach:
> Ok. Looks like the L4 series is another edge case. L4x2 uses ST-USB while L4x5 uses Synopsys.
> 
> I'll change to the MCU being OPT_MCU_STM32[FLG][0-9].
> 
> The driver will be included based on a proper (OPT_MCU_STM32F0 ) || (OPT_MCU_STM32F1 && any (supported CMSIS defines)) || (OPT_MCU_STM32F3).
> 
> I'll only specify F series.

yeah, thanks, I think it is the best to handle specific mcu with #ifdef in the dcd.c file, ST make it too confusing for us :)

pigrew:
Try again? Is the update what you had in mind? the #ifdef is pretty short, actually (for now).
Now that I'm committed the change, perhaps I should just check if "USB_PMAADDR" is defined, in the headers, as that's a good way to tell (at least on the F1).

Should I change to `#if defined(USB_PMAADDR)` instead of `#if defined(stm32f103x6)`?

pigrew:
> > Try again? Is the update what you had in mind? the #ifdef is pretty short, actually (for now).
> > Now that I've committed the change, perhaps I should just check if "USB_PMAADDR" is defined, in the headers, as that's a good way to tell (at least on the F1).
> > Should I change to `#if defined(USB_PMAADDR)` instead of `#if defined(stm32f103x6)`?
> 
> Maybe using USB_PMAADDR is good as well, since it will save us a lot of #ifdef for specific MCU ? Though you know this better than me, I am totally OK either way.

Nevermind, that's hard. I include the CMSIS header inside of the #if statement, so I don't have that define yet to use.... Lets go with what we have now.

pigrew:
> perfect as it is, thank you very much for effort and patient. I will do more test with other F3, F0, Lx . I have several of them on my desk.

Great. To use the boards without a USB port, I crimped a 0.100" pitch header onto a cut-up USB cable. Note that some of them need external series resistors, and others need a pull-up on D+. If the pull-up is external, the MCU has to have started USB not too long after the D+ pull-up is engaged.

hathach:
> > perfect as it is, thank you very much for effort and patient. I will do more test with other F3, F0, Lx . I have several of them on my desk.
> 
> Great. To use the boards without a USB port, I crimped a 0.100" pitch header onto a cut-up USB cable. Note that some of them need external series resistors, and others need a pull-up on D+. If the pull-up is external, the MCU has to have started USB not too long after the D+ pull-up is engaged.

That is too much of hw work for me, I have an [stm32f072disco](https://www.st.com/en/evaluation-tools/32f072bdiscovery.html) which has usb available, I will add its bsp to test with. Other people probably added their fav board as well :) thanks for the head up.

PR #131, opened 2019-09-10
Add PyBoard BSP

cr1901:
Do not merge this yet; this depends on #122 being merged first. I just wanted to get this PR off the ground. This PR adds support for the Micropython [pyboard v1.1](https://www.adafruit.com/product/2390). I have verified that the Mass Storage and CDC parts of the `cdc_msc_hid` example work on this board.

However, I've run into a snag with the LED; no matter what I do, the `SysTickHandler` seems to only be firing approximately 1/10th of the correct speed (and thus the LED only toggles at 1/10th the correct speed), and I can't seem to figure out why. But the clocks are set up correctly; if they weren't, I wouldn't be getting any USB activity at all. I'll need to look closer into this as well before merge.

cc: @ladyada @ptorrone @tannewt


hw/bsp/pyboardv11/pyboardv11.c
```diff
@@ -0,0 +1,200 @@
...
+  RCC_ClkInitStruct.APB1CLKDivider = RCC_HCLK_DIV4;
+  RCC_ClkInitStruct.APB2CLKDivider = RCC_HCLK_DIV2;
+  HAL_RCC_ClockConfig(&RCC_ClkInitStruct, FLASH_LATENCY_5);
+}
+
+void board_init(void)
+{
+#if CFG_TUSB_OS  == OPT_OS_NONE
+  // 1ms tick timer
+  SysTick_Config(SystemCoreClock / 1000);
```
hathach: @cr1901 may  be moving this systick set up after SystemCoreClockUpdate() could help.

cr1901: @hathach I did this earlier today to no effect. I will try again tomorrow (after I sleep) just in case, however.

hathach: @cr1901 maybe you could print out the value of the SystemCoreClock to cdc or debugger to see if it reflect the actually running clock. Systick only take that as arguments, it is the only thing that could cause timing issue I guess. probably 405 has minor diff to 407

cr1901: @hathach That's the thing... SystemCoreClock _is_ correct at 168000000; the HAL will update this as part of `HAL_RCC_ClockConfig`.  I wonder if the SysTick_Handler is getting set up with the wrong clock somehow (16000000 would make sense, since that's about 10 times slower than 168000000).

I'll sleep on it, literally!

hathach: Yeah, I am sure it is sleepy things that cause the issue, good night :smile:

cr1901: Well, this is embarrassing... I set `HSE_VALUE` to 120MHz instead of 12MHz by accident. Fixed now, and ready for merging I think!

hathach: That is so typical when lacking of sleep 😄😄

PR #138, opened 2019-09-11
rename cdc_msc_hid to cdc_msc, hid has its own example hid_composite now

hathach:
remove hid in cdc_msc_hid demo since many of stm32 mcu doesn't have enough endpoints to run this example


hathach:
#95

PR #141, opened 2019-09-11
Stfsdev cleanups

pigrew:
Here is a small bugfix relating to clearing EP registers, but mostly cleanups related to type conversion warnings in ST_FSDEV.

### DIFF
--- a/src/portable/st/stm32_fsdev/dcd_stm32_fsdev.c
+++ b/src/portable/st/stm32_fsdev/dcd_stm32_fsdev.c
@@ -171,7 +171,7 @@ static xfer_ctl_t  xfer_status[MAX_EP_COUNT][2];
 
 static TU_ATTR_ALIGNED(4) uint32_t _setup_packet[6];
 
-static ushort newDADDR; // Used to set the new device address during the CTR IRQ handler
+static uint8_t newDADDR; // Used to set the new device address during the CTR IRQ handler
 
 // EP Buffers assigned from end of memory location, to minimize their chance of crashing
 // into the stack.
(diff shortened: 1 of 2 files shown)


pigrew:
It probably doesn't matter since the driver is only used on ARM and the compiler would likely optimize things anyway.

My thought was relating to the ISA.  AFAIK, ARM cannot do arithmetic using 8-bits or 16-bits of a register. All the math operations use 32-bit values in the registers. There is no advantage in using a shorter data type except for saving space in RAM/flash (especially for loop counters, where the loop counter has to be truncated to 8 or 16 bits at each increment).

C99 guarantees that ushort and uint have>16 bits.

It also reduces some type conversion issues in arithmetic where some intermediate values are 32-bit (since ARM arithematic instructions will store the full 32-bits to the register).  On the other hand, load/store can handle smaller widths (though the register access is only the lowest 8- 16- or 32- bits. There is no way to do a 16-bit load into the upper half of a register)

I didn't check the actual assembly output to see if it makes a difference, though.

My coding preference would be to specify the widths of variables that are defined by the hardware or specification, but leave most of the internal variables as ushort/uint/etc...

Although, we can undo these type changes if you'd like.

(Beware the MSP430 which has a C implementation where there are no 8-bit data types (a char is 16-bits.)

pigrew:
On the other hand the Internet is asserting that MISRA suggests avoiding int and uint.... I'm trying to hunt down their rules to see what they actually say.

pigrew:
(Updated) Another "interesting" tip I found online was to reduce the usage of #define macros, and to use const variables or inline functions, due to them having better type safety and easier for static code analysis. Thoughts?

hathach:
> (Updated) Another "interesting" tip I found online was to reduce the usage of #define macros, and to use const variables or inline functions, due to them having better type safety and easier for static code analysis. Thoughts?

Yeah I know that tips as well, use enum/inline function as much as possible instead of macros. Those are really good tips, Do it if you can, I really like those.

PR #150, opened 2019-09-13
Control dispatch

pigrew:
This refactors the control request dispatch code ,with the following goals:

- Better checking of the validity of endpoint numbers/interface numbers (done in a single place now).
- Prioritizes sending class&vendor request types to the vendor and class drivers.
- First gives the appropriate driver an opportunity to handle standard requests before USBD core handles them. For example, I think the USBTMC driver needs to clear its internal buffer when receiving a CLEAR_FEATURE/ENDPOINT_HALT. The HID driver might need to do special things about get_descriptor, too.

This code compiles, but is untested. I expect it not to work properly, or to have issues.

I'm creating this PR mostly in order to get feedback on if it is the correct approach.

It doesn't depend on the ZLP patch, but this branch already includes that branch, and they should be tested together.


hathach:
I am appreciated that you put effort into refactoring the usbd control interpreter, however it is rather complicated with Host (often windows) does some modification that violates the general rules in the specs.. By the quick glance, I spot a few missing cases in the PR code. 

In general, there are several ways to unroll the request type/recipient/brequest, each usb stack can try to priority one over another. And to be honest, I don't see much of advantage of this pr than the current code. Don't worry much about this usbd control, the SET_IDLE bug is just introduced recently, these can be easily troubleshoot once we put it under usb analyzer.

To sum up, I am sorry to say I won't merge this PR.

pigrew:
Ok,

How about moving the parsing of the itf, EP and driver to the start of the function? I'm a little security paranoid and think we are still missing some bounds checking in some code paths. It seems easier to analyze if we put it all in one place?

hathach:
> Ok,
> 
> How about moving the parsing of the itf, EP and driver to the start of the function? I'm a little security paranoid and think we are still missing some bounds checking in some code paths. It seems easier to analyze if we put it all in one place?
> 
> If USBTMC requires additional changes to the logic, I'll post the particular condition as a bug report.

wValue and wLength are not always itf or endpoint, it depends on the request type, and should be parsed at the branch. Rule of thumb would be using variable at the lowest scope we could.

pigrew:
> wValue and wLength are not always itf or endpoint, it depends on the request type, and should be parsed at the branch. Rule of thumb would be using variable at the lowest scope we could.

I'm not checking wValue or wLength, only wIndex. It checks wIndex based on what the recipient is set to (assumption is with EP recipient, wIndex must be a valid EP; ITF recipient must have a valid interface ID).

Please see updated code on this pull request. I've tried to make the changes to be many fewer.

src/device/usbd.c
```diff
@@ -37,6 +37,8 @@
 #define CFG_TUD_TASK_QUEUE_SZ   16
 #endif
 
+#define ARRAY_LENGTH(ARR) (sizeof(ARR) / sizeof(ARR[0]))
```
pigrew: Perhaps unnecessary. Can remove if desired (or move to a more global header as TU_ARRAY_LENGTH.

src/device/usbd.c
```diff
@@ -59,6 +61,8 @@ typedef struct {
   uint8_t ep2drv[8][2];    // map endpoint to driver ( 0xff is invalid )
 }usbd_device_t;
 
+const uint8_t INVALID_DRV_TOKEN  = 0xFFu;
```
pigrew: This makes it clear what the meaning of 0xFF is, but can be removed if deemed too verbose.

src/device/usbd.c
```diff
@@ -346,12 +350,59 @@ static bool process_control_request(uint8_t rhport, tusb_control_request_t const
 {
   usbd_control_set_complete_callback(NULL);
 
+  // First pre-compute which driver is in charge of the target (and do bounds checking)
+
+  // FIXME: Should itf and ep be 16-bit since the request's value and index are 16-bit?
+  // If so, these need byte swapping on big-endian platforms.
+  uint16_t itf = 0xFFu;
```
pigrew: This code will change the behavior in some cases. It prevents the vendor class from handling a request where the recipient is EP or ITF, and the wIndex is invalid.

src/device/usbd.c
```diff
@@ -346,12 +350,59 @@ static bool process_control_request(uint8_t rhport, tusb_control_request_t const
 {
   usbd_control_set_complete_callback(NULL);
 
+  // First pre-compute which driver is in charge of the target (and do bounds checking)
+
+  // FIXME: Should itf and ep be 16-bit since the request's value and index are 16-bit?
```
pigrew: Do we want to also check the high bytes if they are zero? Do we have a function to swap the bytes from LE to native order?

src/device/usbd.c
```diff
@@ -346,12 +350,59 @@ static bool process_control_request(uint8_t rhport, tusb_control_request_t const
...
+    TU_VERIFY(drvid < USBD_CLASS_DRIVER_COUNT);
+  }
+
   // Vendor request
   if ( p_request->bmRequestType_bit.type == TUSB_REQ_TYPE_VENDOR )
   {
     TU_VERIFY(tud_vendor_control_request_cb);
 
-    if (tud_vendor_control_complete_cb) usbd_control_set_complete_callback(tud_vendor_control_complete_cb);
+    if (tud_vendor_control_complete_cb)
```
pigrew: Adding braces for code readability. No functional change.

src/device/usbd.c
```diff
@@ -484,39 +530,53 @@ static bool process_control_request(uint8_t rhport, tusb_control_request_t const
...
-          uint16_t status = usbd_edpt_stalled(rhport, tu_u16_low(p_request->wIndex)) ? 0x0001 : 0x0000;
-          tud_control_xfer(rhport, p_request, &status, 2);
-        }
-        break;
+        drvHandled = usbd_class_drivers[drvid].control_request(rhport, p_request);
+      }
 
-        case TUSB_REQ_CLEAR_FEATURE:
-          if ( TUSB_REQ_FEATURE_EDPT_HALT == p_request->wValue )
+      if( TUSB_REQ_TYPE_STANDARD == p_request->bmRequestType_bit.type ) {
```
pigrew: Adds condition that the "standard" requests are only handled when the type is standard (after removing the TU_VERIFY of this fact.

src/device/usbd.c
```diff
@@ -484,39 +530,53 @@ static bool process_control_request(uint8_t rhport, tusb_control_request_t const
...
 
-      switch ( p_request->bRequest )
-      {
-        case TUSB_REQ_GET_STATUS:
-        {
-          uint16_t status = usbd_edpt_stalled(rhport, tu_u16_low(p_request->wIndex)) ? 0x0001 : 0x0000;
-          tud_control_xfer(rhport, p_request, &status, 2);
-        }
-        break;
+        drvHandled = usbd_class_drivers[drvid].control_request(rhport, p_request);
```
pigrew: If the driver handles it, still let USBD core handle the request. I'm not sure if the class driver should handle the request before or after the core handles it.

hathach:
hmm, I was writing PR #155 and didn't notice your update on this PR. I did follow some of your changes there. Could you please give me feedback there :)

src/device/usbd.c
hathach: there is already TU_ARRAY_SIZE() macro in the tusb_common.h file :)

src/device/usbd.c
hathach: yeah, I like it, magic number is always bad, I was tentative to use 0xff or (-1) previously. So I did a hardcode number first, which is not really good. After reviewing it, 0xff is better since we don't run into complement 2 implementation of -1 on weird platform.

src/device/usbd.c
hathach: please don't change it, the vendor request will always be forwarded into vendor function. wValue, wIndex has different meaning depending on the request type. In case of Vendor, they can be just about anything (not limited to either ep or itf).

src/device/usbd.c
hathach: yeah I did something similar as well, though I forbid the class driver from stalling if it is standard request. Request can be stalled by returning false on either the `control_request() or control_complete()` callback

src/device/usbd.c
pigrew: I see.... I have not yet discovered a use case of vendor requests. In USBTMC, I'm ignoring them for now (they have their own vendor specification).

pigrew:
Thanks for going through these things. I think that the stack will be better with the changes that we're addressing. Your PR is handling many things that I wasn't aware of.

hathach:
Thanls for your idea and suggestion. In general, stack still has lots of work to do (has been like that for a few years ) 😄😄

PR #155, opened 2019-09-13
fix #154 forward all endpoint recipeint request to class driver

### LINKED ISSUE #154: class-type control request dispatch
USBTMC requires its class driver to handle standard and non-standard control requests to its endpoints.

Currently (type=class, rcpt=EP) control requests are discarded by USBD.

Please implement a call to a class driver's control upon the following occasions:

1) (type=class, rcpt=EP,bRequest=*) control requests
2) (type=standard, rcpt=EP,bRequest=USB_REQ_CLEAR_FEATURE, wValue = ENDPOINT_HALT) control requests. These are needed because "special" buffer processing is needed (as mandated by the USBTMC spec).

It order to implement item 2, one solution would be to first dispatch all (type=standard, rcpt=EP) requests to class drivers before handling them in USBD. The request could then be handled by USBD core (if it supports it). USBD core would return false (stall?) if neither the class nor USBD supports the request. Otherwise, we can special-case just the EP-HALT request.

Code for dispatching callback (and to verify EP is valid):

```
uint8_t drvid=INVALID_DRV_TOKEN;
uint8_t ep = tu_u16_low(p_request->wIndex);
bool drvHandled = false;
if(tu_edpt_number(ep) != 0u)
{
	TU_VERIFY(tu_edpt_number(p_request->wIndex) < ARRAY_LENGTH(_usbd_dev.ep2drv));
	drvid = _usbd_dev.ep2drv[tu_edpt_number(ep)][tu_edpt_dir(ep)];
	TU_VERIFY(drvid != INVALID_DRV_TOKEN);
}

if(drvid != INVALID_DRV_TOKEN) {
	usbd_control_set_complete_callback(usbd_class_drivers[drvid].control_complete);

	drvHandled = usbd_class_drivers[drvid].control_request(rhport, p_request);
}
```

hathach:
fix typo
remove magic number 0xff of driver id

@pigrew would you mind reviewing the changes, this should help to implement USBTMC

### DIFF
--- a/src/common/tusb_common.h
+++ b/src/common/tusb_common.h
@@ -38,7 +38,7 @@
  //--------------------------------------------------------------------+
 // Macros Helper
 //--------------------------------------------------------------------+
-#define TU_ARRAY_SZIE(_arr)   ( sizeof(_arr) / sizeof(_arr[0]) )
+#define TU_ARRAY_SIZE(_arr)   ( sizeof(_arr) / sizeof(_arr[0]) )
 #define TU_MIN(_x, _y)        ( (_x) < (_y) ) ? (_x) : (_y) )
 #define TU_MAX(_x, _y)        ( (_x) > (_y) ) ? (_x) : (_y) )
 
--- a/src/device/usbd.c
+++ b/src/device/usbd.c
@@ -61,6 +61,9 @@ typedef struct {
 
 static usbd_device_t _usbd_dev = { 0 };
 
+// Invalid driver ID in itf2drv[] ep2drv[][] mapping
+enum { DRVID_INVALID = 0xff };
+
 //--------------------------------------------------------------------+
 // Class Driver
 //--------------------------------------------------------------------+
@@ -144,7 +147,7 @@ static usbd_class_driver_t const usbd_class_drivers[] =
   #endif
 };
 
-enum { USBD_CLASS_DRIVER_COUNT = TU_ARRAY_SZIE(usbd_class_drivers) };
+enum { USBD_CLASS_DRIVER_COUNT = TU_ARRAY_SIZE(usbd_class_drivers) };
 
 //--------------------------------------------------------------------+
 // DCD Event
@@ -213,8 +216,8 @@ static void usbd_reset(uint8_t rhport)
 {
   tu_varclr(&_usbd_dev);
 
-  memset(_usbd_dev.itf2drv, 0xff, sizeof(_usbd_dev.itf2drv)); // invalid mapping
-  memset(_usbd_dev.ep2drv , 0xff, sizeof(_usbd_dev.ep2drv )); // invalid mapping
+  memset(_usbd_dev.itf2drv, DRVID_INVALID, sizeof(_usbd_dev.itf2drv)); // invalid mapping
+  memset(_usbd_dev.ep2drv , DRVID_INVALID, sizeof(_usbd_dev.ep2drv )); // invalid mapping
 
   usbd_control_reset(rhport);
 
@@ -287,19 +290,19 @@ void tud_task (void)
         {
           // Invoke the class callback associated with the endpoint address
           uint8_t const ep_addr = event.xfer_complete.ep_addr;
-          uint8_t const epnum = tu_edpt_number(ep_addr);
-          uint8_t const dir   = tu_edpt_dir(ep_addr);
+          uint8_t const epnum   = tu_edpt_number(ep_addr);
+          uint8_t const ep_dir  = tu_edpt_dir(ep_addr);
 
-          _usbd_dev.ep_busy_map[dir] = (uint8_t) tu_bit_clear(_usbd_dev.ep_busy_map[dir], epnum);
+          _usbd_dev.ep_busy_map[ep_dir] = (uint8_t) tu_bit_clear(_usbd_dev.ep_busy_map[ep_dir], epnum);
 
-          if ( 0 == tu_edpt_number(ep_addr) )
+          if ( 0 == epnum )
           {
             // control transfer DATA stage callback
             usbd_control_xfer_cb(event.rhport, ep_addr, event.xfer_complete.result, event.xfer_complete.len);
           }
           else
           {
-            uint8_t const drv_id = _usbd_dev.ep2drv[tu_edpt_number(ep_addr)][tu_edpt_dir(ep_addr)];
+            uint8_t const drv_id = _usbd_dev.ep2drv[epnum][ep_dir];
             TU_ASSERT(drv_id < USBD_CLASS_DRIVER_COUNT,);
 
             usbd_class_drivers[drv_id].xfer_cb(event.rhport, ep_addr, event.xfer_complete.result, event.xfer_complete.len);
@@ -435,7 +438,7 @@ static bool process_control_request(uint8_t rhport, tusb_control_request_t const
     case TUSB_REQ_RCPT_INTERFACE:
     {
       uint8_t const itf = tu_u16_low(p_request->wIndex);
-      TU_VERIFY(itf < TU_ARRAY_SZIE(_usbd_dev.itf2drv));
+      TU_VERIFY(itf < TU_ARRAY_SIZE(_usbd_dev.itf2drv));
 
       uint8_t const drvid = _usbd_dev.itf2drv[itf];
       TU_VERIFY(drvid < USBD_CLASS_DRIVER_COUNT);
@@ -465,18 +468,16 @@ static bool process_control_request(uint8_t rhport, tusb_control_request_t const
           default:
             // forward to class driver: "STD request to Interface"
             // GET HID REPORT DESCRIPTOR falls into this case
-            usbd_control_set_complete_callback(usbd_class_drivers[drvid].control_complete);
-
             // stall control endpoint if driver return false
+            usbd_control_set_complete_callback(usbd_class_drivers[drvid].control_complete);
             TU_ASSERT(usbd_class_drivers[drvid].control_request(rhport, p_request));
           break;
         }
       }else
       {
         // forward to class driver: "non-STD request to Interface"
-        usbd_control_set_complete_callback(usbd_class_drivers[drvid].control_complete);
-
         // stall control endpoint if driver return false
+        usbd_control_set_complete_callback(usbd_class_drivers[drvid].control_complete);
         TU_ASSERT(usbd_class_drivers[drvid].control_request(rhport, p_request));
       }
     }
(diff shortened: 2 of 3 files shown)


src/device/usbd.c
```diff
@@ -61,6 +61,9 @@ typedef struct {
 
 static usbd_device_t _usbd_dev = { 0 };
 
+// Invalid driver ID in itf2drv[] ep2drv[][] mapping
+enum { DRVID_INVALID = 0xff };
```
pigrew: I'm not sure if enum is good or bad; I guess it's a matter of opinion.
In C, we can't declare the enum of any particular type; I think it ends up being an integer or unsigned integer? Would just declaring it as a "uint8_t const" work better since it would be typed? (I don't have a strong opinion on this)

src/device/usbd.c
```diff
@@ -465,56 +468,86 @@ static bool process_control_request(uint8_t rhport, tusb_control_request_t const
...
+          case TUSB_REQ_SET_FEATURE:
+            if ( TUSB_REQ_FEATURE_EDPT_HALT == p_request->wValue )
+            {
+              usbd_edpt_stall(rhport, ep_addr);
+            }
+            tud_control_status(rhport, p_request);
+          break;
+
+          // Unknown/Unsupported request
+          default: TU_BREAKPOINT(); return false;
```
pigrew: Should this be a BKPT or a TU_VERIFY(false)?

src/device/usbd.c
pigrew: How about if we also explicitly declare 0xff as unsigned by using 0xFFu? (This is creating implicit type conversion warnings)

src/device/usbd.c
```diff
@@ -287,19 +290,19 @@ void tud_task (void)
...
+          if ( 0 == epnum )
           {
             // control transfer DATA stage callback
             usbd_control_xfer_cb(event.rhport, ep_addr, event.xfer_complete.result, event.xfer_complete.len);
           }
           else
           {
-            uint8_t const drv_id = _usbd_dev.ep2drv[tu_edpt_number(ep_addr)][tu_edpt_dir(ep_addr)];
+            uint8_t const drv_id = _usbd_dev.ep2drv[epnum][ep_dir];
             TU_ASSERT(drv_id < USBD_CLASS_DRIVER_COUNT,);
```
pigrew: This can happen if the host is being bad and not due to a bug in tinyusb. Does that mean we should use TU_VERIFY?

src/common/tusb_common.h
```diff
@@ -38,7 +38,7 @@
  //--------------------------------------------------------------------+
 // Macros Helper
 //--------------------------------------------------------------------+
-#define TU_ARRAY_SZIE(_arr)   ( sizeof(_arr) / sizeof(_arr[0]) )
+#define TU_ARRAY_SIZE(_arr)   ( sizeof(_arr) / sizeof(_arr[0]) )
```
pigrew: Can I add face-palm emoji onto comments? I didn't even see this define.

src/device/usbd.c
```diff
@@ -465,56 +468,86 @@ static bool process_control_request(uint8_t rhport, tusb_control_request_t const
...
       }
     }
     break;
 
     //------------- Endpoint Request -------------//
     case TUSB_REQ_RCPT_ENDPOINT:
-      // Non standard request is not supported
-      TU_VERIFY( TUSB_REQ_TYPE_STANDARD == p_request->bmRequestType_bit.type );
+    {
+      uint8_t const ep_addr = tu_u16_low(p_request->wIndex);
```
pigrew: We maybe need to add a byte swapping to wIndex (here and many other places) in the event that the MCU is big-endian? I can create a patch to add swapping LE to host to the compiler definition file, and we can deal with this later.

src/device/usbd.c
```diff
@@ -465,56 +468,86 @@ static bool process_control_request(uint8_t rhport, tusb_control_request_t const
...
+      uint8_t const ep_num  = tu_edpt_number(ep_addr);
+      uint8_t const ep_dir  = tu_edpt_dir(ep_addr);
 
-      switch ( p_request->bRequest )
+      TU_ASSERT(ep_num < TU_ARRAY_SIZE(_usbd_dev.ep2drv) );
```
pigrew: Should this be TU_VERIFY since it can be caused by a host mistake, and not a TUSB bug?

src/device/usbd.c
```diff
@@ -465,56 +468,86 @@ static bool process_control_request(uint8_t rhport, tusb_control_request_t const
...
+
+      uint8_t const drv_id = _usbd_dev.ep2drv[ep_num][ep_dir];
+      TU_ASSERT(drv_id < USBD_CLASS_DRIVER_COUNT);
```
pigrew: We need to allow the case that epnum=0, in the case there is a clear halt on EP0, we still need to continue.
TU_VERIFY((ep_num == 0) || (drv_id < USBD_CLASS_DRIVER_COUNT))

src/device/usbd.c
```diff
@@ -465,56 +468,86 @@ static bool process_control_request(uint8_t rhport, tusb_control_request_t const
...
+
+      // Some classes such as TMC needs to clear/re-init its buffer when receiving CLEAR_FEATURE request
+      // We will forward all request targeted endpoint to its class driver
+      // - For non-standard request: driver can ACK or Stall the request by return true/false
+      // - For standard request: usbd decide the ACK stage regardless of driver return value
+      bool ret;
+
+      if ( TUSB_REQ_TYPE_STANDARD != p_request->bmRequestType_bit.type )
```
pigrew: A type of 0x03u is invalid (always, AFAIK), but allowed through here. Do we just want to compare if type is class?

src/device/usbd.c
```diff
@@ -465,56 +468,86 @@ static bool process_control_request(uint8_t rhport, tusb_control_request_t const
...
       {
-        case TUSB_REQ_GET_STATUS:
-        {
-          uint16_t status = usbd_edpt_stalled(rhport, tu_u16_low(p_request->wIndex)) ? 0x0001 : 0x0000;
-          tud_control_xfer(rhport, p_request, &status, 2);
-        }
-        break;
+        // complete callback is also capable of stalling/acking the request
+        usbd_control_set_complete_callback(usbd_class_drivers[drv_id].control_complete);
```
pigrew: In the vendor callback, we also only set the complete_callback if the cb is defined. Do we want to do that here too, as it is a weak function?

src/device/usbd.c
```diff
@@ -465,56 +468,86 @@ static bool process_control_request(uint8_t rhport, tusb_control_request_t const
...
+        // force return true for standard request
+        ret = true;
+
+        switch ( p_request->bRequest )
+        {
+          case TUSB_REQ_GET_STATUS:
           {
-            usbd_edpt_stall(rhport, tu_u16_low(p_request->wIndex));
+            uint16_t status = usbd_edpt_stalled(rhport, ep_addr) ? 0x0001 : 0x0000;
+            tud_control_xfer(rhport, p_request, &status, 2);
```
pigrew: (perhaps worry about this later)
Can the xfer fail?

src/device/usbd.c
```diff
@@ -465,56 +468,86 @@ static bool process_control_request(uint8_t rhport, tusb_control_request_t const
...
 
-        case TUSB_REQ_CLEAR_FEATURE:
-          if ( TUSB_REQ_FEATURE_EDPT_HALT == p_request->wValue )
-          {
-            usbd_edpt_clear_stall(rhport, tu_u16_low(p_request->wIndex));
-          }
-          tud_control_status(rhport, p_request);
-        break;
+      // Invoke class driver first
+      ret = usbd_class_drivers[drv_id].control_request(rhport, p_request);
```
pigrew: Make this conditional on drv_id != INVALID so that things will work on EP0.

src/device/usbd.c
```diff
@@ -287,19 +290,19 @@ void tud_task (void)
...
-          uint8_t const epnum = tu_edpt_number(ep_addr);
-          uint8_t const dir   = tu_edpt_dir(ep_addr);
+          uint8_t const epnum   = tu_edpt_number(ep_addr);
+          uint8_t const ep_dir  = tu_edpt_dir(ep_addr);
 
-          _usbd_dev.ep_busy_map[dir] = (uint8_t) tu_bit_clear(_usbd_dev.ep_busy_map[dir], epnum);
+          _usbd_dev.ep_busy_map[ep_dir] = (uint8_t) tu_bit_clear(_usbd_dev.ep_busy_map[ep_dir], epnum);
 
-          if ( 0 == tu_edpt_number(ep_addr) )
+          if ( 0 == epnum )
```
pigrew: Do we need to call the USBD control_xfer_fb  CLEAR _FEATURE, GET_STATUS, or SET_FEATURE that we need to handle for EP other than 0? (so, call the local one if the driver didn't handle the request?) If both we and the driver handle a request, who gets to set the status?

src/device/usbd.c
hathach: uint8_t const is still addressable and will take up 1 byte of flash. Not much but It isn't worthy it. Did your compiler complain, we can change it to macro instead

hathach: Yes, it should have u suffix. Thanks

pigrew: Since it's const, the compiler knows it won't change. It won't do any load/store for it; the immediate value is directly encoded in the opcode. It's treated the same exact way as the #define by the compiler; just it would have type checking.

src/device/usbd.c
hathach: All Control requests are alway sent to ep0. The setup packet data contains the targeted endpoint

src/device/usbd.c
hathach: Right, we often run into these assert due to mis-calculation (high low byte) or buffer overflow than bad host :)

pigrew: I think you can set TU_VERIFY to create breakpoints, too, for debugging?

src/device/usbd.c
hathach: Big endian mcu is rarely now. There are tons of  changes required to get it work with the stack. Therefore, we should only worry about BE MCU when we actually port one

src/device/usbd.c
pigrew: My mistake. I was confused with the wIndex/target ep=0, not the ep over which the packet came.

src/device/usbd.c
hathach: Host should never issue clear feature request on ep0. Even it does the drvid should be 0xff which is larger than the count so the condition check still cover it :)

src/device/usbd.c
hathach: Yeah, we should check it, you seem have no trust in host at all right 😁😁

src/device/usbd.c
hathach: That is not a really valid request, we can just reject it as it is. Ep0 is never halted, only protocol stall response.

src/device/usbd.c
hathach: If control xfer failed, we won't probably get here. It will be mostly dcd porting issue. But yeah, the stack still needs quite a bit of work

src/device/usbd.c
hathach: Breakpoint, you may think it is host issue. Though throughout my dev time with the stack, it is request that we haven't supported/known it exists. Buffer overflow and incorrect parsing (e.g BE mcu, miscalculation). Mostly it is our fault

src/device/usbd.c
hathach: It is still located in addressable flash and occupy storage. Enum and macro define is translated to immediate constant and won't cost anything. I will change the enum to unsigned as you suggested.

src/device/usbd.c
hathach: No, assert cause breakpoint, verify does not, we will keep it that way

src/device/usbd.c
hathach: Vendor callback is optional but driver callback is mandatory for now. Though it may change in the future if we decide to optimize the callback flow. Though we are far from speed optimization to be honest.

src/device/usbd.c
pigrew: I think you're right (about clear halt). If EP0 isn't working, it can't sendthe message, anyway.

src/device/usbd.c
pigrew: I'm treating it as the attacker. The MCU must not fail. It must not crash. Maybe I can invest in a USB fuzzer?

src/device/usbd.c
hathach: PS: Ah I see, yeah it should be checked in case the vendor callback are the one is targeted here.

PR #157, opened 2019-09-14
TI compiler quirks, and LE byte swapping functions.

pigrew:
This adds some definitions for the TI optimizing C compilers. I've built the codebase with a TI TM4C123GXL target (using a null DCD driver). I hope that it will also work with the MSP430.

I'm also adding functions to swap byte orders between LE and host orders. This would be necessary if the library is ported to BE MCUs (nobody is itching to do a ColdFire port, are they?). 

In addition it changes the byteswap macro to be an inline function. This enables compiler type checking, perhaps finding bugs in the future.

(feel free to reject some or any of this)

### DIFF
--- a/src/common/tusb_compiler.h
+++ b/src/common/tusb_compiler.h
@@ -80,7 +80,7 @@
     return __builtin_bswap16(u16);
   }
 
-  static inline uint16_t tu_bswap32(uint16_t u32)
+  static inline uint32_t tu_bswap32(uint32_t u32)
   {
     return __builtin_bswap32(u32);
   }
@@ -108,7 +108,7 @@
     return __builtin_bswap16(u16);
   }
 
-  static inline uint16_t tu_bswap32(uint16_t u32)
+  static inline uint32_t tu_bswap32(uint32_t u32)
   {
     return __builtin_bswap32(u32);
   }
--- a/src/common/tusb_compiler.h
+++ b/src/common/tusb_compiler.h
@@ -75,17 +75,8 @@
     #define TU_BYTE_ORDER TU_BIG_ENDIAN
   #endif
 
-  static inline uint16_t tu_bswap16(uint16_t u16)
-  {
-    return __builtin_bswap16(u16);
-  }
-
-  static inline uint32_t tu_bswap32(uint32_t u32)
-  {
-    return __builtin_bswap32(u32);
-  }
-
-  #define TU_BSWAP16
+#define TU_BSWAP16(u16) (__builtin_bswap16(u16))
+#define TU_BSWAP32(u32) (__builtin_bswap32(u32))
 
 #elif defined(__TI_COMPILER_VERSION__)
   #define TU_ATTR_ALIGNED(Bytes)        __attribute__ ((aligned(Bytes)))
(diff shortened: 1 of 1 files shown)


src/common/tusb_compiler.h
```diff
@@ -67,20 +70,80 @@
...
+  }
+
+  static inline uint16_t tu_bswap32(uint16_t u32)
+  {
+    return __builtin_bswap32(u32);
+  }
+
+  #define TU_BSWAP16
+
+#elif defined(__TI_COMPILER_VERSION__)
```
cr1901: I'm guessing this isn't the GCC compiler, but FWIW I develop using `msp430-elf-gcc`, and it _has_ `__BYTE_ORDER__` defined:

```
William@William-THINK MINGW64 ~/src/tinyusb
$ msp430-elf-gcc -dM -E - < /dev/null | grep __BYTE_ORDER__
#define __BYTE_ORDER__ __ORDER_LITTLE_ENDIAN__

William@William-THINK MINGW64 ~/src/tinyusb
$ msp430-elf-gcc -v
Using built-in specs.
COLLECT_GCC=C:\msys64\opt\msp430-elf-toolchain\bin\msp430-elf-gcc.exe
COLLECT_LTO_WRAPPER=c:/msys64/opt/msp430-elf-toolchain/bin/../libexec/gcc/msp430-elf/5.3.0/lto-wrapper.exe
Target: msp430-elf
Configured with: /srv/net/msp430/jenkins/workspace/msp430-pkg-gnu/gcc.subtree/configure --build=x86_64-redhat-linux --host=i686-w64-mingw32 --target=msp430-elf --enable-languages=c,c++ --disable-nls --enable-target-optspace --enable-newlib-nano-formatted-io --with-pkgversion='SOMNIUM Technologies Limited - msp430-gcc 5.3.0.219'
Thread model: single
gcc version 5.3.0 (SOMNIUM Technologies Limited - msp430-gcc 5.3.0.219)
```

Of course, being GCC, the `__GNUC__` logic will trigger :).

pigrew: As intended. I think.... I wasn't sure if you were going to use the TI or gcc compiler. I mostly use CCS for programming.

src/common/tusb_compiler.h
```diff
@@ -67,20 +70,80 @@
...
+  {
+    return __builtin_bswap16(u16);
+  }
+
+  static inline uint32_t tu_bswap32(uint32_t u32)
+  {
+    return __builtin_bswap32(u32);
+  }
+
+  #define TU_BSWAP16
```
pigrew: I think this line is a mistake.... need to remove it.

src/common/tusb_compiler.h
```diff
@@ -67,20 +70,80 @@
...
-    #define tu_htons(u16)  __builtin_bswap16(u16)
-    #define tu_ntohs(u16)  __builtin_bswap16(u16)
+    #define TU_BYTE_ORDER TU_LITTLE_ENDIAN
   #else
-    #define tu_htonl(u32)  (u32)
-    #define tu_ntohl(u32)  (u32)
+    #define TU_BYTE_ORDER TU_BIG_ENDIAN
+  #endif
+
+  static inline uint16_t tu_bswap16(uint16_t u16)
```
hathach: __builtin_bswap16/32 is already function that has type checking we can define tu_bswap16/32 a macros here.

src/common/tusb_compiler.h
hathach: @pigrew Does the CCS's TI compiler define `__GNUC__` as well, everything looks pretty much the same as GNUC. If yes, perhaps we could merge them together ?

src/common/tusb_compiler.h
```diff
@@ -67,20 +70,80 @@
...
+  #define tu_le32toh(x) (x)
+
+#elif (TU_BYTE_ORDER == TU_BIG_ENDIAN)
+  #define tu_htonl(u32)  (x)
+  #define tu_ntohl(u32)  (x)
+
+  #define tu_htons(u16)  (x)
+  #define tu_ntohs(u16)  (x)
+
+  #define tu_htole16(x) tu_bswap16(u32)
```
hathach: x and u32, u16 doesn't match in the macros

pigrew: TI defines __TI_GNU_ATTRIBUTE_SUPPORT__, but not __GNUC__. TI has at least three different compilers for the different CPUs they use (ARM, MSP430, C2000, ...), and each has its own quirks.... It's your call, but I'd tend to keep them separate as it does have little quirks.

src/common/tusb_compiler.h
pigrew: No, it does not define __GNUC__.

The logic of endianness selection is different, but they do support most of the GCC macros

[TI ARM compiler docs](http://www.ti.com/lit/ug/spnu151r/spnu151r.pdf)
[MSP430 compiler docs](http://www.ti.com/lit/ug/slau132r/slau132r.pdf)

src/common/tusb_compiler.h
pigrew: x, u32, u16 should be consistent now.

src/common/tusb_compiler.h
pigrew: Changed.

src/common/tusb_compiler.h
pigrew: Removed.

pigrew:
@hathach, I've implemented your feedback, and reverified that the codebase can build a simple example using the TI ARM compiler.

PR #158, opened 2019-09-14
Off-by-one error in Synopsys assertions. (and move some braces to the…

### LINKED ISSUE #156: Off by one error in Synopsys driver
Issue found by Coverity:

dcd_synopsys.c:273:

```
  // Unsupported endpoint numbers/size.
  if((desc_edpt->wMaxPacketSize.size > 64) || (epnum > EP_MAX)) {
    return false;
  }

```

I believe this is an off-by-one, and it should be `epnum >= EP_MAX`.

However, I'd suggest using TU_ASSERT with the inverse logic:

```
TU_ASSERT((desc_edpt->wMaxPacketSize.size <= 64);
TU_ASSERT(epnum < EP_MAX);
```

This could also be made more clear by using names like `EP_COUNT` instead of `EP_MAX`.

pigrew:
Fixes #156 

(I don't have the real hardware here, so it's untested)

### DIFF
--- a/src/portable/st/synopsys/dcd_synopsys.c
+++ b/src/portable/st/synopsys/dcd_synopsys.c
@@ -268,21 +268,22 @@ bool dcd_edpt_open (uint8_t rhport, tusb_desc_endpoint_t const * desc_edpt)
 
   uint8_t const epnum = tu_edpt_number(desc_edpt->bEndpointAddress);
   uint8_t const dir   = tu_edpt_dir(desc_edpt->bEndpointAddress);
-
-  // Unsupported endpoint numbers/size.
-  if((desc_edpt->wMaxPacketSize.size > 64) || (epnum > EP_MAX)) {
-    return false;
-  }
-
+  
+  TU_ASSERT((desc_edpt->wMaxPacketSize.size <= 64);
+  TU_ASSERT(epnum < EP_MAX);
+  
   xfer_ctl_t * xfer = XFER_CTL_BASE(epnum, dir);
   xfer->max_size = desc_edpt->wMaxPacketSize.size;
 
-  if(dir == TUSB_DIR_OUT) {
+  if(dir == TUSB_DIR_OUT)
+  {
     out_ep[epnum].DOEPCTL |= (1 << USB_OTG_DOEPCTL_USBAEP_Pos) | \
       desc_edpt->bmAttributes.xfer << USB_OTG_DOEPCTL_EPTYP_Pos | \
       desc_edpt->wMaxPacketSize.size << USB_OTG_DOEPCTL_MPSIZ_Pos;
     dev->DAINTMSK |= (1 << (USB_OTG_DAINTMSK_OEPM_Pos + epnum));
-  } else {
+  }
+  else
+  {
     // "USB Data FIFOs" section in reference manual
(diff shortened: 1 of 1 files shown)


src/portable/st/synopsys/dcd_synopsys.c
```diff
@@ -268,21 +268,22 @@ bool dcd_edpt_open (uint8_t rhport, tusb_desc_endpoint_t const * desc_edpt)
...
-
-  // Unsupported endpoint numbers/size.
-  if((desc_edpt->wMaxPacketSize.size > 64) || (epnum > EP_MAX)) {
-    return false;
-  }
-
+  
+  TU_ASSERT(desc_edpt->wMaxPacketSize.size <= 64);
+  TU_ASSERT(epnum < EP_MAX);
+
```
cr1901: I don't care for this. Considering that `dcd_edpt_open` returns a bool, I assume there's a reason in the future that the upper levels of the stack would use that information (the current behavior _will_ `TU_BREAKPOINT` eventually).

pigrew: I've been using TU_ASSERT and TU_VERIFY based on if the condition would ever be valid behavior of the MCU's firmware. My reading of the code is that if these conditions happen, it's a bug in tinyusb or the user of tinyusb. Up to now, there's no dynamic allocation of endpoints, so this should be caught the first time the program runs in the debugger.

Are you saying that these are conditions that could happen in normal code flow?

I imagine returning false from the function implies that (for example) there was a USB protocol error or the host disconnected/reset, or something like that. Perhaps the PHY/MAC can report HW errors related to the particular EP which should cause the particular EP to stall. But, as you said, these are not handled at the moment.

Since I believe these conditions can only happen if there is a bug in the MCU firmware (or cosmic rays), it should create a breakpoint (at least in DEBUG mode), so the bug can be fixed. Release mode currently will evaluate the condition and return false (without creating a breakpoint) if the condition fails.

cr1901: >Are you saying that these are conditions that could happen in normal code flow (such as if the host sends a bad packet)?

No, the only time I had this happen was when diagnosing #95, which was a logic error.

>Release mode currently will evaluate the condition and return false (without creating a breakpoint) if the condition fails.

Oh I wasn't aware of this. That makes me feel better, though I admit I never really could use `TU_ASSERT` and `TU_VERIFY` properly lol.
Oh, then that's fine then.

pigrew: @cr1901 Though on ARM M0, the breakpoint macro does nothing...
I'll try to work up a PR which updates the documentation to describe when to use which.  I may also create a TU_DEBUG_ASSERT() which does disappear....
Another related issue I just thought of is that coverity-scan probably  doesn't realize that TU_ASSERT is assert(). I should have been asking it to treat TU_BREAKPOINT as a function which should never be called.... maybe it'll find more issues that way.

hathach: @pigrew I dont really like the idea of DEBUG_ASSERT which expand to nothing in release mode. All the condition is dynamic, we can't say for sure we can omit it in release mode. Let's just leave it as it is. I don't want to follow the std assert.h path

PR #165, opened 2019-09-16
Fsdev small packet

### LINKED ISSUE #162: ST FS Dev driver: typo & EP <64 bytes
The ST FSDev driver has a rx/tx typo, and also doesn't properly support EP <64 bytes.

These are mostly fixed, see pigrew/tinyusb#1, but I don't think that it's ready for merge yet.

Do we have any sort of device test which we can run on a computer to verify that an driver does what it's supposed to do? I saw that the USB-IF has a (USB20-CV)[https://www.usb.org/usb2tools] tool to check if things work well for HID, MSC, video, and device. But, my laptop doesn't support it....

I also need to think about if any locking is needed in the driver, not sure yet, or if some of the globals need to be marked as volatile, etc. Can we guarantee that there won't be simultaneous requests on a single EP (both In and OUT) from multiple tasks?

Progress on patch is on a "draft pull request" (so it doesn't pollute the list in the main repository).

pigrew:
ST FSDEV driver updates
* Small typo fix
* Convert many of the macros to use functions. Keeping lots of stuff as `unsigned int` let the number of type conversions be reduced, reducing the final binary size by about 350 bytes (with optimization enabled). Wow. This also makes the utility functions MUCH more readable. Before the rewrite of them, I couldn't understand what they were doing.
* Implement support for EP less than 64 bytes.
* Add memory/instruction barriers after disabling interrupts
* Implement remote wakeup

These changes have been separated into three commits to make review easier.

I still need to figure out if there are thread/locking issues regarding the globals, or how EPs can be accessed by multiple threads simultaneously. Are memory barriers needed? However, these issues could be moved into a future changeset.

Fixes #162.

I've tested this by using it with the USBTMC class, but not much more than this.

I realize I used `unsigned integer` again... My understanding is that arithematic and logic operations will be up-converted to `unsigned integer` on all platforms (but uint32_t only on 32-bit platforms), so especially for passing register values, it would make sense to use `unsigned integer`? I will create a new commit to `s/unsigned integer/uint32_t`.

### DIFF
--- a/src/portable/st/stm32_fsdev/dcd_stm32_fsdev.c
+++ b/src/portable/st/stm32_fsdev/dcd_stm32_fsdev.c
@@ -121,7 +121,6 @@
 #undef USE_HAL_DRIVER
 
 #include "device/dcd.h"
-#include "bsp/board.h"
 #include "portable/st/stm32_fsdev/dcd_stm32_fsdev_pvt_st.h"
 
 
@@ -174,6 +173,7 @@ static xfer_ctl_t* xfer_ctl_ptr(uint32_t epnum, uint32_t dir)
 static TU_ATTR_ALIGNED(4) uint32_t _setup_packet[6];
 
 static uint8_t newDADDR; // Used to set the new device address during the CTR IRQ handler
+static uint8_t remoteWakeCountdown; // When wake is requested
 
 // EP Buffers assigned from end of memory location, to minimize their chance of crashing
 // into the stack.
@@ -235,7 +235,7 @@ void dcd_init (uint8_t rhport)
   {
     pma[PMA_STRIDE*(DCD_STM32_BTABLE_BASE + i)] = 0u;
   }
-  USB->CNTR |= USB_CNTR_RESETM | USB_CNTR_SOFM | USB_CNTR_CTRM | USB_CNTR_SUSPM | USB_CNTR_WKUPM;
+  USB->CNTR |= USB_CNTR_RESETM | USB_CNTR_SOFM | USB_CNTR_ESOFM | USB_CNTR_CTRM | USB_CNTR_SUSPM | USB_CNTR_WKUPM;
   dcd_handle_bus_reset();
 
   // And finally enable pull-up, which may trigger the RESET IRQ if the host is connected.
@@ -304,17 +304,9 @@ void dcd_set_config (uint8_t rhport, uint8_t config_num)
 void dcd_remote_wakeup(uint8_t rhport)
 {
   (void) rhport;
-  uint32_t start;
 
   USB->CNTR |= (uint16_t)USB_CNTR_RESUME;
-  /* Wait 1 to 15 ms */
-  /* Busy loop is bad, but the osal_task_delay() isn't implemented for the "none" OSAL */
-  start = board_millis();
-  while ((board_millis() - start) < 2)
-  {
-    ;
-  }
-  USB->CNTR &= (uint16_t)(~USB_CNTR_RESUME);
+  remoteWakeCountdown = 4u; // required to be 1 to 15 ms, ESOF should trigger every 1ms.
 }
 
 // I'm getting a weird warning about missing braces here that I don't
@@ -565,6 +557,18 @@ static void dcd_fs_irqHandler(void) {
   if(int_status & USB_ISTR_SOF) {
     reg16_clear_bits(&USB->ISTR, USB_ISTR_SOF);
     dcd_event_bus_signal(0, DCD_EVENT_SOF, true);
+  }
+
+  if(int_status & USB_ISTR_ESOF) {
+    if(remoteWakeCountdown == 1u)
+    {
+      USB->CNTR &= (uint16_t)(~USB_CNTR_RESUME);
+    }
+    if(remoteWakeCountdown > 0u)
+    {
+      remoteWakeCountdown--;
+    }
+    reg16_clear_bits(&USB->ISTR, USB_ISTR_ESOF);
   }
 }
 
--- a/src/portable/st/stm32_fsdev/dcd_stm32_fsdev.c
+++ b/src/portable/st/stm32_fsdev/dcd_stm32_fsdev.c
@@ -168,11 +168,6 @@ static xfer_ctl_t xfer_status[MAX_EP_COUNT][2];
 
 static xfer_ctl_t* xfer_ctl_ptr(uint32_t epnum, uint32_t dir)
 {
-#ifndef NDEBUG
-  TU_ASSERT(epnum < MAX_EP_COUNT);
-  TU_ASSERT(dir < 2u);
-#endif
-
   return &xfer_status[epnum][dir];
 }
 
@@ -587,31 +582,9 @@ bool dcd_edpt_open (uint8_t rhport, tusb_desc_endpoint_t const * p_endpoint_desc
   uint8_t const dir   = tu_edpt_dir(p_endpoint_desc->bEndpointAddress);
   const uint16_t epMaxPktSize = p_endpoint_desc->wMaxPacketSize.size;
   // Isochronous not supported (yet), and some other driver assumptions.
-#ifndef NDEBUG
+
   TU_ASSERT(p_endpoint_desc->bmAttributes.xfer != TUSB_XFER_ISOCHRONOUS);
   TU_ASSERT(epnum < MAX_EP_COUNT);
-
-  switch(p_endpoint_desc->bmAttributes.xfer) {
-  case TUSB_XFER_CONTROL:
-    // USB 2.0 spec on FS packets, 5.5.3 (control)
-    TU_ASSERT((epMaxPktSize == 8) ||(epMaxPktSize == 16) || (epMaxPktSize == 32) || (epMaxPktSize == 64));
-    break;
-  case TUSB_XFER_ISOCHRONOUS: // FIXME: Not yet supported
-    TU_ASSERT(epMaxPktSize <= 1023);
-    break;
-  case TUSB_XFER_BULK:
-    // USB 2.0 spec on FS packets, 5.8.3 (bulk)
-    TU_ASSERT((epMaxPktSize == 8) ||(epMaxPktSize == 16) ||(epMaxPktSize == 32) ||(epMaxPktSize == 64));
-    break;
-  case TUSB_XFER_INTERRUPT:
-    // USB 2.0 spec on FS packets, 5.5.3 (interrupt); interestingly 0 is allowed.
-    TU_ASSERT(epMaxPktSize <= 64);
-    break;
-  default:
-    TU_ASSERT(false);
-    return false;
-  }
-#endif
 
   // Set type
   switch(p_endpoint_desc->bmAttributes.xfer) {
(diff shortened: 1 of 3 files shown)


pigrew:
Also note that using uint32_t for intermediate values is safe IF only AND & OR operations are used. Using right shifts and subtraction could cause issues because the upper bits would pollute the lowest 16-bits (which are used in the register).

pigrew:
(Added the callbacks for doing a remote wakeup)

src/portable/st/stm32_fsdev/dcd_stm32_fsdev.c
```diff
@@ -258,12 +258,8 @@ void dcd_int_enable (uint8_t rhport)
 {
   (void)rhport;
 #if defined(STM32F0)
-  NVIC_SetPriority(USB_IRQn, 0);
   NVIC_EnableIRQ(USB_IRQn);
```
pigrew: Should we be enabling/disabling only the USB IRQ? Or all of them? Do we have a rule that application code's interrupts cannot call USB-related functions? I think that the RTOS will use global interrupt disabling to avoid issue, so it only applies to the NO OS option.

src/portable/st/stm32_fsdev/dcd_stm32_fsdev.c
```diff
@@ -258,12 +258,8 @@ void dcd_int_enable (uint8_t rhport)
...
 #elif defined(STM32F3)
-  NVIC_SetPriority(USB_HP_CAN_TX_IRQn, 0);
-  NVIC_SetPriority(USB_LP_CAN_RX0_IRQn, 0);
```
pigrew: The priority shouldn't have changed, so no need to reset it here.

hathach: yes, the priority should be set by application/bsp to suite their need. E.g in my BLE work, the ble task take higher priority than usb. Some rtos such as freeRTOS reserved highest IRQ prio for kernel etc... So it is best to leave that to user.

src/portable/st/stm32_fsdev/dcd_stm32_fsdev.c
hathach: tinyUSB doesn't and couldn't impose that rule, those tinyUSB will only use dcd_int_enable/disable by itself. Usually for dcd_init() and  accessing tud task event queue (mutex). DCD should only enable/disable those only needed to get it worked.

src/portable/st/stm32_fsdev/dcd_stm32_fsdev.c
```diff
@@ -302,6 +309,17 @@ void dcd_set_config (uint8_t rhport, uint8_t config_num)
 void dcd_remote_wakeup(uint8_t rhport)
 {
   (void) rhport;
+  uint32_t start;
+
+  USB->CNTR |= (uint16_t)USB_CNTR_RESUME;
+  /* Wait 1 to 15 ms */
+  /* Busy loop is bad, but the osal_task_delay() isn't implemented for the "none" OSAL */
+  start = board_millis();
```
hathach: please remove, this API is not available at all in this level. Why should it need to be delayed for too long here ?

src/portable/st/stm32_fsdev/dcd_stm32_fsdev.c
```diff
@@ -566,47 +585,77 @@ bool dcd_edpt_open (uint8_t rhport, tusb_desc_endpoint_t const * p_endpoint_desc
   (void)rhport;
   uint8_t const epnum = tu_edpt_number(p_endpoint_desc->bEndpointAddress);
   uint8_t const dir   = tu_edpt_dir(p_endpoint_desc->bEndpointAddress);
-
+  const uint16_t epMaxPktSize = p_endpoint_desc->wMaxPacketSize.size;
   // Isochronous not supported (yet), and some other driver assumptions.
+#ifndef NDEBUG
```
hathach: please remove NDEBUG, I still prefer to check all parameters from application. We can use `CFG_TUSB_DEBUG` later on for this if user is desperate for flash. So far, there is no call for flash size just yet.

src/portable/st/stm32_fsdev/dcd_stm32_fsdev.c
```diff
@@ -721,16 +771,17 @@ void dcd_edpt_clear_stall (uint8_t rhport, uint8_t ep_addr)
...
 static bool dcd_write_packet_memory(uint16_t dst, const void *__restrict src, size_t wNBytes)
 {
-  uint32_t n =  ((uint32_t)((uint32_t)wNBytes + 1U)) >> 1U;
+  uint32_t n =  ((uint32_t)wNBytes + 1U) >> 1U;
   uint32_t i;
   uint16_t temp1, temp2;
   const uint8_t * srcVal;
 
-#ifdef DEBUG
+#ifndef NDEBUG
```
hathach: please remove `NDEBUG`

src/portable/st/stm32_fsdev/dcd_stm32_fsdev_pvt_st.h
```diff
@@ -83,42 +83,85 @@
 
 // For purposes of accessing the packet
 #if ((PMA_LENGTH) == 512u)
-#  define PMA_STRIDE (2u)
+const size_t PMA_STRIDE = 2u;
```
hathach: please either use macros or enum, I still prefer those two for pure constant expression. We could worry about the compiler complain later.

src/portable/st/stm32_fsdev/dcd_stm32_fsdev.c
```diff
@@ -121,6 +121,7 @@
 #undef USE_HAL_DRIVER
 
 #include "device/dcd.h"
+#include "bsp/board.h"
```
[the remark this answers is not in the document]
pigrew: How do you suggest implementing a delay in the driver?  In order to send a remote wakeup on this hardware, one has to assert a bit, wait 1 to 15 ms, then deassert it.

src/portable/st/stm32_fsdev/dcd_stm32_fsdev.c
pigrew: The remote wakeup is required to be 1 to 15 ms.

src/portable/st/stm32_fsdev/dcd_stm32_fsdev.c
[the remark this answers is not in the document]
hathach: please remove it, the `tinyusb/src` is self-contained, it won't need anything else except those register definitions. If we have no solution, we either drop the feature or re-organize the stack or API. But I think we can address this.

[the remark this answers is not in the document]
pigrew: There are not SOF during suspend mode, so we can't use USB timing. Perhaps a busy-loop looking at the systick counter values... assuming it is set for 1 kHz. I'll continue to think about it.

src/portable/st/stm32_fsdev/dcd_stm32_fsdev.c
hathach: `board_millis()` can easily be replaced with USB_FNR, each USB FS frame is 1 ms precisely. However, we still shouldn't wait here for 15 ms. That is a huge amount of time, lots of things can happen in between. We can use the little trick provided by usbd 
https://github.com/hathach/tinyusb/blob/master/src/device/usbd_pvt.h#L59

`usbd_defer_func()`, enable the bit here, then submit an function to the usbd event queue task

`usbd_defer_func( remote_wakeup_clear_bit,  USB_FNR number, false)`

`remote_wakeup_clear_bit()` will be schedule to be invoked  by tud_task(), when it is invoked, if the USB_FNR hasn't increased enough (e.g 15 frame) then we continue to queue this. The function will be called back-to-back until the we reach our desired frame. For a fast mcu, it will call like hundreds of time, however, it doesn't blocking other task.

hathach: ah, right, there is no SOF when suspended, almost forgot that. Let's me think of another way hmmm

hathach: RM say we can use ESOF interrupt to time this, let's use it for this purpose, enabled that interrupt when dcd_remote_wakeup() is called, then disable it when we clear the bit. We don't need `usbd_defer_func` function at all, still working without any blocking time.

image

src/portable/st/stm32_fsdev/dcd_stm32_fsdev.c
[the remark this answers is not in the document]
hathach: @pigrew systick is only available for ARM, and we are requiring application to enable that. There is tickless system like one I am working with on BLE repo, that is not really cross-platform approach. Check out my comment below, it may work out.

src/portable/st/stm32_fsdev/dcd_stm32_fsdev.c
hathach: Another way is adding uint8_t to dcd_remote_wakeup(), and application must pass '1' to imitate the remote and '0' to finish it ( after 1-15 ms). There maybe other mcu need this API to work with. Though like other API, I will postpone the API changes until I got to work on actual hardware requiring that.

pigrew:
I don't have time to fully test it right now; the wakeup test doesn't work without the GET_CONFIG/SET_CONFIG changes in the other patch, so it isn't as easy to validate the changes here.

I think I've made all requested changes.

pigrew:
Ok, I've tested this driver combined with usbd reconfiguration branch. The wakeup is working (passes the validator test).

src/portable/st/stm32_fsdev/dcd_stm32_fsdev.c
```diff
@@ -233,7 +235,7 @@ void dcd_init (uint8_t rhport)
   {
     pma[PMA_STRIDE*(DCD_STM32_BTABLE_BASE + i)] = 0u;
   }
-  USB->CNTR |= USB_CNTR_RESETM | USB_CNTR_SOFM | USB_CNTR_CTRM | USB_CNTR_SUSPM | USB_CNTR_WKUPM;
+  USB->CNTR |= USB_CNTR_RESETM | USB_CNTR_SOFM | USB_CNTR_ESOFM | USB_CNTR_CTRM | USB_CNTR_SUSPM | USB_CNTR_WKUPM;
```
hathach: I think `USB_CNTR_ESOFM` should only be enabled by `dcd_remote_wakeup()` and clear when unused. It will prevent MCU is interrupted when in suspended mode, and haven't got any need to wake up yet. This will help to application with low power design. Sorry for being picky, If you don't have time, I will merge and change it myself, not a problem at all. Your work is already fantastic

src/portable/st/stm32_fsdev/dcd_stm32_fsdev.c
```diff
@@ -147,31 +147,33 @@
...
   uint16_t total_len;
   uint16_t queued_len;
+  uint16_t max_packet_size;
 } xfer_ctl_t;
 
-static xfer_ctl_t  xfer_status[MAX_EP_COUNT][2];
-#define XFER_CTL_BASE(_epnum, _dir) &xfer_status[_epnum][_dir]
+static xfer_ctl_t xfer_status[MAX_EP_COUNT][2];
+
+static xfer_ctl_t* xfer_ctl_ptr(uint32_t epnum, uint32_t dir)
```
hathach: should be static inline, since it is called quite a lot.

PR #169, opened 2019-09-17
Implement HID desc request.

### LINKED ISSUE #168: HID class get HID descriptor request
The HID class fails to reply when a request for the HID descriptor is made:

hid_device.c:208
```
    if (p_request->bRequest == TUSB_REQ_GET_DESCRIPTOR && desc_type == HID_DESC_TYPE_REPORT)
    {
      uint8_t const * desc_report = tud_hid_descriptor_report_cb();
      tud_control_xfer(rhport, p_request, (void*) desc_report, p_hid->reprot_desc_len);
    }
    else
    {
      return false; // stall unsupported request
    }
```

Based on the HID spec, it must also support HID_DESC_TYPE_HID.

This causes the HID checks in the USB validator to fail.

pigrew:
Fixes #168

### DIFF
--- a/src/class/hid/hid_device.c
+++ b/src/class/hid/hid_device.c
@@ -46,12 +46,13 @@ typedef struct
   uint8_t boot_protocol; // Boot mouse or keyboard
   bool    boot_mode;     // default = false (Report)
   uint8_t idle_rate;     // up to application to handle idle rate
-  uint16_t reprot_desc_len;
+  uint16_t report_desc_len;
 
   CFG_TUSB_MEM_ALIGN uint8_t epin_buf[CFG_TUD_HID_BUFSIZE];
   CFG_TUSB_MEM_ALIGN uint8_t epout_buf[CFG_TUD_HID_BUFSIZE];
 
-}hidd_interface_t;
+  tusb_hid_descriptor_hid_t const * hid_descriptor;
+} hidd_interface_t;
 
 CFG_TUSB_MEM_SECTION static hidd_interface_t _hidd_itf[CFG_TUD_HID];
 
@@ -167,8 +168,8 @@ bool hidd_open(uint8_t rhport, tusb_desc_interface_t const * desc_itf, uint16_t 
 
   //------------- HID descriptor -------------//
   p_desc = tu_desc_next(p_desc);
-  tusb_hid_descriptor_hid_t const *desc_hid = (tusb_hid_descriptor_hid_t const *) p_desc;
-  TU_ASSERT(HID_DESC_TYPE_HID == desc_hid->bDescriptorType);
+  p_hid->hid_descriptor = (tusb_hid_descriptor_hid_t const *) p_desc;
+  TU_ASSERT(HID_DESC_TYPE_HID == p_hid->hid_descriptor->bDescriptorType);
 
   //------------- Endpoint Descriptor -------------//
   p_desc = tu_desc_next(p_desc);
@@ -178,7 +179,7 @@ bool hidd_open(uint8_t rhport, tusb_desc_interface_t const * desc_itf, uint16_t 
 
   p_hid->boot_mode = false; // default mode is REPORT
   p_hid->itf_num   = desc_itf->bInterfaceNumber;
-  memcpy(&p_hid->reprot_desc_len, &desc_hid->wReportLength, 2);
+  memcpy(&p_hid->report_desc_len, &(p_hid->hid_descriptor->wReportLength), 2);
 
   *p_len = sizeof(tusb_desc_interface_t) + sizeof(tusb_hid_descriptor_hid_t) + desc_itf->bNumEndpoints*sizeof(tusb_desc_endpoint_t);
 
@@ -205,6 +206,7 @@ bool hidd_control_request(uint8_t rhport, tusb_control_request_t const * p_reque
     if (p_request->bRequest == TUSB_REQ_GET_DESCRIPTOR && desc_type == HID_DESC_TYPE_HID)
     {
       // FIXME: Should check which is the active configuration, but no callback in usbd currently exists to do that
+      TU_VERIFY(p_hid->hid_descriptor != NULL);
       tusb_desc_configuration_t const* desc_cfg =
           (tusb_desc_configuration_t const*) tud_descriptor_configuration_cb(0);
       uint8_t const * p_desc   = ((uint8_t const*) desc_cfg) + sizeof(tusb_desc_configuration_t);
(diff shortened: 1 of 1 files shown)


src/class/hid/hid_device.c
```diff
@@ -202,7 +202,29 @@ bool hidd_control_request(uint8_t rhport, tusb_control_request_t const * p_reque
     uint8_t const desc_index = tu_u16_low (p_request->wValue);
     (void) desc_index;
 
-    if (p_request->bRequest == TUSB_REQ_GET_DESCRIPTOR && desc_type == HID_DESC_TYPE_REPORT)
+    if (p_request->bRequest == TUSB_REQ_GET_DESCRIPTOR && desc_type == HID_DESC_TYPE_HID)
+    {
+      // FIXME: Should check which is the active configuration, but no callback in usbd currently exists to do that
```
hathach: @pigrew we can have an pointer to store the HID Descriptor in `hidd_open()` at line 169 and used it to return here. That will solve this issue

src/class/hid/hid_device.c
```diff
@@ -202,10 +203,33 @@ bool hidd_control_request(uint8_t rhport, tusb_control_request_t const * p_reque
     uint8_t const desc_index = tu_u16_low (p_request->wValue);
     (void) desc_index;
 
-    if (p_request->bRequest == TUSB_REQ_GET_DESCRIPTOR && desc_type == HID_DESC_TYPE_REPORT)
+    if (p_request->bRequest == TUSB_REQ_GET_DESCRIPTOR && desc_type == HID_DESC_TYPE_HID)
+    {
+      // FIXME: Should check which is the active configuration, but no callback in usbd currently exists to do that
+      TU_VERIFY(p_hid->hid_descriptor != NULL);
+      tusb_desc_configuration_t const* desc_cfg =
```
hathach: @pigrew we can just return the p_hid->hid_descriptor now, a simple 
`tud_control_xfer(rhport, p_request, (void*) p_hid->hid_descriptor, p_hid->hid_descriptor->bLength)`

will work

pigrew: I remember typing that line... I must have gotten distracted.

pigrew:
I found another issue with spurious TU_ASSERT that I had fixed in my develop branch... adding that here, too.

src/class/hid/hid_device.c
```diff
@@ -192,6 +193,10 @@ bool hidd_open(uint8_t rhport, tusb_desc_interface_t const * desc_itf, uint16_t
 // return false to stall control endpoint (e.g unsupported request)
 bool hidd_control_request(uint8_t rhport, tusb_control_request_t const * p_request)
 {
+  if (p_request->bmRequestType_bit.recipient != TUSB_REQ_RCPT_INTERFACE)
```
hathach: the recipient can be device in some cases, please revert this

pigrew: Something still seems wrong. The next lines try to lookup the interface based on wIndex. If it's a device target, that won't be right. Of the top of my head, I'm not sure what wIndex means for device requests.

I'll revert this, and let you figure out the proper checks?

hathach:
> I found another issue with spurious TU_ASSERT that I had fixed in my develop branch... adding that here, too.

Did you push it, or maybe I miss it somehow.

pigrew:
> > I found another issue with spurious TU_ASSERT that I had fixed in my develop branch... adding that here, too.
> 
> Did you push it, or maybe I miss it somehow.

I mean that if statement on the interface RCPT that I just reverted.

I just also commented that it unconditionally calls get_interface_by_itfnum(wIndex), which seems wrong if the RCPT isn't the interface. 

It then calls TU_ASSERT(p_hid);. This will cause an assertion to be called whenever a EP request is made where wIndex != 0.

hathach:
> > > I found another issue with spurious TU_ASSERT that I had fixed in my develop branch... adding that here, too.
> > 
> > 
> > Did you push it, or maybe I miss it somehow.
> 
> I mean that if statement on the interface RCPT that I just reverted.
> 
> I just also commented that it unconditionally calls get_interface_by_itfnum(wIndex), which seems wrong if the RCPT isn't the interface.
> 
> It then calls TU_ASSERT(p_hid);. This will cause an assertion to be called whenever a EP request is made where wIndex != 0. (unless wIndex happens to match a valid value)

Hmm, You are totally right, though I did see Host trying to send this request to Device as recipient, especially if it is HID-only device. Hmm, OK, let's reject the Device rcpt, this will force host to re-try with Interface rcpt.

pigrew:
> > > > I found another issue with spurious TU_ASSERT that I had fixed in my develop branch... adding that here, too.
> > > 
> > > 
> > > Did you push it, or maybe I miss it somehow.
> > 
> > 
> > I mean that if statement on the interface RCPT that I just reverted.
> > I just also commented that it unconditionally calls get_interface_by_itfnum(wIndex), which seems wrong if the RCPT isn't the interface.
> > It then calls TU_ASSERT(p_hid);. This will cause an assertion to be called whenever a EP request is made where wIndex != 0. (unless wIndex happens to match a valid value)
> 
> Hmm, You are totally right, though I did see Host trying to send this request to Device as recipient, especially if it is HID-only device. Hmm, OK, let's reject the Device rcpt, this will force host to re-try with Interface rcpt.

Does that mean I should revert the reversion?

hathach:
> > > > > I found another issue with spurious TU_ASSERT that I had fixed in my develop branch... adding that here, too.
> > > > 
> > > > 
> > > > Did you push it, or maybe I miss it somehow.
> > > 
> > > 
> > > I mean that if statement on the interface RCPT that I just reverted.
> > > I just also commented that it unconditionally calls get_interface_by_itfnum(wIndex), which seems wrong if the RCPT isn't the interface.
> > > It then calls TU_ASSERT(p_hid);. This will cause an assertion to be called whenever a EP request is made where wIndex != 0. (unless wIndex happens to match a valid value)
> > 
> > 
> > Hmm, You are totally right, though I did see Host trying to send this request to Device as recipient, especially if it is HID-only device. Hmm, OK, let's reject the Device rcpt, this will force host to re-try with Interface rcpt.
> 
> Does that mean I should revert the reversion?

yeah, though, let's me check the spec a bit :)

pigrew:
I feel live I'm overworking Travis-CI by running the same codebase many times in a row.

hathach:
> I feel live I'm overworking Travis-CI by running the same codebase many times in a row.

Sorry for asking you revert the reverion. The spec say it should by Interface rcpt. Although i see device as rcpt previously. This will probably force the host to use interface rcpt instead. Thank you very much.
image

PR #172, opened 2019-09-20
Ported stm32 L0 and tested with L0358 discovery board

hathach:
- also added upcoming Feather stm32f405


src/portable/st/stm32_fsdev/dcd_stm32_fsdev.c
```diff
@@ -165,7 +165,7 @@ typedef struct
 
 static xfer_ctl_t xfer_status[MAX_EP_COUNT][2];
 
-static xfer_ctl_t* xfer_ctl_ptr(uint32_t epnum, uint32_t dir)
+static inline xfer_ctl_t* xfer_ctl_ptr(uint32_t epnum, uint32_t dir)
```
hathach: @pigrew inline furnction since this is quick and called often. If you have time, please let me know if you are OK with other changes of fsdev as well

pigrew: Probably would be optimized out by the compiler, anyway, but looks fine.

README.md
```diff
@@ -42,23 +53,12 @@ Support multiple device configurations by dynamically changing usb descriptors.
 
 ## OS Abtraction layer
 
-Currently the following OS are supported with tinyusb out of the box with a simple change of **CFG_TUSB_OS** macro.
+TinyUSB is completely thread-safe by pushing all ISR events into a central queue, then process it later in the non-ISR context. It also uses semphore/mutex to access shared resource such as CDC fifo. Therefore the stack needs to use some of OS's basic APIs. Following OSes are already supported out of the box.
```
pigrew: Capitalize  FIFO

src/portable/st/stm32_fsdev/dcd_stm32_fsdev.c
hathach: Yeah, attribute always inline will probably make it truely inline. I used to put those attr but removed since it may not be portable with different compiler.

PR #175, opened 2019-09-21
Clear feature ordering

### LINKED ISSUE #174: Effect of dcd_edpt_clear_stall
I'm hunting down a bug with my USBTMC & FSDEV driver, which I believe is due to the handling of stalls.

**Should the `dcd_edpt_clear_stall` function always set the EDPT to NAK, or return it to its previous state (NAK or ACK)? Normally it will change the toggle to DATA0, if the EDPT is in NAK or ACK state, does it still reset the toggle to DATA0?**

What should happen in USBTMC is:

1. Error on endpoint (bad command, timeout, etc)
2. Class requests an EDPT stall on its bulk out EP
3. host sends CLEAR FEATURE(EP HALT)
4. device must clear the stall, and also accept new transactions on the bulk OUT EP.

So, the final question: **Does the class need to initiate a new BULK IN transfer after an EP halt, or does the core remember the previous request?** If the former, it seems the clear feature needs to be sent to the class AFTER clear_stall is called by USBD; however the class is presently getting the event before the USBD core is clearing the stall.

I see a few solutions, but I'd like input on how it is designed to work. I tried reading the various drivers, but it's hard for me to efficiently understand them.

ADDED LATER: Thinking about it a bit more, the classes should have a way to abort pending transactions, and stalling a port seems like a reasonable way to request that that port be cleared. Thus, I think the drivers should restore the port to NAK once stall is cleared. Furthermore, there needs to be a way for the class drivers to open a transaction after the endpoint has cleared the stall condition. I don't know what the "proper" fix will be, but for my personal branch, I'll modify the handling in USBD core such that if the class driver return true, the standard handler will not be called.

pigrew:
This fixes #174 (and the related functions of the ST FSDEV driver).

I broke the commit into many different steps. If any seem controversial, I'm happy to revert/change particular commits.

It's been lightly tested with the USBTMC (with other patches) and hid_composite examples.

### DIFF
--- a/src/device/usbd.c
+++ b/src/device/usbd.c
@@ -804,7 +804,7 @@ bool usbd_open_edpt_pair(uint8_t rhport, uint8_t const* p_desc, uint8_t ep_count
   {
     tusb_desc_endpoint_t const * desc_ep = (tusb_desc_endpoint_t const *) p_desc;
 
-    TU_VERIFY(TUSB_DESC_ENDPOINT == desc_ep->bDescriptorType && xfer_type == desc_ep->bmAttributes.xfer);
+    TU_ASSERT(TUSB_DESC_ENDPOINT == desc_ep->bDescriptorType && xfer_type == desc_ep->bmAttributes.xfer);
     TU_ASSERT(dcd_edpt_open(rhport, desc_ep));
 
     if ( tu_edpt_dir(desc_ep->bEndpointAddress) == TUSB_DIR_IN )
--- a/src/device/usbd.c
+++ b/src/device/usbd.c
@@ -633,7 +633,7 @@ static void mark_interface_endpoint(uint8_t ep2drv[8][2], uint8_t const* p_desc,
       ep2drv[tu_edpt_number(ep_addr)][tu_edpt_dir(ep_addr)] = driver_id;
     }
 
-    len   += tu_desc_len(p_desc);
+    len   = (uint16_t)(len + tu_desc_len(p_desc));
     p_desc = tu_desc_next(p_desc);
   }
 }
--- a/src/device/usbd.c
+++ b/src/device/usbd.c
@@ -551,7 +551,7 @@ static bool process_control_request(uint8_t rhport, tusb_control_request_t const
 
       if ( usbd_class_drivers[drv_id].control_request )
       {
-        ret = ret | usbd_class_drivers[drv_id].control_request(rhport, p_request);
+        ret = ret || usbd_class_drivers[drv_id].control_request(rhport, p_request);
       }
       return ret;
     }
--- a/src/device/usbd.c
+++ b/src/device/usbd.c
@@ -549,9 +549,10 @@ static bool process_control_request(uint8_t rhport, tusb_control_request_t const
       //                           must not call tud_control_status(), and return value will have no effect
       // class driver is invoked last, so that EP already has EP stall cleared (in event of clear feature EP halt)
 
-      if ( usbd_class_drivers[drv_id].control_request )
+      if ( usbd_class_drivers[drv_id].control_request &&
+           usbd_class_drivers[drv_id].control_request(rhport, p_request))
       {
-        ret = ret || usbd_class_drivers[drv_id].control_request(rhport, p_request);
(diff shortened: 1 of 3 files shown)


src/device/usbd.c
```diff
@@ -633,7 +633,7 @@ static void mark_interface_endpoint(uint8_t ep2drv[8][2], uint8_t const* p_desc,
       ep2drv[tu_edpt_number(ep_addr)][tu_edpt_dir(ep_addr)] = driver_id;
     }
 
-    len   += tu_desc_len(p_desc);
+    len   = (uint16_t)(len + tu_desc_len(p_desc));
```
hathach: how about `len += (uint16_t) tu_desc_len(p_desc)

src/device/usbd.c
```diff
@@ -552,7 +542,17 @@ static bool process_control_request(uint8_t rhport, tusb_control_request_t const
...
+      // Some classes such as TMC needs to clear/re-init its buffer when receiving CLEAR_FEATURE request
+      // We will forward all request targeted endpoint to its class driver
+      // For class-type requests:  must (call tud_control_status(); return true) or (return false)
+      // For std-type requests:    non-std request codes are already discarded.
+      //                           must not call tud_control_status(), and return value will have no effect
+      // class driver is invoked last, so that EP already has EP stall cleared (in event of clear feature EP halt)
 
+      if ( usbd_class_drivers[drv_id].control_request )
+      {
+        ret = ret | usbd_class_drivers[drv_id].control_request(rhport, p_request);
```
hathach: it should be logic or `||` instead of bitwise `|`, or you can use `|=`

src/device/usbd.c
```diff
@@ -878,6 +879,7 @@ void usbd_edpt_clear_stall(uint8_t rhport, uint8_t ep_addr)
   uint8_t const dir   = tu_edpt_dir(ep_addr);
 
   dcd_edpt_clear_stall(rhport, ep_addr);
+  _usbd_dev.ep_busy_map[dir] = (uint8_t) tu_bit_clear(_usbd_dev.ep_busy_map[dir], epnum);
```
hathach: great catch on this issue :+1:

src/device/usbd_control.c
```diff
@@ -122,7 +122,7 @@ bool usbd_control_xfer_cb (uint8_t rhport, uint8_t ep_addr, xfer_result_t result
   }
 
   _control_state.total_transferred += xferred_bytes;
-  _control_state.buffer += xferred_bytes;
+  _control_state.buffer = ((uint8_t*)_control_state.buffer) + xferred_bytes;
```
hathach: good catch for this warning as well.  I need to add more warnings option to my makefile for sure :)

src/device/usbd.c
pigrew: My compiler complains on pedantic mode... The issue is that the (+) operation widens the result to integer (32-bit), but it's being stored to a 16-bit value, which means that it could be an overflow.

I'll revert this.

src/device/usbd.c
pigrew: I keep doing that... bad habit, though bool should be 0 or 1, so it would work... will change.

src/device/usbd.c
```diff
@@ -549,9 +549,10 @@ static bool process_control_request(uint8_t rhport, tusb_control_request_t const
       //                           must not call tud_control_status(), and return value will have no effect
       // class driver is invoked last, so that EP already has EP stall cleared (in event of clear feature EP halt)
 
-      if ( usbd_class_drivers[drv_id].control_request )
+      if ( usbd_class_drivers[drv_id].control_request &&
+           usbd_class_drivers[drv_id].control_request(rhport, p_request))
       {
-        ret = ret || usbd_class_drivers[drv_id].control_request(rhport, p_request);
+        ret = true;
```
hathach: indeed, this is what I worried about when going to bed yesterday, this is much better :)

src/device/usbd.c
hathach: compiler nowadays is even smarter than us :). I am fine with either way.

PR #176, opened 2019-09-22
Stall EP if class xfer cb returns false

pigrew:
Should an EP be stalled automatically if a class' xfer_cb return false?

Or should the class itself set the stall condition?

### DIFF
--- a/src/device/usbd.c
+++ b/src/device/usbd.c
@@ -305,7 +305,11 @@ void tud_task (void)
             uint8_t const drv_id = _usbd_dev.ep2drv[epnum][ep_dir];
             TU_ASSERT(drv_id < USBD_CLASS_DRIVER_COUNT,);
 
-            usbd_class_drivers[drv_id].xfer_cb(event.rhport, ep_addr, event.xfer_complete.result, event.xfer_complete.len);
+            if(!usbd_class_drivers[drv_id].xfer_cb(event.rhport, ep_addr, event.xfer_complete.result, event.xfer_complete.len))
+            {
+              // Stall EP on failure
+              usbd_edpt_stall(event.rhport,ep_addr);
+            }
           }
         }
       break;


src/device/usbd.c
```diff
@@ -305,7 +305,11 @@ void tud_task (void)
             uint8_t const drv_id = _usbd_dev.ep2drv[epnum][ep_dir];
             TU_ASSERT(drv_id < USBD_CLASS_DRIVER_COUNT,);
 
-            usbd_class_drivers[drv_id].xfer_cb(event.rhport, ep_addr, event.xfer_complete.result, event.xfer_complete.len);
+            if(!usbd_class_drivers[drv_id].xfer_cb(event.rhport, ep_addr, event.xfer_complete.result, event.xfer_complete.len))
```
hathach: great idea, usbd should stall endpoint if it is physical error. However, this is not **quite right**. The `xfer_cb` mean the transfer is already physically done in case xfer result is XFER_RESULT_SUCCESS (0). If there is error, it is only class protocol error, which the class should stall the endpoint itself. 

Though we can still stall the endpoint if the dcd report `event.xfer_complete.result` is `XFER_RESULT_FAILED` (1), which is an indicator of physical error. I would suggest to change the code to

```
if ( event.xfer_complete.result == XFER_RESULT_FAILED ) 
{
  usbd_edpt_stall(event.rhport,ep_addr);

  // TODO should we forward the failed physical xfer_cb to class driver as well ? 
}else
{
  usbd_class_drivers[drv_id].xfer_cb(event.rhport, ep_addr, event.xfer_complete.result, event.xfer_complete.len)
}


```

pigrew:
At least in some of the classes the result is already checked within xfer-cb, so I think it should continue to be unconditionally called.

Maybe we leave the code unchanged, but I'm not sure what it means and what to do if the xfer-cb returns false.

hathach:
I will close this for now, leaving the code as it is.

Issue #171, opened 2019-09-19
`dcd_event_timeout`?


cr1901:
**Is your feature request related to a problem? Please describe.**
One thing I don't handle in the Synopsys driver is the fact that IN xfers can time out. Currently, I don't have code to handle time out even though I have the timeout interrupt enabled, per the manual recommendations (yes, _this is a bug!_, but read on).

If the driver times out, it will eventually [die](https://github.com/hathach/tinyusb/blob/master/src/portable/st/synopsys/dcd_synopsys.c#L594) in `TU_BREAKPOINT` in the RX FIFO handler, which obscures the origin of the bug.

As a stopgap have added code to the end of the IRQ handler to handle timeouts by breakpointing immediately.

```
if(in_ep[n].DIEPINT & USB_OTG_DIEPINT_TOC) {
    TU_BREAKPOINT();
}
```

Of course, this doesn't work in release mode, but makes a timeout readily apparent while debugging.

**Describe the solution you'd like**
I think perhaps tinyusb should have a timeout event to propagate that an IN xfer failed, that the upper layers of the stack can use (or ignore, if isochronous is ever supported), such as redoing the IN xfer.

**Describe alternatives you've considered**
A timeout condition is something not worth recovering from, and worth a stack reset, so perhaps called `dcd_init` (within the ISR) to force the whole stack to reset when the bus reset event triggers.

**Additional context**
None.

pigrew:
The stm32 IP has a nebulous "ERROR" flag which can be due to a huge number of things, and there is no way to tell what caused an issue. (And the manual says to ignore the errors since the hardware will likely retry the transfer). In the USBTMC protocol, there is a set of error handling actions that the host can request in order to clear stalled/NAK EP, or abort transfers.. I'm not sure if other classes usually depend on the host for error handling. 

I don't see much that the stack can do, but if we start signalling errors, it probably should have a "generic" error condition when it doesn't know which EP is causing an issue.

I didn't notice anywhere that these were used in the ST Micro HAL code. Can you find examples where these flags are used?

hathach:
@cr1901 If it is transfer error, we could report back to the stack with 
`dcd_event_xfer_complete(ep_addr, xferred_bytes, result)`

where result = 0 is success, 1 is error, 2 for stalled. However, the stack doesn't handle such an error, primarily since I haven't encountered it while developing. Though the stack could re-queue/retry the transfer if it has error. 

[...]

Issue #139, opened 2019-09-11
hid_composite example doesn't work with STM32 fsdev + synopsis


hathach:
**Describe the bug**
The new hid_composite example doesn't work with any of stm32 chip. The example is hid only device with only 1 input endpoint. It works on other mcu just fine 

**Set up (please complete the following information):**
 - OS: Ubuntu 18.04
 - Board: stm32f072disco, stm32f407disco
 - Firmware Code: examples/device/hid_composite

cr1901:
FWIW, I was able to get `hid_composite` working on my stm32f407 discovery board by waiting about 30 seconds before pressing the user button.

pigrew:
I build it in STM32CubeIDE... seems there is an invalid descriptor (as reported by usbtreeview):

```        ---------------- Interface Descriptor -----------------
bLength                  : 0x09 (9 bytes)
bDescriptorType          : 0x04 (Interface Descriptor)
bInterfaceNumber         : 0x00
bAlternateSetting        : 0x00
bNumEndpoints            : 0x01 (1 Endpoint)
bInterfaceClass          : 0x03 (HID - Human Interface Device)
bInterfaceSubClass       : 0x00 (None)
bInterfaceProtocol       : 0x00 (None)
iInterface               : 0x00 (No String Descriptor)
Data (HexDump)           : 09 04 00 00 01 03 00 00 00                        .........

        ------------------- HID Descriptor --------------------
bLength                  : 0x09 (9 bytes)
bDescriptorType          : 0x21 (HID Descriptor)
bcdHID                   : 0x0111 (HID Version 1.11)
bCountryCode             : 0x00 (00 = not localized)
bNumDescriptors          : 0x01
Data (HexDump)           : 09 21 11 01 00 01 22 90 00                        .!...."..
Descriptor 1:
bDescriptorType          : 0x22 (Class=Report)
wDescriptorLength        : 0x0090 (144 bytes)
Error reading descriptor : ERROR_INVALID_PARAMETER
```

hathach:
> I built it in STM32CubeIDE... seems there is an invalid descriptor (as reported by usbtreeview):
> 
> ```
> bLength                  : 0x09 (9 bytes)
> bDescriptorType          : 0x04 (Interface Descriptor)
> bInterfaceNumber         : 0x00
> bAlternateSetting        : 0x00
> bNumEndpoints            : 0x01 (1 Endpoint)
[...]

PR #177, opened 2019-09-22
dcd_synopsys: Fix off-by-one error in FIFO allocation.

cr1901:
In a [comment](https://github.com/hathach/tinyusb/issues/139#issuecomment-531095383) in #139, I said the following about the `hid_composite` demo:

>In addition, if the keyboard and mouse fail to initialize, I am able to crash the demo eventually by repeatedly pressing the user button on my Nucleo 743. According to gdb, this is because one of the IN endpoints triggers a TimeOut Condition (USB_OTG_DIEPINT_TOC) interrupt, which I have enabled, but don't actually handle in code right now :P. I'll open a separate issue re: this soon.

While the crash disappeared when #139 was fixed, I suspected it was hiding another bug, so I started some [digging](https://github.com/hathach/tinyusb/issues/171#issuecomment-533410701).

>According to my USB captures, an HID IN xfer gets sent down the control pipe as the Data Phase to a "Get Device Descriptor" SETUP xfer. Not a typo. Meaning somehow I managed to get tinyusb into a state where it's sending data meant for EP 1 down EP 0. I'll look into it more after I sleep.

Turns out I had a nasty off-by-one error that was never triggered _except_ as a side effect of the `hid_composite` demo being broken. Before this PR, IN EP 0 and IN EP 1 were sharing the same TX FIFO. This means it was possible to the core to send data meant for EP 1 down EP 0 and vice versa.

I'll write up a more detailed analysis later, but this patch "fixes" the crash I described from when `hid_composite` was broken. Of course, the demo remains broken if using a commit before #151, like I am ;)...

EDIT: What makes this interesting to me is that the docs say:

>Bits 25:22 TXFNUM: TxFIFO number
These bits specify the FIFO number associated with this endpoint. Each active IN endpoint
must be programmed to a separate FIFO number.

However, you already _have_ to use different addresses for different endpoints to write data into the FIFOs; EP IN/OUT 0 is at peripheral offset 0x1000, IN/OUT 1 at 0x2000, etc. Without seeing the HDL (fat chance!), I don't know why Synopsys wants you to assign both a FIFO number and use a specific address to write data for each IN endpoint. But clearly it makes a difference...

### DIFF
--- a/src/portable/st/synopsys/dcd_synopsys.c
+++ b/src/portable/st/synopsys/dcd_synopsys.c
@@ -268,10 +268,10 @@ bool dcd_edpt_open (uint8_t rhport, tusb_desc_endpoint_t const * desc_edpt)
 
   uint8_t const epnum = tu_edpt_number(desc_edpt->bEndpointAddress);
   uint8_t const dir   = tu_edpt_dir(desc_edpt->bEndpointAddress);
-  
+
   TU_ASSERT(desc_edpt->wMaxPacketSize.size <= 64);
   TU_ASSERT(epnum < EP_MAX);
-  
+
   xfer_ctl_t * xfer = XFER_CTL_BASE(epnum, dir);
   xfer->max_size = desc_edpt->wMaxPacketSize.size;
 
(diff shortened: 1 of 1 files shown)


hathach:
> In a [comment](https://github.com/hathach/tinyusb/issues/139#issuecomment-531095383) in #139, I said the following about the `hid_composite` demo:
> 
> > In addition, if the keyboard and mouse fail to initialize, I am able to crash the demo eventually by repeatedly pressing the user button on my Nucleo 743. According to gdb, this is because one of the IN endpoints triggers a TimeOut Condition (USB_OTG_DIEPINT_TOC) interrupt, which I have enabled, but don't actually handle in code right now :P. I'll open a separate issue re: this soon.
> 
> While the crash disappeared when #139 was fixed, I suspected it was hiding another bug, so I started some [digging](https://github.com/hathach/tinyusb/issues/171#issuecomment-533410701).
> 
> > According to my USB captures, an HID IN xfer gets sent down the control pipe as the Data Phase to a "Get Device Descriptor" SETUP xfer. Not a typo. Meaning somehow I managed to get tinyusb into a state where it's sending data meant for EP 1 down EP 0. I'll look into it more after I sleep.
> 
> Turns out I had a nasty off-by-one error that was never triggered _except_ as a side effect of the `hid_composite` demo being broken. Before this PR, IN EP 0 and IN EP 1 were sharing the same TX FIFO. This means it was possible to the core to send data meant for EP 1 down EP 0 and vice versa.

Great works, this is difficult to trace down. I guess the reason this bug is well hidden due to EP1 IN is CDC notification in cdc_msc example, which we didn't use at all. The hid composite use EP1 IN for reporting, that increases the chance for conflict :) 

> I'll write up a more detailed analysis later, but this patch "fixes" the crash I described from when `hid_composite` was broken. Of course, the demo remains broken if using a commit before #151, like I am ;)...

It is already clear enough, it is OK if you don't have the time for the writeup :) 

> EDIT: What makes this interesting to me is that the docs say:
> 
> > Bits 25:22 TXFNUM: TxFIFO number
> > These bits specify the FIFO number associated with this endpoint. Each active IN endpoint
> > must be programmed to a separate FIFO number.
> 
> However, you already _have_ to use different addresses for different endpoints to write data into the FIFOs; EP IN/OUT 0 is at peripheral offset 0x1000, IN/OUT 1 at 0x2000, etc. Without seeing the HDL (fat chance!), I don't know why Synopsys wants you to assign both a FIFO number and use a specific address to write data for each IN endpoint. But clearly it makes a difference...

Yeah, I agreee that is too many arguments than needed, IP design guys seems always make thing harder for firmware guys :smile:

PR #181, opened 2019-09-26
ST FSDEV EP0 left in stall

pigrew:
The ST FSDEV driver has been sometimes not completing USB enumeration (perhaps 10% of the time). I have not looked at it with a USB analyzer, but my feeling is that it is leaving the EP in a still condition after the host requests the device qualifier (which does not exist), and it's never leaving stall.

This patch removes some hardware validation of control status packets being 0-length, and also resets EP0 (both directions) to NAK upon receiving a SETUP packet.

This patch does not seem correct to me, but seems like it's working better than the original code. I'm not sure where the proper place to reset EP0 from stall is, or maybe this is the correct patch for that. I'll have to study the USB spec more, and if the hardware will automatically NAK control in/out after receiving a SETUP (even though the EP was set to stall).

This patch also adds a check to only pass setup packets to the stack if they are exactly 8 bytes long.

The hardware manual suggests setting the opposite direction of the transfer to stall during all but the last packet. It's not immediately clear what to do to the code, so I'm submitting this as an interim patch to improve the driver while thinking about the correct fix (perhaps in days, perhaps in weeks).

If desired, I can create an issue in the tracker to remind us to improve the driver to use EP_KIND and better stall/unstall.

### DIFF
--- a/src/portable/st/stm32_fsdev/dcd_stm32_fsdev.c
+++ b/src/portable/st/stm32_fsdev/dcd_stm32_fsdev.c
@@ -252,7 +252,9 @@ void dcd_init (uint8_t rhport)
 void dcd_int_enable (uint8_t rhport)
 {
   (void)rhport;
-
+  // Member here forces write to RAM before allowing ISR to execute
+  __DSB();
+  __ISB();
 #if CFG_TUSB_MCU == OPT_MCU_STM32F0 || CFG_TUSB_MCU == OPT_MCU_STM32L0
   NVIC_EnableIRQ(USB_IRQn);
 #elif CFG_TUSB_MCU == OPT_MCU_STM32F3
@@ -276,10 +278,7 @@ void dcd_int_disable(uint8_t rhport)
 #else
   #error Unknown arch in USB driver
 #endif
-  // I'm not convinced that memory synchronization is completely necessary, but
-  // it isn't a bad idea.
-  __DSB();
-  __ISB();
+  // CMSIS has a membar after disabling interrupts
 }
 
 // Receive Set Address request, mcu port must also include status IN response
@@ -440,10 +439,10 @@ static uint16_t dcd_ep_ctr_handler(void)
           }
 
           /* Process Control Data OUT status Packet*/
-          if(EPindex == 0u && xfer->total_len == 0u)
+          /*if(EPindex == 0u && xfer->total_len == 0u)
           {
              pcd_clear_ep_kind(USB,0); // Good, so allow non-zero length packets now.
-          }
+          }*/
           dcd_event_xfer_complete(0, EPindex, xfer->total_len, XFER_RESULT_SUCCESS, true);
 
           pcd_set_ep_rx_cnt(USB, EPindex, CFG_TUD_ENDPOINT0_SIZE);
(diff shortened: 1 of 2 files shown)


hathach:
@pigrew could you tell me steps to reproduce the issue, I will try to capture the usb to see what is the exact cause. Sometime we make changes and the issue goes away, but we may only fix 90% of the cause, and it may still re-occur in the future.

pigrew:
It's hard to reproduce. It has been happening about 10% of the time while plugging in the USBTMC demo. WireShark shows that the device is stalling descriptor requests, so I'm fairly sure that it's due to a stall condition and/or the EP_KIND flag not being reset on EP0, after an error. We may be able to introduce the bug by sending an illegal control transaction (bad request number, etc), and then trying to make a proper request, to see if the stall goes away.

I won't be able to do that tonight, but maybe tomorrow I can get you a procedure to reproduce it.

Come to think of it, this would be a good test for other device drivers too.

hathach:
> It's hard to reproduce. It has been happening about 10% of the time while plugging in the USBTMC demo. WireShark shows that the device is stalling descriptor requests, so I'm fairly sure that it's due to a stall condition and/or the EP_KIND flag not being reset on EP0, after an error. We may be able to introduce the bug by sending an illegal control transaction (bad request number, etc), and then trying to make a proper request, to see if the stall goes away.
> 
> I won't be able to do that tonight, but maybe tomorrow I can get you a procedure to reproduce it.
> 
> Come to think of it, this would be a good test for other device drivers too.

Anytime you feel convenient, though I think we don't need to generate an protocol error to test with the stack. USB stack is very established, all the hosts should follows that. It is a bit over engineered and may force us to write never-run code :) . I am happy with trouble-shooting when the bug arises :)

pigrew:
Sorry, I've been consumed with babysitting antenna simulations, so have not had a chance to reply here.

I'm attaching a WireShark log, showing that the device is stalling every transaction. I don't have a way to consistently tigger the bug, just plug it in and unplug it a bunch of times with the USBTMC demo. I don't know if it also is triggered by other classes... have not tried them as much.
[stall_log.pcapng.gz](https://github.com/hathach/tinyusb/files/3663227/stall_log.pcapng.gz)

src/portable/st/stm32_fsdev/dcd_stm32_fsdev.c
```diff
@@ -419,10 +418,15 @@ static uint16_t dcd_ep_ctr_handler(void)
           uint8_t userMemBuf[8];
           /* Get SETUP Packet*/
           count = pcd_get_ep_rx_cnt(USB, EPindex);
-          //TU_ASSERT_ERR(count == 8);
-          dcd_read_packet_memory(userMemBuf, *pcd_ep_rx_address_ptr(USB,EPindex), 8);
+          if(count == 8) // Setup packet should always be 8 bits. If not, ignore it, and try again.
```
hathach: I guess you mean 8 bytes not bits. Le'ts me know when you think PR is ready for merging.

pigrew:
@hathach , I'm happy with the patch now. I also added a testcase to the USBTMC test script in order to verify that EP0 can unstall.

Before the patch, the test fails. After the patch, it passes.

src/portable/st/stm32_fsdev/dcd_stm32_fsdev.c
pigrew: Yep, 64 bits. Fixed.

PR #192, opened 2019-10-01
2nd attempt to fix #161

### LINKED ISSUE #161: ep_busy_map: concurrency issues
EDIT: I've updated this because I misread the code initially (though the variable was accessed from within interrupts). I believe this bug only applies if using a preemptive RTOS.

The `ep_busy_map`:
* Is not declared volatile
* Is set by the `usbd_edpt_xfer()` function, which is called from many tasks (such as the USBD core task, USBD control handler, and application tasks).
* Is reset in USBD core task upon receiving the DCD_EVENT_XFER_COMPLETE event.
* Can have its value read using the usbd_edpt_busy() function.

I've noticed two issue with this:

* Because tud_task() never returns, it may not be guaranteed to write the value of  `ep_busy_map` back to main memory (perhaps given sufficient compiler optimization). Setting it to volatile would fix this (but is not a sufficient solution for MCUs with out-of-order execution).
* `usbd_edpt_xfer()` may be called from multiple tasks "simultaneously" (such as both an application task AND the USBD task (handling a control request)). Bit setting/clearing is NOT done using a single-instruction cycle, so:

| Task1                   | Task2                   |
|-------------------------|-------------------------|
| Enter `usbd_edpt_xfer`  |                         |
| `LOAD R1,ep_busy_map`   |                         |
|                         | Enter `usbd_edpt_xfer`  |
|                         | `LOAD R2, ep_busy_map`  |
|                         | `R2 &#124;= EP1_bit          |
|                         | `STORE ep_busy_map, R2` |
| `R1&#124;= EP3_bit          |                         |
| `STORE ep_busy_map, R1` |                         |

The same issues may also apply to `ep_busy_map`, but I've not analyzed that yet.

hathach:
@pigrew I think this should be sufficient enough, each endpoint has its own status bit. This will solve the issue when `1 bit should be set but not`.  The conflict where `it should be 0 but 1` is not critical since the **edpt_busy()** will be called again later on (polling).

### DIFF
--- a/src/class/hid/hid_device.c
+++ b/src/class/hid/hid_device.c
@@ -74,7 +74,7 @@ bool tud_hid_ready(void)
 {
   uint8_t itf = 0;
   uint8_t const ep_in = _hidd_itf[itf].ep_in;
-  return tud_ready() && (ep_in != 0) && !usbd_edpt_busy(TUD_OPT_RHPORT, ep_in);
+  return tud_ready() && (ep_in != 0) && usbd_edpt_ready(TUD_OPT_RHPORT, ep_in);
 }
 
 bool tud_hid_report(uint8_t report_id, void const* report, uint8_t len)
--- a/src/device/usbd.c
+++ b/src/device/usbd.c
@@ -52,11 +52,16 @@ typedef struct {
     uint8_t self_powered          : 1; // configuration descriptor's attribute
   };
 
-  volatile uint8_t ep_busy_map[2];  // bit mask for busy endpoint
-  volatile uint8_t ep_stall_map[2]; // bit map for stalled endpoint
-
   uint8_t itf2drv[16];     // map interface number to driver (0xff is invalid)
   uint8_t ep2drv[8][2];    // map endpoint to driver ( 0xff is invalid )
+
+  struct TU_ATTR_PACKED
+  {
+    volatile bool busy    : 1;
+    volatile bool stalled : 1;
+
+    // TODO merge ep2drv here, 4-bit should be sufficient
+  }ep_status[8][2];
 }usbd_device_t;
 
 static usbd_device_t _usbd_dev;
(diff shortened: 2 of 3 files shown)


pigrew:
@hathach I've not had enough time to think about the patch. I'll analyze it once I get a chance.... I'm hoping to close out the STM32 F0 stall patch tonight.

The new patch uses read-modify-write, so that makes it more vulnerable. Perhaps using a enum-type state variable (0=closed, 1=stalled, 2=nak, 3=busy) would be easier since one could directly write `state=stall` instead of having the bit fields where the compiler will load the current value, and then write back the new value.

I also need to read up on what volatile bit of a bit field means.... does having a single bit volatile make the entire bit field volatile? Or not? Another note is that C99 does not support anonymous structs.  They were added in C11.... I've been going under the understanding that we're trying to target C99?

hathach:
> @hathach I've not had enough time to think about the patch. I'll analyze it once I get a chance.... I'm hoping to close out the STM32 F0 stall patch tonight.
> 
> The new patch uses read-modify-write, so that makes it more vulnerable. Perhaps using a enum-type state variable (0=closed, 1=stalled, 2=nak, 3=busy) would be easier since one could directly write `state=stall` instead of having the bit fields where the compiler will load the current value, and then write back the new value.

Yeah, using enum state is also good as well, but I don't think we could entirely eliminate the race condition without a real `mutex`. E.g edpt_xfer() is about to set state to BUSY, but got stalled by Host. Then state is set to STALLED (since usbd task preempt user task), then when resuming back to edpt_xfer() it will set the state back to BUSY (dropping stalled).

Although there is more chance of vulnerable as you said, but the case of conflict is very rare, **edpt_stall()** is not used very often (probably not at all in most cases), when it does it is a managed data flow. I think It is relatively safe enough for now. Maybe we can revisit this later when it becomes a bigger issue. 

> I also need to read up on what volatile bit of a bit field means.... does having a single bit volatile make the entire bit field volatile? Or not? 

It will cause the whole storage unit as volatile in this case, the whole byte where the bit is part of. 

> Another note is that C99 does not support anonymous structs. They were added in C11.... I've been going under the understanding that we're trying to target C99?

You are right, but all compilers seem to be happy with this :D. I use it quite a lot. Rewriting to comfort the spec is easy enough, but I am lazy now :D . Maybe we could start to look at anonymous struct/union when there is a compiler that failed to support this :) .

Issue #128, opened 2019-09-10
Add ability for application layer to define IRQ handler


pigrew:
Could we add a dcd method for invoking the USB interrupt handler? Or would suggest how to proceed with interrupts on the ST port?

On some STM32F3, the interrupt is shared by CAN and USB, so the USB stack can't directly define the interrupt handler since both the CAN and USB handlers must be called. My board_test() module is currently failing the CI build because it isn't including the tusb drivers. Perhaps you could look at the code and suggest a fix?

It'd probably involve changing all the existing drivers, but I think the bsp should define the interrupt handler and call the dcd handler as needed. Prehaps we should add "void tud_handle_irq(uint8_t port)" which would call the appropriate "void dcd_handle_interrupt(uint8_t port)"? The port is for devices whose driver would have multiple instances (like one for FS and another for HS).

Another note is that TI-RTOS will intercept all of the interrupts, and then re-dispatch them (drivers are not allowed to directly handle the interrupts). We should think about how that would be handled, too (especially as I see someone wants to work on a MSP430 port, which would be good to be compatible with TI-RTOS).

cr1901:
Adding my own opinions before I forget...

>Another note is that TI-RTOS will intercept all of the interrupts, and then re-dispatch them (drivers are not allowed to directly handle the interrupts).

This is ultimately how Rust [real time solutions](https://github.com/japaric/cortex-m-rtfm) handle interrupts as well- setup and teardown code is required before calling the actual interrupt code because the compiler requires it.

On the other hand, it seems that FreeRTOS can happily deal with interrupts without requiring boilerplate code, so this change will make it a bit less convenient for FreeRTOS users.

rajivr:
In TockOS (non-realtime Rust OS), we do something [similar](https://github.com/tock/tock/blob/b7e7eeecd75b5479b002bf62a22b454f1ad7b02c/arch/cortex-m3/src/lib.rs#L60-L115).

I am still coming up to speed with Tinyusb code base. 

My current understanding is that interrupt enabling (`dcd_int_enable`) and disabling (`dcd_int_disable`) at NVIC is only used prevent interrupt handler from manipulating event queue, while thread mode de-queuing is happening in `tud_task`.

[...]

PR #193, opened 2019-10-02
ported stm32f1, tested with f103 blue pill

hathach:


pigrew:
Nice, it's great to have the Blue Pill ported (though one usually has to substitute the D+ pull-up resistor).

The IRQ handlers need some thought (eventually). I think think we can just enable/disable one of the IRQs which would handle all three (or something like that). Or we were once talking about having the BSP or OS implement the handler, and call into the driver....

hathach:
> Nice, it's great to have the Blue Pill ported (though one usually has to substitute the D+ pull-up resistor).
> 
> The IRQ handlers need some thought (eventually). I think think we can just enable/disable one of the IRQs which would handle all three (or something like that). Or we were once talking about having the BSP or OS implement the handler, and call into the driver....

Yeah, lots of #ifdef already, though. We could wait until we tested with all the STM32 and decide later on. We also should wait until #128 is resolved as well :)

PR #194, opened 2019-10-03
[WIP] MSP430 Support

### LINKED ISSUE #184: TinyUSB devices will not enumerate with 8-byte max packet size.
**Describe the bug**
While working on the MSP430 driver, I noticed that tinyusb would get into a state where the host would start the Status Phase of a Control Read _before_ tinyusb was done sending the packet of the data phase. This only happens during a Get Device Descriptor request immediately after a device is plugged in (or the firmware starts).

Page 39 (PDF page 67) of the USB 2.0 spec states:

>In order to determine the maximum packet size for the Default Control Pipe, the USB System Software reads the device descriptor. The host will read the first eight bytes of the device descriptor. The device always responds with at least these initial bytes in a single packet. After the host reads the initial part of the device descriptor, it is guaranteed to have read this default pipe’s wMaxPacketSize field (byte 7 of the device descriptor). It will then allow the correct size for all subsequent transactions. For all other control endpoints, the maximum data payload size is known after configuration so that the USB System Software can ensure that no data payload will be sent to the endpoint that is larger than the supported size.

I interpret this as "the first packet ever sent to the host after a device is connected connection is _negotiated_ by the host and device to be `bMaxPacketSize0`". What tinyusb does instead is schedule to send the whole device descriptor, which is always 18 bytes.

This works fine for 64 and 32-byte packet sizes, because the whole descriptor fits into a single packet. For 8 and 16-byte packet sizes, the descriptor doesn't fit into 1 packet. After the first packet is sent, the OS thinks the transaction is done and goes to the Status Phase of the control xfer- in other words, starts sending OUT ZLPs. Tinyusb doesn't know this, so it keeps trying to send the rest of the descriptor- and failing, because the hardware is waiting for an IN token the host will never send!

I made a [branch](https://github.com/cr1901/tinyusb/tree/ctl-8b) to try and accommodate 8/16 byte packets on the control pipe on STM32, but my patch does not seem to be correct. @hathach, perhaps you have an idea what's wrong because I am personally stumped right now.

**Set up (please complete the following information):**
 - OS: Windows 7
 - Board: stm32f407disco
 - Firmware Code: examples/device/cdc_msc

**To Reproduce**
1. Check out my [branch](https://github.com/cr1901/tinyusb/tree/ctl-8b) of tinyusb which modifies the cdc_msc demo and Synopsys driver to use only 8-byte packets on EP 0.
2. Compile the cdc_msc demo and upload it to the device.
3. The firmware should breakpoint in an xfer that is marked as complete after only sending 32 out of 64 bytes (`xfer->queued_size` will say 24; this is due to how `transmit_packet` works, but 32 bytes were queued).

**Expected behavior**
Upon device insertion/firmware init, tinyusb should expect the OS to terminate a Get Device Descriptor xfer after sending only one packet, _regardless_ of packet size. This only applies to the first Get Device Descriptor request.

Also, my attempted fix for the behavior described above should not break the Synopsys driver :P.

**Screenshots**
TODO

**Additional context**
I know that Windows will always go to the Status Phase of the xfer after a single packet is sent after device connection. I am unsure what Linux or other OSes do.

cr1901:
Do not merge this yet. 

I got `cdc_msc` working properly yesterday on the msp-exp430f5529 launchpad. However, there are a few issues I need help fixing in tinyusb before I feel comfortable having this merged:

- [x] Fix #184. I have a [stopgap solution](https://github.com/hathach/tinyusb/issues/184#issuecomment-536221082) I've been using for the time being. However, my patch breaks when combined with #149, as I described below.

When testing the `hid_composite` demo, I was getting a number of STALLs that appear to have been hardware generated (traces/pics upon request). I've since discovered that the msp430 USB core will STALL if there's a mismatch between the `DIR` bit in the `USBCTL` and the packet actually received. Thanks to msp430's 8 byte `bMaxPacketSize` on the Control Pipe, I was able to find not one, but _two_ separate bugs!

- [ ] In [`tud_control_xfer`](https://github.com/hathach/tinyusb/blob/master/src/device/usbd_control.c#L90), the Data Phase xfer size needs to be set based on the minimum of `len` and `wLength` in the Setup packet. If not, it is possible to schedule a data xfer with `wLength` of 0, which should go straight to the Status Phase. #149 will fix this, but locally I'm using commit b0b737b.

- [ ] Merge #149 with a fix for [EP 0](https://github.com/hathach/tinyusb/pull/149#issuecomment-537784737). There's no way around this in the `hid_composite` demo. One of the descriptors is 144 bytes, and 144 is unfortunately an exact multiple of 8. #149 also [breaks](https://github.com/hathach/tinyusb/issues/184#issuecomment-537774101) my proposed solution for #184.

There's still things I can do while I wait for fixes:

- [ ] Implement suspend/detach logic. I can't test detach logic, as there is only a single cable on the msp-exp430f5529 launchpad (a [USB hub](http://www.ti.com/product/TUSB2046B) separates the debug/programmer logic from application logic, but I should implement it anyway.

- [x] Implement UART support. The debugger chip on this board has a CDC interface completely separate from tinyusb for, well, debugging! You feed data in/out to the debugger chip using the UART peripheral. I know the UART is mainly for host debugging, but this could be helpful for device debugging in the future.

- [x] Make sure travis builds work. Ubuntu's msp430 gcc version is too old, so I'll likely have to untar a copy of the [official gcc compiler](http://www.ti.com/tool/MSP430-GCC-OPENSOURCE) and cache it: IME, TI doesn't like repeated downloads of the compiler too much, as would happen on CI builds. Don't ask how I know :P.

- [x] Fix race condition in `board_millis`. `system_ticks` is a 32-bit var, msp430 is 16-bit. Need to disable timer interrupt while updating `system_ticks` so the update appears atomic from timer interrupt's POV.

That being said, the `cdc_msc` demo works wonderfully, although for whatever reason on Windows, writes to the text file are honored despite me defining `READONLY`. Ahhh well...


hathach:
Superb !!!! This is exciting, thank you very much for another great port. I will get an MSP430 board to test with. Maybe I could help to address some of your outlined issues :)

cr1901:
>I will get an MSP430 board to test with.

http://www.ti.com/tool/MSP-EXP430F5529LP

This is the board I used.

>Maybe I could help to address some of your outlined issues :)

I would really appreciate that :D!

pigrew:
> Implement suspend/detach logic. I can't test detach logic, as there is only a single cable on the msp-exp430f5529 launchpad (a USB hub separates the debug/programmer logic from application logic, but I should implement it anyway.

I think you can pull the "5V" jumper off of the board to simulate a bus detach.

cr1901:
>I think you can pull the "5V" jumper off of the board to simulate a bus detach.

Just checked. You are correct, with the caveat that about half the time I knock the debugger MCU off the bus too :'D.

cr1901:
@hathach Travis support for MSP430 is implemented. I'm not sure why, but Travis is not checking out the correct commit; 5ee1070 should've fixed this. Could you restart the build/figure out what's going on?

hathach:
> @hathach Travis support for MSP430 is implemented. I'm not sure why, but Travis is not checking out the correct commit; [e4a88bc](https://github.com/hathach/tinyusb/commit/e4a88bc826a389b452fa0795292f39456585ed05) should've fixed the failing build. Could you restart the build/figure out what's going on?

I just trigger the re-build, travis is happy now. My board also arrives, I am getting started with msp430 for now. Haven't used the mcu before :)

hathach:
@cr1901 phew, I got into lots of trouble with flashing msp430 and the missing libmsp430.so on Linux. But finally fixed it thanks to the precompiled library from energia. Have to use use MSP430Flasher from TI  http://www.ti.com/tool/MSP430-FLASHER on windows to recover from an failed FET upgrade. Phew :)

```
MSPDebug version 0.25 - debugging tool for MSP430 MCUs
Copyright (C) 2009-2017 Daniel Beer <dlbeer@gmail.com>
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
Chip info database from MSP430.dll v3.13.0.601 Copyright (C) 2013 TI, Inc.

tilib_api: can't find libmsp430.so: libmsp430.so: cannot open shared object file: No such file or directory
```

hw/bsp/msp_exp430f5529lp/msp_exp430f5529lp.c
```diff
@@ -0,0 +1,205 @@
...
+
+#include "msp430.h"
+
+#define LED_PORT              P1OUT
+#define LED_PIN               BIT0
+#define LED_STATE_ON          1
+
+#define BUTTON_PORT           P1IN
+#define BUTTON_PIN            BIT1
+#define BUTTON_STATE_ACTIVE   1
```
hathach: active button state is 0

hw/bsp/msp_exp430f5529lp/msp_exp430f5529lp.c
```diff
@@ -0,0 +1,205 @@
...
+  }
+  else
+  {
+    LED_PORT &= ~LED_PIN;
+  }
+}
+
+uint32_t board_button_read(void)
+{
+  return (P1IN & BIT1);
```
hathach: this should be `return ((P1IN & BIT1) >> 1) == BUTTON_STATE_ACTIVE;`
To test this you could run the board_test example, the example does nothing but test the led and button. LED blink normal when button is not active, and blink faster when pressing the button.

src/tusb_option.h
```diff
@@ -67,6 +67,9 @@
 #define OPT_MCU_STM32L1       308 ///< ST STM32L1
 #define OPT_MCU_STM32L4       309 ///< ST STM32L4
 
+#define OPT_MCU_MSP430x5xx    400 ///< TI MSP430x5xx
```
hathach: the conflict is caused by the newly merged sony port which also use 400. Please increase this to 500 to resolve the conflict

cr1901:
> @cr1901 phew, I got into lots of trouble with flashing msp430 and the missing libmsp430.so on Linux. But finally fixed it thanks to the precompiled library from energia. Have to also use use MSP430Flasher from TI http://www.ti.com/tool/MSP430-FLASHER on windows to recover from an failed FET upgrade. Phew :)
> 
> ```
> MSPDebug version 0.25 - debugging tool for MSP430 MCUs
> Copyright (C) 2009-2017 Daniel Beer <dlbeer@gmail.com>
> This is free software; see the source for copying conditions.  There is NO
> warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
> Chip info database from MSP430.dll v3.13.0.601 Copyright (C) 2013 TI, Inc.
> 
> tilib_api: can't find libmsp430.so: libmsp430.so: cannot open shared object file: No such file or directory
> ```

Oh yea, I should've mentioned that lol. Yea, TI provides a library for talking to the FET and EEM and this is required to program the launchpad.

Another thing worth noting... `mspdebug` _does_ support watchpoints, but for whatever reason you need to use the `gdb` monitor command to set them properly: `monitor setwatch [address]`. Simply doing `watch [address]` from a `gdb` prompt will _not_ work.

hathach:
> > @cr1901 phew, I got into lots of trouble with flashing msp430 and the missing libmsp430.so on Linux. But finally fixed it thanks to the precompiled library from energia. Have to also use use MSP430Flasher from TI http://www.ti.com/tool/MSP430-FLASHER on windows to recover from an failed FET upgrade. Phew :)
> > ```
> > MSPDebug version 0.25 - debugging tool for MSP430 MCUs
> > Copyright (C) 2009-2017 Daniel Beer <dlbeer@gmail.com>
> > This is free software; see the source for copying conditions.  There is NO
> > warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
> > Chip info database from MSP430.dll v3.13.0.601 Copyright (C) 2013 TI, Inc.
> > 
> > tilib_api: can't find libmsp430.so: libmsp430.so: cannot open shared object file: No such file or directory
> > ```
> 
> Oh yea, I should've mentioned that lol. Yea, TI provides a [library](http://www.ti.com/tool/MSPDS) for talking to the FET and EEM on some msp430 devices and this is required to program the USB Launchpad. I'm not sure if the versions of `libmsp430` on Github are the mirrored TI-provided library or a reimplementation.
> 
> Another thing worth noting... `mspdebug` _does_ support watchpoints, but for whatever reason you need to use the `gdb` monitor command to set them properly: `monitor setwatch [address]`. Simply doing `watch [address]` from a `gdb` prompt will _not_ work.
> 
> I will incorporate your requested changes as I have time.

Thanks for the debug tip, it is helpful since I am new to msp430 :). It is not urgent, make the changes whenever you got time .

cr1901:
@hathach When you get the chance, could you read this post by me on TI's website for context? https://e2e.ti.com/support/microcontrollers/msp430/f/166/p/845259/3129735#3129735

I recently learned about and semi-fixed a bug with dealing with SETUP packets on MSP430. Receiving a SETUP packet on MSP430 _requires_ both:

1. EP0 NAK bits must be cleared before receiving. 
2. _Immediately_ after a SETUP packet is received, the EP0 NAK bits must be _manually_, only unsetting/resetting them to handle Data and Status Phase. 

If both the above steps aren't followed, you get a nasty race condition. Well, I can confirm not doing #2 causes a race condition where SETUP packets will spuriously STALL. I have not actually seen a case where not doing #1 causes any problems.

While my fix works to my satisfaction, I should still follow the manual, thus my fix is incomplete. What I _should_ be doing after a Status Phase is complete is "scheduling" another SETUP packet to be accepted by manually clearing the EP0 NAK bits (I already do this during `bus_reset`). TinyUSB at present doesn't have a "hook" to prepare to receive a new SETUP packet (by clearing the EP0 NAK bits) after the Status Phase of the previous control xfer has completed. Since msp430 is a special case, I think this hook should be optional, but how would you go about adding a hook?

hathach:
> @hathach When you get the chance, could you read this post by me on TI's website for context? https://e2e.ti.com/support/microcontrollers/msp430/f/166/p/845259/3129735#3129735
> 
> I recently learned about and semi-fixed a bug with dealing with SETUP packets on MSP430. Receiving a SETUP packet on MSP430 _requires_ both:
> 
> 1. EP0 NAK bits must be cleared before receiving.
> 2. _Immediately_ after a SETUP packet is received, the EP0 NAK bits must be _manually_, only unsetting/resetting them to handle Data and Status Phase.
> 
> If both the above steps aren't followed, you get a nasty race condition. Well, I can confirm not doing 2. causes a race condition where SETUP packets will spuriously STALL. I have not actually seen a case where not doing 1. causes any problems. While my fix works to my satisfaction, I should still follow the manual, thus my fix is incomplete.
> 
> What I _should_ be doing after a Status Phase is complete is "scheduling" another SETUP packet to be accepted by manually clearing the EP0 NAK bits (I already do this during `bus_reset`). TinyUSB at present doesn't have a "hook" to prepare to receive a new SETUP packet (by clearing the EP0 NAK bits) after the Status Phase of the previous control xfer has completed. Since msp430 is a special case, I think this hook should be optional, but how would you go about adding a hook?

Thanks @cr1901 ,  these are new to me, I am reading USB module from msp430 specs, doing a bit of testing. Though I don't see any STALL response, msp just doesn't response with the status (zero-length packet) possibly you fixed it in the recent changes. I will work more on this to get a better picture with msp then decide if we need an additional `dcd_setup_prepare()` API 

image

cr1901:
@hathach 
>Though I don't see any STALL response, msp just doesn't response with the status (zero-length packet) possibly you fixed it in the recent changes.

This is an unrelated bug (#184) to the one I was discussing in the TI thread. If you keep capturing packets for a few more seconds, you should see a successful enumeration. Are you using commit d0146be?

What I was seeing before that commit was spurious hardware-induced STALLs on EP0 (in addition to the the bug I opened in #184). _As of d0146be, these hardware-induced STALLs went away (but not #184)._ However the fix I committed is technically incomplete; after the Status Phase of a control xfer ends, I really should be clearing the NAK bits on EP0; right now, tinyusb doesn't provide a way for me to do that.

hathach:
> @hathach
> 
> > Though I don't see any STALL response, msp just doesn't response with the status (zero-length packet) possibly you fixed it in the recent changes.
> 
> This is an unrelated bug (#184) to the one I was discussing in the TI thread. If you keep capturing packets for a few more seconds, you should see a successful enumeration. Are you using commit [d0146be](https://github.com/hathach/tinyusb/commit/d0146be00b4dd41826f29e36bb10e0ecafce5e47)?

Yeah, I tried with this latest commit of yours.

> What I was seeing before that commit was spurious hardware-induced STALLs on EP0 (in addition to the the bug I opened in #184). _As of [d0146be](https://github.com/hathach/tinyusb/commit/d0146be00b4dd41826f29e36bb10e0ecafce5e47), these hardware-induced STALLs went away (but not #184)._ However the fix I committed is technically incomplete; after the Status Phase of a control xfer ends, I really should be clearing the NAK bits on EP0. Right now, tinyusb doesn't provide a way for me to do that, because there is no "Status Phase Complete callback" that can call into the DCD layer.

I start to get a clearer picture with msp430, though I will need to spend a bit more hands-on + testing time with it. I am currently in a middle of other work with nRF52. Will check this out as soon as I could.

hathach:
back to testing this and realize TI just release msp430 gcc 8.3 with installer for linux + debug stack + usb drivers, which is more convenient since last time I tried :)

cr1901:
@hathach Please take your time trying to fix issues :D! I'm not in a particular hurry to merge since I've listed everything in my initial message that needs to be fixed.

If you can't get to my issues right now, I'll eventually take another look later.

hathach:
I find out that if I enable the debug log, which is just added to master to help out with this port. msp430 will get passed the enumeration process, show up as both msc and cdc. So it is probably something out of order or need a bit of timing wait for a event. I am reading USB Module from TI specs.

I have a quick question though, how could I push update to your msp430 branch now 🤔🤔, maybe I push my changes to a another branch then submit pr to yours 🤔🤔

cr1901:
@hathach 

>I have a quick question though, how could I push update to your msp430 branch now?

I believe you can push changes to this PR directly by doing something like:

```
git remote add cr1901 https://github.com/cr1901/tinyusb
git fetch cr1901 msp430f5529
git checkout msp430f5529
*Make your changes*
git commit -m "Changes"
git push cr1901 msp430f5529
```

After this, I'll make sure my own local copy matches yours :).

Feel free to merge if you want when you're done w/ your changes; just let me know if #149 and #184 still need fixing.

hathach:
@cr1901 I check out your branch, make some modification and pushed it here https://github.com/hathach/tinyusb/tree/cr1901-msp430f5529

It includes the fix for the #184  as well, but as you said, we need to also address the ZLP issue #149 . I will work on that and keep you update with the progress.

cr1901:
@hathach Thank you very much for your help!

I never actually committed changes that address your [review](https://github.com/hathach/tinyusb/pull/194#pullrequestreview-298557364). I'm gonna merge your branch back into mine, make appropriate changes, and make sure the demos work.

cr1901:
@hathach I am done with my own changes on my branch- all issues should be resolved. Have fun :D!

hathach:
> @hathach I am done with my own changes on my branch- all issues should be resolved. Have fun :D!

Thank @cr1901 , I am working on ZLP2 patch,  hopefully tinyusb could have better support for msp430 :)

cr1901:
>I am working on ZLP2 patch,
Looking forward to a fix very much :D. 

>Hopefully tinyusb could have better support for msp430 :)
The C code seems portable enough that any arch with a `gcc` port will work.

Re: "suspend/resume" and detach logic, I'd rather get the examples working first and I'll add support for that later via a PR. Detach logic is difficult to test on `msp_exp430f5529lp` and I've not had much luck creating a self-powered application. Is that okay?

hathach:
> Looking forward to a fix very much :D.

Just keep your patient :D 

> The C code seems portable enough that any arch with a `gcc` port will work.

It is the goal of this project to try supporting all ports :)

> Re: "suspend/resume" and detach logic, I'd rather get the examples working first and I'll add support for that later via a PR. Detach logic is difficult to test on `msp_exp430f5529lp` and I've not had much luck creating a self-powered application. Is that okay?

Yeah, suspend/resume is a bit advanced features, not all current ports are tested as well. We can have it as separated PR.

hathach:
I push the PR that could fix the ZLP. We can do PR to PR, once it is merged, I will do something similar to your branch as well. Making collaborate easier, especially we don't always spend time on this PR. Also there is some basic logging added with UART for debugging with the stack, you can just enable the `TUSB_CFG_DEBUG=2` for it, or make with DEBUG=1 e.g 
> make DEBUG=1 BOARD=pca10056

since it print uart inside the usbd task non-isr, it is safe enough, but it slows down the USB speed to Uart rate. But it is still useful, it isn't print much now, I may add more in the future, if it become too slow, we will only print more with higher level of DEUBG e.g TUSB_CFG_DEBUG=3.

a log example
```
USBD init
USBD: event SETUP_RECEIVED
  000:  80 06 00 01 00 00 40 00                         | ......@.
  Get Descriptor
USBD: event XFER_COMPLETE
  Endpoint: 0x80, Bytes: 18
USBD: event SETUP_RECEIVED
  000:  80 06 00 01 00 00 40 00                         | ......@.
  Get Descriptor
```

cr1901:
>I push the PR that could fix the ZLP. We can do PR to PR, once it is merged, I will do something similar to your branch as well. Making collaborate easier, especially we don't always spend time on this PR. 

Indeed, I agree. Right now, I won't be touching my branch until you're done making changes to it.

>Also there is some basic logging added with UART for debugging with the stack, you can just enable the TUSB_CFG_DEBUG=2 for it, or make with DEBUG=1

What a great idea :D! I personally think this should be separate from `DEBUG=1`, i.e. opt-in. I can think of a situation where you want debug symbols but not UART activity. What's the difference between `TUSB_CFG_DEBUG` and `DEBUG`?

hathach:
`TUSB_CFG_DEBUG` is one of option defined in `tusb_config.h` like others, the make `DEBUG` only make it easier when we want to enable that option. We can of course change it to other option like `LOG=` if needed. there is no requirement between TUSB_CFG_DEBUG and DEBUG except this line in the makefile https://github.com/hathach/tinyusb/blob/master/examples/make.mk#L82

hathach:
@cr1901 I submit an update an hacking code to get first 8 byte device descriptor play nicely with ZLP patch. https://github.com/cr1901/tinyusb/pull/1 

I am finally seeing msp430 enumerated nicely on my Linux machine. cdc_hid work great, kudo, great work :) . Will test with hid_composite later on.

hathach:
testing hid_composite on Linux, sometime msp430 will STALL the request from host for no reason
image

PS:  Somehow msp430 work flawlessly with my virtual machine windows, no STALL and It responded with ZLP pretty well when requested by windows (on Linux, driver set the wLength to 144), doing more test later on. 

image

cr1901:
>testing hid_composite on Linux, sometime msp430 will STALL the request from host for no reason

Hrm... unfortunately I am unable to test this on Linux right now. The only thing I can think of is:

>However the fix I committed is technically incomplete; after the Status Phase of a control xfer ends, I really should be clearing the NAK bits on EP0. Right now, tinyusb doesn't provide a way for me to do that, because there is no "Status Phase Complete callback" that can call into the DCD layer.

I'm not actually clearing the NAK bits after a Status Phase ends; it works on Windows, but maybe it breaks on Linux? I think perhaps the DCD should have an _optional_ (in other words, `weak` with the default to do nothing) "Status Phase Complete" callback to clear the NAK bits...

hathach:
I added the optional `dcd_control_status_complete()` which will be called when the status phase is complete https://github.com/cr1901/tinyusb/pull/2 . @cr1901 when you have free time, please add the implementation of clearing NAK bit for them. It can be used to prepare for the next setup as well if needed.

cr1901:
@hathach I'm still a bit overwhelmed, so I'll try to get to this by Wednesday. If I'm not able to, the next time I'll be able to feasible work on this is a week from today (Sunday).

hathach:
No problem at all, I understand, I got a bit overwhelmed as well, do it whenever you have time :)

cr1901:
@hathach Wanted to give an update:

>If I'm not able to, the next time I'll be able to feasible work on this is a week from today (Sunday).

This applies _this_ week (as well :P). After this Sunday or Monday, things should start going back to normal and I'll have time to look into the Linux errors...

hathach:
> @hathach Wanted to give an update:
> 
> > If I'm not able to, the next time I'll be able to feasible work on this is a week from today (Sunday).
> 
> This applies _this_ week (as well :P). After this Sunday or Monday, things should start going back to normal and I'll have time to look into the Linux errors...

not a problem at all, feel free to do it whenever you like. When I have some time, I will try to clear NAK bit as you suggested as well to see if it is a quick fix :)

cr1901:
I will hopefully circle back to this soon. It seems like @pigrew and a few others have been busy dealing with ZLP and "xfer divisible by packet size" fun recently. This most likely means that my hot patches in this branch no longer apply.

Therefore, my first order of business would be to just make sure an msp430 example compiles, even if enumeration fails. Then I can work on making the examples work, and most likely dealing with hardware-initiated STALLs again :).

hathach:
@cr1901 I could help with merging and syncing this current PR with master, if you haven't done it already. We already come very close to getting msp430 ported already.

cr1901:
@hathach I would appreciate the help if you can sync and merge this PR for me. Perhaps it might be alright to merge as-is, and debug issues as they come up (especially wrt ZLPs and multiple of packet size xfers)?

>I'm not actually clearing the NAK bits after a Status Phase ends; it works on Windows, but maybe it breaks on Linux? I think perhaps the DCD should have an optional (in other words, weak with the default to do nothing) "Status Phase Complete" callback to clear the NAK bits...

Windows worked fine last I checked, I recall you having issues with Linux. I've not tested whether `dcd_control_status_complete()` fixes Linux STALLs since I last worked on this in October.

hathach:
@cr1901 ok, let's leave that sync work for me, I will do overall testing again and let you know the result as wel :v:

cr1901:
@hathach Thank you very much, I appreciate the help more than you know. It's not been a good week for me (and I can _honestly_ say for most people), and I'm trying to cope. But I realize I need some help to accomplish my goals. This will help me _a lot_.

hathach:
Yeah, it is a worst week for the world in general. I am working on this pr this weekend. Will also trouble any issues with it as well.

hathach:
@cr1901 I have merged and submit PR to your branch https://github.com/cr1901/tinyusb/pull/3 . everything is working after synced with the master. We must have fixed an ZLP or control transfer issue with other PRs !!!. I have tested

cdc_msc
hid_composite
msc_dual_lun

It should be ready to merge to the master. 

PS: Apparently I didn't resolve conflict correctly (skip over the instruction from github). However, I did manual merge locally from master --> your branch. If you could merge/push --force that PR, it will work just fine. Sorry for that, I have never used rebase, and a bit hesitate to do so.

cr1901:
@hathach This PR is ready. I added a `dcd_edpt0_status_complete` dummy handler, per [your comment](https://github.com/hathach/tinyusb/pull/194#issuecomment-549125674). It doesn't work properly right now, and it's not required despite what the manual says. However, it would be nice to do things the right way eventually.

What probably has happened is that the device-independent parts of tinyusb changed enough in the past 6 months that my `dcd_edpt0_status_complete` violates some assumptions I make in the msp430 driver. We'll figure it out eventually.

examples/rules.mk
```diff
@@ -30,10 +30,15 @@ INC += $(TOP)/src
 
 #
 CFLAGS += $(addprefix -I,$(INC))
-LDFLAGS += $(CFLAGS) -fshort-enums -Wl,-T,$(TOP)/$(LD_FILE) -Wl,-Map=$@.map -Wl,-cref -Wl,-gc-sections -specs=nosys.specs -specs=nano.specs
+ifeq ($(BOARD), msp_exp430f5529lp)
+  LDFLAGS += $(CFLAGS) -fshort-enums -Wl,-T,$(TOP)/$(LD_FILE) -Wl,-Map=$@.map -Wl,-cref -Wl,-gc-sections
+else
+  LDFLAGS += $(CFLAGS) -fshort-enums -Wl,-T,$(TOP)/$(LD_FILE) -Wl,-Map=$@.map -Wl,-cref -Wl,-gc-sections -specs=nosys.specs -specs=nano.specs
+endif
+LDFLAGS += $(addprefix -L,$(LDINC))
```
hathach: I think we should move `LDFLAGS += $(addprefix -L,$(LDINC))` to board.mk like https://github.com/hathach/tinyusb/blob/master/hw/bsp/pca10056/board.mk#L24 
since it is not as nearly universal like C header include path

hw/bsp/msp_exp430f5529lp/board.mk
```diff
@@ -0,0 +1,32 @@
...
+INC += $(TOP)/hw/mcu/ti/msp430/msp430-gcc-support-files/include
+
+# For TinyUSB port source
+VENDOR = ti
+CHIP_FAMILY = msp430x5xx
+
+# Allow user to use mspdebug if set on make command line.
+ifdef MSPDEBUG
+# flash target using mspdebug.
+flash: $(BUILD)/$(BOARD)-firmware.elf
```
hathach: aha, I thought people would like to use MSP430Flasher more since it is from TI. But not at all, maybe just define it with `flash-mspdebug` target to skip entering that via command line.

cr1901: Oh right, the reason I didn't do `flash-mspdebug` is that the `which` line spits out a bunch of errors if `MSP430Flasher` is _not_ on your path. I'd like to avoid noise from that :P.

hathach:
> @hathach This PR is ready. I added a `dcd_edpt0_status_complete` dummy handler, per [your comment](https://github.com/hathach/tinyusb/pull/194#issuecomment-549125674). It doesn't work properly right now, and it's not required despite what the manual says. However, it would be nice to do things the right way eventually.
> 
> What probably has happened is that the device-independent parts of tinyusb changed enough in the past 6 months that my `dcd_edpt0_status_complete` violates some assumptions I make in the msp430 driver. We'll figure it out eventually.

Great idea, yeah I agree, it is much better to put some comment there, that will save us lots of reading time & headache in the future. Lots of things sure changes, but it is for the good though :) . There is somehow still `.travis.yml.bk` conflict. You could just delete that file now, we don't use travis as ci anymore.

hw/bsp/msp_exp430f5529lp/board.mk
cr1901: Actually, let's gate the `LD_LIBRARY_PATH` line to not run on Windows instead... that should work fine.

hathach: yeah, let's do that, much nicer, I still don't know why it could pickup the libmsp430.so when it is located in the same folder with MSP430Flasher in Linux :thinking: but anyway !!

cr1901:
Thank you for your help and patience :)! I agree this is an important milestone and am happy we finally got it merged.

>This is the first port with non 32-bit MCU. Which means TinyUSB is also good for even smaller chip.
Yes, besides being most familiar w/ msp430, this was another reason I did the port- to prove that the code can target smaller devices. Would love to possibly see an 8051/FX2 or AVR port eventually. The 16/8-bits still have some fight left in them!

hathach:
AVR has a very popular stack called LUFA, I am not sure if people want to switch. Although it nice to get more ports with them and 8051. It wouldn't be my focus at the moment. Though I am more than happy to help any others that want to do the dcd work :) . Great work 🎉🎊

PR #195, opened 2019-10-04
Add Spresense board

kamtom480:
I am currently porting CircuitPython to a Spresense board, which requires tinyUSB porting.

Refer to [developer.sony.com/develop/spresense/](https://developer.sony.com/develop/spresense/) for information about this board.

### DIFF
--- a/examples/make.mk
+++ b/examples/make.mk
@@ -80,5 +80,9 @@ CFLAGS += \
 ifeq ($(DEBUG), 1)
   CFLAGS += -O0 -ggdb -DCFG_TUSB_DEBUG=1
 else
+ifneq ($(BOARD), spresense)
   CFLAGS += -flto -Os
+else
+  CFLAGS += -Os
+endif
 endif
--- a/src/portable/sony/cxd56/dcd_cxd56.c
+++ b/src/portable/sony/cxd56/dcd_cxd56.c
@@ -141,12 +141,16 @@ static void dcd_suspend(FAR struct usbdevclass_driver_s *driver, FAR struct usbd
 {
   (void) driver;
   (void) dev;
+
+  dcd_event_bus_signal(0, DCD_EVENT_SUSPEND, true);
 }
 
 static void dcd_resume(FAR struct usbdevclass_driver_s *driver, FAR struct usbdev_s *dev)
 {
   (void) driver;
   (void) dev;
+
+  dcd_event_bus_signal(0, DCD_EVENT_RESUME, true);
 }
 
 void dcd_init(uint8_t rhport)
--- a/src/portable/sony/cxd56/dcd_cxd56.c
+++ b/src/portable/sony/cxd56/dcd_cxd56.c
@@ -1,0 +1,321 @@
+/* 
+ * The MIT License (MIT)
+ *
+ * Copyright 2019 Sony Semiconductor Solutions Corporation
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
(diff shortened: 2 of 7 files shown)


examples/make.mk
```diff
@@ -80,5 +80,9 @@ CFLAGS += \
 ifeq ($(DEBUG), 1)
   CFLAGS += -O0 -ggdb -DCFG_TUSB_DEBUG=1
 else
+ifneq ($(BOARD), spresense)
```
hathach: does **spresense** have issue with `lto` option. Do you know what could cause the option failed on this port

src/portable/sony/cxd56/dcd_cxd56.c
```diff
@@ -0,0 +1,321 @@
...
+}
+
+// Enable device interrupt
+void dcd_int_enable(uint8_t rhport)
+{
+  (void) rhport;
+}
+
+// Disable device interrupt
+void dcd_int_disable(uint8_t rhport)
```
hathach: Can USB IRQ on this port be disabled ? There could be a race condition between events generator such as `dcd_setup()` and event consumer like device task handling `tud_task()`

src/portable/sony/cxd56/dcd_cxd56.c
```diff
@@ -0,0 +1,321 @@
...
+{
+  (void) driver;
+
+  dcd_event_bus_signal(0, DCD_EVENT_BUS_RESET, true);
+  DEV_CONNECT(dev);
+}
+
+static void dcd_suspend(FAR struct usbdevclass_driver_s *driver, FAR struct usbdev_s *dev)
+{
+  (void) driver;
```
hathach: we should call `dcd_event_bus_signal(0, DCD_EVENT_SUSPEND, true)` here

src/portable/sony/cxd56/dcd_cxd56.c
```diff
@@ -0,0 +1,321 @@
...
+  (void) dev;
+}
+
+static void dcd_resume(FAR struct usbdevclass_driver_s *driver, FAR struct usbdev_s *dev)
+{
+  (void) driver;
```
hathach: should also call we should call `dcd_event_bus_signal(0, DCD_EVENT_RESUME, true)` here

examples/make.mk
kamtom480: Spresense uses the static library (libsdk.a) that was built without this flag. As far as I know, lto requires that all source files use this flag. For this reason, Spresense cannot use this flag.

src/portable/sony/cxd56/dcd_cxd56.c
kamtom480: I didn't have any problems but you are right. I added disabling and enabling interrupts.

src/portable/sony/cxd56/dcd_cxd56.c
kamtom480: Added!

src/portable/sony/cxd56/dcd_cxd56.c
kamtom480: Added

kamtom480:
I added missing items to PR.

Good to hear that you want to test Spresense. :)

You will also need an extension board to test this. The Extension USB is connected to the USB controller, but the Main USB is not. (https://developer.sony.com/develop/spresense/docs/introduction_en.html#_extension_board)

examples/make.mk
hathach: Ah I see, LTO require all sources must be compile with the flags. Linking with precompiled library without lto can be troublesomes. I may modify makefile to make this as option for `board.mk`, it will be a cleaner solution.

src/portable/sony/cxd56/dcd_cxd56.c
hathach: @kamtom480 could we only disable the USB IRQ, this seems to disable all interrupts ? This could affect other modules/driver. Sorry for the question, since I am not familiar with this mcu.

hathach:
> I added missing items to PR.
> 
> Good to hear that you want to test Spresense. :)
> 
> You will also need an extension board to test this. The Extension USB is connected to the USB controller, but the Main USB is not. (https://developer.sony.com/develop/spresense/docs/introduction_en.html#_extension_board)

Ah thanks, I didn't know that. I will order it as well. It is useful to have since I tried to run all examples on every ports every time releasing the library to make sure modification doesn't break :).

src/portable/sony/cxd56/dcd_cxd56.c
kamtom480: That's right. This disables all interrupts. I changed this to disable only USB interrupt.

PR #212, opened 2019-11-04
fix #211

### LINKED ISSUE #211: Provide compilation option not to inline `dcd_*` callback functions
**Is your feature request related to a problem? Please describe.**

Currently `dcd_*` callback functions are inlined. This causes the corresponding linker symbols not to be generated. 

This means we _cannot_ call them from Rust.

**Describe the solution you'd like**

It would be good if we can get a compilation option so the callback functions are not inlined.

**Describe alternatives you've considered**

I am currently using [this](https://gist.github.com/rajivr/b31277f710860cdd7c30dcd7e9eb7b3c) patch as a workaround.

hathach:
@rajivr will merge after passing travis

### DIFF
--- a/src/device/dcd.h
+++ b/src/device/dcd.h
@@ -120,50 +120,20 @@ void dcd_edpt_stall       (uint8_t rhport, uint8_t ep_addr);
 void dcd_edpt_clear_stall (uint8_t rhport, uint8_t ep_addr);
 
 //--------------------------------------------------------------------+
-// Event API
+// Event API (Implemented by device stack)
 //--------------------------------------------------------------------+
 
 // Called by DCD to notify device stack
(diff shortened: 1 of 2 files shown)


pigrew:
@hathach :

I might be out of the loop, but do we need to declare the functions `xtern`?

Would this degrade results if LTO is enabled?

My understanding was that bug #211 only requested that the `static inline` attribute be removed?

hathach:
@pigrew It shouldn't be an issue, function header is default to be **extern** I remembered. `extern` just one way to tell dcd that these API are implemented by the stack. But the main reason is to keep CMock (part of the unit testing framework) happy.

rajivr:
@hathach Thank you for the changes. :-)

I tested with master (`3c49ff15`) and it works as expected.

Issue #207, opened 2019-11-01
Order of irq handling matters?


xobs:
It appears as though the order that irq events are handled matters, and this order is undocumented.

I'm porting tinyusb to a new platform, and right now it's running slowly. This architecture can handle IN and OUT endpoints at the same time, and as a result its possible to handle both inside the isr (provided it's running slowly enough).

As a result, I get odd behavior in examples such as the MSC example. The following sequence of events appears to trigger the issue:

1. The Device sends a Status Transport response (IN). The packet transmits while the Device resumes execution.
2. The host sends a Command Transport request (OUT). 
3. The host begins sending Data Transport requests (IN).

My ISR and processes OUT packets first, since I thought OUT and IN endpoints were separate and order doesn't matter. However, it appears to matter at least in this case.

During (2),the Device queues 512 bytes of data to be sent to the Host. This blocks, since the IN packet from (1) hasn't been finished. Then, execution continues and we process IN packets. The Device sees that the 13 byte transfer from (1) has finished, so it calls xfer_complete.

The odd behavior is that tinyusb will now queue 499 bytes to be sent, on top of the 512 bytes it's already sent.

Does the order matter? Should that be documented anywhere? I notice that many drivers handle OUT packets last, but is that merely coincidence or is it a requirement?

I will post traces and logs as soon as my home Internet connection gets fixed.

xobs:
More information:

Here's a gdb trace of the last 8 queued transfer sizes and endpoints, along with the order in which `xfer_complete()` was called:

```
(gdb) p xfer_log_offset
$38 = 52 '4'
(gdb) p queue_log_offset
$39 = 8 '\b'
(gdb) p/x xfer_log[45]@7
$40 = {{
    ep_num = 0x83,
    size = 0x200
  }, {
    ep_num = 0x83,
    size = 0xd
  }, {
    ep_num = 0x3,
    size = 0x1f
  }, {
    ep_num = 0x83,
    size = 0x200
  }, {
    ep_num = 0x3,
    size = 0x1f
  }, {
    ep_num = 0x83,
    size = 0xd
  }, {
    ep_num = 0x83,
    size = 0x200
  }}
(gdb) p/x queue_log[0]@7
$41 = {{
    ep_num = 0x83,
    size = 0x200
  }, {
    ep_num = 0x83,
    size = 0xd
  }, {
    ep_num = 0x3,
    size = 0x1f
  }, {
    ep_num = 0x83,
    size = 0x200
  }, {
[...]

PR #217, opened 2019-11-13
WIP: Add Valentyusb eptri

xobs:
This adds support for the new `eptri` interface, as described at https://rm.fomu.im/usb.html

This has been tested on a Fomu EVT board with SPI debugging, with both Linux (Raspberry Pi) and Windows hosts.

So far I have tested `cdc_msc`, `msc_dual_lun`, and `usbtmc`, proving that interleved transactions work.

If you have a Fomu that's running eptri (which has not yet been released as part of an official firmware), you can build the examples by running:

```
make CROSS_COMPILE=riscv64-unknown-elf- BOARD=fomu
```

I have several questions:

* What needs to be done to get this merged?
* Is this an acceptable approach for multi-platform?
* What sorts of tests should I run to ensure this is reliable?
* There are lots of "WIP commits" here.  Should I squash them?
* How should we do testing?  `CROSS_COMPILE` could be `riscv64-unknown-elf-` if you're using a multiarch toolchain, or `riscv32-unknown-elf-` if you're using a self-built toolchain.


src/tusb_option.h
```diff
@@ -70,6 +70,8 @@
 
 #define OPT_MCU_CXD56         400 ///< SONY CXD56
 
+#define OPT_MCU_VALENTYUSB_EPTRI    500 ///< Fomu eptri config
```
hathach: not a big deal, but it is best to change this to 600, since there is WIP PR for msp430 that use this value https://github.com/hathach/tinyusb/pull/194/files#diff-73307617739bffdbee43cd26855ee4c0R72

hathach:
> This adds support for the new `eptri` interface, as described at https://rm.fomu.im/usb.html
> 
> This has been tested on a Fomu EVT board with SPI debugging, with both Linux (Raspberry Pi) and Windows hosts.
> 
> So far I have tested `cdc_msc`, `msc_dual_lun`, and `usbtmc`, proving that interleved transactions work.
> 
> If you have a Fomu that's running eptri (which has not yet been released as part of an official firmware), you can build the examples by running:
> 
> ```
> make CROSS_COMPILE=riscv64-unknown-elf- BOARD=fomu
> ```

Superb !! Thank you very much for the PR, this fomu fpga + risc V board is entirely new to me. It will take a bit of time for me to learn this awesome work.

> I have several questions:
> 
> * What needs to be done to get this merged?

Normally I only need travis build to pass, I will also try to run local test with the hw to see if I could spot any issues. Currently the travis build failed due to lack of toolchain 
https://travis-ci.com/hathach/tinyusb/builds/136267447#L356
Can you update the `.travis.yml` to download the toolchain and `make.mk` to use the correct compiler based on the board. I will do a better board-to-compiler mapping when the ifeq grow bigger. You may want to check out this also WIP PR for msp430 for reference, it also download and run ti msp430 compiler https://github.com/hathach/tinyusb/pull/194/files#diff-73307617739bffdbee43cd26855ee4c0R72

> * Is this an acceptable approach for multi-platform?

Yes, I don't see any issues with this approach. As long as it doesn't break existing code, we could always enhance it later on.

> * What sorts of tests should I run to ensure this is reliable?

Testing with existing examples is good enough, it shows that DCD is working as expected from device stack usbd. There may be corner case or race condition issues when transferring with multiple endpoints etc .. but that is not a problem for now. 

> * There are lots of "WIP commits" here.  Should I squash them?

I have zero issues with long list of commits, I would suggest to leave it as it is to save our time.

> * How should we do testing?  `CROSS_COMPILE` could be `riscv64-unknown-elf-` if you're using a multiarch toolchain, or `riscv32-unknown-elf-` if you're using a self-built toolchain.

I am new to riscv, please pick what you think is best or what most of people would use. 

PS: Please also try with the PR #208 to see if it fix your #207 issue

xobs:
All tests now pass,  I opted for the xPack toolchain, so `CROSS_COMPILE` doesn't need to be specified anymore.

I'll try and come up with a bitstream you can load onto a Fomu device, so you can test it out.

hathach:
> All tests now pass, I opted for the xPack toolchain, so `CROSS_COMPILE` doesn't need to be specified anymore.
> 
> I'll try and come up with a bitstream you can load onto a Fomu device, so you can test it out.

Thanks, that will be useful, I will definitely need to learn this cool stuffs :)

siddacious:
I <3 this PR

xobs:
There are a few things I'd like to do still:

* Check the `VERSION_MAJOR` register and only run if it's `2` (http://rm.fomu.im/version.html#version-major) -- this is because the currently-released version `1` uses `epfifo` and not `eptri`
* Maybe rename `fomu` to `fomu-eptri`, in case we do add support later for `fomu-epfifo`?

hathach:
> There are a few things I'd like to do still:
> 
> * Check the `VERSION_MAJOR` register and only run if it's `2` (http://rm.fomu.im/version.html#version-major) -- this is because the currently-released version `1` uses `epfifo` and not `eptri`
> * Maybe rename `fomu` to `fomu-eptri`, in case we do add support later for `fomu-epfifo`?

Yeah, these are totally up to you to decide, though IMHO, since `eptri` support 16 EP while `epfifo` only support 2 EPs, I don't see reason why people wouldn't want to upgrade from `epfifo` to `eptri`. Unless the upgrade progress is kind of a PITA :) . But we can always add `fomu-epfifo` later on if needed, maybe with a separate PR. Let's me know when you are happy with the code, I will do the merge.

xobs:
Go ahead and merge please! I'll go through the process of getting it working in Fomu core so that it can be used in core Fomu.

hathach:
Merged :)

PR #232, opened 2019-11-29
valentyusb: eptri: coalesce resets

xobs:
Sorry, found another small issue.  I was overflowing the queue with reset events.  This prevented it from working on Intel boards under Linux.

### DIFF
--- a/src/portable/valentyusb/eptri/dcd_eptri.c
+++ b/src/portable/valentyusb/eptri/dcd_eptri.c
@@ -42,6 +42,8 @@
 #include "irq.h"
 void fomu_error(uint32_t line);
 
+static uint8_t last_ev;
+
 #if LOG_USB
 struct usb_log {
   uint8_t ep_num;
@@ -568,7 +570,11 @@ static void handle_reset(void)
 
   // This event means a bus reset occurred.  Reset everything, and
   // abandon any further processing.
-  dcd_reset();
+
+  // The core can sometimes send us lots of resets, particularly if
+  // it's a prolonged SE0 state.  Ignore these subsequent resets.
+  if (last_ev != (1 << CSR_USB_NEXT_EV_RESET_OFFSET))
+    dcd_reset();
 }
 
 static void handle_setup(void)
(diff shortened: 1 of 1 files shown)


xobs:
I think the issue might be considered a hardware bug.

For starters, the CPU can be incredibly slow, especially if it's executing out of XIP SPI flash.

Next, the `usb_reset` hardware signal basically indicates whether the device is _currently in reset_.  That is, that signal goes high when a reset starts, and goes low when a reset finishes. https://github.com/im-tomu/valentyusb/blob/master/valentyusb/usbcore/rx/pipeline.py#L48-L58

This signal is exposed to `eptri, and is directly wired into an `EventSourcePulse` as the trigger: https://github.com/im-tomu/valentyusb/blob/97ac21d6ea4f9b632773a833476906c1ab95a730/valentyusb/usbcore/cpu/eptri.py#L497

However, the way `EventSourcePulse` appears to be implemented, if the signal is long and not just a short pulse, it becomes a level-triggered interrupt instead of an edge-triggered one: https://github.com/enjoy-digital/litex/blob/master/litex/soc/interconnect/csr_eventmanager.py#L73

So perhaps this should be fixed in the hardware as well.

hathach:
@xobs yeah, I think this counts as hw issue, it is much better if you could fix this issue in the core. The interrupt should only fired at the end of reset sequence. Although the specs say SE0 ~ 10ms is counted as reset. Host controller driver (windows/linux) normally can prolong it to typically 50-100 ms. Though the software walk-around is still enough for now, we can revert this back later on :)

image

hathach:
Let's me know if you want me to merge the PR for now, or decide to fix the hardware instead :)

xobs:
I think better to fix it in hardware, since the device is effectively unusable when the host has it in reset.  With this patch, tinyusb will no longer give asserts, it'll just stop working.

hathach:
Let's fix the hardware then :)

xobs:
Fixed in https://github.com/im-tomu/valentyusb/commit/b262b994fd9e6568b257b8424fe808cfc403e652

PR #233, opened 2019-12-01
Fix mynewt osal queue definition

kasjer:
Queue table has pointers instead of data os_event structs.
This resulted in crashes when elements were put to queue and
overwritten variables that were just after mpool desiged
for queue.


hathach:
Thanks for pr. I am currently on vacation, will check it out in a week or so.

PR #234, opened 2019-12-11
move VBUS sense out of dcd_synosys to bsp

hathach:
Hardware VBUS sensing with PA9 is optional in case application want to force either Host or Device most (no OTG). Since this is a valid use case,which is the case of low-cost stm32f411 **blackpill** https://github.com/adafruit/circuitpython/pull/2361 .  `dcd_synosys` should not enforce VBUS sensing and should leave this to board initialization. 

@cr1901 @tannewt  @hierophect for review

image

image

- Also add VBUS sense enable for all current boards.


cr1901:
Looks fine to me.

I put VBUS sensing into `dcd_synopsys` because it was the configuration that _worked_, and I thought it was universal. Apparently not :).

hathach:
thanks @cr1901 and @hierophect for reviewing

PR #236, opened 2019-12-11
Update dcd_stm32_fsdev.c with note about F042Fx remapping

jeremyherbert:
On the STM32F042Fx, there is a special pin remapping step required. It is only used in the low pincount packages and uses a register which is not inside the GPIO group. This just adds a note in the comments about it so hopefully someone else does not waste as much time as I did on this ;)

### DIFF
--- a/src/portable/st/stm32_fsdev/dcd_stm32_fsdev.c
+++ b/src/portable/st/stm32_fsdev/dcd_stm32_fsdev.c
@@ -30,7 +30,7 @@
 
 /**********************************************
  * This driver has been tested with the following MCUs:
- *  - F070, F072, L053
+ *  - F070, F072, L053, F042F6
  *
  * It also should work with minimal changes for any ST MCU with an "USB A"/"PCD"/"HCD" peripheral. This
  *  covers:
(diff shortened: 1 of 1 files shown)


pigrew:
Looks like good comments to add.

I think I remember STM32CubeMX doing the pin remapping for me, as part of its GPIO initialization. 

On the other hand, it's good to hear that the code works in crystal-less mode; I only tested it with an external oscillator.

hathach:
@jeremyherbert thanks for the note, are you using an ST devkit (or any easy-to-buy board). It would be great if you could make an board support PR for it :)

jeremyherbert:
I'm just using the IC on a breadboard right now. I did think of making a BSP but it's kind of hard to do for just this. But I will hopefully be making a small open source board soon so I will add one at that point. Here is my rough code anyway in case anyone comes from google: https://gist.github.com/jeremyherbert/45fbcd32e6bd187375459ad6402dd0bb

On the CubeMX, it does set things up properly, but it's in a bit of a confusing way. If I remember correctly, it puts the pinswap code in the HAL_MspInit function and not in the GPIO init function, so if you just copy the init code into a different project it doesn't come with it. Even worse is that when you enable the D+ pullup, it will still work even if the pinswap was not configured correctly. So if you hook up an analyser to the bus, the host will still detect a new device, but it will never respond because the USB subsystem is not actually connected to the D+/D- pins (see below picture)

_Untitled_-_Total_Phase_Data_Center_v6_73_007

hathach:
Great to hear you plan to make opensource hardware, I am looking forward to your PR when it is out to the market :)

PR #237, opened 2019-12-16
[tusb]fix typo in comment

samveen:
fix typo in comment in `src/class/hid/hid.h`


samveen:
@hathach I was actually searching for that line, and it didn't match, thus ... :D

PR #240, opened 2019-12-23
implement multiple interfaces support

majbthrd:
Instances of "TODO Support multiple interfaces" in the CDC, HID, MIDI, and VENDOR device classes were replaced with code that implemented multiple interfaces.

### DIFF
--- a/examples/device/cdc_dual_ports/Makefile
+++ b/examples/device/cdc_dual_ports/Makefile
@@ -1,0 +1,12 @@
+include ../../../tools/top.mk
+include ../../make.mk
+
+INC += \
+  src \
+  $(TOP)/hw \
+
+# Example source
+EXAMPLE_SOURCE += $(wildcard src/*.c)
+SRC_C += $(addprefix $(CURRENT_PATH)/, $(EXAMPLE_SOURCE))
+
+include ../../rules.mk
--- a/examples/device/cdc_dual_ports/src/main.c
+++ b/examples/device/cdc_dual_ports/src/main.c
@@ -1,0 +1,98 @@
+/* 
+ * The MIT License (MIT)
+ *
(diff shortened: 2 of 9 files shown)


cr1901:
I think I'm confused by that TODO: I thought CDC by definition required multiple interfaces (with one endpoint each)?

majbthrd:
You are not wrong; CDC does require multiple interfaces.

This patch is a class driver improvement.  There is nothing to cause additional CDC ports to sprout on the existing examples.

To take advantage of this added functionality, there are assorted changes that must be made to the usb_descriptors.c (or equivalent) and tusb_config.h, not to mention the user code that leverages the additional ports.

CFG_TUD_CDC must be increased to the number of ports, corresponding elements must be added to the ITF_* enum, additional TUD_CDC_DESCRIPTOR elements must be added to the desc_configuration (including valid and unique endpoint addresses), etc., etc.

This is all typical of any usage of tinyusb.

hathach:
> I think I'm confused by that TODO: I thought CDC by definition required multiple interfaces (with one endpoint each)?

My bad, that TODO should said to support multiple instance of CDCs (ACM port) instead. @majbthrd thanks for the PR, I will check it out as soon as  I could. I will probably create an additional `cdc_dual_ports` example to make sure it works as proof of concept and just to show user how to use the `tud_cdc_n` API

majbthrd:
@hathatch, to save you the hassle, I've updated the PR to include a cdc_dual_ports example that I whipped up, and this demonstrates the tud_cdc_n API.

Like one of your existing examples, it implements an echo/loopback.  The gimmick with this code is the way that it demonstrates the different serial ports: all echoed text on the first port is always in lower case, and the second port is always echoed in upper case.

Issue #241, opened 2019-12-23
TinyUSB  with cdc_msc causes macbook pro to not wake from sleep


jeremyherbert:
While trying to investigate #179 further, I have discovered that a certain combination of devices causes my laptop to never wake from sleep. As soon as I unplug the tinyUSB device, the wakeup works fine.

**Set up:**
 - OS: Mac OS Catalina
 - Board: STM32F042F6 on a breadboard
 - Firmware Code: cdc_msc example

**To Reproduce**
I have a MBP (16"  late 2019) connected to a thunderbolt 3 dock (Caldigit TS3 Plus). I then have a Orico 10 port USB3 hub (A3H10, VL812 chipset I believe) connected to the dock and the tinyUSB device is plugged into the hub via a USB analyser.

On a separate PC I have a Beagle 480 analyser (which is inline with the tinyUSB device) and an ST-Link which is connected to the STM32F042 via SWD. This PC also has a USB to serial converter which I have connected to the STM32F042 and enabled level 2 debugging.

To trigger the problem, I put the laptop to sleep, wait a minute or two for it to actually go to sleep, and then try to wake it (pressing keys on the keyboard, etc). Laptop is totally unresponsive until I unplug the tinyUSB device, after which everything goes back to normal.

**Expected behavior**
The MBP should be able to wake up.

**Additional context**
The log output roughly shows this (note that I have commented out the SETUP_RECEIVED dump as it was taking too long to print)

```
USBD: event SUSPEND
USBD: event RESUME
USBD: event SETUP_RECEIVED
    Get Status
process_control_request 620: assert failed
                                            Stall EP0
USBD: event BUS_RESET
USBD: event SETUP_RECEIVED
    Set Address
USBD: event XFER_COMPLETE
  Endpoint: 0x80, Bytes: 0
  EP Addr = 0x80, len = 0
USBD: event SETUP_RECEIVED
    Get Descriptor
...
```
full log is here: https://gist.github.com/jeremyherbert/adf841bfdad76d621fea0cdd7e43ce20

tannewt:
I saw an issue where CircuitPython (which uses TinyUSB) would actually crash Catalina due to a timeout trying to set the SCSI power state. To fix it I had to make sure and respond to a start command with an error. This is done in a callback so I didn't actually need to change TinyUSB. The fix for CircuitPython is here: https://github.com/adafruit/circuitpython/pull/2406/files#diff-880d63a1c107dd299a7ed80251505a9eL200

[...]

Issue #213, opened 2019-11-04
stm32f0 (fsdev) seems skip setup packet


hathach:
**Set up (please complete the following information):**
 - OS: Ubuntu 18.04
 - Board: stm32f072disco
 - Firmware Code: examples/device/cdc_msc

**Describe the bug**
found this issue while merging #149, 

fsdev somehow miss a setup packet, it doesn't report the setup packet that it doesn't response to. 
sometime it misses a setup packet, then completely non-response to that packet as following screenshot nd log. As shown in the log, the setup packet is as follows **string index 0, len = 255, then string index 2 len =2. It completely skip the string index = 2, len = 255**

I have no idea how removing above line could cause this issue, it is possibly something related to the fsdev timing thing ? Not sure though hmm. I was thinking may be event queue is overflow and increased `CFG_TUD_TASK_QUEUE_SZ` but it doesn't help

You can get the similar log with new `LOG=`, LOG value will set the `CFG_TUSB_DEBUG` option.

> make BOARD=stm32f072disco LOG=1 clean all flash

![image](https://user-images.githubusercontent.com/249515/68081510-9b19cf00-fe41-11e9-98e9-6c3135acf768.png)

```
000:  80 06 00 02 00 00 62 00                         | ......b.
  EP Addr = 0x80, len = 64                                                        
  EP Addr = 0x80, len = 34                                                        
  EP Addr = 0x00, len = 0                                                         

000:  80 06 00 03 00 00 FF 00                         | ........
  EP Addr = 0x80, len = 4                                                         
  EP Addr = 0x00, len = 0                                                         
000:  80 06 02 03 09 04 02 00                         | ........
  EP Addr = 0x80, len = 2                                                         
  EP Addr = 0x00, len = 0                                                         
000:  80 06 02 03 09 04 1E 00                         | ........
  EP Addr = 0x80, len = 30                                                        
  EP Addr = 0x00, len = 0                                                         

000:  80 06 01 03 09 04 FF 00                         | ........
  EP Addr = 0x80, len = 16                                                        
  EP Addr = 0x00, len = 0                                                         
```

PR #251, opened 2019-12-27
Let device respond to endpoint requests on EP0

jeremyherbert:
Since endpoint 0 is used for control requests, it doesn't have a class driver attached to it. As such, the corresponding `_usbd_dev.ep2drv` entry points to driver `0xFF`, which is invalid and this makes the `TU_ASSERT(drvid < USBD_CLASS_DRIVER_COUNT);` line fail. This eventually causes an unconditional endpoint stall. So as-is the stack cannot properly respond to any endpoint requests on endpoint 0.

However, standard requests on endpoint 0 do not need a class driver to produce a valid response. This commit changes the order of execution so that the assert is only checked if the endpoint is not 0. Alternatively, if a custom driver has been attached to endpoint 0 it still passes the events through (I have no idea if this is actually allowed in this codebase though).

Fixes https://github.com/hathach/tinyusb/issues/241 and maybe related to https://github.com/hathach/tinyusb/issues/213

### DIFF
--- a/src/device/usbd.c
+++ b/src/device/usbd.c
@@ -618,7 +618,6 @@ static bool process_control_request(uint8_t rhport, tusb_control_request_t const
       TU_ASSERT(ep_num < TU_ARRAY_SIZE(_usbd_dev.ep2drv) );
 
       uint8_t const drvid = _usbd_dev.ep2drv[ep_num][ep_dir];
-      TU_ASSERT(drvid < USBD_CLASS_DRIVER_COUNT);
 
       bool ret = false;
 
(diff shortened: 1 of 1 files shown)


src/device/usbd.c
```diff
@@ -618,7 +618,6 @@ static bool process_control_request(uint8_t rhport, tusb_control_request_t const
       TU_ASSERT(ep_num < TU_ARRAY_SIZE(_usbd_dev.ep2drv) );
 
       uint8_t const drvid = _usbd_dev.ep2drv[ep_num][ep_dir];
-      TU_ASSERT(drvid < USBD_CLASS_DRIVER_COUNT);
```
hathach: Yeah, this maybe possible. But I am curious which request the macos use to run into this issue. Since the only request targeted endpoint is halt/clear halt. Which will never be used on Control Endpoint anyway

jeremyherbert: It’s a GET_STATUS on EP0.

jeremyherbert: According to the USB2.0 spec, this is definitely valid. In fact, in the address state the *only* valid GET_STATUS endpoint is EP0. See section 9.4.5. The last paragraph also basically says that GET_STATUS should never stall: 

“It is neither required nor recommended that the Halt feature be implemented for the Default Control Pipe. However, devices may set the Halt feature of the Default Control Pipe in order to reflect a functional error condition.  If the feature is set to one, the device will return STALL in the Data and Status stages of each standard request to the pipe **except GetStatus()**, SetFeature(), and ClearFeature() requests.”

hathach: Right, right, I forgot the  the GET_STATUS request, these are rarely used on control endpoints. Thank you very much for the explanation.

src/device/usbd.c
```diff
@@ -658,13 +657,17 @@ static bool process_control_request(uint8_t rhport, tusb_control_request_t const
...
       }
 
-      // Some classes such as USBTMC needs to clear/re-init its buffer when receiving CLEAR_FEATURE request
-      // We will forward all request targeted endpoint to class drivers after
-      // - For class-type requests: driver is fully responsible to reply to host
-      // - For std-type requests  : driver init/re-init internal variable/buffer only, and
-      //                            must not call tud_control_status(), driver's return value will have no effect.
-      //                            EP state has already affected (stalled/cleared)
-      if ( invoke_class_control(rhport, drvid, p_request) ) ret = true;
+      if (ep_num != 0 || (ep_num == 0 && drvid != 0xFF)) {
```
hathach: I think the if condition is a bit overkill, a simple
`if ( drvid < USBD_CLASS_DRIVER_COUNT )`
is sufficient enough. Else we just simply  do nothing :)

jeremyherbert: Ok, changed as you wish. This is what I had originally but thought it might be a bit too broad.

PR #258, opened 2020-01-02
MIDI packet interface 

kaysievers:


cr1901:
@hathach This PR might've fell through the cracks... think we should take a look at trying to get this merged? Looks like it doesn't touch much of the code that's being hacked on right now.

Unfortunately, I don't know enough about MIDI (right now!) to do a review, otherwise I would do it.

hathach:
@cr1901 indeed, I feel so terrible for losing track of this PR as well. Sorry @kaysievers , you have been submitting MIDI related PR to this core, and Arduino repo. I always tell myself to pull out the MIDI spces to check but got sunken into lots of stuff. I will definitely check this out this weekend. Thank you for reminding me.

PR #259, opened 2020-01-04
add Nuvoton NUC121/NUC125/NUC126

majbthrd:
this includes:

- a new USB device driver for the Nuvoton NUC121, NUC125, and NUC126
- BSP for three Nuvoton eval boards (NuTiny-NUC121S, NuTiny-NUC125S, and NuTiny-NUC126V)
- adds a git submodule to ./hw/mcu pointing to https://github.com/majbthrd/nuc_driver to obtain include files, etc.


hathach:
@majbthrd ah hah, gcc being very smart and catch the error in the NUC std driver, where the register is using the wrong mask  UI2C_PROTSTS_NACKIF_Msk, should be UI2C_PROTSTS_STORIF_Msk :D 
https://github.com/majbthrd/nuc_driver/blob/master/nuc126/StdDriver/src/usci_i2c.c#L571

Since TinyUSB example only use gpio, and uart. We can safely exclude this from SRC_C :)

hathach:
@majbthrd I submit an PR to your fork https://github.com/majbthrd/tinyusb/pull/1 to fix ci build since you may not be familiar with it. Please review it, once it merged on your fork, the changes will be propagated to this PR as well.

hathach:
Superb!! Merged, thank you very much 👍👍

hathach:
@majbthrd my NUC board arrived today, I spent a hour or so to find a way to flash it  with command line in Linux (ubuntu), but couldn't. Look like Nuvoton only provide tools for windows !? I would be appreciated if you could give me a hint to use nulink on Linux :)

majbthrd:
@hathatch, sorry about that, I presumed that you would be using a Segger J-Link (as this is what the TinyUSB Makefiles do).

I think it unfortunate that Nuvoton has a proprietary interface with only IAR and Keil drivers as well as a Windows-only utility.  They do, however, have a fork of OpenOCD that I have not tried:

https://github.com/OpenNuvoton/OpenOCD-Nuvoton

When I bought the boards, my immediate approach was to break off their proprietary Nu-Link-Me along the PCB break-line, solder a connector, and use an external debugger.  The signals are marked in silkscreen on the bottom of the PCB.

Their Nu-Link-Me is based on a "NUC12SRE3DE", which seems to be a NUC120-series variant (device ID 0x4001_2009 instead of 0x0001_2009).  I've contemplating porting my ["Dapper Miser" CMSIS-DAP implementation](https://github.com/majbthrd/DapperMiser/), but the Nu-Link-Me firmware update over USB mechanism is proprietary.  So, the only way to load the new code would be to use an external debugger, and if the user already has an external debugger, there isn't much incentive for them to have interest to reprogram the Nu-Link-Me.

hathach:
@majbthrd You have nothing to sorry about, I tried with Nuvoton openocd fork and it work very well, hopefully they submit an PR to upstream openocd in the near future. Only need to compile it, but it is chore under Ubuntu :) . I have updated the NUC BSP with LED, and flash target here https://github.com/hathach/tinyusb/pull/270

Yeah, the makefile have a universal `flash-jlink` since jlink support most of the mcu on the market, `flash` is the typical/default way that most user will be using to upload firmware (dfu/ stlink etc...). Don't worry about it, I am just new to Nuc micros :)

PR #260, opened 2020-01-04
CDC device: fix behavior for transfers that are a whole multiple of endpoint buffer

majbthrd:
A proper implementation of the CDC class should append a Zero Length Packet to transfers that are a whole multiple of the endpoint buffer.

To demonstrate the current incorrect behavior, take the device examples "webusb_serial" or "cdc_dual_ports".  Take this line in main.c:

      tud_cdc_write_char(buf[i]);

and modify it to be:

      for (int j = 0; j < 32; j++)
        tud_cdc_write_char(buf[i]);

Re-compile and download.  Now, when you run your favorite terminal emulator and connect to the emulated serial port, every single keystroke you make will result in 32 copies.  This is good so far.

However, change the "32" to "64", re-compile, download, and try the terminal emulator again.  Depending on your OS and driver, you might get some characters out sometimes, but clearly something is wrong.  I believe what is wrong is the absence of a ZLP in the CDC class driver implementation.

I believe the patch is the most expedite way to achieve proper operation.  I have tested this on STM32F072, SAMD32, and NUC121 platforms.


pigrew:
I'm not in a position to test the code, but what about if CDC is used to send a multiple of the packet length? e.g., 64 bytes with a packet length of 32?

pigrew:
Will this send a spurious ZLP when transmitting a multiple of the packet length? (e.g., 64 byte message with a 32 byte packet length)

I'm not currently able to test the code... I'm only reading it.

majbthrd:
Thanks so much for the review and critique, but I think it is good in this situation.  To be absolutely sure, I just tried it to confirm.

In this context, "xferred_bytes" is not accumulative.  TinyUSB hands a transfer size to the driver, and then driver generates a callback once those bytes have been sent.

Having just written a Nuvoton driver, this is all fresh in my memory.  In TinyUSB, EP0 and EP1 get treated specially and are fragmented in the driver; I don't believe there are ever EP0/EP1 transfers more than the endpoint size.  For the other endpoints, however, transfers can be smaller or larger than the endpoint size.  The same value is passed back through the CDC callback.

So, a transfer of 32 + 32 gets two callbacks, each with an "xferred_bytes" of 32.  In neither of those cases is "(0 == (xferred_bytes % CFG_TUD_CDC_EPSIZE))".

majbthrd:
@pigrew, alas, I managed to misread your described scenario when first responding.  The situation is still OK, but the explanation is different.

When transmitting a multiple of the packet length, a theoretical 64 byte transfer with a 32 byte packet length would result in a single callback with xferred_bytes = 64, causing a single ZLP to be generated.

I say "theoretical" because the current tud_cdc_n_write_flush() implementation caps all transfers to the packet length (CFG_TUD_CDC_EPSIZE).  It does this, if for no other reason, because the destination of the data copy (epin_buf) is also capped at that same size.

It is possible to hack up the code to confirm your scenario works, though, and I've done this for certainty.  However, that made me appreciate that the current code could be improved to ensure consistency between the code and the size of the buffers that it operates on, so I took the liberty of making an additional commit.

src/class/cdc/cdc_device.c
```diff
@@ -402,10 +402,10 @@ bool cdcd_xfer_cb(uint8_t rhport, uint8_t ep_addr, xfer_result_t result, uint32_
   // Data sent to host, we could continue to fetch data tx fifo to send.
   // But it will cause incorrect baudrate set in line coding.
   // Though maybe the baudrate is not really important !!!
-//  if ( ep_addr == p_cdc->ep_in )
-//  {
-//
-//  }
+  if ( ep_addr == p_cdc->ep_in )
+  {
+    if ( xferred_bytes && (0 == (xferred_bytes % CFG_TUD_CDC_EPSIZE)) ) usbd_edpt_xfer(TUD_OPT_RHPORT, p_cdc->ep_in, NULL, 0);
```
kasjer: This fixes the problem of endpoint-size transmission that is not perceived on host side due to missing zero length packet.
But as @pigrew noticed zero length packets will be send in more cases than needed.

Changing condition to:
```if ( xferred_bytes && (0 == (xferred_bytes % CFG_TUD_CDC_EPSIZE)) && p_cdc->tx_ff.count == 0)```
...could skip sending ZLP when there is something more in fifo already and will be sent in next flush.

majbthrd: Thanks @kasjer ; since this is a now dated branch, I've merged it with the presently master branch, and have implemented your suggestion.

rymanluk: @hathach anything is missing here or is it possible to merge it. I'm also interested in that patch. Thanks

hathach: I lack the motivation, since the current implementation works well and has no issue with any driver at all. Although this PR is the right implementation, however since it doesn't "fix" any problems. I am reluctant to test & merge.

cr1901: AFAIK, it's unfortunately class-specific whether a ZLP is sent if a transfer size is a multiple of the packet size.

The USB spec itself punts on the issue, saying "the two sides must have agreed on how much data is to be sent ahead of time. When that total amount of data is sent, _or any packet that is less than max packet size is sent_, the transfer ends immediately. Sending more data than agreed upon ahead of time is undefined."

This definition allows a transfer to end on a packet multiple _without_ sending a subsequent ZLP, since the two sides could've agreed ahead of time that "`64`, `128`, `192`, etc bytes will be the full transfer, we'll both stop after that". This is the behavior the [control pipe](https://github.com/hathach/tinyusb/issues/139#issuecomment-530954118) uses when the device can send the host _all_ its requested data.

I think the above definition _also_ allows the ZLP case. But I couldn't tell you which device classes do this. What does the CDC spec say on this?

majbthrd: > I lack the motivation, since the current implementation works well and has no issue with any driver at all. Although this PR is the right implementation, however since it doesn't "fix" any problems. I am reluctant to test & merge.

I thought that I provided a concise demonstration back on Jan 4 that the current implementation does not work when writing 64 characters with tud_cdc_write_char():

https://github.com/hathach/tinyusb/pull/260#issue-359242869

hathach: > > I lack the motivation, since the current implementation works well and has no issue with any driver at all. Although this PR is the right implementation, however since it doesn't "fix" any problems. I am reluctant to test & merge.
> 
> I thought that I provided a concise demonstration back on Jan 4 that the current implementation does not work when writing 64 characters with tud_cdc_write_char():
> 
> [#260 (comment)](https://github.com/hathach/tinyusb/pull/260#issue-359242869)

aha, how could I miss this. I possibly was debugging something else while reading this. I will test it out

hathach:
> A proper implementation of the CDC class should append a Zero Length Packet to transfers that are a whole multiple of the endpoint buffer.
> 
> To demonstrate the current incorrect behavior, take the device examples "webusb_serial" or "cdc_dual_ports". Take this line in main.c:
> 
> ```
>   tud_cdc_write_char(buf[i]);
> ```
> 
> and modify it to be:
> 
> ```
>   for (int j = 0; j < 32; j++)
>     tud_cdc_write_char(buf[i]);
> ```
> 
> Re-compile and download. Now, when you run your favorite terminal emulator and connect to the emulated serial port, every single keystroke you make will result in 32 copies. This is good so far.
> 
> However, change the "32" to "64", re-compile, download, and try the terminal emulator again. Depending on your OS and driver, you might get some characters out sometimes, but clearly something is wrong. I believe what is wrong is the absence of a ZLP in the CDC class driver implementation.
> 
> I believe the patch is the most expedite way to achieve proper operation. I have tested this on STM32F072, SAMD32, and NUC121 platforms.

@majbthrd late response to this, do you still remember which OS that gives the incorrect output windows/mac/linux ?

kasjer:
> > A proper implementation of the CDC class should append a Zero Length Packet to transfers that are a whole multiple of the endpoint buffer.
> > To demonstrate the current incorrect behavior, take the device examples "webusb_serial" or "cdc_dual_ports". Take this line in main.c:
> > ```
> >   tud_cdc_write_char(buf[i]);
> > ```
> > 
> > 
> > and modify it to be:
> > ```
> >   for (int j = 0; j < 32; j++)
> >     tud_cdc_write_char(buf[i]);
> > ```
> > 
> > 
> > Re-compile and download. Now, when you run your favorite terminal emulator and connect to the emulated serial port, every single keystroke you make will result in 32 copies. This is good so far.
> > However, change the "32" to "64", re-compile, download, and try the terminal emulator again. Depending on your OS and driver, you might get some characters out sometimes, but clearly something is wrong. I believe what is wrong is the absence of a ZLP in the CDC class driver implementation.
> > I believe the patch is the most expedite way to achieve proper operation. I have tested this on STM32F072, SAMD32, and NUC121 platforms.
> 
> @majbthrd late response to this, do you still remember which OS that gives the incorrect output windows/mac/linux ?

I saw this behavior on Linux and Windows. But person who informed me about strange behavior in the first place, is using Mac, so all systems are affected.

majbthrd:
> @majbthrd late response to this, do you still remember which OS that gives the incorrect output windows/mac/linux ?

I was primarily using Linux, with some Windows usage.

hathach:
> > > A proper implementation of the CDC class should append a Zero Length Packet to transfers that are a whole multiple of the endpoint buffer.
> > > To demonstrate the current incorrect behavior, take the device examples "webusb_serial" or "cdc_dual_ports". Take this line in main.c:
> > > ```
> > >   tud_cdc_write_char(buf[i]);
> > > ```
> > > 
> > > 
> > > and modify it to be:
> > > ```
> > >   for (int j = 0; j < 32; j++)
> > >     tud_cdc_write_char(buf[i]);
> > > ```
> > > 
> > > 
> > > Re-compile and download. Now, when you run your favorite terminal emulator and connect to the emulated serial port, every single keystroke you make will result in 32 copies. This is good so far.
> > > However, change the "32" to "64", re-compile, download, and try the terminal emulator again. Depending on your OS and driver, you might get some characters out sometimes, but clearly something is wrong. I believe what is wrong is the absence of a ZLP in the CDC class driver implementation.
> > > I believe the patch is the most expedite way to achieve proper operation. I have tested this on STM32F072, SAMD32, and NUC121 platforms.
> > 
> > 
> > @majbthrd late response to this, do you still remember which OS that gives the incorrect output windows/mac/linux ?
> 
> I saw this behavior on Linux and Windows. But person who informed me about strange behavior in the first place, is using Mac, so all systems are affected.

Thanks, I will check it out soon enough. Can't imagine why I missed the whole thing. I guess you should not check other PR while debugging :D

hathach:
I am feeling so good now, this PR actually fixed an issue with lots of print out with circuitpython here https://github.com/adafruit/circuitpython/issues/1639 . I should be more diligent with PR. Thanks again :)

PR #269, opened 2020-01-14
dcd_eptri: Fix rx_buffer/tx_buffer volatile annotation

mateusz-holenko:
This makes `rx_buffer` and `tx_buffer` *pointers*
volatile in order to avoid caching them in a register.

The original notation meant "a pointer to a volatile value"
(equivalent of `volatile uint8_t *`). This resulted in
`while(rx_buffer[ep_num] != NULL) ;` loop to get stuck
forever, even though the IRQ handler set the `rx_buffer[ep_num] = NULL`.


hathach:
Thanks @mateusz-holenko  for submitting PR and Thanks @xobs reviewing it :)

Issue #268, opened 2020-01-14
Nuvonton NUC121,125,126 MSC doesn't work


hathach:
**Describe the bug**
MSC class doesn't work with NUC mcu 

**Set up (please complete the following information):**
 - OS: Ubuntu 18.04
 - Board: nutiny nuc121s, 125s 126v
 - Firmware Code: examples/device/cdc_msc

**To Reproduce**
`TinyUSB MSC` disk show up but failed to read its contents

majbthrd:
I've submitted a PR to address an issue with the driver handling large OUT transfers.

However, things still go awry later on in the execution, and I'm not terribly familiar with the MSC class.

With the PR applied, in both a D21 reference and the NUC121, the USB analyzer shows the same behavior through a read LBA 0 followed by a read LBA 2.

In the D21, the next transaction is a write LBA 0.  (No IN until the STATUS after lots of OUT.)

However, with the NUC121, after the read LBA 2, the host immediately attempts to do an unexpected IN on EP 3, and since the MSC driver hasn't requested such a transfer, it gets NAKed.

It has got me scratching my head.

PR #271, opened 2020-01-15
nuc121: fix handling of OUT transfers larger than max packet size

majbthrd:
Usage with MSC example demonstrated that OUT transfers larger than the max packet size were not being correctly handled.

### DIFF
--- a/src/portable/nuvoton/nuc121/dcd_nuc121.c
+++ b/src/portable/nuvoton/nuc121/dcd_nuc121.c
@@ -395,9 +395,11 @@ void USBD_IRQHandler(void)
           xfer->remaining_bytes -= available_bytes;
           xfer->data_ptr += available_bytes;
 
-          /* when the transfer is finished, alert TinyUSB */
+          /* when the transfer is finished, alert TinyUSB; otherwise, accept more data */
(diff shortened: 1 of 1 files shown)


src/portable/nuvoton/nuc121/dcd_nuc121.c
```diff
@@ -395,9 +395,11 @@ void USBD_IRQHandler(void)
           xfer->remaining_bytes -= available_bytes;
           xfer->data_ptr += available_bytes;
 
-          /* when the transfer is finished, alert TinyUSB */
+          /* when the transfer is finished, alert TinyUSB; otherwise, accept more data */
           if ( (0 == xfer->remaining_bytes) || (available_bytes < xfer->max_packet_size) )
             dcd_event_xfer_complete(0, ep_addr, available_bytes, XFER_RESULT_SUCCESS, true);
```
hathach: the `available_bytes` in `dcd_event_xfer_complete()` should be an accumulated value so far, ideally it is the `original len - remaining_bytes`

majbthrd:
Thank you, @hathatch!  I realized my error and have added a second commit.  This now fixes issue #268.

PR #272, opened 2020-01-16
nuc121: enhanced driver and hopefully added readability

majbthrd:
It has been gnawing on me all day that there ought to be a hopefully more elegant way to the conflicting tracking requirements of IN vs OUT endpoints.  This uses an unnamed union to allow the same location in RAM to have two names, each with usage appropriate for the data direction.

### DIFF
--- a/src/portable/nuvoton/nuc121/dcd_nuc121.c
+++ b/src/portable/nuvoton/nuc121/dcd_nuc121.c
@@ -77,8 +77,11 @@ static bool active_ep0_xfer;
 /* RAM table needed to track ongoing transfers performed by dcd_edpt_xfer(), dcd_in_xfer(), and the ISR */
 static struct xfer_ctl_t
 {
-  uint8_t *data_ptr;         /* collectively, data_ptr and remaining_bytes track progress of endpoint transfers */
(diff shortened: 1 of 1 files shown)


src/portable/nuvoton/nuc121/dcd_nuc121.c
```diff
@@ -77,8 +77,11 @@ static bool active_ep0_xfer;
 /* RAM table needed to track ongoing transfers performed by dcd_edpt_xfer(), dcd_in_xfer(), and the ISR */
 static struct xfer_ctl_t
 {
-  uint8_t *data_ptr;         /* collectively, data_ptr and remaining_bytes track progress of endpoint transfers */
-  uint16_t remaining_bytes;
+  uint8_t *data_ptr;         /* data_ptr tracks where to next copy data to (for OUT) or from (for IN) */
+  union {
+    uint16_t in_remaining_bytes; /* for IN endpoints, we track how many bytes are left to transfer */
```
hathach: thanks for the PR, normally I just used the **xferred_bytes**, as long as you feel good. Any names is OK :)

PR #273, opened 2020-01-16
nuc121: survive USB bus reset

majbthrd:
Previous versions of the NUC121 driver didn't survive a USB bus reset (such as might happen when rebooting a PC or when done programmatically).


hathach:
@majbthrd there is a slight conflict, I have resolved it using github web editor. It looks good so far, once the ci complete, we could merge it :). Thank you very much for putting more effort into the nuc port :)

PR #275, opened 2020-01-17
rt10xx: correct max endpoint count is 8

hathach:
@tannewt @arturo182


arturo182:
Nice, I can test on my Feather later today!

hathach:
> Nice, I can test on my Feather later today!

Thanks for spotting the issue in the first place :)

PR #276, opened 2020-01-25
added NUC120 support (with caveats)

majbthrd:
I thought this was going to be a quick and easy addition, but it proved to be much more of a pain.

IMPORTANT: the NUC120 only has six endpoints in total, two of which get used for the control endpoints.  Given that a CDC port, for example, takes three of those, and many of the TinyUSB examples won't even run.

At least with the particular NUC120LE3 on the NuTiny-SDK-NUC120 that I bought, certain registers didn't work.  The Nuvoton errata sheet for the part shows a glowing situation of only one minor erratum, and this was not my experience.  (Pre-production part?)  CLK->CLKSTATUS always reads zero, forcing me to diagnose the problem and re-write portions of their clk.c driver.  The GPIO peripheral has registers to individually address particular pins, and that didn't work either (forcing an alternate solution in the ./hw/bsp/ board driver).

Given that Nuvoton's FAQ makes mention of one of these, perhaps the errata sheet is not kept current:

https://www.nuvoton.com/support/technical-support/faq/d5f1bbab-04f1-11ea-b113-05daf7eedf03/

So, given the choice of designing in a part, I would choose the newer NUC121/NUC125/NUC126 over the older NUC120 in a heartbeat.  The newer parts support crystal-less USB operation to boot.

There is also allegedly a NuTiny-NUC120V board, but this is not stocked by the US distributor that I ordered from.

P.S.: msc_dual_lun fails the build, but this is only because that example requires more RAM than the part has.

### DIFF
--- a/hw/bsp/nutiny_sdk_nuc120/board.mk
+++ b/hw/bsp/nutiny_sdk_nuc120/board.mk
@@ -3,6 +3,7 @@ CFLAGS += \
   -mthumb \
   -mabi=aapcs-linux \
   -mcpu=cortex-m0 \
+  -DCFG_EXAMPLE_MSC_READONLY \
   -DCFG_TUSB_MCU=OPT_MCU_NUC120
 
 # All source paths should be relative to the top level.
--- a/hw/bsp/nutiny_sdk_nuc120/board.mk
+++ b/hw/bsp/nutiny_sdk_nuc120/board.mk
@@ -1,0 +1,56 @@
+CFLAGS += \
+  -flto \
+  -mthumb \
+  -mabi=aapcs-linux \
+  -mcpu=cortex-m0 \
+  -DCFG_TUSB_MCU=OPT_MCU_NUC120
+
+# All source paths should be relative to the top level.
+LD_FILE = hw/bsp/nutiny_sdk_nuc120/nuc120_flash.ld
+
+SRC_C += \
+  hw/mcu/nuvoton/nuc100_120/Device/Nuvoton/NUC100Series/Source/system_NUC100Series.c \
+  hw/mcu/nuvoton/nuc100_120/StdDriver/src/acmp.c \
+  hw/mcu/nuvoton/nuc100_120/StdDriver/src/adc.c \
+  hw/mcu/nuvoton/nuc100_120/StdDriver/src/clk.c \
+  hw/mcu/nuvoton/nuc100_120/StdDriver/src/crc.c \
+  hw/mcu/nuvoton/nuc100_120/StdDriver/src/fmc.c \
+  hw/mcu/nuvoton/nuc100_120/StdDriver/src/gpio.c \
+  hw/mcu/nuvoton/nuc100_120/StdDriver/src/i2c.c \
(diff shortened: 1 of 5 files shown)


hathach:
Thanks @majbthrd ,  I am celebrating Lunar New Year aka TET. I will check this out asap.

hathach:
Thanks for another great port, sorry for the late response, I am just back from Lunar New Year holiday

> I thought this was going to be a quick and easy addition, but it proved to be much more of a pain.
> 
> IMPORTANT: the NUC120 only has six endpoints in total, two of which get used for the control endpoints. Given that a CDC port, for example, takes three of those, and many of the TinyUSB examples won't even run.

Six endpoints is decent as well, Atmel and ST also have 6 on some of their low cost mcu as well. User can get creative by switching configuration descriptor using switches :) 

> At least with the particular NUC120LE3 on the NuTiny-SDK-NUC120 that I bought, certain registers didn't work. The Nuvoton errata sheet for the part shows a glowing situation of only one minor erratum, and this was not my experience. (Pre-production part?) CLK->CLKSTATUS always reads zero, forcing me to diagnose the problem and re-write portions of their clk.c driver. The GPIO peripheral has registers to individually address particular pins, and that didn't work either (forcing an alternate solution in the ./hw/bsp/ board driver).
> 
> Given that Nuvoton's FAQ makes mention of one of these, perhaps the errata sheet is not kept current:
> 
> https://www.nuvoton.com/support/technical-support/faq/d5f1bbab-04f1-11ea-b113-05daf7eedf03/

That is a really troublesome to debug. I am glad you figure it out. However, in my experience, we shouldn't try to modify the official BSP, that will cause lots of maintenance effort in the future, especially upgrading the BSP. If possible, we could just copy and paste the actual code to the BSP of affected NUC or file issue on Nuvoton for them to fix it. Though If you don't have time to do that now, we can merge it as it is and open an issue to do that later as reminder. As long as It works, it is good enough to merge, we can always enhance later on :).

> So, given the choice of designing in a part, I would choose the newer NUC121/NUC125/NUC126 over the older NUC120 in a heartbeat. The newer parts support crystal-less USB operation to boot.

Yeah, if the cost is not much difference, later part is always better :).

> There is also allegedly a NuTiny-NUC120V board, but this is not stocked by the US distributor that I ordered from.

It is available from Nuvoton Direct, https://direct.nuvoton.com/en/nutiny-nuc120v they ship world wide for 6.99$ I will order one to test with this PR. Let me know if you have any other NUC board suggestion that I may need to order as well.  

> P.S.: msc_dual_lun fails the build, but this is only because that example requires more RAM than the part has.

I saw you fixed this already with `CFG_EXAMPLE_MSC_READONLY`, not really well documented on my side :)

hathach:
I am finally has the nutiny-nuc120v to test with this PR, unfortunately the nuc120v doesn't work quite well. it has usb initialized but somehow STALL the set_configuration request, I guess maybe one of the endpoint is failed to open, or it maybe specific to this board. Since it is different board to your nutiny-sdk-nuc120. Anyway, since it is not preferred in the new design where NUC121/125 is a better choice. I will merge this and just file an issue regarding this as reminder for us later on, should we decide to enhance the nuc120 port

image

PR #280, opened 2020-02-08
add Nuvoton NUC505

majbthrd:
This adds the Nuvoton NUC505, a USB 2.0 high speed device, as well as board support for the NuTiny-SDK-NUC505 (an eval board for the part).

The NUC505 has four boot mode pins, and these are provided as DIP switches on the NuTiny-SDK-NUC505.  These boot modes are mutually exclusive, so if you enable "ICE" (allow external debugger), it won't execute your code at hardware reset or power up.  (On the plus side, hopefully this means that you can recover the part with an external debugger even if the programmed image is invalid.)  There is also a "USB" boot mode that provides emulated USB Mass Storage to which you can copy a .bin image (but skip to the end for the caveat).  If you use the board much, you'll get used to repeatedly changing the DIP switches and pressing the hardware reset button.  (I know that I did!)  The bottom silkscreen of the board provides a table for the different modes.

Beyond the USB 2.0 HS capability, what makes this part interesting is that it has an internal SPI flash (up to 2MBytes) and 128kBytes of SRAM.

It memory maps the SPI flash (and does this for booting), but it is painfully slow to execute code this way.  Running a USB stack this way doesn't work with any degree of reliability.

So, the only proper solution is to copy, at boot, the code out of flash and into RAM and then branch to the RAM.  (Maddeningly, Nuvoton doesn't provide a linker script for this, despite it being utterly essential for valid part operation and emphasized in the documentation!) So, this is what the enhanced linker script in the provided bsp directory (and tweaked startup code) does.

The emulated "USB" Mass Storage boot mode ought to make the part very hobbyist friendly, as the part comes from the factory with this, allowing it to be reprogrammed easily.  Alas, it looks like they only tested it in Windows.  In all Linux distributions that I've tried, Linux panics about the FAT table not being what it expects, so it mounts the drive as read-only out of caution.  I don't have access to a MacOS device to verify whether or not it works there.

Just for completeness, below is what the Linux error messages indicate when using the part's built-in USB Mass Storage bootloader:

[ 8172.072009] usb 2-1.1: new high-speed USB device number 4 using ehci-pci
[ 8172.181791] usb 2-1.1: New USB device found, idVendor=0416, idProduct=b195
[ 8172.181796] usb 2-1.1: New USB device strings: Mfr=1, Product=2, SerialNumber=1
[ 8172.181800] usb 2-1.1: Product: MassStorage
[ 8172.181802] usb 2-1.1: Manufacturer: Nuvoton
[ 8172.181804] usb 2-1.1: SerialNumber: Nuvoton
[ 8172.216162] usb-storage 2-1.1:1.0: USB Mass Storage device detected
[ 8172.218708] scsi host4: usb-storage 2-1.1:1.0
[ 8172.218843] usbcore: registered new interface driver usb-storage
[ 8172.222228] usbcore: registered new interface driver uas
[ 8173.245190] scsi 4:0:0:0: Direct-Access     ICE-USB  Mass Storage 2.0 0    PQ: 0 ANSI: 0 CCS
[ 8173.245728] sd 4:0:0:0: Attached scsi generic sg1 type 0
[ 8173.246270] sd 4:0:0:0: [sdb] 32001 512-byte logical blocks: (16.4 MB/15.6 MiB)
[ 8173.246760] sd 4:0:0:0: [sdb] Write Protect is off
[ 8173.246764] sd 4:0:0:0: [sdb] Mode Sense: 03 00 00 00
[ 8173.247388] sd 4:0:0:0: [sdb] Write cache: enabled, read cache: enabled, doesn't support DPO or FUA
[ 8173.253890]  sdb:
[ 8173.256000] sd 4:0:0:0: [sdb] Attached SCSI removable disk
[ 8179.158013] FAT-fs (sdb): error, fat_get_cluster: invalid cluster chain (i_pos 0)
[ 8179.158020] FAT-fs (sdb): Filesystem has been set read-only


hathach:
@majbthrd thank you very much for yet another awesome work, and detailed instruction for using nutiny  NUC505, it seems to require a bit more work than other NUC. I am very thankful, but currently is on tied schedule with other works. Please give me a few more days to do a full testing and review.

hathach:
@majbthrd Sorry for the huge delay so far, thank you very much for huge amount of work on this PR. I feel bad keeping you waiting. I am just able to pull out the board to test with, as you mentioned the board is a bit difficult to work with. Its flashing speed is surprisingly slow with openocd (required SW to set to ICE+SPI)

```
Info : Nuvoton NuMicro: Flash Write ...
Info : Have written 100%
wrote 7916 bytes from file _build/build-nutiny_sdk_nuc505/nutiny_sdk_nuc505-firmware.elf in 106.501045s (0.073 KiB/s)
```

Afterwards, I change DIP switch to internal SPI and reset + power off/on but the code doesn't seem to run. LED at PC3 doesn't blink with board_test example, as well as no USB detected on cdc_msc. Maybe I did something wrong, I will try to test again tomorrow.

majbthrd:
@hathach, there's no reason to feel bad.  Your considerable TinyUSB efforts are greatly appreciated by many.

The three most important DIP switch modes are:

1111 "SPI" (run from internal flash)
1110 "USB" (mass storage emulator that accepts a .bin file)
0111 "ICE-SPI" (allow external debugger access)

I would encourage you to start by trying the "USB" mode.  At least seeing a mass storage device appear (even if it is read-only) would confirm that the components are functioning well enough to work as a USB device.  It also helps to confirm the correct orientation of the DIP switches.

The slow flash write speed sounds odd, although I must admit that I broke off the Nuvoton Nu-Link-Me debugger as soon as I got the boards.  I wouldn't dismiss the possibility that their openocd fork had issues; their Nu-Link-Me firmware may also not be up-to-date, and their Windows programming software is the only method that I'm aware of that will recognize and update the Nu-Link-Me firmware.

The simplest method to update firmware is by copying the .bin files to the built-in "USB" mode MSC bootloader.

In my own development, I use Crossworks for ARM with a Segger J-Link, combined with my own custom flash loader (95kB/s).  Neither Rowley nor Segger has a flash loader for the NUC505; the code and binaries for my Rowley solution are available in my "MCUmisfits" project.

This evening, I tried both my NUC505 boards.  Whether I wrote images by using an external debugger in "ICE-SPI" mode or by copying .bin files on a Windows host in "USB" mode, when I switched to "SPI" mode and did a hardware reset (or unplugged and plugged the device), it ran the "board_test" and "cdc_msc" code without issue (including LED blinking).... and I went back and forth between these several times to confirm that I wasn't simply running a holdover image from earlier development.

I've emailed links to precompiled images via email, but these should be functionally the same as anything you build locally.

hathach:
Thank you @majbthrd , I will give it a try again. I prefer sticking with the on-board nulink-me since it is what most people will use to flash it when using the board. Maybe the openocd for nuc505 is not correct yet. Hopefully, we can close this one soon enough.

majbthrd:
@hathach, here's a workaround for using the "USB" boot mode bootloader with a Linux machine.

Attach the NUC505 target in its "USB" boot mode.  A 16MB storage device should appear.

Use your favorite method to determine what the device name of this device is, e.g. /dev/sdX.  The mount command would be one example.

Be CERTAIN that you have correctly identified the device name; if you misidentify it, the next command will corrupt an unintended drive.

Given a NUC505 that appears as "/dev/sdX" (replace with the name of your device) and a firmware image named firmware.bin (replace with the name of your firmware image):

``
sudo dd if=firmware.bin of=/dev/sdX seek=384
``

It works because it effectively writes a file to the region after the emulated FAT table, and that is all that such MSC bootloaders care about.

hathach:
@majbthrd thank you for your helpful instruction to work with NUC505. I am finally starting to understand it a bit more. The reason why I cannot flash it using openocd is because NUC505 is in ICE mode `0111`, which seems to be Emulation only. To be able to program the Internal SPI flash, it need to boot with ICP mode `1011`. However, in ICP mode, opencod cannot establish connection to the mcu. Therefore, i guess there is no easy command line flash for NUC505. The `dd` command is indeed too dangerous to be included in makefile as target. 

Luckily fire Windows VM could easily solve this problem,  I could just forward nulink and click couple of more buttons. Finally, I am able to work with the actual code !!!
 
image

hw/bsp/nutiny_sdk_nuc505/nutiny_sdk_nuc505.c
```diff
@@ -0,0 +1,118 @@
...
+//--------------------------------------------------------------------+
+// Board porting API
+//--------------------------------------------------------------------+
+
+void board_led_write(bool state)
+{
+  uint32_t current = (state) ? LED_STATE_ON : (1-LED_STATE_ON);
+  current <<= LED_PIN;
+  uint32_t irq_state = __get_PRIMASK();
+  __disable_irq();
```
hathach: May I ask why we need to disable irq and to write to GPIO. Sorry if the question is so obvious with this mcu

pigrew: The LED is being changed via a read-modify-write cycle. This would cause a concurrency issue if the task were switched out to another task (or interrupt) that also writes to this IO port. It looks like the NUC505's GPIO doesn't have a 'bit set' or 'bit reset' register, so this is required.

hathach: @pigrew wow, that is new to me, thank you very much for the detailed explanation. NUC505 is definitely a bit more difficult to work with :D. I could only imagine to use this mcu bit banging an protocol like i2c :face_with_head_bandage:

PR #287, opened 2020-03-03
add CDC-ECM/RNDIS/CDC-EEM network device class with example

majbthrd:
Given a USB microcontroller with enough RAM (tens of kilobytes), this acts like a USB Ethernet adapter connected to a virtual network.

My past experience has been that people look down their noses at a USB network class implementation unless it also just happens to have a complete TCP/IP implementation, so the example has an embedded DHCP server, DNS server, and HTTP (web) server (courtesy of lwIP).

Since there are three different specifications for a USB network adapter floating out there, I wound up implementing all three just for good measure! :)

RNDIS worked in my testing with Linux and Windows 7 / Windows 10.  CDC-ECM worked with Linux and ought to with MacOS (although I lack a Mac to test with).  As far as I'm aware, CDC-EEM never caught on and is just a historical curiosity, but it seems to function in Linux.

The Atmel/Microchip SAMD21 and Nuvoton NUC126 both have sufficient resources (16-20kBytes of RAM and 40+ kBytes of flash).  I was very happy with operation of all three implementations variants on both of these MCUs.

I also tried the ST STM32F072, and this didn't work so well.  I'm not sure if it is case of exercising the driver in a new way or using the TinyUSB API in a manner that it wasn't intended but which the SAMD21 and NUC126 drivers just happened to accept.

### DIFF
--- a/lib/networking/rndis_protocol.h
+++ b/lib/networking/rndis_protocol.h
@@ -1,0 +1,307 @@
+/**
+ * \file rndis_protocol.h
+ *         RNDIS Defines
+ *
+ * \author
+ *         Colin O'Flynn <coflynn@newae.com>
+ *
+ * \addtogroup usbstick
+ */
+
+/* Copyright (c) 2008  Colin O'Flynn
+
+   Redistribution and use in source and binary forms, with or without
+   modification, are permitted provided that the following conditions are met:
+
+   * Redistributions of source code must retain the above copyright
+     notice, this list of conditions and the following disclaimer.
+   * Redistributions in binary form must reproduce the above copyright
+     notice, this list of conditions and the following disclaimer in
+     the documentation and/or other materials provided with the
+     distribution.
+   * Neither the name of the copyright holders nor the names of
+     contributors may be used to endorse or promote products derived
+     from this software without specific prior written permission.
+
+  THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+  AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+  IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
+  ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT OWNER OR CONTRIBUTORS BE
+  LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
+  CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+  SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+  INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+  CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+  ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+  POSSIBILITY OF SUCH DAMAGE.
+*/
+
+#ifndef _RNDIS_H
+#define _RNDIS_H
+
+/** 
+  \addtogroup RNDIS
+  @{
+  */
+
+#include <stdint.h>
+
+#define RNDIS_MAJOR_VERSION	1
+#define RNDIS_MINOR_VERSION 0
+
+#define RNDIS_STATUS_SUCCESS            0X00000000
+#define RNDIS_STATUS_FAILURE            0XC0000001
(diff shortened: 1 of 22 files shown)


src/device/usbd.c
```diff
@@ -181,6 +181,24 @@ static usbd_class_driver_t const _usbd_driver[] =
...
   },
   #endif
+
+  #if CFG_TUD_NET
+  {
+      .class_code       = 
+#if CFG_TUD_NET == OPT_NET_RNDIS
+                          TUD_RNDIS_ITF_CLASS,
+#else
+                          TUSB_CLASS_CDC,
```
hathach: I definitely need to add subclass here, to allow using both ACM and ECM at the same time :). Dont't worry it is my homework :)

hathach:
Superb, awesome and amazing works, this is an excited class driver addition. In my previous experience working with RNDIS (I tried to add host driver), this is very hard to pull off since it require knowledge of both USB and network stack. That is very impressive that you could glue TinyUSB and lwip together :) 

As far as I know RNDIS is the most common protocol, even Android use RNDIS when enable cellular tethering. This will open lots of great network related application. Don't worry about STM32, once it works with samd and nuc, it is only minor issue somewhere in the codebase :). I will help to fix it later on :)

hathach:
work on the first try with samd21 samd51, also failed with stm32f4, due to the lacking response to `REMOTE_NDIS_INITIALIZE_MSG`, nrf52840 stall other request. It is possibly specific to those dcd driver port. We will trouble shoot it later :). This is already awesome work to get it running already, bravo !!!!

image

src/class/net/net_device.h
```diff
@@ -0,0 +1,84 @@
...
+#ifdef __cplusplus
+ extern "C" {
+#endif
+
+//--------------------------------------------------------------------+
+// Application API
+//--------------------------------------------------------------------+
+
+// client must provide this: initialize any network state back to the beginning
+void network_init_callback(void);
```
hathach: For convention purpose, public API of TinyUSB Device should be prefix with `tud_`, and callback is suffix with `_cb` if possible please change these to
- tud_network_init_cb()
- tud_network_recv_cb()
- tud_network_recv_renew()
- tud_network_can_xmit()
- tud_network_xmit()

If you don't, don't worry, I will do it later on

lib/networking/rndis_protocol.h
```diff
@@ -0,0 +1,307 @@
+/**
+ * \file rndis_protocol.h
```
hathach: these rndis protocol parser is very handy, maybe we should include them into the stack as well. Though I don't have enough knowledge at the moment to decide. We could always move files around later on.

majbthrd:
Thanks!  I've added a commit with the requested naming changes.

On customizing the web server content, there is a Perl script in ./lib/lwip/src/apps/http/makefsdata that can inhale a given subdirectory's file system and provide a "fsdata.c".

If that resultant fsdata.c is renamed to fsdata_custom.c, placed in ./examples/device/lwip_webserver/src/, and the Makefile is modified so that CUSTOM_FS_DATA=1 (rather than CUSTOM_FS_DATA=0), the next clean build will use that in preference to the default web content.

Also, as far as use possibilities go, there can be "CGI" interaction between the brower and MCU via the http server.  My [D21ecm](https://github.com/majbthrd/D21ecm), [D21rndis](https://github.com/majbthrd/D21rndis), and [D21eem](https://github.com/majbthrd/D21eem) demonstrate this (albeit at the cost of using even more flash).  Credit goes to [lrndis](https://github.com/fetisov/lrndis) for originally showing what is possible with lwIP; the lrndis web code actually has a picture of a STM32F4 Discovery board and when the user clicks in the browser, the corresponding LEDs turn on and off on the Discovery board.

hathach:
> Thanks! I've added a commit with the requested naming changes.
Yeah thank you very much

> On customizing the web server content, there is a Perl script in ./lib/lwip/src/apps/http/makefsdata that can inhale a given subdirectory's file system and provide a "fsdata.c".
> 
> If that resultant fsdata.c is renamed to fsdata_custom.c, placed in ./examples/device/lwip_webserver/src/, and the Makefile is modified so that CUSTOM_FS_DATA=1 (rather than CUSTOM_FS_DATA=0), the next clean build will use that in preference to the default web content.

Great, I will try to learn more about rndis and customize the page later on, when checking the example with all other ports (hopefully more will work than won't :D )

> Also, as far as use possibilities go, there can be "CGI" interaction between the brower and MCU via the http server. My [D21ecm](https://github.com/majbthrd/D21ecm), [D21rndis](https://github.com/majbthrd/D21rndis), and [D21eem](https://github.com/majbthrd/D21eem) demonstrate this (albeit at the cost of using even more flash). Credit goes to [lrndis](https://github.com/fetisov/lrndis) for originally showing what is possible with lwIP; the lrndis web code actually has a picture of a STM32F4 Discovery board and when the user clicks in the browser, the corresponding LEDs turn on and off on the Discovery board.

Yeah, it is very helpful, I did some similar webserver long time ago with Broadcom WICED using ajax to interact with device. We could definitely add button + led interaction in the webserver as well. Thank for great info.

hathach:
Merged !!! Great work, thank you very much again for this amazing PR.

PR #290, opened 2020-03-05
usbnet RNDIS correction (STM32 RNDIS now valid)

majbthrd:
rndis_report[] resized from 128 bytes to 120 bytes  With hindsight, I believe this is a more correct size for the RNDIS protocol.  As a pleasant side-effect, the driver doesn't need to do a ZLP on EP0.

Added testing successes:

STM32F072 RNDIS

Existing platforms verified as still valid:

SAMD21 CDC-ECM
SAMD21 RNDIS
SAMD21 CDC-EEM
NUC126 CDC-EDM
NUC126 RNDIS
NUC126 CDC-EEM
STM32 CDC-EEM

### DIFF
--- a/src/class/net/net_device.c
+++ b/src/class/net/net_device.c
@@ -198,7 +198,17 @@ bool netd_control_request(uint8_t rhport, tusb_control_request_t const * request
   TU_VERIFY (_netd_itf.itf_num == request->wIndex);
 
 #if CFG_TUD_NET == OPT_NET_RNDIS
-  tud_control_xfer(rhport, request, rndis_buf, sizeof(rndis_buf));
+  if (request->bmRequestType_bit.direction == TUSB_DIR_IN)
+  {
+    rndis_generic_msg_t *rndis_msg = (rndis_generic_msg_t *)rndis_buf;
+    uint32_t msglen = tu_le32toh(rndis_msg->MessageLength);
+    TU_ASSERT(msglen <= sizeof(rndis_buf));
+    tud_control_xfer(rhport, request, rndis_buf, msglen);
+  }
+  else
+  {
+    tud_control_xfer(rhport, request, rndis_buf, sizeof(rndis_buf));
+  }
 #else
   (void)rhport;
 #endif
--- a/src/class/net/net_device.c
+++ b/src/class/net/net_device.c
@@ -64,7 +64,7 @@ CFG_TUSB_MEM_SECTION CFG_TUSB_MEM_ALIGN static uint8_t received[CFG_TUD_NET_PACK
 CFG_TUSB_MEM_SECTION CFG_TUSB_MEM_ALIGN static uint8_t transmitted[CFG_TUD_NET_PACKET_PREFIX_LEN + CFG_TUD_NET_MTU + CFG_TUD_NET_PACKET_PREFIX_LEN];
 
 #if CFG_TUD_NET == OPT_NET_RNDIS
-  CFG_TUSB_MEM_SECTION CFG_TUSB_MEM_ALIGN static uint8_t rndis_buf[128];
+  CFG_TUSB_MEM_SECTION CFG_TUSB_MEM_ALIGN static uint8_t rndis_buf[120];
 #endif
 
 //--------------------------------------------------------------------+


hathach:
I have retested and It works great with nrf52840. Look like ZLP response is the root cause of this issue. However, I would like to make usbd fully compliant to response with ZLP as well rather than reply on shortened transfer even if 120 is more right than 128. Therefore I suggest we keep this PR open, and try to fix the ZLP first with usbd. Then merge later, it is good chance, since we don't generally got reproducible issue this easy.

hathach:
PS: STM32F4 still doens't work, it is highly due to its own issue I guess.

hathach:
@cr1901 no worries, I will troubleshoot it when I have time :)

pigrew:
While this patch does make things work, I believe that it's more correct to only send back the message bytes in `net_device.c:netd_control_request`? The disadvantage is that the code is more complicated.

```
// Handle class control request
bool netd_control_request(uint8_t rhport, tusb_control_request_t const * request)
{
  // Handle class request only
  TU_VERIFY(request->bmRequestType_bit.type == TUSB_REQ_TYPE_CLASS);

  TU_VERIFY (_netd_itf.itf_num == request->wIndex);
#if CFG_TUD_NET == OPT_NET_RNDIS
  if (request->bmRequestType_bit.direction == TUSB_DIR_IN)
  {
    uint32_t msglen = tu_le32toh(*((uint32_t*)(rndis_buf+4)));
    TU_ASSERT(msglen < sizeof(rndis_buf));
    tud_control_xfer(rhport, request, rndis_buf, msglen);
  }
  else
  {
    tud_control_xfer(rhport, request, rndis_buf, sizeof(rndis_buf));
  }
#else
  (void)rhport;
#endif

  return true;
}
```

majbthrd:
@pigrew, I like it!  I made two small additional changes: one was to tweak the TU_ASSERT to use "<=" rather than "<" to allow the full buffer to be used, and the other was to leverage an existing struct in the include files that makes it more apparent to the user why it is peering an offset into the message:

```
#if CFG_TUD_NET == OPT_NET_RNDIS
  if (request->bmRequestType_bit.direction == TUSB_DIR_IN)
  {
    rndis_generic_msg_t *rndis_msg = (rndis_generic_msg_t *)rndis_buf;
    uint32_t msglen = tu_le32toh(rndis_msg->MessageLength);
    TU_ASSERT(msglen <= sizeof(rndis_buf));
    tud_control_xfer(rhport, request, rndis_buf, msglen);
  }
  else
  {
    tud_control_xfer(rhport, request, rndis_buf, sizeof(rndis_buf));
  }
#else
  (void)rhport;
#endif
```

pigrew:
@majbthrd 

Looks good, I didn't look at the code well enough to notice the structs.... I didn't test it so much (while I've been trying to debug the stm32 issues, which seem to be related to enforcing an expectation of DAT1 for the status stage.).

majbthrd:
> PS: STM32F4 still doens't work, it is highly due to its own issue I guess.

The fingerprint seems to be the USB descriptor length; both cdc_dual_ports and net_lwip_webserver are affected.  #299

pigrew:
@majbthrd : Could you update this PR with the extra changes? Or should I create a PR with the updated code?

On stm32f0, I found that removing three driver lines seems to (mostly) fix the driver. It's now able to provide DHCP to MS Windows (the computer gets an IP adress), but I'm unable to connect to the webserver.

https://github.com/hathach/tinyusb/blob/a0eb8a26517040850a0e4185aef398183b04ddc7/src/portable/st/stm32_fsdev/dcd_stm32_fsdev.c#L399-L402

However, I think there are still some issues with the driver. I started rewriting the transfer handling code (greatly simplifying it). I had originally based it on the ST Micro drivers, but I must have broken it in translation. We could immediately remove the bad three lines or commit my current revised driver, but I'm leaning towards waiting for a more complete fix (hopefully after we have a solution to #300 which I found in the process of debugging).

majbthrd:
@pigrew : the PR has been updated.  Thanks!

pigrew:
@hathach:

I'm happy for this PR to be merged (perhaps after you verify the webserver still works on the nrf52840?), as I've been able to identify the code at fault in stm32fsdev. However, even with this PR applied, it doesn't work right on my F070...  I get DHCP, but no DNS or webserver. I maybe should try a different computer...

I do have a partial fix for the stm32fsdev driver (deleting the three lines mentioned above). Let me know if you want me to submit that as a PR (or you can...). I think I'll need a week or two to get a more complete fix. I've been extending my USBTMC python test scripts to better detect the driver failures (and exercise various length control transfers). We may want to create a new class (like libopencm3's gadget0) to do more full stack testing. Because libopencm3 is LGPL licensed, I don't think we can directly use their test code.

hathach:
@pigrew @majbthrd great discussion and analysis so far. I learned a lot from you guys. There is no rush to submit temp PR just for the sake of testing. However, If it is possible, you could create an [WIP] PR for stm32fsdev when you feel it is relatively good enough. So that I and other could also participate in the progress of troubleshooting as well, should you need my help. But that is totally up to you.

For USBTMC python test, yeah it is a great idea, we can put it into `test/`currently, there is only ceedling unit-test suite there, however it is only running on host PC to simulate DCD driver to test the usbd/class driver behavior. There is no test to run on the actual controller just yet. We can move the current ceedling test into a subfolder `test/unittest` and have another subfolder for actual controller testing. That would help immensely for testing.

PR #301, opened 2020-03-15
NUC505 : fix operation with net_lwip_webserver

majbthrd:
Three changes were made to allow the net_lwip_webserver example to work with the Nuvoton NUC505 (in all CDC-ECM, RNDIS, and CDC-EEM variants):

The TinyUSB convention is to have CFG_TUSB_RHPORT0_MODE defined in each and every example's tusb_config.h.  This means that every single example must be updated to conditionally add OPT_MODE_HIGH_SPEED whenever a new USB high-speed device driver is added to TinyUSB.  The NUC505 PR was submitted before the net PR, so the net_lwip_webserver example  managed to escape this update.

There was a bug in ./src/portable/nuvoton/nuc505/dcd_nuc505.c in the way that it handled EP0 OUT transfers; it wasn't issuing a callback to TU, and this resulted in gummed-up class requests.

Lastly, I disabled DMA in the NUC505 driver.  It was a flaky USB-specific peripheral from the get-go, but in testing net_lwip_webserver, I found that it wouldn't finish transfers larger than 64 bytes than were not a multiple of 4 bytes.  This didn't present a problem for MSC (always 512 bytes), but is very much an issue for network packets that are by design not a fixed size.  Even if the driver were re-written to do the last 0-3 bytes manually, the DMA peripheral didn't leave the endpoint buffers in a fit state to allow this.  Nuvoton's own example code doesn't use DMA, and I'm realizing there is a good reason for that.


src/portable/nuvoton/nuc505/dcd_nuc505.c
```diff
@@ -465,6 +467,7 @@ void USBD_IRQHandler(void)
 
     if (bus_state & USBD_BUSINTSTS_DMADONEIF_Msk)
     {
+#if 0
```
hathach: We should either remove these DMA code completely if DMA issue is not solvable, or defined an macro e.g `USE_DMA = 0`, with a bit of description  in which situation it doesn't work, like you put in the PR description. It will help us a lot after a while looking back at the DMA issue.

examples/device/net_lwip_webserver/src/tusb_config.h
```diff
@@ -39,7 +39,7 @@
   #error CFG_TUSB_MCU must be defined
 #endif
 
-#if CFG_TUSB_MCU == OPT_MCU_LPC43XX || CFG_TUSB_MCU == OPT_MCU_LPC18XX || CFG_TUSB_MCU == OPT_MCU_MIMXRT10XX
+#if CFG_TUSB_MCU == OPT_MCU_LPC43XX || CFG_TUSB_MCU == OPT_MCU_LPC18XX || CFG_TUSB_MCU == OPT_MCU_MIMXRT10XX || CFG_TUSB_MCU == OPT_MCU_NUC505
```
hathach: the highspeed opt is indeed not a ideal solution, though some vendor has dual ports like lpc18xx, which only has a builtin HS PHY for 1 port, the other port is only has FS PHY. I will try a better way to describe the port number/speed selection later on.

majbthrd:
> PS: it seems that 5005 and 100 series is not really good design ;D. The later series like 121/125 and hopefully other are much better to work with :).

@hathach, I don't think there is anything wrong per se with these chips.  Software programmers have the luxury of being able to hide their mistakes in updates; those of us who do hardware, however, must live with our errata forever.*  We are all human and can make mistakes, and all chips tend to have some errata (documented or not).

In this case, it is just that the level of effort needed to try to devise a workaround for the NUC505 DMA hardware limitations (and the added complexity of the workaround) is more onerous than the perceived benefit.

*Herbert Hoover (former US president and engineer) said: "The great liability of the engineer compared to men of other professions is that his works are out in the open where all can see them. His acts, step by step, are in hard substance. He cannot bury his mistakes in the grave like the doctors. He cannot argue them into thin air or blame the judge like the lawyers. He cannot, like the politicians, screen his shortcomings by blaming his opponents and hope the people will forget. The engineer simply cannot deny he did it. If his works do not work, he is damned."

hathach:
> > PS: it seems that 5005 and 100 series is not really good design ;D. The later series like 121/125 and hopefully other are much better to work with :).
> 
> @hathach, I don't think there is anything wrong per se with these chips. Software programmers have the luxury of being able to hide their mistakes in updates; those of us who do hardware, however, must live with our errata forever.* We are all human and can make mistakes, and all chips tend to have some errata (documented or not).
> 
> In this case, it is just that the level of effort needed to try to devise a workaround for the NUC505 DMA hardware limitations (and the added complexity of the workaround) is more onerous than the perceived benefit.
> 
> *Herbert Hoover (former US president and engineer) said: "The great liability of the engineer compared to men of other professions is that his works are out in the open where all can see them. His acts, step by step, are in hard substance. He cannot bury his mistakes in the grave like the doctors. He cannot argue them into thin air or blame the judge like the lawyers. He cannot, like the politicians, screen his shortcomings by blaming his opponents and hope the people will forget. The engineer simply cannot deny he did it. If his works do not work, he is damned."

so true :+1: those are such wise words. I am not familiar enough with the NUC series, and found it more difficult to work with. However, they have their own advantage as well :)

PR #304, opened 2020-03-22
STM32FSDEV: Rewrite transfer ISR

pigrew:
This rewrites the transfer ISR code for STM32FSDEV. The TX logic is unchanged, but the RX logic is greatly simplified.

There were multiple issues, particularity with EP0:
* DTOG was set wrongly.
* EP0 must always be ready to receive status packets, when CTR flag low.
* Use same logic for short packets for all transfers (both control and non-control).

While I cannot find any issues with the code in particular, I'm still unable to connect to the RNDIS webserver (I'm using Windows 10).

I've run my USBTMC test routines, and there seems to be no regressions in it, but I'd appreciate additional testing.

### DIFF
--- a/hw/bsp/stm32f070rbnucleo/stm32f070rbnucleo.c
+++ b/hw/bsp/stm32f070rbnucleo/stm32f070rbnucleo.c
@@ -185,10 +185,7 @@ void HardFault_Handler (void)
   */
 void assert_failed(char *file, uint32_t line)
 { 
-  /* USER CODE BEGIN 6 */
-  /* User can add his own implementation to report the file name and line number,
-     tex: printf("Wrong parameters value: file %s on line %d\r\n", file, line) */
-  /* USER CODE END 6 */
+  TU_LOG1("Assertion failed (%s:%ld)\r\n", file, line);
 }
 #endif /* USE_FULL_ASSERT */
 
--- a/src/portable/st/stm32_fsdev/dcd_stm32_fsdev.c
+++ b/src/portable/st/stm32_fsdev/dcd_stm32_fsdev.c
@@ -184,7 +184,7 @@ static void dcd_handle_bus_reset(void);
 static bool dcd_write_packet_memory(uint16_t dst, const void *__restrict src, size_t wNBytes);
 static bool dcd_read_packet_memory(void *__restrict dst, uint16_t src, size_t wNBytes);
 static void dcd_transmit_packet(xfer_ctl_t * xfer, uint16_t ep_ix);
-static uint16_t dcd_ep_ctr_handler(void);
+static void dcd_ep_ctr_handler(void);
 
 
 // Using a function due to better type checks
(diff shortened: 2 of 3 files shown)


majbthrd:
@pigrew, I can confirm the issue with Windows 10 (and 7) using a STM32F072-DISCO, and this was also present before the PR.  The behavior that the PC gets a DHCP lease but is unable to ping the USB device nor access the web server.  This problem happens with the TinyUSB build and ARM's 2019-q4 gcc, but doesn't when using a different tool chain (Rowley Crossworks with the same ARM gcc 2019-q4).  Stack sizes in the linker scripts are the same.  Operation under Linux is not affected.  I have Beagle USB logs to compare between the two, but there isn't a clear explanation as to the problem.

pigrew:
@majbthrd : Thanks for the checking. Being toolchain related, this seems like an optimization related thing. In fact, I just tried with `-O1` optimization in GCC, and it works fine.  We should be able to enable optimization file-by-file, to get a better idea of the source of the issue, though I don't know an efficient way to do that....

pigrew:
It's looking like it's optimization of usb_descriptors.c which breaks things... which is odd.

majbthrd:
Adding "__attribute__((optimize("O0")))" to tud_descriptor_string_cb() will result in it functioning, but "O1" won't.  This feels like one of those problems where there is sensitivity of memory placement for a particular object, and usb_descriptor.c happens to be placed prior in memory to whatever object is sensitive.

pigrew:
I just also realized that LTO is enabled. Disabling it (`hw/bsp/stm32f070rbnucleo/board.mk`) also made the compile work. There is also a compile flag used to disable warnings about "variable-size enums" mismatched between the pre-compiled libnosys.a & libc_nano.a and the LTO-enabled source files.

Then again, it could also easily be a coincidence that disabling LTO makes it work.

hathach:
I could confirm this PR make stm32f072disco rndis work great with Linux, but have issue with windows 10. Although windows recognize stm32 as rndis device (showing "the driver is working properly") but couldn't connect to lwip webserver.  

also verify that adding __attribute__((optimize("O0"))) to `tud_descriptor_string_cb()` get the webserver work :thinking: :thinking: 

Screenshot from 2020-03-23 11-37-36

hathach:
this PR is perfect, the issue is in rndis get mac address response. Which is fixed by #305. 
The optimization option for **tud_descriptor_string_cb** accidentally allow the existence of tud_network_mac_address. GCC LTO is very smart to detect that tud_network_mac_address isn't used since the OID_802_3_CURRENT_ADDRESS, OID_802_3_PERMANENT_ADDRESS only use the pointer address only.

Again, Linux driver is more forgiving on this matter than Windows :D. Once 305 is merged, we will sync this with master for double check. If everything is OK, we could merge this one :)

pigrew:
If it looks good, I'm happy for you to merge this into master. I believe it moves things in the correct direction.

hathach:
@pigrew @majbthrd merged !!! everything is on track now, thank you very much for your enhancing work to the dcd. I really enjoy collaborating with awesome developers as you

PR #305, opened 2020-03-23
fix mac address for rndis response

hathach:
- remove cflags -fshort-enums
- clean up webserver descriptor

this PR will fix the issue with optimization option with tud_descriptor_string_cb() for #304 . The option accidentally allow the existence of `tud_network_mac_address`. GCC is very smart to detect that tud_network_mac_address isn't used since the OID_802_3_CURRENT_ADDRESS, OID_802_3_PERMANENT_ADDRESS only use the pointer address only.

Again, Linux driver is more forgiving on this matter than Windows :D


lib/networking/rndis_reports.c
```diff
@@ -120,8 +120,8 @@ static void rndis_query(void)
   {
     case OID_GEN_SUPPORTED_LIST:         rndis_query_cmplt(RNDIS_STATUS_SUCCESS, OIDSupportedList, 4 * OID_LIST_LENGTH); return;
     case OID_GEN_VENDOR_DRIVER_VERSION:  rndis_query_cmplt32(RNDIS_STATUS_SUCCESS, 0x00001000);  return;
-    case OID_802_3_CURRENT_ADDRESS:      rndis_query_cmplt(RNDIS_STATUS_SUCCESS, &station_hwaddr, 6); return;
-    case OID_802_3_PERMANENT_ADDRESS:    rndis_query_cmplt(RNDIS_STATUS_SUCCESS, &permanent_hwaddr, 6); return;
+    case OID_802_3_CURRENT_ADDRESS:      rndis_query_cmplt(RNDIS_STATUS_SUCCESS, station_hwaddr, 6); return;
```
hathach: LTO is super smart to know that mac address isn't used at all.

examples/device/net_lwip_webserver/src/usb_descriptors.c
```diff
@@ -155,15 +155,16 @@ static uint16_t _desc_str[32];
 // Application return pointer to descriptor, whose contents must exist long enough for transfer to complete
 uint16_t const* tud_descriptor_string_cb(uint8_t index, uint16_t langid)
 {
-  (void)langid;
+  (void) langid;
 
-  unsigned chr_count = 0;
```
pigrew: I'm not a fan of uint8_t for counter variables, since it may require extra instructions to truncate the register values to 8-bit (since the result of all arithmetic is 32-bit). But, in this case, this change doesn't make any difference in the optimized firmware image (since the compiler already knows that the variable is limited to small numbers, I think it'll use 32-bit, anyway).

I vote to leave this as unsigned, but no functional issue with either choice (since they compile to the same thing).

examples/device/net_lwip_webserver/src/usb_descriptors.c
```diff
@@ -155,15 +155,16 @@ static uint16_t _desc_str[32];
...
+  uint8_t chr_count = 0;
 
   if (STR_LANGID == index)
   {
     memcpy(&_desc_str[1], string_desc_arr[0], 2);
     chr_count = 1;
   }
+#if CFG_TUD_NET == OPT_NET_ECM
   else if (STR_MAC == index)
```
pigrew: Maybe also add an ifdef for STR_MAC on line 87 in the enum definition?

lib/networking/rndis_reports.c
pigrew: Those ampersands are so sneaky.

examples/rules.mk
```diff
@@ -30,7 +30,7 @@ INC += $(TOP)/src
 
 #
 CFLAGS += $(addprefix -I,$(INC))
-LDFLAGS += $(CFLAGS) -fshort-enums -Wl,-T,$(TOP)/$(LD_FILE) -Wl,-Map=$@.map -Wl,-cref -Wl,-gc-sections -specs=nosys.specs -specs=nano.specs
```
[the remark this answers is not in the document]
hathach: that is so confusing, I will do a quick commit history check to see if it is there on purpose.

examples/device/net_lwip_webserver/src/usb_descriptors.c
majbthrd: Amen.  I'm also not a fan of uint8_t for all the same reasons.

hathach: I do prefer the fixed size integer since we don't know which MCU are running. Yesterday I attempted to compile lwip_webserver with MSP430 (16-bit), almost get it running. But lwip throw out too many integer overflow warning e.g `1 << 24` so I have to skip it. Though since it is  uint8_t, it make no difference, we can keep it as it is if you like it that way, though I would suggested to put it `unsigned int` as I like to be explicit although it is the same.

examples/device/net_lwip_webserver/src/usb_descriptors.c
hathach: yes, it is a better. Though, thanks to this we are able to catch the `&` issue :)

lib/networking/rndis_reports.c
hathach: sneaky sneaky :snake:

examples/device/net_lwip_webserver/src/usb_descriptors.c
majbthrd: This has uncovered a coding goof on my part.  My intent was there, but I never finished improving the code.  I do NOT like 'magic numbers' in code, and my intent was for the enum to replace these 'magic numbers'; it makes things self-documenting.  The enum definition should have gone towards the top of the code, and then it would get used in "desc_device".  .iSerialNumber would get assigned STR_MAC (not the present erroneous 3 == STR_ITFNAME).  With that corrected, there is no point in the #if in tud_descriptor_string_cb(), as it is always used.  Also, "string_desc_arr[0]" should become "string_desc_arr[STR_LANGID]".

I'm not a git guru; is there a way for me to submit these changes to this PR, or will the above verbiage have to suffice?  Thanks!

examples/device/net_lwip_webserver/src/usb_descriptors.c
pigrew: Hmm. maybe `unsigned int`, then. It'll be 16-bit on MSP430, and 32-bit elsewhere, which should be fine.

examples/rules.mk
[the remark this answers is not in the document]
pigrew: It looks like the ARM standard might be no-short-enums, but gcc uses short-enums. I'm not sure why the warnings were popping up sometimes, but not other times.

examples/device/net_lwip_webserver/src/usb_descriptors.c
```diff
@@ -155,15 +155,16 @@ static uint16_t _desc_str[32];
...
 
-  unsigned chr_count = 0;
+  uint8_t chr_count = 0;
 
   if (STR_LANGID == index)
   {
     memcpy(&_desc_str[1], string_desc_arr[0], 2);
     chr_count = 1;
   }
+#if CFG_TUD_NET == OPT_NET_ECM
```
majbthrd: This code should not be #if conditional.  It only seems to be prudent because I failed to do this in desc_device:

```
    .iManufacturer      = STR_MANUFACTURER,
    .iProduct           = STR_PRODUCT,
    .iSerialNumber      = STR_MAC,
```

examples/device/net_lwip_webserver/src/usb_descriptors.c
hathach: @majbthrd no hurry, once this is merged, you could just do another PR. It is much easier that way. I am not good with git as well :) . Currently in the middle of working on another branch, will switch back and finish this asap.

pigrew: No great way for you to write to this PR.... It'd require that @hathach give you write permissions to his PR branch. You could, however, branch off of this PR branch, and link to that or create a PR from that.

examples/rules.mk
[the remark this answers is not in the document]
hathach: remove both `-Wno-error=lto-type-mismatch` and `-fshort-enums` , I don't encounter any warnings with gcc 9. I guess we can safely remove them for now.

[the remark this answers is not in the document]
pigrew: I'm using gcc 7.3.1 (as distributed with ST STM32CubeIDE). I'll let you know if anything comes up, but we can leave it as is for now.

examples/device/net_lwip_webserver/src/usb_descriptors.c
hathach: Seem like I have no ways to play lazy anymore.

examples/device/net_lwip_webserver/src/usb_descriptors.c
```diff
@@ -159,13 +157,12 @@ uint16_t const* tud_descriptor_string_cb(uint8_t index, uint16_t langid)
 
   unsigned int chr_count = 0;
 
-  if (STR_LANGID == index)
+  if (STRID_LANGID == index)
   {
     memcpy(&_desc_str[1], string_desc_arr[0], 2);
```
majbthrd: Could "string_desc_arr[0]" become "string_desc_arr[STRID_LANGID]"?

hathach: ah right, I did missed that

pigrew:
@hathach 
Just updated to gcc 9.2.1 toolchain from the ARM website, and still getting Warnings. Ugh.

Adding `-fshort-enums` to CFLAGS in `board.mk` "fixes" it, but I don't know why I'm getting the warning, but you are not.

hathach:
@pigrew I have no idea, I installed it from the xpac https://github.com/xpack-dev-tools/arm-none-eabi-gcc-xpack since there is no official apt any more. It is the same as ci which doesn't throw out warnings as well. What is your warning message  ?

pigrew:
> @pigrew I have no idea, I installed it from the xpac https://github.com/xpack-dev-tools/arm-none-eabi-gcc-xpack since there is no official apt any more. It is the same as ci which doesn't throw out warnings as well. What is your warning message?

Perhaps I should do a fresh checkout of tinyusb...

Bunches of lines like:
```c:/program files (x86)/gnu tools arm embedded/9 2019-q4-major/bin/../lib/gcc/arm-none-eabi/9.2.1/../../../../arm-none-eabi/bin/ld.exe: warning: c:/program files (x86)/gnu tools arm embedded/9 2019-q4-major/bin/../lib/gcc/arm-none-eabi/9.2.1/../../../../arm-none-eabi/lib/thumb/v6-m/nofp\libnosys.a(fstat.o) uses variable-size enums yet the output is to use 32-bit enums; use of enum values across objects may fail
c:/program files (x86)/gnu tools arm embedded/9 2019-q4-major/bin/../lib/gcc/arm-none-eabi/9.2.1/../../../../arm-none-eabi/bin/ld.exe: warning: c:/program files (x86)/gnu tools arm embedded/9 2019-q4-major/bin/../lib/gcc/arm-none-eabi/9.2.1/../../../../arm-none-eabi/lib/thumb/v6-m/nofp\libnosys.a(isatty.o) uses variable-size enums yet the output is to use 32-bit enums; use of enum values across objects may fail
c:/program files (x86)/gnu tools arm embedded/9 2019-q4-major/bin/../lib/gcc/arm-none-eabi/9.2.1/../../../../arm-none-eabi/bin/ld.exe: warning: c:/program files (x86)/gnu tools arm embedded/9 2019-q4-major/bin/../lib/gcc/arm-none-eabi/9.2.1/../../../../arm-none-eabi/lib/thumb/v6-m/nofp\libnosys.a(lseek.o) uses variable-size enums yet the output is to use 32-bit enums; use of enum values across objects may fail
```

hathach:
> > @pigrew I have no idea, I installed it from the xpac https://github.com/xpack-dev-tools/arm-none-eabi-gcc-xpack since there is no official apt any more. It is the same as ci which doesn't throw out warnings as well. What is your warning message?
> 
> Perhaps I should do a fresh checkout of tinyusb... (ADDENDUM: didn't help)
> 
> Bunches of lines like:
> 
> ```
> c:/program files (x86)/gnu tools arm embedded/9 2019-q4-major/bin/../lib/gcc/arm-none-eabi/9.2.1/../../../../arm-none-eabi/bin/ld.exe: warning: c:/program files (x86)/gnu tools arm embedded/9 2019-q4-major/bin/../lib/gcc/arm-none-eabi/9.2.1/../../../../arm-none-eabi/lib/thumb/v6-m/nofp\libnosys.a(isatty.o) uses variable-size enums yet the output is to use 32-bit enums; use of enum values across objects may fail
> c:/program files (x86)/gnu tools arm embedded/9 2019-q4-major/bin/../lib/gcc/arm-none-eabi/9.2.1/../../../../arm-none-eabi/bin/ld.exe: warning: c:/program files (x86)/gnu tools arm embedded/9 2019-q4-major/bin/../lib/gcc/arm-none-eabi/9.2.1/../../../../arm-none-eabi/lib/thumb/v6-m/nofp\libnosys.a(lseek.o) uses variable-size enums yet the output is to use 32-bit enums; use of enum values across objects may fail
> ```

This is probably the issue with pre-built gcc option when ARM compile that on windows specifically. Let's revert to have short-enum back, we could revisit this later on.

PR #306, opened 2020-03-23
add support for seeeduino xiao

jeremyherbert:
This adds support for the seeeduino xiao


jeremyherbert:
Added the board to the docs list, and I also changed the linker script so it doesn't clobber the bootloader anymore. I personally don't care about the bootloader but I don't want to ruin anyone's day ;)

PR #308, opened 2020-03-26
Modify include of HAL, remove CMSIS clock configuration code, STM32F407 UART

### LINKED ISSUE #204: DHSE_VALUE=8000000 override in STM32 Makefiles
**Here is my question**
Is there a particular reason why this override of the xxx_hal_conf.h file exists in the Makefile Cflags of all the BSP examples? If a user isn't aware of its use, or assumes it is overridden by the HAL instead of the other way around, it will quietly cause the frequency to be wrong across all registers and HAL functions. 

I personally skimmed over it in my initial installation of tusb and then forgot about it, inviting problems down the line. It doesn't seem to serve any purpose since this value is an essential component of hal_conf. Should it be removed?

pigrew:
Fixes #204 

(It has only been tested on the F407, lets hope it passes CI.)

There is still a little bit of non-uniformity to the BSP. This patch leaves some of the individuality.

### DIFF
--- a/hw/bsp/stm32f070rbnucleo/stm32f070rbnucleo.c
+++ b/hw/bsp/stm32f070rbnucleo/stm32f070rbnucleo.c
@@ -26,8 +26,7 @@
 
 #include "../board.h"
 
-#include "stm32f0xx.h"
-#include "stm32f0xx_hal_conf.h"
+#include "stm32f0xx_hal.h"
 
 #define LED_PORT              GPIOA
 #define LED_PIN               GPIO_PIN_5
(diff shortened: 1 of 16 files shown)


pigrew:
@hathach 

I'm getting a warning for H743 that `RCC_OscInitStruct.PLL.PLLR` is not initialized.

Do you know what value it should be? I think PLLR might be unused, but it's bad to leave it unitialized. Perhaps use "/2"?

hathach:
the value is unused according to the comment, value is not matter, best to be zero with `//unused` comment.
Hey, your the time over there flies much faster than mine. Your tomorrow come so soon :D

hathach:
update: you can run ci locally with

```
python3 tools/build_all [example]
```

if example is not provided, it will build all the examples. you need to install other toolchain e.g RISC-V though to pass build on fomu board. Or just ignore its error mesasge.

pigrew:
@hathach 

I just force-pushed the one extra PLLR setting. There are a few other places where the initialization struct fields are uninitialized, but I think we can save them for later. I think we can merge this PR.

Issue #316, opened 2020-03-30
Add ESP32-S2 port


hathach:
**Description**
ESP32-S2 is already supported in ESP-IDF https://github.com/espressif/esp-idf/tree/master/components/tinyusb/port/esp32s2 . Although the IDF is specific to ESP32-S2, it is best to integrate it to current bare-metal build of our repo for consistency across ports and helps 
1. Testing with new features, changes and class driver easier
2. Other projects such as CircuitPython that includes tinyusb as submodule to work with ESP32-S2
3. Easier to maintain for me

Note: since ESP32-S2 use the same Synopsis design ware IP as STM32F4. It makes sense to merge their DCD together if applicable.

@ladyada @tannewt @dhalbert @igrr

cr1901:
@hathach Ahhh yes, @ladyada mentioned this part to me a while back. I think their DCDs (not necessarily interrupts, however) should be the same, though it may make sense for the time being to copy the STM32F4 `dcd_synposys.c` to an `ESP32` directory to see how much code can actually be shared after the port is working.

hathach:
@cr1901 yeah, that makes sense, we will keep them separated for now. And decide later if needed. Also the USB specs of EPS32S2 isn't available yet, it is better to keep it as it is :)

igrr:
At the moment ESP32-S2 port (as included in IDF) uses FIFOs, while the STM port seems to use DMA. IIRC there are some alignment requirements for the USB DMA buffers, and ESP-IDF heap allocator didn't support custom alignment values when we started porting TinyUSB. Alignment support [is implemented now](https://github.com/espressif/esp-idf/commit/d26ddaa644997a8b65dfa52b716e5d9502f69c9e#diff-b43890c7d22cca8d9ae095e020ef36cb), so we could switch to a DMA based as well. This needs some testing, however, so adding the FIFO-based driver first might be a quicker option.

cr1901:
@igrr FWIW, the Synopsys driver in tinyusb doesn't use DMA; it's purely interrupt-driven.

igrr:
You are right, of course; sorry, I must have been looking at some other code for the ST.

hathach:
@cr1901 @igrr yeah, I will just use the dcd as it is for now. We can always improve later :)

hathach:
implemented by PR #319 

@igrr Should you be interested on TinyUSB on Arduino. Adafruit also has an Adruino version of TinyUSB. It is composed of 2 parts
[...]

PR #317, opened 2020-03-30
add provision for alternate interfaces

majbthrd:
Another revision to usbnet CDC-ECM is in the pipeline and operational, but it depends on support for alternate interfaces.  Re-reading the CDC-ECM specification, this is mandatory; although OS X ‘Lion’ and Linux wasn't bothered, the CDC-ECM drivers in more modern MacOS insist upon alternate interfaces.

So, this PR provides provision for this.

### DIFF
--- a/src/device/usbd.c
+++ b/src/device/usbd.c
@@ -55,7 +55,9 @@ typedef struct {
 
   uint8_t itf2drv[16];     // map interface number to driver (0xff is invalid)
   uint8_t ep2drv[8][2];    // map endpoint to driver ( 0xff is invalid )
+#if CFG_TUD_SUPPORT_ALTERNATES
   uint8_t itf2alt[16];     // map interface number to current alternative setting
+#endif
 
   struct TU_ATTR_PACKED
   {
@@ -84,7 +86,9 @@ typedef struct {
   bool (* control_complete ) (uint8_t rhport, tusb_control_request_t const * request);
   bool (* xfer_cb          ) (uint8_t rhport, uint8_t ep_addr, xfer_result_t event, uint32_t xferred_bytes);
   void (* sof              ) (uint8_t rhport);
+#if CFG_TUD_SUPPORT_ALTERNATES
   void (* set_alternate    ) (uint8_t rhport, uint8_t num);
+#endif
 } usbd_class_driver_t;
 
 static usbd_class_driver_t const _usbd_driver[] =
@@ -345,7 +349,9 @@ static void usbd_reset(uint8_t rhport)
 
   memset(_usbd_dev.itf2drv, DRVID_INVALID, sizeof(_usbd_dev.itf2drv)); // invalid mapping
   memset(_usbd_dev.ep2drv , DRVID_INVALID, sizeof(_usbd_dev.ep2drv )); // invalid mapping
+#if CFG_TUD_SUPPORT_ALTERNATES
   memset(_usbd_dev.itf2alt, 0, sizeof(_usbd_dev.ep2drv )); // default mapping
+#endif
 
   usbd_control_reset();
 
@@ -606,7 +612,11 @@ static bool process_control_request(uint8_t rhport, tusb_control_request_t const
         {
           case TUSB_REQ_GET_INTERFACE:
           {
+#if CFG_TUD_SUPPORT_ALTERNATES
             uint8_t alternate = _usbd_dev.itf2alt[itf];
+#else
+            uint8_t alternate = 0;
+#endif
             tud_control_xfer(rhport, p_request, &alternate, 1);
           }
           break;
@@ -615,8 +625,12 @@ static bool process_control_request(uint8_t rhport, tusb_control_request_t const
           {
             uint8_t const alternate = (uint8_t) p_request->wValue;
 
+#if CFG_TUD_SUPPORT_ALTERNATES
             _usbd_dev.itf2alt[itf] = alternate;
             if (_usbd_driver[drvid].set_alternate) _usbd_driver[drvid].set_alternate(rhport, alternate);
+#else
+            TU_ASSERT(alternate == 0);
+#endif
             tud_control_status(rhport, p_request);
           }
           break;
--- a/src/device/usbd.c
+++ b/src/device/usbd.c
@@ -55,6 +55,7 @@ typedef struct {
 
   uint8_t itf2drv[16];     // map interface number to driver (0xff is invalid)
   uint8_t ep2drv[8][2];    // map endpoint to driver ( 0xff is invalid )
+  uint8_t itf2alt[16];     // map interface number to current alternative setting
 
   struct TU_ATTR_PACKED
   {
@@ -83,6 +84,7 @@ typedef struct {
   bool (* control_complete ) (uint8_t rhport, tusb_control_request_t const * request);
   bool (* xfer_cb          ) (uint8_t rhport, uint8_t ep_addr, xfer_result_t event, uint32_t xferred_bytes);
   void (* sof              ) (uint8_t rhport);
+  void (* set_alternate    ) (uint8_t rhport, uint8_t num);
 } usbd_class_driver_t;
 
 static usbd_class_driver_t const _usbd_driver[] =
@@ -343,6 +345,7 @@ static void usbd_reset(uint8_t rhport)
 
   memset(_usbd_dev.itf2drv, DRVID_INVALID, sizeof(_usbd_dev.itf2drv)); // invalid mapping
   memset(_usbd_dev.ep2drv , DRVID_INVALID, sizeof(_usbd_dev.ep2drv )); // invalid mapping
+  memset(_usbd_dev.itf2alt, 0, sizeof(_usbd_dev.ep2drv )); // default mapping
 
   usbd_control_reset();
 
@@ -603,8 +606,7 @@ static bool process_control_request(uint8_t rhport, tusb_control_request_t const
         {
           case TUSB_REQ_GET_INTERFACE:
           {
-            // TODO not support alternate interface yet
-            uint8_t alternate = 0;
+            uint8_t alternate = _usbd_dev.itf2alt[itf];
             tud_control_xfer(rhport, p_request, &alternate, 1);
           }
           break;
@@ -613,8 +615,8 @@ static bool process_control_request(uint8_t rhport, tusb_control_request_t const
           {
             uint8_t const alternate = (uint8_t) p_request->wValue;
 
-            // TODO not support alternate interface yet
-            TU_ASSERT(alternate == 0);
+            _usbd_dev.itf2alt[itf] = alternate;
+            if (_usbd_driver[drvid].set_alternate) _usbd_driver[drvid].set_alternate(rhport, alternate);
             tud_control_status(rhport, p_request);
           }
           break;
--- a/src/tusb_option.h
+++ b/src/tusb_option.h
@@ -183,6 +183,10 @@
 #define CFG_TUSB_OS               OPT_OS_NONE
 #endif
 
+#ifndef CFG_TUD_SUPPORT_ALTERNATES
+  #define CFG_TUD_SUPPORT_ALTERNATES 0
+#endif
+
 //--------------------------------------------------------------------
 // DEVICE OPTIONS
 //--------------------------------------------------------------------


pigrew:
Based on this patch, I looked at how other stacks handle things, and wanted to note some thoughts I had:

* USB 1.1 Spec _9.1.1.5_ states that _Configuring a device or changing an alternate setting causes all of the status and configuration values associated with endpoints in the affected interfaces to be set to their default values._
* Should we impose the restriction that all alternates of a given interface must be handled by the same class driver?  (This seems like a reasonable limitation, but should be documented)

Three implementations come to mind:

**Implementation 0** Call an application-level event callback when  a set_interface has been requested.

**Implementation 1** (what's been proposed) Create a class handler. This handler would need to close and open EPs, verify the new alternate index is valid, reset its own data structures.

**Implementation 2**: No set_interface call is required to be added. The alternate should be selected by pointing to the proper `desc_intf` in the class's `->open(...)` callback. The class driver's `reset()` would need to be called to close the endpoints prior to reopening them with the `open()` call. Perhaps a `->deinit` should be added, too, for the case that class drivers change, to free memory/hardware resources allocated in the class driver?

Implementation 0 could be added with either of the other two. It shouldn't be needed, but could be useful in rare circumstances.

I'm leaning towards _implementation 2_ since it likely allows for the simplest class implementation. However, both 1 and 2 are sound.

For _implementation 1_, the following issues may need to be addressed:

1. How does the class driver find the new interface descriptor? Endpoint numbers may change when changing the interface? Should `set_alternate` have the new interface descriptor as a parameter instead of the alternate number?
2. `set_alternate` can fail (e.g. if the alternate number isn't valid or endpoints are not idle). When it fails, the active alternate number shouldn't change.
3. `TU_VERIFY` that wLength is zero in SET_INTERFACE. Though, USB spec says this would be undefined behavior, so anything could be done.
4. SET_INTERFACE must stall EP0 when in addressed (but not configured) state. In default state, it's undefined what to do (but probably should fail, too). Same for GET_INTERFACE
5. Document that all alternate interfaces must be handled by a particular class driver. (changing alternate interface cannot cause a different class driver to be used). Perhaps add runtime check?
6. The various alternates may have different class/subclass/protocol codes. How is that handled? (One solution would be to create a `bool supports_interface_cb(tusb_desc_interface_t*)` function in the class table. Currently, subclass are not in the class table. This change would have the side effect of also adding support for composite devices where multiple interfaces have the same class but different subclass.)

How do other stacks handle alternates:

* ST USB
  * No stack implementation
* ARM CMSIS 
  * No stack implementation
* TI MSP432E4
  * Has lots of levels of redirection. Perhaps overengineered, but I like it.
  * Stack allows only ONE driver per USB interface. However, there is a "composite" driver that allows multiple drivers to be called (and performs the event dispatching)...
  * stack usbd module itself handles opening and closing endpoints (not the class implementation).
  * It has multiple levels of callbacks. One can add an USB device-class-driver level handler to interface change, or a Device-API level callback (and maybe DriverLib level?)
* MicroChip PIC32
  * Seems to send the setup packet to the class driver, it isn't handled by the usb device module.

majbthrd:
@pigrew, I appreciate the critique and the comparison with other stacks.

>     * Should we impose the restriction that all alternates of a given interface must be handled by the same class driver?  (This seems like a reasonable limitation, but should be documented)

I think so.  TinyUSB already maps a given interface number to a particular driver, and only the driver knows best how to handle this.

> **Implementation 2**: No set_interface call is required to be added. The alternate should be selected by pointing to the proper `desc_intf` in the class's `->open(...)` callback. The class driver's `reset()` would need to be called to close the endpoints prior to reopening them with the `open()` call. Perhaps a `->deinit` should be added, too, for the case that class drivers change, to free memory/hardware resources allocated in the class driver?

I suppose one could just let the class's ->open() callback take care of the reset and deinit that it deemed necessary?

>     1. How does the class driver find the new interface descriptor? Endpoint numbers may change when changing the interface? Should `set_alternate` have the new interface descriptor as a parameter instead of the alternate number?

I admit to having a myopic view at this point, but in the CDC-ECM descriptors, there are either zero endpoints or two.  So, at least in this scenario, endpoint numbers don't change; they either exist or they don't.

>     2. `set_alternate` can fail (e.g. if the alternate number isn't valid or endpoints are not idle). When it fails, the active alternate number shouldn't change.

Yeah, a way to address this problem would be to return a bool that could be checked with TU_VERIFY().  I believe that is similar to project code elsewhere.

I'll add a 2(b), and that is the situation where the set_alternate() pointer is NULL.  I considered putting an else to the if with a TU_VERIFY() to preserve existing behavior, but I'm always torn about unnecessarily adding extra code to a "tiny" implementation.  I already feel bad enough about adding 16 bytes to RAM usage with the itf2alt table.
 
>     3. `TU_VERIFY` that wLength is zero in SET_INTERFACE. Though, USB spec says this would be undefined behavior, so anything could be done.

I wonder if the project would benefit from a 'TU_VERIFY_PEDANTIC' for run-time checks that could get optimized out if the user desired, or perhaps that just adds complexity.  I know that as a user, I would appreciate the option to choose to keep code smaller (at the cost of being ill-behaved with an unlikely standards-non-compliant host)

>     6. The various alternates may have different class/subclass/protocol codes. How is that handled regarding the class callback table in `usbd.c`? (One solution would be to create a `bool supports_interface_cb(tusb_desc_interface_t*)` function in the class table. Currently, subclass are not in the class table. This change would have the side effect of also adding support for composite devices where multiple interfaces have the same class but different subclass.)

Again, I'm looking from the myopic perspective of CDC-ECM:

> "To recover the network aspects of a device to known states, the host will first select the default interface setting (with no endpoints) and then select the appropriate alternate interface setting. In response to this action the device shall"...

Things could always be improved/enhanced/uprooted over time.

pigrew:
> > **Implementation 2**: No set_interface call is required to be added. The alternate should be selected by pointing to the proper `desc_intf` in the class's `->open(...)` callback. The class driver's `reset()` would need to be called to close the endpoints prior to reopening them with the `open()` call. Perhaps a `->deinit` should be added, too, for the case that class drivers change, to free memory/hardware resources allocated in the class driver?
> 
> I suppose one could just let the class's ->open() callback take care of the reset and deinit that it deemed necessary?

Yes, tat would work for most situations. If the stack allows a different class driver to be selected, this wouldn't work (the first would need to be closed before the second opened).

> > ```
> > 1. How does the class driver find the new interface descriptor? Endpoint numbers may change when changing the interface? Should `set_alternate` have the new interface descriptor as a parameter instead of the alternate number?
> > ```
> 
> I admit to having a myopic view at this point, but in the CDC-ECM descriptors, there are either zero endpoints or two. So, at least in this scenario, endpoint numbers don't change; they either exist or they don't.

Yes, it's likely good enough. How would the _open() callback operate if it is passed alternate 0? It would have to scan through all alternates. However, I don't think it can since it doesn't have a pointer to the configuration descriptor, and doesn't know how many alternates there are?

I'm probably overgeneralizing the implementation. There should be a balance somewhere....

> > ```
> > 2. `set_alternate` can fail (e.g. if the alternate number isn't valid or endpoints are not idle). When it fails, the active alternate number shouldn't change.
> > ```
> 
> Yeah, a way to address this problem would be to return a bool that could be checked with TU_VERIFY(). I believe that is similar to project code elsewhere.
> 
> I'll add a 2(b), and that is the situation where the set_alternate() pointer is NULL. I considered putting an else to the if with a TU_VERIFY() to preserve existing behavior, but I'm always torn about unnecessarily adding extra code to a "tiny" implementation. I already feel bad enough about adding 16 bytes to RAM usage with the itf2alt table.

I've also been thinking about this. Perhaps we need a CFG_TUD_SUPPORT_ALTERNATES=1 define to enable support? Also it looks like if alternates are not supported, it's allowed to stall EP0 upon GET_INTERFACE or SET_INTERFACE, so we could just remove these branches completely if not being used.
> 
> > ```
> > 3. `TU_VERIFY` that wLength is zero in SET_INTERFACE. Though, USB spec says this would be undefined behavior, so anything could be done.
> > ```
> 
> I wonder if the project would benefit from a 'TU_VERIFY_PEDANTIC' for run-time checks that could get optimized out if the user desired, or perhaps that just adds complexity. I know that as a user, I would appreciate the option to choose to keep code smaller (at the cost of being ill-behaved with an unlikely standards-non-compliant host)

I care that the device not segfault with a misbehaving host. In this case, the check seems completely optional, but could be helpful in diagnosing misbehaving hosts. No need for the check here, since the spec allows undefined behavior. A comment about this would be helpful.

To make the code smaller, I think we could remove all TU_ASSERT, as there should be called only in the case of a stack bug?

majbthrd:
> I've also been thinking about this. Perhaps we need a CFG_TUD_SUPPORT_ALTERNATES=1 define to enable support?

I like it!  I've gone ahead with another commit that adds this, preserving the existing behavior by default.

cr1901:
>Should we impose the restriction that all alternates of a given interface must be handled by the same class driver? (This seems like a reasonable limitation, but should be documented)

This might be an overly pedantic question, but... I could've sworn that the USB spec explicitly says that an alternate interface _must_ be the same device class as the primary interface. So this might not be a limitation at all, but "complying to spec"?

pigrew:
> > Should we impose the restriction that all alternates of a given interface must be handled by the same class driver? (This seems like a reasonable limitation, but should be documented)
> 
> This might be an overly pedantic question, but... I could've sworn that the USB spec explicitly says that an alternate interface _must_ be the same device class as the primary interface. So this might not be a limitation at all, but "complying to spec"?

Huh, you got me searching more:

1. CDC EEM v1.0 3.2.3 states `If a Communication Class interface appears with multiple alternate settings, all alternate settings for that interface shall have the same bInterfaceClass, bInterfaceSubclass and bInterfaceProtocol codes.`

2. Printer class spec v1.1 3.2 states `If multiple printer interfaces are supported, they shall be implemented as alternate interfaces. The Protocol field in the interface descriptor informs the host of the interface type: 01 is the unidirectional interface, 02 is the bi-directional interface. and 03 is the 1284.4 interface.`

3. USB mass storage, bulk only transfer v1.0 4.3 states `The interface may have multiple alternate settings. The host shall examine each of the alternate settings to look for the bInterfaceProtocol and bInterfaceSubClass it supports optimally`

I can't find anywhere that states that they have to have the same class code.

All of the examples I found show the alternates having the same class code, but perhaps different subclass codes. I can't think of a situation where different class drivers would be required to handle the different alternates of a particular interface.

PanRe:
I am currently working on an audio class implementation (UAC2) for which alternate settings for interfaces is also mandatory (in the same sense that alternate setting 0 is default setting with 0 EPs active and alternate setting 1 is the active setting).
It would be pretty cool if alternate settings are supported by this nice stack soon! :)

PanRe:
@majbthrd 
Without testing your code, there might be an issue with process_set_config() in which alternate settings are not treated so far. Up to my knowledge (but i am a bloody rookie in USB) interfaces with alternate settings show up consecutively with equal interface numbers. Within process_set_config() a check is conducted to avoid interfaces with the same interface number to be handed over to the class driver. So the class driver can not be informed of an alternate setting (or is it intended to let the class driver check for alternate settings?). And further, where do you fill in itf2alt?
Sorry if i understand you implementation wrong, as i said i did not test it so far...

majbthrd:
> Without testing your code, there might be an issue with process_set_config() in which alternate settings are not treated so far. Up to my knowledge (but i am a bloody rookie in USB) interfaces with alternate settings show up consecutively with equal interface numbers. Within process_set_config() a check is conducted to avoid interfaces with the same interface number to be handed over to the class driver.

@PanRe, the author’s reasoning behind the comment in process_set_config() mentioning alternate interfaces is not clear to me.  I think it was just a speculative comment to cause the code to be revisited at a later date.  The check has nothing to do with alternate interfaces and everything to do with configurations.

process_set_config() is invoked by a TUSB_REQ_SET_CONFIGURATION, whereas setting alternate interfaces is caused by TUSB_REQ_SET_INTERFACE.

> And further, where do you fill in itf2alt?

Try searching ./src/device/usbd.c for "itf2alt" or "case TUSB_REQ_SET_INTERFACE".

hathach:
@majbthrd I am currently porting ESP32-S2 #316 , it takes the  priority  since Adafruit's team is waiting for esp32 port to start working on a Circuitpython port. Please be patient in a few days. Thanks a lot :)

FYI: Alternate Interface is required for most of the interface that make use of Isochronous transfer, which is a Constant Data Rate for audio/video. Those interfaces may come with a series of Alt iterface like youtube with 240p, 360p, 480p etc... Depending on the host state and its current bus bandwidth or power or user preference, it will SET_ALTERNATE the appropriate Alt Interface. So far Iso transfer isn't implemented, yeah just because we didn't have chance to work with those classes yet. I will try to look at this asap.

PanRe:
@majbthrd Thanks for the quick reply.
@hathach Thanks a lot! 
Unfortunately, to my understanding process_set_config() still has some short comings in case alternate settings are present. Please correct me if i am wrong! Right now, .open() would need to parse all class related interface descriptors (including all alternate setting interface descriptors). This, however, is not possible since the number of alternate setting interfaces descriptors is arbitrary and unknown in general. Also, the number of interfaces belonging to this class (or better USB function) is unknown. The best possible solution would be let .open() parse through all interface descriptors associated to it, including all alternate settings, by giving it the number of interfaces to be treated (given by a preceeding IAD descriptor) and the total length of the descriptor bytes left to read. Both informations are necessary to be able to parse all successfully.
If this would not be done, an interface belonging to a different class (as in multi-interface devices) following such a pack of interfaces could not be found.

majbthrd:
> Unfortunately, to my understanding process_set_config() still has some short comings in case alternate settings are present. Please correct me if i am wrong! Right now, .open() would need to parse all class related interface descriptors (including all alternate setting interface descriptors). This, however, is not possible since the number of alternate setting interfaces descriptors is arbitrary and unknown in general.

@PanRe, perhaps you could cite a specific descriptor that you tried with the code and which didn't work?

I agree that .open() must parse all class related interface descriptors, and it does that (at least if the class has been written to support alternate descriptors, which I think it necessary since, as you point out, some classes require alternate settings).  I've been running code for the past several days, so it is more than possible,

Just for completeness, here is the interface descriptor that I am using in local testing.  As you can see, it has multiple alternate settings:

```
    Interface Descriptor:
      bLength                 9
      bDescriptorType         4
      bInterfaceNumber        0
      bAlternateSetting       0
      bNumEndpoints           1
      bInterfaceClass         2 Communications
      bInterfaceSubClass      6 Ethernet Networking
      bInterfaceProtocol      0 
      iInterface              4 TinyUSB Network Interface
      CDC Header:
        bcdCDC               1.20
      CDC Union:
        bMasterInterface        0
        bSlaveInterface         1 
      CDC Ethernet:
        iMacAddress                      5 0202846A9600
        bmEthernetStatistics    0x00000000
        wMaxSegmentSize               1514
        wNumberMCFilters            0x0000
        bNumberPowerFilters              0
      Endpoint Descriptor:
        bLength                 7
        bDescriptorType         5
        bEndpointAddress     0x81  EP 1 IN
        bmAttributes            3
          Transfer Type            Interrupt
          Synch Type               None
          Usage Type               Data
        wMaxPacketSize     0x0040  1x 64 bytes
        bInterval               1
    Interface Descriptor:
      bLength                 9
      bDescriptorType         4
      bInterfaceNumber        1
      bAlternateSetting       0
      bNumEndpoints           0
      bInterfaceClass        10 CDC Data
      bInterfaceSubClass      0 Unused
      bInterfaceProtocol      0 
      iInterface              0 
    Interface Descriptor:
      bLength                 9
      bDescriptorType         4
      bInterfaceNumber        1
      bAlternateSetting       1
      bNumEndpoints           2
      bInterfaceClass        10 CDC Data
      bInterfaceSubClass      0 Unused
      bInterfaceProtocol      0 
      iInterface              0 
      Endpoint Descriptor:
        bLength                 7
        bDescriptorType         5
        bEndpointAddress     0x82  EP 2 IN
        bmAttributes            2
          Transfer Type            Bulk
          Synch Type               None
          Usage Type               Data
        wMaxPacketSize     0x0040  1x 64 bytes
        bInterval               0
      Endpoint Descriptor:
        bLength                 7
        bDescriptorType         5
        bEndpointAddress     0x02  EP 2 OUT
        bmAttributes            2
          Transfer Type            Bulk
          Synch Type               None
          Usage Type               Data
        wMaxPacketSize     0x0040  1x 64 bytes
        bInterval               0
```

hathach:
@PanRe your analysis is right, and it is on-purpose as well. Until we has an audio/video class driver, there is no need (for now) to support ALT Interface. I always opt for not writing code that isn't run. The USBD-Class driver can be changed when needed, and that won't affect the application at all. So there is no worry at all.

majbthrd:
> Until we has an audio/video class driver, there is no need (for now) to support ALT Interface.

@hathach, I have a USB-ECM driver that is not an "audio/video class driver", but yet it needs support for alternate interfaces.  This is specifically why I submitted this PR.

hathach:
@majbthrd ah yeah, then we have a reason to implement it now. Sorry, I still haven't checked out this PR just yet. will do asap :sweat_smile:

PanRe:
@majbthrd I am not at all saying your code is not working - no offence intended! Also your PR is fine i would say. It is just .open() what concerns me. How do you know the number of interfaces and alternate setting interfaces to parse by .open() in your case? Is this a fixed number or hard coded?

majbthrd:
> How do you know the number of interfaces and alternate setting interfaces to parse by .open() in your case? Is this a fixed number or hard coded?

@PanRe, there is no substitute for trying the code and seeing how existing implementations work and then adapting this yourself to try to serve your additional purposes.  You'll see that the third argument to open() is a pointer to the interface descriptor length.  This serves as a mechanism for open() to report back how large the interface descriptor is; process_set_config() doesn't have any fixed numbers or hard coded values to tell open() what to expect.  It is incumbent on open() to incrementally seek through the descriptors until it decides it has seen everything pertaining to that class's interface descriptor.

PanRe:
@majbthrd Sorry for the late reply! I honestly would like to test your code, but as far as i can see in your examples you do not have any alternate settings inferfaces defined... neither in TUD_CDC_ECM_DESCRIPTOR nor in TUD_RNDIS_DESCRIPTOR or in TUD_CDC_EEM_DESCRIPTOR. In those descriptors you have at most two interfaces defined, which you parse in a row in your .open() function (this is what i meant with hard coded, i did not mean process_set_config() to contain any hard coded stuff, also your code with your descriptors does not need recursive parsing). Now i see you posted a destriptor with an alternate setting present. Good, but this only proofs that the USB-Host was able to successfully parse your descriptors. Not that your device code parses that successful. Next, the part
```c
// Interface number must not be used already TODO alternate interface
      TU_ASSERT( DRVID_INVALID == _usbd_dev.itf2drv[desc_itf->bInterfaceNumber] );
_usbd_dev.itf2drv[desc_itf->bInterfaceNumber] = drv_id;
```
in process_set_config() should prohibit recursive calling of .open() since once
`
_usbd_dev.itf2drv[desc_itf->bInterfaceNumber] = drv_id;
`
was called TU_ASSERT() should abort if an alternate setting interface with the same bInterfaceNumber comes up. Just from that i am pretty sure your implementation never treats alternate settings. 
Also, you do not even treat multiple interfaces for one class! You only have one netd_interface_t  _netd_itf defined. Imagine multiple interfaces netd_interface_t  _netd_itf[X] to handle. How do you tell how many interfaces should belong to the first driver instance how many to the second. Ok - this might be possible to deduct from the sequence of how descriptors are ordered but an IAD would be more elegant in my opinion.

Maybe i am terribly wrong here somewhere (sorry in that case). Once your PR is through i am going to test your stuff. After all i need the same as you profided here.

majbthrd:
> Just from that i am pretty sure your implementation never treats alternate settings.
> Also, you do not even treat multiple interfaces for one class!

> Maybe i am terribly wrong here somewhere (sorry in that case). Once your PR is through i am going to test your stuff.

@PanRe, I respectfully ask that you extend me the courtesy of waiting until the new CDC-ECM PR is pushed.  I believe you will feel differently when you see the upcoming PR... there are distinct class handlers for each interface and there are alternate interfaces in the descriptors (descriptor preview shown above).  At least with my CDC-ECM class, there is no issue with the TU_ASSERT() you point to.  Perhaps there will be something new with your UAC2 class when it arrives, but I'm confident that you will devise a solution to address any such problems at that time.

This current PR was supposed to be a super simple intermediate step just to get some buy-in on an approach before I pushed a PR that depended on alternate interfaces.

majbthrd:
@PanRe et all, I've uploaded a [usbnet_preview branch](https://github.com/majbthrd/tinyusb/tree/usbnet_preview) to my fork.  I'm hoping this will allay concerns or doubts.

My preferred approach was to submit several incremental PRs, each illustrated a logical step.  However, to just prove that the code is functional, this has several commits, including this PR plus the CDC-ECM changes to use alternate interfaces, merging RNDIS and CDC-ECM into a single class driver, and adding MDNS support (needed for newer Linux and macOS to allow the user to use "tinyusb.local" in their web browser without even knowing the IP address of the USB device).

PanRe:
@majbthrd Please take all the time you need, i did not intend to interfere with your work! I am certain you know what you do! Stay healthy at those times! :)

pigrew:
@majbthrd 

I thought of another implementation, to have the class driver handle the SET_INTERFACE request in the class driver, instead of usbd. This also removes the need for storing the active interface for all interfaces, as the class driver keeps track in its own data structure. One functional change is that it would stall EP0 when SET_INTERFACE is not supported by the driver. This is allowed by the USB spec, but I don't know if it would create any host incompatibilities. Binary size is reduced by 20 bytes for the usbtmc example (which doesn't use alternate interfaces).

https://github.com/pigrew/tinyusb/commits/altAltSettings 

(I made a stub implementation in the net class. If I create a PR, I'd remove that first.)

majbthrd:
> I thought of another implementation, to have the class driver handle the SET_INTERFACE request in the class driver, instead of usbd.

@pigrew, that has teeth.  I believe the executive summary is: by removing the existing TUSB_REQ_USB_INTERFACE handler, execution drops into invoke_class_control() than in turn invokes the class's control_request() handler.  Said handler now services a second purpose as a handler for SET_INTERFACE.

As you say, it does change the behavior for SET_INTERFACE, so I defer to others on whether that is an issue for some platforms.  Alternatively, to keep the existing behavior, every existing class could be modified to act like the existing code (but at the cost of adding n instances of duplicate code).

pigrew:
@hathach & @majbthrd,

I just reread a bit of the USB spec, where it seems to say that all endpoint DTOG must be reset to 0 when SET_ALTERNATE happens. There doesn't seem to be any way for the class driver to direct the USBD driver to do this. There is no `dcd_edpt_close` function. Currently, it seems that only a USB reset can close the endpoints in the drivers.

The class drivers will have to close and open all endpoints when a `SET_INTERFACE` happens. I'm going back to thinking about my implementation 2 above, with each class driver having an `->open()` and `->close()` callback, which would be called by USBD when a `SET_INTERFACE` happens. This maybe could be aligned with `SET_CONFIG` calls, where the config could also be changed by calling `class->clase()` and `class->open()`, with a pointer to the appropriate interface/IAD.

In summary, I think neither of these PR can be merged, pending a resolution of the DTOG issue. I believe adding a `dcd_edpt_close` is the proper solution. I'd appreciate your input on if that is the proper solution.

majbthrd:
I think it is almost a given that any real-world implementations of alternate interfaces, the assorted alternate interface values may cause data to be enabled or disabled and might cause the data rate and type to be different for particular enabled variants.  However, the same endpoints are used throughout.

Isn't the data sequence, aka DTOG, intrinsically reset as part of the call to dcd_edpt_open()?  If so, we might have an existing mechanism to get that 'for free' without a dcd API revision.

Now, I recognize that *IF* the chosen alternate interface has no endpoints...

For OUT endpoints, is the host driver really going to challenge the device implementation by seeing if the endpoint mistakenly accepts the packet?  ("Gotcha!  You were not supposed to ACK that!")  Surely, a reasonable compromise is to simply have the device class driver not do anything with any packets received (that shouldn't be sent by the host to begin with).

For IN endpoints, how would the host driver know that the endpoint had not been closed if the device class driver simply didn't send any data?

In both #317 and #320, the class driver gets a call to inform it of the new alternate setting.  I would suggest that the class simple be on its honor not to generate traffic when instructed so.

Some sort of dcd_edpt_close to de-initialize endpoints would be a worthy long-term goal (and has the added benefit of perhaps enabling the different classes in different configuration concern @pigrew raised in #167).

pigrew:
> Isn't the data sequence, aka DTOG, intrinsically reset as part of the call to dcd_edpt_open()? If so, we might have an existing mechanism to get that 'for free' without a dcd API revision.
>
> Now, I recognize that _IF_ the chosen alternate interface has no endpoints...

The question is if it is allowed to call `dcd_edpt_open()` on an already opened endpoint. I know I wasn't expecting that when I wrote the stm32fsdev driver, but perhaps it's fine, it'd need to be verified for all of the drivers.

I think that the USB dev is supposed to reply with STALL upon a transfer from/to an EP which hasn't been opened? In that case, the class driver could stall the EP when transitioning to an alternate state where the EP doesn't exist. So, we could get around it, even without the `dcd_edpt_close` function.

> In both #317 and #320, the class driver gets a call to inform it of the new alternate setting. I would suggest that the class simple be on its honor not to generate traffic when instructed so.

Yeah, I think that there will be a lot of this gentleman's agreement needed:

1. `edpt_open` cannot be called on a non-stalled endpoint,
2. When going to a different alternate state, all opened EP must be set to stall during the transition.

How do we handle the concurrency issue of if there is an IN bulk transfer in flight when the SET_INTERFACE is received? Could the transfer_complete callback happen after the SET_INTERFACE? Is this an issue?

> Some sort of dcd_edpt_close to de-initialize endpoints would be a worthy long-term goal (and has the added benefit of perhaps enabling the different classes in different configuration concern @pigrew raised in #167).

Yep, I think that it is required for a full multiple-configuration implementation. I only had in mind a single config when I wrote PR #167, with the goal in mind of passing the USB test suite with a single-configuration device..

hathach:
> Implementation 1 (what's been proposed) Create a class handler. This handler would need to close and open EPs, verify the new alternate index is valid, reset its own data structures.

I agree with moving all implementation to class driver, since it is the best approach. I will create an PR for changes for review as well. It is somewhat similar, but I think it could solve the issue at least with USB NET. 

PS: Sorry for the wait, is super busy porting esp32-s2. It is quite demanding port :wink: :wink:

majbthrd:
now obsolete due to #345

PR #319, opened 2020-04-01
port ESP32 S2

hathach:
This PR adds support for ESP32-S2, implement #316, still work in progress. What works is
- [x] work with current example `cdc_msc_freertos` & `hid_composite_freertos`, since ESP32-S2 run freeRTOS natively, an freeRTOS example is required, luckily we already got one.
- [x] CDC seems to work as expected, connect/disconnect and echo
- [x] MSC does work (**Fixed** ) <s>not work yet, looking at bus analyzer, esp32s2 response with more data (128 bytes) than its endpoint size (64) in a transaction. Causing host driver to abort the interface. Unfortunately, there is no USB specs available for S2 just yet. So I have no idea how to fix this. @igrr did your team have any luck with USB MSC so far !!?!</s>
- [x] HID composite (keyboard + mouse) work well

## Steps to build

@ladyada @tannewt @dhalbert @cr1901 

1. https://docs.espressif.com/projects/esp-idf/en/latest/esp32s2/get-started/index.html#step-4-set-up-the-environment-variables
2. Install to toolchain, make sure you run hello world example first.
3. Add USB breakout D+ (GPIO20), D- (GPIO19) and an LED (GPIO21) for indicator. (note: saola board has an neopixel on board, but I will skip that one for now).
4. Only `cdc_msc_freertos` and `hid_composite_freertos` could work, so let's cd to it
5. build and flash with `idf.py build flash`


cr1901:
@hathach I don't know if this will solve your issues, but this stands out to me:

https://github.com/hathach/tinyusb/pull/319/files#diff-371e8160b188a80c1a3821c75d194adfR579-R584

```
    case 0x06: { // Setup packet recvd
        s_setup_phase = 0;
        ESP_EARLY_LOGV(TAG, "TUSB IRQ - setup_phase 0"); // new setup process
        // For some reason, it's possible to get a mismatch between
        // how many setup packets were received versus the location
        // of the Setup packet done word. This leads to situations
        // where stale setup packets are in the RX FIFO that were received
        // after the core loaded the Setup packet done word. Workaround by
        // only accepting one setup packet at a time for now.
        _setup_packet[0] = (USB0.grxstsp);
        _setup_packet[1] = (USB0.grxstsp);
        ESP_EARLY_LOGV(TAG, "TUSB IRQ - RX : Setup packet : 0x%08x 0x%08x",
                       _setup_packet[0], _setup_packet[1]);
    }
```

I [fixed](https://github.com/hathach/tinyusb/commit/d4302dacc584c4488cf1905caa822f96bd64930d) this bug while doing the H7 port. So it looks like the `dcd_synopsys.c` source you're using for the ESP32 port isn't up to date. There's other bug fixes too I can't think of right now in that time period. Bringing the source you're using for ESP32 up to date _may_ or _may not_ fix the > 64 bytes response bug.

ladyada:
```
.Guru Meditation Error: Core  0 panic'ed (IntegerDivideByZero). Exception was unhandled.
.
.Core  0 register dump:
.PC      : 0x40080cc3  PS      : 0x00060330  A0      : 0x80080e01  A1      : 0x3ffc1d30  
.A2      : 0x00000000  A3      : 0x00000001  A4      : 0x00000000  A5      : 0x3ffbd450  
.A6      : 0x00000001  A7      : 0x3ffc1d30  A8      : 0x80080cb1  A9      : 0x3ffc1d20  
.A10     : 0x3ffc1de0  A11     : 0x08000000  A12     : 0x00000000  A13     : 0x3ffc1de0  
.A14     : 0x3ffbcd28  A15     : 0x00000000  SAR     : 0x00000020  EXCCAUSE: 0x00000006  
.EXCVADDR: 0x00000000  LBEG    : 0x00000000  LEND    : 0x3ffc1de0  LCOUNT  : 0x400257fd  
.
.Backtrace:0x40080cc0:0x3ffc1d30 0x40080dfe:0x3ffc1d60 0x40082875:0x3ffc1da0 0x40087669:0x3ffc1de0 0x4002a771:0x3ffc1e50
.
.
.ELF file SHA256: 7c80004780167331
```
it enumerates in win10 for me, but i cannot connect/echo without a crash. still - it does enumerate!

hathach:
@ladyada windows driver can disable a whole device when receiving above error with msc, which is out of usb specs. Linux driver is the most forgiving, therefore it could still work with cdc. I will ping you when there is more to test

@cr1901 thank you for pointing that out. It is very helpful, I will do a side by side comparison with stm dcd and sync them up :)

tools/build_all.py
```diff
@@ -50,10 +50,18 @@ def build_size(example, board):
...
-    for skip_file in glob.iglob(ex_dir + '/.skip.MCU_*'):
-        mcu_cflag = '-DCFG_TUSB_MCU=OPT_' + os.path.basename(skip_file).split('.')[2]
-        with open(board_mk) as mk:
-            if mcu_cflag in mk.read():
+
+    with open(board_mk) as mk:
+        mk_contents = mk.read()
+
+        # Skip ESP32-S2 board if example is not FreeRTOS one
+        if 'freertos' not in example and 'CROSS_COMPILE = xtensa-esp32-elf-' in mk_contents:
```
igrr, line 58: Cross compiler should be xtensa-esp32s2-elf for ESP32-S2.

examples/device/cdc_msc_freertos/sdkconfig
```diff
@@ -0,0 +1,1060 @@
...
+CONFIG_ESP32_PTHREAD_TASK_STACK_SIZE_DEFAULT=3072
+CONFIG_ESP32_PTHREAD_STACK_MIN=768
+CONFIG_ESP32_PTHREAD_TASK_CORE_DEFAULT=-1
+CONFIG_ESP32_PTHREAD_TASK_NAME_DEFAULT="pthread"
+CONFIG_SPI_FLASH_WRITING_DANGEROUS_REGIONS_ABORTS=y
+# CONFIG_SPI_FLASH_WRITING_DANGEROUS_REGIONS_FAILS is not set
+# CONFIG_SPI_FLASH_WRITING_DANGEROUS_REGIONS_ALLOWED is not set
+CONFIG_SUPPRESS_SELECT_DEBUG_OUTPUT=y
+CONFIG_SUPPORT_TERMIOS=y
+# End of deprecated options
```
igrr, line 1060: It is recommended to only check in sdkconfig.defaults files for example projects, not the full sdkconfig files.

tools/build_all.py
hathach: yeah, it is, this is only a ci script to skip non-freertos example in the repo :)

examples/device/cdc_msc_freertos/sdkconfig
hathach: AH thanks, I didn't know that, will fix this in the next push.

src/portable/espressif/esp32s2/dcd_esp32s2.c
```diff
@@ -0,0 +1,749 @@
...
+    // here.
+    if (dir == TUSB_DIR_IN) {
+        // A full IN transfer (multiple packets, possibly) triggers XFRC.
+        int bytes2fifo_left = total_bytes;
+        uint32_t val; // 32 bit val from 4 buff addresses
+
+        USB0.in_ep_reg[epnum].diepint = ~0U; // clear all ints
+        USB0.in_ep_reg[epnum].dieptsiz = (num_packets << USB_D_PKTCNT0_S) | total_bytes;
+        USB0.in_ep_reg[epnum].diepctl |= USB_D_EPENA1_M | USB_D_CNAK1_M; // Enable | CNAK
+        while (bytes2fifo_left > 0) {  // TODO move it to ep_in_handle (IDF-1475)
```
hathach, line 329: @cr1901 this seems to be an issue, the setup you pointed out shouldn't cause the overrun endpoint size.

cr1901: Indeed, I wasn't sure whether the setup problem could cause overrun or not, but it should still be fixed.

I don't think `CFG_TUD_ENDOINT0_SIZE` can be < 8 per spec. 

In my synopsys port, it looks like I have [a guard](https://github.com/hathach/tinyusb/blob/master/src/portable/st/synopsys/dcd_synopsys.c#L530-L538) for this. It's also in [this file](https://github.com/hathach/tinyusb/pull/319/files/748cc88769ffd2039fcb0f4db9bb2cb4bc84965f#diff-371e8160b188a80c1a3821c75d194adfR522-R528) as well.

Instead of doing an initial xfer in `dcd_edpt_xfer`, I opt to instead deliberately [trigger](https://github.com/hathach/tinyusb/blob/master/src/portable/st/synopsys/dcd_synopsys.c#L356) an IN endpoint ISR to start the transfer. This avoids me having to write out that code twice :P.

Is there any particular constraint on this version of the Synopsys core that makes kickstarting the ISR impossible?

hathach: @cr1901 it is comment out on purpose from ESP, there is possibly some configuration difference comparing to stm32. I tried to enable the use of  `dtknqr4_fifoemptymsk` ( ESP's name for DIEPEMPMSK) in both edpt_xfer and handle_epin_ints but it results in no data transfer at all --> failed to get descriptor. Still doing trials and errors :sweat_smile: :sweat_smile:

src/portable/espressif/esp32s2/dcd_esp32s2.c
```diff
@@ -0,0 +1,749 @@
...
+            // IN XFER complete (entire xfer).
+            if (USB0.in_ep_reg[n].diepint & USB_D_XFERCOMPL0_M) {
+                ESP_EARLY_LOGV(TAG, "TUSB IRQ - IN XFER complete!");
+                USB0.in_ep_reg[n].diepint = USB_D_XFERCOMPL0_M;
+                // USB0.dtknqr4_fifoemptymsk &= ~(1 << n); // Turn off TXFE b/c xfer inactive.
+                dcd_event_xfer_complete(0, n | TUSB_DIR_IN_MASK, xfer->total_len, XFER_RESULT_SUCCESS, true);
+            }
+
+            // XFER FIFO empty
+            if (USB0.in_ep_reg[n].diepint & USB_D_XFERCOMPL0_M) {
```
hathach, line 657: @cr1901 spotted anything suspicious :thinking: :thinking:  It took me a day with tons of trials and reading specs back and forth to catch this copy-paste typo. Man, that is so typical, phew doing some clean up. there is still a bit of issue with other thing. But It is quite a progress for today :)

cr1901: Copy and paste typos happen. Good that you caught it though- because I certainly didn't :P.

hathach: phew, I feel so lucky catching this typo :)

.github/workflows/build.yml
```diff
@@ -36,15 +36,23 @@ jobs:
...
         echo "::add-path::`echo $HOME/opt/xPacks/@xpack-dev-tools/riscv-none-embed-gcc/*/.content/bin`"
+
+        # TI MSP430 GCC
+        wget http://software-dl.ti.com/msp430/msp430_public_sw/mcu/msp430/MSPGCC/8_3_0_0/exports/msp430-gcc-8.3.0.16_linux64.tar.bz2 -O /tmp/msp430-gcc.tar.bz2
+        tar -C $HOME -xaf /tmp/msp430-gcc.tar.bz2
         echo "::add-path::`echo $HOME/msp430-gcc-*_linux64/bin`"
-      
+
+        # ESP IDF
+        git clone https://github.com/espressif/esp-idf.git $HOME/esp-idf
```
pigrew, line 52: To make the download smaller, perhaps do a shallow clone `git --depth 1 clone $(URL) `. In the future, some particular release should be tagged. As a side note, it looks like they have one branch (4.1) that supports the beta hardware, and v4.2 will support production.

hathach: good idea, though this is purely for CI only. Since the repo currently doesn't include ESP-IDF as submodules since it is not simply mcu driver like others. And people probably has their own ESP-IDF setup somewhere already.

pigrew: Agreed; I feel bad for the CI servers doing extra work.

hathach: I will apply your suggestion in the next push, Faster CI is what we want as well :)

src/portable/espressif/esp32s2/dcd_esp32s2.c
hathach: funny thing, correct the typo make thing worse, esp32s2 doesn't pass enumeration now. It now only responds with exactly 1 packet. I better to rest now, probably made some typo/mistake in the progress :sweat_smile: :sweat_smile:

.github/workflows/build.yml
pigrew: Related to that, it looks like we can also use the `action/cache` module to cache copies of the MSP430 downloads. I don't know if that would quicken things or not.

hathach: Yeah, with Travis, cache is needed for downloaded file like msp430 gcc. Though for action i it is super fast, I guess it does have some caching scheme in the background

igrr: By the way, [there is a docker image](https://docs.espressif.com/projects/esp-idf/en/latest/esp32/api-guides/tools/idf-docker-image.html) of ESP-IDF which you could use for CI: `espressif/idf`. Typically downloading that image in CI is faster than cloning IDF plus installing the tools.

hathach:
Update: the msc and cdc example work perfect now, please test it out if you have time. I will start to add more freertos example for hid/midi for testing with those class later on. It may need several more push for clean up, changes.

src/portable/espressif/esp32s2/dcd_esp32s2.c
```diff
@@ -0,0 +1,750 @@
...
+                   USB_USBRSTMSK_M |
+                   USB_ENUMDONEMSK_M |
+                   USB_IEPINTMSK_M |
+                   USB_OEPINTMSK_M |
+                   USB_RESETDETMSK_M |
+                   USB_DISCONNINTMSK_M;
+
+    USB0.daintmsk |= USB_OUTEPMSK0_M | USB_INEPMSK0_M;
+    USB0.doepmsk |= USB_SETUPMSK_M | USB_XFERCOMPLMSK;
+    USB0.diepmsk |= USB_TIMEOUTMSK_M | USB_DI_XFERCOMPLMSK_M /*| USB_INTKNTXFEMPMSK_M*/;
```
hathach, line 128: @cr1901 while troubleshoting esp32s2, I notice that this **Received IN token when FIFO is empty** (USB_INTKNTXFEMPMSK_M) mask is not enabled at the global endpoint IN. I tried to enable it and it breaks the whole driver, same thing happens with stm32 (USB_OTG_DIEPMSK_ITTXFEMSK). 
https://github.com/hathach/tinyusb/blob/master/src/portable/st/synopsys/dcd_synopsys.c#L111
May be it got triggered when data is not ready or something 

However, I am curious how we got the interrupt when it is not enabled as current code !!! According to STM32 Interrupt tree, the mask is needed !!! Maybe I misread the datasheet due to long hour of debugging, If you still remember the detail, please help to explain it. If it is indeed needed we can enable the mask and fixed the issue with it (data not ready or endpoint not enabled I guess).

image

tools/build_all.py
hathach: ops, I just realize my mistake, sorry, change to esp32s2 now

.github/workflows/build.yml
hathach: docker is a great option, though, I am not familiar with docker, and spent too much time tweaking CI. I separated esp32s2 from other builds matrix and it build much quicker now. Happy with the result for the time being. We will visit the docker later on :)

hathach:
This PR is ready for review & merged, if anyone else could confirm the 2 example  `cdc_msc_freertos` & `hid_composite_freertos` work then I think we could merge this. There are probably more bugs, but we could fix it later on.

igrr:
I have tried cdc_msc_freertos, and occasionally I get the following issue when opening the CDC port:

```
Guru Meditation Error: Core  0 panic'ed (Interrupt wdt timeout on CPU0). 

Core  0 register dump:
PC      : 0x4008518a  PS      : 0x00050034  A0      : 0x40025849  A1      : 0x3ffbd610  
0x4008518a: dcd_int_handler at dcd_esp32s2.c:?

0x40025849: _xt_lowint1 at /Users/ivan/e/esp-idf/components/freertos/xtensa/xtensa_vectors.S:1105

A2      : 0x00000000  A3      : 0x040400a0  A4      : 0x800825b0  A5      : 0x4002aa40  
0x4002aa40: _frxt_int_enter at /Users/ivan/e/esp-idf/components/freertos/xtensa/portasm.S:119

A6      : 0x00000001  A7      : 0x3ffbd610  A8      : 0x80085209  A9      : 0x3ffbd5e0  
A10     : 0x00000007  A11     : 0x00000021  A12     : 0x00000001  A13     : 0x00000000  
A14     : 0x00000001  A15     : 0x3ffbd5e0  SAR     : 0x0000001a  EXCCAUSE: 0x00000005  
EXCVADDR: 0x00000000  LBEG    : 0x00000001  LEND    : 0x00000000  LCOUNT  : 0x40024e6c  
0x40024e6c: xt_highint4 at /Users/ivan/e/esp-idf/components/esp_system/port/esp32s2/dport_panic_highint_hdl.S:58

Core  0 was running in ISR context:
EPC1    : 0x400976e3  EPC2    : 0x00000000  EPC3    : 0x00000000  EPC4    : 0x4008518a
0x400976e3: uart_hal_write_txfifo at /Users/ivan/e/esp-idf/components/soc/src/hal/uart_hal_iram.c:35

0x4008518a: dcd_int_handler at dcd_esp32s2.c:?


Backtrace:0x40085187:0x3ffbd610 0x40025846:0x3ffbd640 0x400825b3:0x3ffc1f80 0x40081371:0x3ffc1fc0 0x4008791f:0x3ffc2000 0x4002a8bd:0x3ffc2020
0x40085187: dcd_int_handler at dcd_esp32s2.c:?

0x40025846: _xt_lowint1 at /Users/ivan/e/esp-idf/components/freertos/xtensa/xtensa_vectors.S:1105

0x400825b3: usbd_control_xfer_cb at ??:?

0x40081371: tud_task at ??:?

0x4008791f: usb_device_task at /private/tmp/lcovr/tinyusb/examples/device/cdc_msc_freertos/build/../src/main.c:117 (discriminator 1)

0x4002a8bd: vPortTaskWrapper at /Users/ivan/e/esp-idf/components/freertos/xtensa/port.c:143


```

"Interrupt watchdog" likely indicates that either we are constantly re-entering the interrupt, or 
have entered an infinite loop in the interrupt handler.

hathach:
@igrr thanks for testing, this is the same issue that @ladyada got as well. How long will WDT trigger if we don't reset it. The example doesn't  delay much to give time for idle hook to reset wdt.

igrr:
ESP-IDF has two watchdogs: interrupt watchdog and task watchdog.

Interrupt watchdog checks that FreeRTOS tick interrupt runs periodically. The default timeout is 300ms. If another interrupt enters an infinite loop, or an interrupt is triggered repeatedly (because the handler doesn't clear the interrupt status), or the code enters a portENTER_CRITICAL for a long time, interrupt watchdog will trigger. In this case it seems that USB interrupt starts getting triggered repeatedly, probably because some interrupt status bit is not getting cleared.

Task watchdog checks that a specified task periodically feeds the watchdog. By default, we enable task watchdog for the idle task, and the timeout is 5 seconds. So we need the application to yield to the idle task at least once in a while.

igrr:
Have tested `hid_composite_freertos`, no issue observed there.

hathach:
Thank you for your wdt explanation, I disable the SOF interrupt, which occurs very often at 1ms interval. Also TinyUSB stack doesn't make use of it for now. Please try again to see it is the one that causes wdt issue with repeatedly interrupts.

igrr:
Have pulled the latest branch, I'm afraid the behavior looks the same. Modified the panic handler to print out the latest DCD interrupt status and mask, it says:

DCD status: 0x040400a0
DCD mask: 0x208c3c12

hathach:
Thanks for testing it out.Can I ask which OS you are using (windows/linux/mac) @ladyada has the same issue on windows, but somehow I haven't got into any with Linux. BTW, do you see the TinyUSB MSC disk appear, and could you read the README.TXT file there when there is no issue. I will try with other PC that have around to reproduce this.

igrr:
I am testing on macOS now. I can see the MSC disk and can open README file, no issue there. The issue happens only when i connect to the CDC port (using miniterm.py)

hathach:
thanks for confirming, Limor actually got worse, she got instant reset when enumerating (Linux > macOS > windows :laughing: ).  I will try miniterm.py as well, have been always using minicom so far. 

BTW I tried with led_strip example on esp-idf to get the on-board WS2812 as indicator but it doesnt seem to work h. Do you have any hint for this 
ttps://github.com/espressif/esp-idf/tree/master/examples/peripherals/rmt/led_strip

igrr:
One known workaround is to add a delay at a certain spot in `read_rx_fifo`:

```c++
    case 0x06: { // Setup packet recvd
        ets_delay_us(300);               // <--- BUG? it is needed for some reason
        s_setup_phase = 0;
        ESP_EARLY_LOGV(TAG, "TUSB IRQ - setup_phase 0"); // new setup process
```

which seems to indicate that there is a race condition somewhere...

hathach:
> One known workaround is to add a delay at a certain spot in `read_rx_fifo`:
> 
> ```c++
>     case 0x06: { // Setup packet recvd
>         ets_delay_us(300);               // <--- BUG? it is needed for some reason
>         s_setup_phase = 0;
>         ESP_EARLY_LOGV(TAG, "TUSB IRQ - setup_phase 0"); // new setup process
> ```
> 
> which seems to indicate that there is a race condition somewhere...

oh, this particular setup issue is fixed by @cr1901 , he did mentioned it above. When connecting to CDC, it does send out an control transfer, so it is probably relevant. I will sync dcd with current stm32 driver to see if that helps :)

suda-morris:
> BTW I tried with led_strip example on esp-idf to get the on-board WS2812 as indicator but it doesnt seem to work h. Do you have any hint for this

Which GPIO that consumed by the LED, it should have the silk to indicate (e.g. it's GPIO18 from my side, it could also be GPIO17 in some early boards).

hathach:
> > BTW I tried with led_strip example on esp-idf to get the on-board WS2812 as indicator but it doesnt seem to work h. Do you have any hint for this
> 
> Which GPIO that consumed by the LED, it should have the silk to indicate (e.g. it's GPIO18 from my side, it could also be GPIO17 in some early boards).

Ah, right, You are spot on, it is GPIO17. I didn't notice the silk and only use GPIO18 as specified by the schematics. It works now, time to update the BSP :)

igrr:
In the final version of the board it will be GPIO18, i think. Sorry, you got the not-really-final revision :)

igrr:
I have tried to apply https://github.com/hathach/tinyusb/commit/d4302dacc584c4488cf1905caa822f96bd64930d to the esp32 driver, but unfortunately this doesn't fix the problem, still getting an interrupt loop with

DCD status: 0x040400a0
DCD mask: 0x208c3c12

hathach:
> In the final version of the board it will be GPIO18, i think. Sorry, you got the not-really-final revision :)

No problem at all, as long as I worked.

> I have tried to apply [d4302da](https://github.com/hathach/tinyusb/commit/d4302dacc584c4488cf1905caa822f96bd64930d) to the esp32 driver, but unfortunately this doesn't fix the problem, still getting an interrupt loop with
> 
> DCD status: 0x040400a0
> DCD mask: 0x208c3c12

Thanks for your info, I couldn't reproduce this. So it is kind of hard to troubleshoot, I will try my best to fix it, will ping you if there is update.

igrr:
By the way, do you know of any capture/replay tools for USB? I was thinking, if I could somehow capture the USB packets (into a .pcap file), send the file to you, would you be able to reproduce this issue?

hathach:
> By the way, do you know of any capture/replay tools for USB? I was thinking, if I could somehow capture the USB packets (into a .pcap file), send the file to you, would you be able to reproduce this issue?

It could be helpful, yes. most people I know use Beagle USB analyzer https://www.totalphase.com/beagle-products/ it is quite popular and run on all OS. Though I use an less known analyzer Lecroy Mercury https://teledynelecroy.com/protocolanalyzer/usb/mercury-t2 since it is what I used to work with in the previous company. Anything you find available is fine.

Meanwhile I am trying to set up another PCs to test with.

src/portable/espressif/esp32s2/dcd_esp32s2.c
cr1901: @hathach Interesting, I agree it should be fixed. Indeed, I never enabled this interrupt on STM32. I don't have ESP32 hardware to test.

Are you saying on ESP32, the "Received IN token when FIFO is empty" interrupt triggers anyway _without_ the mask being enabled (whereas since I don't explicitly enable on STM32, the interrupt never fires)? And when you _do_ enable the mask on STM32, the driver _also_ breaks?

hathach: @cr1901 yeah, when I enabled it on STM32 it cannot pass enumerataion, it failed somewhere. I didn't look at the detail. I think I should ask you first before troubleshoot it when esp32s2 is more reliable.  Though I think on both STM32 and ESP, the interrupt is fired even without this mask is enabled.

cr1901: >Though I think on both STM32 and ESP, the interrupt is fired even without this mask is enabled.

Of course it does... :P. Why would documentation match reality ;)?

Worth opening an issue? I can check sometime after April 15th on H7 board. I still have not replaced my fried F4 board yet.

hathach: I open issue for it https://github.com/hathach/tinyusb/issues/331 . Feel free to check it whenever you have time, since we have absolute no issue with current code  :D. H7 is probably the same as F4, only faster :D.

igrr:
I haven't managed to capture in .pcap format yet, but here is the decoded log i got from Saleae Logic, in text format. I open the CDC port around 17.4 s time, and then around 17.8s ESP32 restarts.

Judging by the density of the trace, there is quite some traffic after the port is opened:
Screenshot 2020-04-09 at 19 15 12
[cdc_log.zip](https://github.com/hathach/tinyusb/files/4457648/cdc_log.zip)

hathach:
@igrr thanks, don't worry about the pcap, it is not that useful, I am able to guess (kind of) the issue and revising usb manual and code now. Though it can be helpful if you could dump the internal debug log of TinyUSB here by
-  uncomment and change CFG_TUSB_DEBUG to 2 https://github.com/hathach/tinyusb/blob/add-esp32s2/examples/device/cdc_msc_freertos/src/tusb_config.h#L51
-  bump the USBD_STACK_SIZE to 4096 so that the printf won't overlfow the task https://github.com/hathach/tinyusb/blob/add-esp32s2/examples/device/cdc_msc_freertos/src/main.c#L59
- The compile will try to complain current code use "%lu" and "%lX" for uint32_t, just simply change it to %u and %X. It is due to the fact that armgcc define uint32_t as unsigned long. I have fixed it, but my current code is a mess now.

Once you done just run with monitor, and paste your Long here from start to the point when it is stuck.

igrr:
Unfortunately enabling the log makes the issue disappear. CDC works nicely with logging enabled. I suppose that logging acts in a similar way as the `ets_delay(300)` workaround I have mentioned above.

I also had to increase CDC_STACK_SZIE (sic.) to 4096, since configMINIMAL_STACK_SIZE doesn't leave much room for nested function calls, and stack overflow happens. By the way, i recommend setting `CONFIG_FREERTOS_WATCHPOINT_END_OF_STACK=y` when debugging — it helps detect stack overflows when they happen, rather than later when FreeRTOS goes about task switching.

hathach:
> Unfortunately enabling the log makes the issue disappear. CDC works nicely with logging enabled. I suppose that logging acts in a similar way as the `ets_delay_us(300)` workaround I have mentioned above.
> 
> I also had to increase CDC_STACK_SZIE (sic.) to 4096, since configMINIMAL_STACK_SIZE doesn't leave much room for nested function calls, and stack overflow happens. By the way, i recommend setting `CONFIG_FREERTOS_WATCHPOINT_END_OF_STACK=y` when debugging — it helps detect stack overflows when they happen, rather than later when FreeRTOS goes about task switching.
> 
> Edit: i've also tried setting UART baud rate to 2Mbps to make logging take less time, but that still introduces enough delay somewhere to prevent the issue from happening.

thanks for trying, I am not pretty sure where the issue could be now. Will try to fix it asap. Thanks for the tip about `CONFIG_FREERTOS_WATCHPOINT_END_OF_STACK`, I will add it into the sdkconfig.default. The stack task is configured as minimal  as possible to get the example running on much smaller chip (16KB of SRAM ).

cr1901:
@igrr

>CDC works nicely with logging enabled.

At this point I would selectively start removing some of the function calls/statements inside the logging function to find a minimum set of changes (relative to what you _do_ have working) that causes the issue. You can iterate from there to create a minimum working example like I did for [this](https://e2e.ti.com/support/microcontrollers/msp430/f/166/p/845259/3128051#3128051) _wonderful_ issue when debugging MSP430 :).

Also, this is some amusing crosstalk... I know the maintainer of usbrply and have collaborated with him on [open-tl866](https://github.com/ProgHQ/open-tl866). He says he's up to help if you need help getting `usbrply` up and running, and can most easily be found under the name `digshadow` on Freenode IRC.

hathach:
> @igrr
> 
> > CDC works nicely with logging enabled.
> 
> At this point I would selectively start removing some of the function calls/statements inside the logging function to find a minimum set of changes (relative to what you _do_ have working) that causes the issue. You can iterate from there to create a minimum working example like I did for [this](https://e2e.ti.com/support/microcontrollers/msp430/f/166/p/845259/3128051#3128051) _wonderful_ issue when debugging MSP430 :).
> 
> Also, this is some amusing crosstalk... I know the maintainer of usbrply and have collaborated with him on [open-tl866](https://github.com/ProgHQ/open-tl866). He says he's up to help if you need help getting `usbrply` up and running, and can most easily be found under the name `digshadow` on Freenode IRC (`/join #proghq` and `/join #usbrply`).

Thanks William, I am able to guess where is the issue, I am experiencing here and there. Actually, I made the problems worse so that I could always occur on my machine then troubleshoot from there.

igrr:
@hathach latest commit works on my side without workarounds!

igrr:
One thing I wanted to ask, is there a blocking alternative to tud_cdc_read (which would block on a queue in FreeRTOS case)? The following code in the example:

```c++
  while ( 1 )
  {
    if ( tud_cdc_connected() )
    {
      if ( tud_cdc_available() )
      {
        // read and echo back
      }
    }
  }
}

```
is an anti-pattern for an RTOS environment. I understand that TinyUSB also needs to run on bare metal, but do you think some queue-based blocking functions could be added?

hathach:
> @hathach latest commit works on my side without workarounds!

Yeah !! Finally we can wrap this up, once Adafruit team @ladyada and @tannewt confirms it works on their end as well.

> One thing I wanted to ask, is there a blocking alternative to tud_cdc_read (which would block on a queue in FreeRTOS case)? The following code in the example:
> 
> ```c++
>   while ( 1 )
>   {
>     if ( tud_cdc_connected() )
>     {
>       if ( tud_cdc_available() )
>       {
>         // read and echo back
>       }
>     }
>   }
> ```
> 
> is an anti-pattern for an RTOS environment (especially in a configMAX_PRIORITIES-2 task!). I understand that TinyUSB also needs to run on bare metal, but do you think some queue-based blocking functions could be added?

RTOS sync is already in-place, cdc data use internal fifo (size = CFG_TUD_CDC_RX_BUFSIZE)
https://github.com/hathach/tinyusb/blob/master/src/common/tusb_fifo.h#L46
Which also use rtos mutex when an rtos is used. Therefore it could prevent conflict with isr pushing item.

Ah, I just realized you may mean a different thing in regard of the data available. There is an rx callback you can use to signal/post semaphore for a task to start reading. Most of stock example is very minimal. Let's me know if you still have question.

```
// Invoked when CDC interface received data from host
void tud_cdc_rx_cb(uint8_t itf)
{
  (void) itf;
}
``` 

PS: I will start to push update for clean up

igrr:
Yes, i think tud_cdc_rx_cb is okay — it can put characters into a queue/ring buffer/message buffer and then the task can do blocking reads on the buffer. Thanks for the tip.

hathach:
> Yes, i think tud_cdc_rx_cb is okay — it can put characters into a queue/ring buffer/message buffer and then the task can do blocking reads on the buffer. Thanks for the tip.

@igrr could you confirm the amount of dedicated SRAM for USB FIFO of esp32s2, on stm32 implementation, it is different depending on the series, could range from 512 to 4KB bytes

image

Spritetm:
The dedicated packet buffer shared between all FIFOs is 1KBytes. It can be dynamically partitioned so the amount each FIFO gets is customizable.

Spritetm:
For completeness sake, these are the rest of the parameters:
usbdc_esp32s2

hathach:
> For completeness sake, these are the rest of the parameters:
> usbdc_esp32s2

Thanks @Spritetm that is very helpful, I notice the `Number of Device Mode active IN Endpoints including Control Endpoint 0` = 5, does that means out of 7 IN endpoints (including EP0), there are only max 5 IN that can be enabled ?

Spritetm:
...possibly? Sorry, it has been ages since I looked at these specs, I'm not 100% sure about their definitions anymore, but it should be in the databook. (I assume you have access to the dwc-otg databook? Otherwise I can look it up.)

hathach:
> ...possibly? Sorry, it has been ages since I looked at these specs, I'm not 100% sure about their definitions anymore, but it should be in the databook. (I assume you have access to the dwc-otg databook? Otherwise I can look it up.)

Unfortunately, I don't have that access, I look at stm32 user manual the entire time assuming they are the same, which is quite true so far :wink:

Spritetm:
Up to a point, probably. The DWC OTG IP is pretty configurable, I imagine STM 'flattened' the databook to exclude the options that are not in their chips; we plan to do mostly the same for the ESP32S2 IIRC.

The databook agrees with your guess, btw. It specifically states that this affects the amount of device mode TX fifos in the IP.

hathach:
> Up to a point, probably. The DWC OTG IP is pretty configurable, I imagine STM 'flattened' the databook to exclude the options that are not in their chips; we plan to do mostly the same for the ESP32S2 IIRC.
> 
> The databook agrees with your guess, btw. It specifically states that this affects the amount of device mode TX fifos in the IP.

thanks for confirmation, to sum up out of 7 endpoints EP0-EP6 we can use up to
- 7 OUT: EP0 - EP6 
- 5 IN: and since EP0 in must always available, there is only 4 lef for class drirver

@tannewt the number of IN endpoints can limit some certain classes usaged in circuitpython.

hathach:
@igrr I clean up the code and pushed the update, please double check it again whenever you could. Best to do a fullclean just in case it caches something. If it works, then we can mark this PR as ready again.

Spritetm:
@hathach To my understanding this is correct. Also, can you elaborate a bit about how that would limit you? If I know the use cases, I can perhaps advocate for more IN endpoints in future silicon.

hathach:
> @hathach To my understanding this is correct. Also, can you elaborate a bit about how that would limit you? If I know the use cases, I can perhaps advocate for more IN endpoints in future silicon.

Yeah, I am happy to do so. Each USB class require a certain number of endpoints, most often we are short of IN endpoint  than OUT.
- an CDC class instance requires 2 endpoint IN (notification + data in) and 1 endpoint out (Data out).
- MSC requires 1 endpoint IN and 1 endpoint out
- MIDI requires 1 endpoint In and 1 endpoint out
- HID requires 1 endpoint IN, (endpoint OUT is optional)
- USBNET is an variant of CDC therefore also need 2 IN and 1 OUT.

Therefore it does limit the number of classes we can bundle into a composite device :). STM32 also has family with lower number of endpoints supported as well. As far I could guess, your config is about 5 (including EP0). Still decent enough though :) 

image

igrr:
@hathach Latest version still works on my side! Have tested on Windows and macOS.

src/portable/espressif/esp32s2/dcd_esp32s2.c
```diff
@@ -0,0 +1,751 @@
...
+
+  USB0.gahbcfg |= USB_NPTXFEMPLVL_M | USB_GLBLLNTRMSK_M; // Global interruptions ON
+  USB0.gusbcfg |= USB_FORCEDEVMODE_M;                    // force devmode
+
+  USB0.gotgctl &= ~(USB_BVALIDOVVAL_M | USB_BVALIDOVEN_M | USB_VBVALIDOVVAL_M); //no overrides
+#ifdef CONFIG_IDF_TARGET_ESP32S2BETA // needed for beta chip only
+  //C. chip 7.2.2 hack
+  ESP_LOGV(TAG, "DCD init - chip ESP32-S2 beta hack");
+  USB0.gotgctl = (0 << USB_BVALIDOVVAL_S); //B override value
+  ets_delay_us(20);
```
hathach, line 177: @igrr since removing delay at the end of this function work. Maybe remove this couple of delay_us(20) also work as well. However, there is a any comment on this particular code with BETA mcu. Since I don't have a beta mcu (I am not sure if I have one). Would you mind confirming whether we could safely remove this with Beta, or Beta requires this to work.

hathach:
> @hathach Latest version still works on my side! Have tested on Windows and macOS.

Yeaah, let's wait for Adafruit team to confirm as well. I re-mark this PR as Ready.

src/portable/espressif/esp32s2/dcd_esp32s2.c
Spritetm: From what I remember, beta needed this as there was a bug in the phy logic that made the chip reset as soon as it transmitted anything. Setting the B override value made it ignore resets (any resets, generated by the faulty logic or not), which 'fixed' the problem well enough to test usb with it. For the final silicon, this is not needed.

hathach: AH thanks, then we leave it here, however, do you know any ID register we can check to know it is a beta mcu and run this walkaround instead of depending on user to define the `CONFIG_IDF_TARGET_ESP32S2BETA` macro. If not then never mind, we can leave it as it is.

Spritetm: Within esp-idf, this define is generated (in sdkconfig.h, which is only indirectly included here... could use a direct inclusion tbh) when the user compiles for the ESP32S2. As firmwares for the beta and the final silicon are not interchangable in the first place (iirc, @igrr correct me on that if I'm wrong there) there is no need for runtime figuring-this-out-ness anyway. 

(Also, do note that the beta silicon run was very small and software support for it is not in mainstream esp-idf, as such you may consider phasing out support for it alltogether somewhere in the future.)

src/portable/espressif/esp32s2/dcd_esp32s2.c
```diff
@@ -0,0 +1,751 @@
...
+      ESP_EARLY_LOGV(TAG, "TUSB IRQ - RX : Setup packet done");
+      USB0.out_ep_reg[epnum].doeptsiz |= USB_SUPCNT0_M;
+      break;
+
+    case 0x06: { // Step1: Setup data packet received
+      volatile uint32_t *rx_fifo = USB0.fifo[0];
+
+      // We can receive up to three setup packets in succession, but
+      // only the last one is valid. Therefore we just overwrite it
+      _setup_packet[0] = (*rx_fifo);
```
hathach, line 575: @cr1901 I found that the number of setup packet in doeptsiz of esp32 is inconsistent. It can still correctly extract the setup packet and have the SETUP count = 3 !!! I found that just simply overwrite the setup[0-1] with latest from rx_fifo work best and have the behave the same as specs ( process the last setup) still consume less memory 4x2 vs 4x6 of `_setup_packet`

src/portable/espressif/esp32s2/dcd_esp32s2.c
```diff
@@ -0,0 +1,751 @@
...
+    uint16_t queued_len;
+    uint16_t max_size;
+    bool short_packet;
+} xfer_ctl_t;
+
+static const char *TAG = "TUSB:DCD";
+static intr_handle_t usb_ih;
+
+
+static uint32_t _setup_packet[2];
```
hathach, line 69: reduce setup_packet by simply overwrite it.

src/portable/espressif/esp32s2/dcd_esp32s2.c
hathach: Ah thanks, got it, let's just leave it as it for now, we could safely remove it later on when all beta users got their hand on production.

igrr: Can safely remove the code related to the beta chip. In case we need it, can restore from git history.

ladyada:
tested, now working great on windows 10 here :) nice work everyone!

src/portable/espressif/esp32s2/dcd_esp32s2.c
cr1901: @hathach I say do the optimization for now, and if there are setup packet handling issues, revert this change and see if the bug is fixed.

I plan on leaving dcd_synopsys as-is for now

src/portable/espressif/esp32s2/dcd_esp32s2.c
hathach: @igrr Ok, let's remove it then :)

src/portable/espressif/esp32s2/dcd_esp32s2.c
hathach: @cr1901 on paper it should, but I found out it is not, it still 3 (max) and still received an correct SETUP packet, which cause the racing issue as discussed. OK we will leave the stm32 dcd as it is for now. Should it have some kind of issues with control transfer, we could apply this tip if needed :)

cr1901: @hathach Change the following [line](https://github.com/hathach/tinyusb/pull/319/files#diff-371e8160b188a80c1a3821c75d194adfR334) from:

```
USB0.out_ep_reg[epnum].doeptsiz = USB_PKTCNT0_M |
        ((xfer->max_size & USB_XFERSIZE0_V) << USB_XFERSIZE0_S);
```

to

```
USB0.out_ep_reg[epnum].doeptsiz |= USB_PKTCNT0_M |
        ((xfer->max_size & USB_XFERSIZE0_V) << USB_XFERSIZE0_S);
```

Note the ORing. See if the bug goes away and you can save all 3 setup packets again. This seems quite similar to a [bug](https://github.com/hathach/tinyusb/commit/d4302dacc584c4488cf1905caa822f96bd64930d) I fixed from my original STM32 port.

hathach:
Since it works with all people on all OS. We can wrap this up now (Finally Phew). Thank you everyone for participating in the PR 🎉🎉

UnexpectedMaker:
Congrats on the fixes and merge!
So until this is merged into IDF, are there any manual steps published to "hack it in"?
I'd love to see if my ProS2 works with this :)

src/portable/espressif/esp32s2/dcd_esp32s2.c
hathach: @cr1901 ooppps, I think you mentioned it at the first review. I did a bi-sect but man, that ORing is difficult to spot !!! :eyes: Shame on me. I have a good excuse those, the indent mess up the meld side-by-side comparison. Thanks, I will make an pr fix for this.

cr1901: >I did a bi-sect but man, that ORing is difficult to spot !!! 👀 Shame on me.

No shame at all... I mean, I was the one who wrote the bug in the first place :P. I wrote my initial comments while still sleepy. It took me an hour to realize: "wait... I've seen this before". Anyways I left feedback/requested changes on your PR. I would like to merge the two as well, but hope there's some consistency in register names (or an easy way to define the reg names to vendor-neutral names).

hathach: Yeah, thanks for pointing it out, we will merge them together later on :).

PR #320, opened 2020-04-02
Implement alternate settings, mostly handled by class driver. (draft)

pigrew:
This is an alternative to #317. Both will work, both have advantages.

This patch lets the class driver handle SET_INTERFACE request, instead of usbd. This also removes the need for storing the active interface for all interfaces, as the class driver keeps track in its own data structure. One functional change is this would stall EP0 when SET_INTERFACE is not supported by the driver. This is allowed by the USB spec, but I don't know if it would create any host incompatibilities. Binary size is reduced by 20 bytes (vs master) for the usbtmc example (which doesn't use alternate interfaces), due to the SET_INTERFACE handler being removed.

(The example stubs I put in the net class driver should be removed before merging. They'll break the class driver.)


hathach:
thank you for the PR, I will check out this issue as soon as I could (there is quite a bit of PR running in parallel now).  Just want to tell you the tip, that we can create a DRAFT PR when submitting. 

image

It will display as gray PR, and cannot be merged until marked as `Ready for review`, here is my example https://github.com/hathach/tinyusb/pull/319

Too bad, you cannot mark an normal PR as draft, it is once way ticket :D

pigrew:
@hathach, I'll look for it next time, though I don't think I can do a draft while posting it to `hathach/tinyusb`. I might only be able to post my drafts to `pigrew/tinyusb`, where it wouldn't get as much visibility.

I realized that this patch doesn't work since there isn't a way to set DTOG to zero from the class driver. See https://github.com/hathach/tinyusb/pull/317#issuecomment-609979240

hathach:
@pigrew hmm that is weird you can create draft PR, maybe there is some setting I have to configure !!! I will look it up later on.

hathach:
github allow to convert normal to draft under the review session, just found out about this yesterday
image

src/class/net/net_device.c
```diff
@@ -44,6 +44,7 @@ typedef struct
   uint8_t ep_notif;
   uint8_t ep_in;
   uint8_t ep_out;
+  uint_fast8_t active_alt_setting;
```
hathach, line 47: please use uint8_t, I like the type is exactly fixed width on all platform.

pigrew: Actually, I'll remove all changes to that file before merging. (The changes don't work, anyway)

src/device/usbd.c
```diff
@@ -83,6 +83,7 @@ typedef struct {
   bool (* control_complete ) (uint8_t rhport, tusb_control_request_t const * request);
   bool (* xfer_cb          ) (uint8_t rhport, uint8_t ep_addr, xfer_result_t event, uint32_t xferred_bytes);
   void (* sof              ) (uint8_t rhport);
+  uint8_t (* get_alt_setting ) (uint8_t rhport, tusb_control_request_t const * request); //< (optional) Returns active alternate setting of an opened interface.
```
hathach, line 86: I have a better idea for getting Alt interface, we will define an common usbd_interface_t which contains
- itf_num
- itf_alt
- ep_num
- ep[]

each interface will start with this first, I have been wanting to do this (but being too busy or too lazy), that's why you see all class driver itf start with these. That will also allow to remove the itf2drv and ep2drv mapping. Let's me make an PR for that. This will help with adding Alt Interface.

pigrew: At the moment, I don't understand what you're suggesting, but I'll await the PR, and I'm sure all will become clear.

hathach: yeah, wait for it ...

pigrew:
@hathach 

I just merged in master, and also removed the example changes to netd.

PR #321, opened 2020-04-03
Misc. tweaks, and fix for USBTMC descriptors without an INT EP

pigrew:
This is a set of random fixes (from my old development branch):

* gcc interprets values as signed in enums, and MAX_DELAY was causing a pedantic warning. Might as well get rid of the issue
* USBTMC descriptors were not properly set when the interrupt endpoint was disabled.
* Typo of usmtmc instead of usbtmc.
* Make some `#defines` functions use the `do{}while(0)` syntax, so they are treated more like normal C functions.

### DIFF
--- a/src/osal/osal.h
+++ b/src/osal/osal.h
@@ -40,8 +40,9 @@ enum
 {
   OSAL_TIMEOUT_NOTIMEOUT    = 0,      // return immediately
   OSAL_TIMEOUT_NORMAL       = 10,     // default timeout
-  OSAL_TIMEOUT_WAIT_FOREVER = 0xFFFFFFFFUL
 };
+
+static const uint32_t  OSAL_TIMEOUT_WAIT_FOREVER = 0xFFFFFFFFUL;
 
 #define OSAL_TIMEOUT_CONTROL_XFER  OSAL_TIMEOUT_WAIT_FOREVER
 
--- a/src/osal/osal.h
+++ b/src/osal/osal.h
@@ -36,13 +36,12 @@
 
 #include "common/tusb_common.h"
 
-enum
-{
-  OSAL_TIMEOUT_NOTIMEOUT    = 0,      // return immediately
-  OSAL_TIMEOUT_NORMAL       = 10,     // default timeout
-};
-
-static const uint32_t  OSAL_TIMEOUT_WAIT_FOREVER = 0xFFFFFFFFUL;
+// Return immediately
+#define OSAL_TIMEOUT_NOTIMEOUT (0)
+// Default timeout
+#define OSAL_TIMEOUT_NORMAL       (10)
+// Wait forever
+#define OSAL_TIMEOUT_WAIT_FOREVER  (UINT32_MAX)
 
 #define OSAL_TIMEOUT_CONTROL_XFER  OSAL_TIMEOUT_WAIT_FOREVER
 
--- a/examples/device/usbtmc/src/usb_descriptors.c
+++ b/examples/device/usbtmc/src/usb_descriptors.c
@@ -109,18 +109,18 @@ uint8_t const * tud_hid_descriptor_report_cb(void)
      TUD_USBTMC_IF_DESCRIPTOR(_itfnum, _bNumEndpoints,  /*_stridx = */ 4u, TUD_USBTMC_PROTOCOL_USB488), \
      TUD_USBTMC_BULK_DESCRIPTORS(/* OUT = */0x01, /* IN = */ 0x81, /* packet size = */USBTMCD_MAX_PACKET_SIZE)
 
-#if defined(CFG_TUD_USBTMC_ENABLE_INT_EP)
+#if CFG_TUD_USBTMC_ENABLE_INT_EP
 // Interrupt endpoint should be 2 bytes on a FS USB link
 #  define TUD_USBTMC_DESC(_itfnum) \
      TUD_USBTMC_DESC_MAIN(_itfnum, /* _epCount = */ 3), \
      TUD_USBTMC_INT_DESCRIPTOR(/* INT ep # */ 0x82, /* epMaxSize = */ 2, /* bInterval = */16u )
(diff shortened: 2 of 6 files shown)


src/osal/osal.h
```diff
@@ -40,9 +40,10 @@ enum
 {
   OSAL_TIMEOUT_NOTIMEOUT    = 0,      // return immediately
   OSAL_TIMEOUT_NORMAL       = 10,     // default timeout
-  OSAL_TIMEOUT_WAIT_FOREVER = 0xFFFFFFFFUL
 };
 
+static const uint32_t  OSAL_TIMEOUT_WAIT_FOREVER = 0xFFFFFFFFUL;
```
hathach, line 45: I think it is better to have OSAL_TIMEOUT_WAIT_FOREVER as macro than constant

pigrew:
@hathach, updated per your request.

Somewhat related, I learned about `uint_fast16_t` the other day. It evaluates to uint16_t on 16-bit micros, but 32-bit on ARM. It should be the appropriate thing to use for loop counters (with limited range), or maybe the compiler automatically uses it, and we don't need to worry about it. I'm not sure.

src/osal/osal.h
pigrew: Modified.

cr1901:
@hathach I'm still uncomfortable with `OSAL_TIMEOUT_WAIT_FOREVER` having a different type (`uint32_t`) than the other `OSAL_TIMEOUT_*` constants (`enums` are `int`, which is not only _signed_, but 32-bit on ARM/RISCV, 16-bit on MSP430).

I don't have any concrete examples, but I feel like this could lead to some unexpected issues down the road.

pigrew:
@cr1901, very good point. Perhaps the `enum` shouldn't be used at all, since we can't declare what type it is.

Can we use:

```
uint_fast16_t timeout;
static const uint_fast16_t OSAL_TIMEOUT_NOTIMEOUT  = 0U;
static const uint_fast16_t OSAL_TIMEOUT_NORMAL       = 10U;
static const uint_fast16_t OSAL_TIMEOUT_WAIT_FOREVER  = UINT_FAST16_MAX;
```

Or is there a better method?

cr1901:
I don't think the enum is the problem, honestly, as long as the constants contained within have a consistent type for the platform in question.

What I would _personally_ do is (don't bother w/ the fast types. _I_ personally don't care, but you get odd looks from ppl when you use them IME :P):

```
typedef enum osal_timeout {
  OSAL_TIMEOUT_NOTIMEOUT  = 0,
  OSAL_TIMEOUT_NORMAL       = 10,
  OSAL_TIMEOUT_WAIT_FOREVER  = INT_MAX
} osal_timeout_t;
```

And any function that requires a timeout will accept an `osal_timeout_t` in its arguments.

hathach:
> @hathach, updated per your request.
> 
> Somewhat related, I learned about `uint_fast16_t` the other day. It evaluates to uint16_t on 16-bit micros, but 32-bit on ARM. It should be the appropriate thing to use for loop counters (with limited range), or maybe the compiler automatically uses it, and we don't need to worry about it. I'm not sure.

> @cr1901, very good point. Perhaps the `enum` shouldn't be used at all, since we can't declare what type it is.
> 
> Can we use:
> 
> ```
> uint_fast16_t timeout;
> static const uint_fast16_t OSAL_TIMEOUT_NOTIMEOUT  = 0U;
> static const uint_fast16_t OSAL_TIMEOUT_NORMAL       = 10U;
> static const uint_fast16_t OSAL_TIMEOUT_WAIT_FOREVER  = UINT_FAST16_MAX;
> ```
> 
> Or is there a better method?

`uint_fast16_t` is new to me as well, though it could raise questions from people. I still prefer the fixed width numbered, it is more deterministic. The timeout constant need to be at least 32-bit (can be 64-bit for 64-bit mcu), shouldn't be 16-bit. We should either declare them all as enum or define.

cr1901:
>The timeout constant need to be at least 32-bit (can be 64-bit for 64-bit mcu), shouldn't be 16-bit.

Ahh good to know. Then `enum`s are out of the question then, since they can be 16-bit (as that is a legal width of an `int`).

>I still prefer the fixed width numbered, it is more deterministic.

Almost every system should have `uint32_t`, even DSPs with weird `CHAR_BIT`. If by some miracle tinyusb gets ported to a DSP without `uint32_t`, then `uint_least32_t` (which is guaranteed to exist in C99) can be used.

hathach:
> > The timeout constant need to be at least 32-bit (can be 64-bit for 64-bit mcu), shouldn't be 16-bit.
> 
> Ahh good to know. Then `enum`s are out of the question then, since they can be 16-bit (as that is a legal width of an `int`).
> 
> > I still prefer the fixed width numbered, it is more deterministic.
> 
> Almost every system should have `uint32_t`, even DSPs with weird `CHAR_BIT`. If by some miracle tinyusb gets ported to a DSP without `uint32_t`, then `uint_least32_t` (which is guaranteed to exist in C99) can be used.

yeah, I think it is all thanks to the short-enum, however, gcc should still force it to 32-bit due to the FOREVER assign to max uint32 UL. But yeah, it is probably better to just use define here.

pigrew:
@hathach,

What is the consensus? Do we merge this PR as-is?

Go back to what I had proposed with "static const uint32_t" for the three constants, since that very explicitly states their type?

Or something else?

hathach:
> @hathach,
> 
> What is the consensus? Do we merge this PR as-is?
> 
> Go back to what I had proposed with "static const uint32_t" for the three constants, since that very explicitly states their type?
> 
> Or something else?

Please change it to using all macro define, since I don't like to put variable declaration in header, even it is static.

```
#define OSAL_TIMEOUT_NOTIMEOUT 0
#define OSAL_TIMEOUT_NORMAL       10
#define OSAL_TIMEOUT_FOREVER     UINT32_MAX
```

pigrew:
@hathach,
> Please change it to use all macro define, since I don't like to put variable declaration in header, even it is static.

Done. (I'm now waiting on CI)

PR #323, opened 2020-04-04
Cache MSP430 toolchain and NPM modules.

pigrew:
This uses github actions to cache xPack and MSP430 downloads.

It looks like there's no great way to derive a cache hash key from the downloads, so I'm manually setting the hash keys with a `-0` serial number. When package revisions are updated, the serial number should be incremented. I explicitly mark the xPack revisions to use so that multiple versions of an xPack package won't be installed simultaneously (the cached and the latest). 

This PR also limits the depth of git submodule checkouts to `1`. The LWIP repository can't handle checking out arbitrary commits with a depth of `1`, so it has to be special-cased.  Referencing a tagged commit (a release commit), would work, but the latest release is 16 months ago. Except for this special case, the `checkout` action could automatically also checkout submodules, by setting `submodules` to true.

Overall, this reduces the CI runtime by somewhat more than 1 minute (if the the toolchains are already in the cache). I don't know if it's worth the extra complexity or not.

This PR conflicts with the ESP32 changes, so I'm marking this as draft. I'll update this commit once ESP32 changes are merged.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -31,15 +31,6 @@ jobs:
     - name: Setup Python
       uses: actions/setup-python@v1
     
-    - name: Cache xPacks
-      uses: actions/cache@v1
-      env:
-        cache-name: cache-xpacks-modules
-      with:
-        path: ~/opt/xPacks # xpm cache files are installed in `~/opt/xPacks` on Linux.
-        # Update key with new serial number when packages are updated
-        key: xpm-${{ runner.os }}-build-${{ env.cache-name }}-0
-    
     - name: Cache MSP430 Toolchain
       id: cache-msp430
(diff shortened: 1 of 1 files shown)


hathach:
@pigrew I think it isn't worth the effort to cache the xpack dev, let's just install the latest version for xapck. That will help us to keep up with latest change from armgcc. For msp430 we can make use of the cache, since it needed to be updated manually anyway.

pigrew:
> @pigrew I think it isn't worth the effort to cache the xpack dev, let's just install the latest version for xapck. That will help us to keep up with latest change from armgcc. For msp430 we can make use of the cache, since it needed to be updated manually anyway.

Xpack has been removed from caching.

Issue #322, opened 2020-04-03
Callback from ISR?


tannewt:
Could we add a callback from the `dcd_isr` or indirectly when an event is queued to the buffer? Calling into TinyUSB and other polling tasks repeatedly adds up to a significant overhead. By having TinyUSB call back when queueing a task, we'll know exactly when it needs to be called and do so once we're back on the main thread of execution.

majbthrd:
Ideally, it seems undesirable for TinyUSB to callback every time it adds something to the event queue; I think what is needed is a single determination at the end of the ISR as to whether there is anything queued.

Knowing whether TinyUSB has queued anything should be as straightforward as adding something like this to ./src/device/usbd.c

bool usbd_events_are_queued(void)
{
  return !tu_fifo_empty(&_usbd_q->ff);
}

If the last thing done by any dcd_isr() was to call some routine that invoked a check like the above, that routine could then conditionally invoke a callback if events were queued.  There could be a weak declaration of the callback that did nothing for users that didn't need this mechanism.  However, users that did need it could use it to do whatever more efficient mechanism was needed (set a semaphore?) to cause a task to run tud_task().

pigrew:
Thanks for pointing out `dcd_isr`. I didn't use it in stm32... now I gotta go update the driver to use it.

The library was intended for use with a RTOS where TinyUSB was a task. The RTOS would handle the waiting on the event queue, so no polling is necessary. With FreeRTOS, `tud_task()` will handle events forever, and never return.

On the other hand, it sounds like you want to be able to wait on a semaphore/event_flags shared by usbtmc and other libraries (for micropython?), and you want to release that shared semaphore in TinyUSB.

The cleanest way might be to create a new "OSAL" header for yourself, so you can intercept the queueing functions. A more hackish way would be to add your callback at the end of the `src/device/usbd.c:dcd_event_handler(...)`. Note that this function can be called from interrupt handlers and also from normal code.

--

Currently the OSAL functions are implemented as inline functions in a `.h` file, and there's no config option to include a custom OSAL. I think that the solution to this issue would be to change the headers so that a custom OSAL could be selected.

In the NO_OS case, `tud_task()` loops over all events until there is no event left. Perhaps it's better for you if it were changed to handling a single event, and then returning true/false based on if it did any work.

tannewt:
@pigrew I think `dcd_isr` is just a samd thing. I'm not sure it applies to STM.

[...]

Issue #279, opened 2020-02-07
Notes from porting TinyUSB


rajivr:
Hi @hathach 

Belated Lunar New Year Greetings and thanks again for TinyUSB. :-)

I managed to get a basic port of TinyUSB working. In our case the DCD layer is in Rust. There are some issues that came up during the port that I wanted to share with you. 

I don't know a good way to create a PR for these changes, so I am providing links to the patches that we currently are maintaining in the tree.

1. We needed to remove `static inline` from `tu_edpt_*` functions. The patch is [here](https://gist.github.com/rajivr/556d63448acf39cd21e8bb97e0d401d3).

    These functions are called from Rust DCD code.

2. We needed to remove `TU_ATTR_WEAK` from a few functions. The patch is [here](https://gist.github.com/rajivr/25ba9904edc93c2e5fd527251f61ca70)

    What was happening in this case was that the linker was picking up the weak definition from TinyUSB instead of the function defined in our crate (Rust's equivalent of a library).

3. Introduce `tud_task_is_queue_empty()` function. The patch is [here](https://gist.github.com/rajivr/2c15302993eceefbfc3104358ffa1a9d)

    We introduced this function because in the OS main loop, we needed some way to determine when TinyUSB did not have any more work left to do, so we can safely execute `wfi` instruction.

4. Because Rust does not have native bitfield support, I had to a workaround by having a `dcd_edpt_open` function written in C that extracted out data from endpoint descriptor before calling Rust DCD function. The code is [here](https://gist.github.com/rajivr/bdbcf473ab288e9505cfe4e5fc19be23)

I hope this is going to be helpful to other TinyUSB/Rust users.

cr1901:
I'm not going to profile the static code (mainly because I don't know how :P), but I would rather see a Rust-api specific define (call it `TU_RUST_API`) that [changes](https://github.com/hathach/tinyusb/blob/7b20b80dc0694cfe49b63222f47a8c3d18906e41/src/common/tusb_compiler.h#L67) the definition of `TU_ATTR_WEAK` to nothing, e.g:

```
#ifdef TU_RUST_API
#define TU_ATTR_WEAK      
#else
#define TU_ATTR_WEAK                  __attribute__ ((weak))
#endif
```

And gates whether currently `static` functions remain `static`, e.g. in `tusb_types.h`:

```
#ifndef TU_RUST_API
static inline tusb_dir_t tu_edpt_dir(uint8_t addr)
{
[...]

PR #324, opened 2020-04-05
allow application to implement its own os

hathach:
after discussion in #322 , this allow application to provide its own cutom OS implementation if so desired. 

Note: the osal API() is not finalized nor clean yet. Since part of the API is used by host stack, which is way behind device stack for now.


hathach:
thank you both of you for reviewing.

cr1901:
@hathach A bit too late for me for me to comment, but will you still accept other OS ports in the future (c.f. #279)? Or will new OSes now be required to implement the custom interface?

hathach:
@cr1901 we will still accept more OS in the future if people submit PR making it into builtin support list. The custom is simply a way for people that want to alternate the implementation of existing os or testing out their new OS implementation without changing the stack at the time of testing.

PR #325, opened 2020-04-06
move irqhandler to application

### LINKED ISSUE #128: Add ability for application layer to define IRQ handler

hathach:
as discussed in #322 irq handler is moved to application, `tud_isr()` must be called to  forward the irqhandle to the stack. This PR is marked as draft and only modify the samd for reviewing purpose. If you guys agree with the changes, I will apply it to all other ports. Ports got changes so far

- [x] Microchip SAMD
- [x] Microchip SAMG
- [x] Nordic nRF5x
- [x] Nuvoton NUC120
- [x] Nuvoton NUC121/125/126
- [x] Nuvoton NUC505
- [x] NXP LPC3511
- [x] NXP LPC17/40
- [x] NXP Transdimension (18/ iMX RT)
- Sony CXD56, no need to change since it has its own interrupt installation
- [x] STM32 FSDev
- [x] STM32 Synopsys
- [x] TI MSP430
- [x] Fomu

@pigrew @tannewt @majbthrd 

will fix #128


majbthrd:
Letting the board support code map the board-specific IRQ handler name into the handler and @pigrew's notion of the extreme case of the app calling the handler to reduce latency both seem like non-trivial and worthwhile benefits.  It seems like a win to me.

A minor downside might be a tiny bit of extra CPU cycles and stack usage due to pushing and popping an extra LR onto the stack.  However, compilers are always improving and this may already be a non-issue.

pigrew:
(I think this needs more attention that I can pay in the next few hours, but here are my initial thoughts):

The patch here looks good, but I wonder if we _also_ need the "enable_interrupt" and "disable_interrupt" functions to be part of the BSP. Maybe not? Perhaps we can merge this patch, and see how it is used, and move the these functions to the BSP when/if there is a need? Are there any other libraries from which inspiration can be drawn?

>A minor downside might be a tiny bit of extra CPU cycles and stack usage due to pushing and popping an extra LR onto the stack. However, compilers are always improving and this may already be a non-issue.

For most of the targets, LTO is enabled, so the I'd expect there to be no change in object code.

hathach:
> Letting the board support code map the board-specific IRQ handler name into the handler and @pigrew's notion of the extreme case of the app calling the handler to reduce latency both seem like non-trivial and worthwhile benefits. It seems like a win to me.
> 
> A minor downside might be a tiny bit of extra CPU cycles and stack usage due to pushing and popping an extra LR onto the stack. However, compilers are always improving and this may already be a non-issue.

It is not much, hah, I am sure I have wasted a lot more cycles somewhere :D.   

> (I think this needs more attention that I can pay in the next few hours, but here are my initial thoughts):
> 
> The patch here looks good, but I wonder if we _also_ need the "enable_interrupt" and "disable_interrupt" functions to be part of the BSP. Maybe not? Perhaps we can merge this patch, and see how it is used, and move the these functions to the BSP when/if there is a need? Are there any other libraries from which inspiration can be drawn?

Indeed interrupt enable/disable are also relevant to this, but it is less common to see those changes. Let's just keep it as it is until it becomes a problems. This PR start with circuitpython and TI-OS and probably TockOS (if I remembered correctly) wanting to change it. So they are all practical problems.

hathach:
moved all the irq handler to bsp for all ports. That is a lots more work than it sounds. That would help if anyone could look at dcd implementation to see if I missed anything. I did hardware testing 1 board for each dcd though. 

Update: This PR should only be merged after the #319 (which is almost complete), that will allow circuitpython (and other projects) to start porting ESP32S2 without the need to update other ports to match this. Although it is minor work, it is still work to be done.

examples/device/board_test/Makefile
```diff
@@ -9,4 +9,10 @@ INC += \
 EXAMPLE_SOURCE += $(wildcard src/*.c)
 SRC_C += $(addprefix $(CURRENT_PATH)/, $(EXAMPLE_SOURCE))
 
+# board_test example is special example that doesn't enable device or host stack
+# This can cause some TinyUSB API missing, this hack to allow us to fill those API
+# to pass the compilation process
+CFLAGS +=	\
+	-D"tud_irq_handler(x)= " \
```
cr1901, line 16: Why does this cause symbols to be missing? It's not as nice-looking but I feel like an internal include guard (`$*INTERNAL_PREFIX)_IN_BOARD_TEST_EXAMPLE`) would indicate intent better.

cr1901:
LGTM once I get clarification on my question!

examples/device/board_test/Makefile
hathach: > So just so I understand... before this PR, TinyUSB had four special symbols: `tud_isr` and `tuh_isr`, `dcd_isr`, and `hcd_isr` (with [defines](https://github.com/hathach/tinyusb/commit/bbec47b647ab7b2e38644ffff60f45b54b791d42#diff-82aaa3cc625cca0e786868f5c6660f0cR51) relating the first two to the latter two).
> 
> Before this PR, `tud_isr` and `tuh_isr` were only used if a microcontroller supported _both_ host and device mode. If the microcontroller only supported device mode, `tud_isr` was ignored and we directly implemented the IRQ handler in the `dcd` implementations.

before this PR, tud_isr and tuh_isr still are the same as this PR, I just start doing this with port that is added later on after discussion on this matter in an issue

> As of this PR:
> 
> * `tuh_isr`/`hcd_isr` are left alone.

Yes, I will rename it when reworking with host. I am a bit lazy to pull out an host board for physical testing now. They should be renamed to _irq_hanlder() as well though.

> * `dcd_isr` was renamed to `dcd_irq_handler`.
> * `tud_isr` was renamed to `tud_irq_handler`.

Yes

> * `dcd files no longer implement the IRQ handler directly, even if they are device-only. It is the board file's responsibility to called `tud_irq_handler`.
> 
> Is this accurate?

Yes, it has another benefit as well, some mcus has the same IRQ Handler for host and device, when both enabled, it could just still work if application call both tud/tuh irq_handler

hathach: > Why does this cause symbols to be missing? It's not as nice-looking but I feel like an internal include guard (`$*INTERNAL_PREFIX)_IN_BOARD_TEST_EXAMPLE`) would indicate intent better.

The board_test example has `CFG_TUSB_RHPORT0_MODE = OPT_MODE_NONE` which cause the dcd_irq_handler() not compiled at all, it throw out lot errors. Instead of having macro to check if device is enabled for each bsp ( which is not really needed when application use the stack). I just hack it into command line define.

hathach:
> Looks good, though a note could be added to the porting.md file about interrupt handlers.
> 
> (I didn't test the code, but it looks correct.)

I tried my best, feel free to correct my bad English :stuck_out_tongue_winking_eye:

hathach:
merged, thank you everyone for reviewing the changes.

PR #326, opened 2020-04-06
GET/SET_CONFIGURATION for multiple configurations

majbthrd:
This PR cherry-picks aspects of #167.  That PR has several objectives; this has just one: allow GET_CONFIGURATION/SET_CONFIGURATION to facilitate multiple configurations.

Behavior should be as the USB spec says.  A zero configuration value is only accepted when in Address or Configured state.  A non-zero configuration value is accepted only when it matches the current configuration or when the device is in Address state.

Testing can be accomplished with [this suggested procedure](https://github.com/hathach/tinyusb/pull/167#issuecomment-609480555).

### DIFF
--- a/src/class/net/net_device.c
+++ b/src/class/net/net_device.c
@@ -273,23 +273,26 @@ static void handle_incoming_packet(uint32_t len)
   }
 #endif
 
+  bool accepted = false;
+
   if (size)
   {
     struct pbuf *p = pbuf_alloc(PBUF_RAW, size, PBUF_POOL);
-    bool accepted = true;
 
     if (p)
     {
       memcpy(p->payload, pnt, size);
       p->len = size;
       accepted = tud_network_recv_cb(p);
-    }
 
-    if (!p || !accepted)
-    {
-      /* if a buffer couldn't be allocated or accepted by the callback, we must discard this packet */
-      tud_network_recv_renew();
+      if (!accepted) pbuf_free(p);
     }
+  }
+
+  if (!accepted)
+  {
+    /* if a buffer was never handled by user code, we must renew on the user's behalf */
+    tud_network_recv_renew();
   }
 }
 
--- a/src/device/usbd.c
+++ b/src/device/usbd.c
@@ -555,8 +555,7 @@ static bool process_control_request(uint8_t rhport, tusb_control_request_t const
           }
           else
           {
-            TU_VERIFY( (0 == _usbd_dev.config_num) || (cfg_num == _usbd_dev.config_num) );
-            if (0 == _usbd_dev.config_num) TU_ASSERT( process_set_config(rhport, cfg_num) );
+            if (!_usbd_dev.config_num) TU_VERIFY( process_set_config(rhport, cfg_num) );
           }
 
           _usbd_dev.config_num = cfg_num;
--- a/src/device/usbd.c
+++ b/src/device/usbd.c
@@ -45,13 +45,14 @@ typedef struct {
   {
     volatile uint8_t connected    : 1;
     volatile uint8_t addressed    : 1;
-    volatile uint8_t configured   : 1;
     volatile uint8_t suspended    : 1;
 
     uint8_t remote_wakeup_en      : 1; // enable/disable by host
     uint8_t remote_wakeup_support : 1; // configuration descriptor's attribute
     uint8_t self_powered          : 1; // configuration descriptor's attribute
   };
+
+  volatile uint8_t selected_config;
 
   uint8_t itf2drv[16];     // map interface number to driver (0xff is invalid)
   uint8_t ep2drv[8][2];    // map endpoint to driver ( 0xff is invalid )
@@ -294,7 +295,7 @@ static char const* const _tusb_std_request_str[] =
 //--------------------------------------------------------------------+
 bool tud_mounted(void)
 {
-  return _usbd_dev.configured;
+  return _usbd_dev.selected_config;
 }
 
 bool tud_suspended(void)
@@ -337,19 +338,23 @@ bool tud_init (void)
   return true;
 }
 
-static void usbd_reset(uint8_t rhport)
+static void usbd_reset_config(uint8_t rhport)
 {
-  tu_varclr(&_usbd_dev);
-
   memset(_usbd_dev.itf2drv, DRVID_INVALID, sizeof(_usbd_dev.itf2drv)); // invalid mapping
   memset(_usbd_dev.ep2drv , DRVID_INVALID, sizeof(_usbd_dev.ep2drv )); // invalid mapping
-
-  usbd_control_reset();
 
   for (uint8_t i = 0; i < USBD_CLASS_DRIVER_COUNT; i++)
   {
     if ( _usbd_driver[i].reset ) _usbd_driver[i].reset( rhport );
   }
+}
+
+static void usbd_reset(uint8_t rhport)
+{
+  tu_varclr(&_usbd_dev);
+  usbd_control_reset();
+
+  usbd_reset_config(rhport);
 }
 
 /* USB Device Driver task
@@ -532,7 +537,7 @@ static bool process_control_request(uint8_t rhport, tusb_control_request_t const
 
         case TUSB_REQ_GET_CONFIGURATION:
         {
-          uint8_t cfgnum = _usbd_dev.configured ? 1 : 0;
+          uint8_t cfgnum = _usbd_dev.selected_config;
           tud_control_xfer(rhport, p_request, &cfgnum, 1);
         }
         break;
@@ -543,9 +548,18 @@ static bool process_control_request(uint8_t rhport, tusb_control_request_t const
 
           dcd_set_config(rhport, cfg_num);
 
-          if ( !_usbd_dev.configured && cfg_num ) TU_ASSERT( process_set_config(rhport, cfg_num) );
+          if (0 == cfg_num)
+          {
+            usbd_reset_config(rhport);
+            if (tud_umount_cb) tud_umount_cb();
+          }
+          else
+          {
+            TU_VERIFY( (0 == _usbd_dev.selected_config) || (cfg_num == _usbd_dev.selected_config) );
+            if (0 == _usbd_dev.selected_config) TU_ASSERT( process_set_config(rhport, cfg_num) );
+          }
 
-          _usbd_dev.configured = cfg_num ? 1 : 0;
+          _usbd_dev.selected_config = cfg_num;
 
           tud_control_status(rhport, p_request);
         }
@@ -876,8 +890,8 @@ void dcd_event_handler(dcd_event_t const * event, bool in_isr)
     case DCD_EVENT_UNPLUGGED:
       _usbd_dev.connected  = 0;
       _usbd_dev.addressed  = 0;
-      _usbd_dev.configured = 0;
       _usbd_dev.suspended  = 0;
+      _usbd_dev.selected_config = 0;
       osal_queue_send(_usbd_q, event, in_isr);
     break;
 
--- a/src/device/usbd.c
+++ b/src/device/usbd.c
@@ -52,7 +52,7 @@ typedef struct {
     uint8_t self_powered          : 1; // configuration descriptor's attribute
   };
 
-  volatile uint8_t selected_config;
+  volatile uint8_t config_num;
 
   uint8_t itf2drv[16];     // map interface number to driver (0xff is invalid)
   uint8_t ep2drv[8][2];    // map endpoint to driver ( 0xff is invalid )
@@ -295,7 +295,7 @@ static char const* const _tusb_std_request_str[] =
 //--------------------------------------------------------------------+
 bool tud_mounted(void)
 {
-  return _usbd_dev.selected_config;
+  return _usbd_dev.config_num;
 }
 
 bool tud_suspended(void)
@@ -537,7 +537,7 @@ static bool process_control_request(uint8_t rhport, tusb_control_request_t const
 
         case TUSB_REQ_GET_CONFIGURATION:
         {
-          uint8_t cfgnum = _usbd_dev.selected_config;
+          uint8_t cfgnum = _usbd_dev.config_num;
           tud_control_xfer(rhport, p_request, &cfgnum, 1);
         }
         break;
@@ -555,11 +555,11 @@ static bool process_control_request(uint8_t rhport, tusb_control_request_t const
           }
           else
           {
-            TU_VERIFY( (0 == _usbd_dev.selected_config) || (cfg_num == _usbd_dev.selected_config) );
-            if (0 == _usbd_dev.selected_config) TU_ASSERT( process_set_config(rhport, cfg_num) );
+            TU_VERIFY( (0 == _usbd_dev.config_num) || (cfg_num == _usbd_dev.config_num) );
+            if (0 == _usbd_dev.config_num) TU_ASSERT( process_set_config(rhport, cfg_num) );
           }
 
-          _usbd_dev.selected_config = cfg_num;
+          _usbd_dev.config_num = cfg_num;
 
           tud_control_status(rhport, p_request);
         }
@@ -891,7 +891,7 @@ void dcd_event_handler(dcd_event_t const * event, bool in_isr)
       _usbd_dev.connected  = 0;
       _usbd_dev.addressed  = 0;
       _usbd_dev.suspended  = 0;
-      _usbd_dev.selected_config = 0;
+      _usbd_dev.config_num = 0;
       osal_queue_send(_usbd_q, event, in_isr);
     break;
 
--- a/src/portable/nuvoton/nuc120/dcd_nuc120.c
+++ b/src/portable/nuvoton/nuc120/dcd_nuc120.c
@@ -223,6 +223,15 @@ void dcd_set_config(uint8_t rhport, uint8_t config_num)
 {
   (void) rhport;
   (void) config_num;
+
+  /* reset context of all non-control endpoints */
+
+  for (enum ep_enum ep_index = PERIPH_EP2; ep_index < PERIPH_MAX_EP; ep_index++)
+  {
+    USBD->EP[ep_index].CFG = 0;
+  }
+
+  bufseg_addr = PERIPH_EP2_BUF_BASE;
 }
 
 void dcd_remote_wakeup(uint8_t rhport)
--- a/src/portable/nuvoton/nuc121/dcd_nuc121.c
+++ b/src/portable/nuvoton/nuc121/dcd_nuc121.c
@@ -229,6 +229,15 @@ void dcd_set_config(uint8_t rhport, uint8_t config_num)
 {
   (void) rhport;
   (void) config_num;
+
+  /* reset context of all non-control endpoints */
+
+  for (enum ep_enum ep_index = PERIPH_EP2; ep_index < PERIPH_MAX_EP; ep_index++)
+  {
+    USBD->EP[ep_index].CFG = 0;
+  }
+
+  bufseg_addr = PERIPH_EP2_BUF_BASE;
 }
 
 void dcd_remote_wakeup(uint8_t rhport)
--- a/src/portable/nuvoton/nuc505/dcd_nuc505.c
+++ b/src/portable/nuvoton/nuc505/dcd_nuc505.c
@@ -112,8 +112,6 @@ static struct
   uint16_t total_bytes;
 } ctrl_in_xfer;
 
-static volatile bool configuration_changed;
-
 static volatile struct xfer_ctl_t *current_dma_xfer;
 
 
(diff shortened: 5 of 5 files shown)


src/device/usbd.c
```diff
@@ -543,9 +548,18 @@ static bool process_control_request(uint8_t rhport, tusb_control_request_t const
 
           dcd_set_config(rhport, cfg_num);
```
hathach, line 549: for this to work: `dcd_set_config(rhport, 0)` must be implemented by all DCD to reset all the opened endpoints when confignum is 0

src/device/usbd.c
```diff
@@ -543,9 +548,18 @@ static bool process_control_request(uint8_t rhport, tusb_control_request_t const
...
 
-          if ( !_usbd_dev.configured && cfg_num ) TU_ASSERT( process_set_config(rhport, cfg_num) );
+          if (0 == cfg_num)
+          {
+            usbd_reset_config(rhport);
+            if (tud_umount_cb) tud_umount_cb();
+          }
+          else
+          {
+            TU_VERIFY( (0 == _usbd_dev.selected_config) || (cfg_num == _usbd_dev.selected_config) );
```
hathach, line 558: I am not sure why we need `cfg_num == _usbd_dev.selected_config` in the TU_VERIFY(), if it is the case, we will just skip doing anything since host SET the same config. If they are different and both are not zero, we will purposely invoke `dcd_set_config(rhport, 0)` first for dcd to close all opened endpoints (as mentioned above), then call `dcd_set_config(rhport, cfg_num)` again, `then process_set_config()`

Again: this requires all DCD to implement dcd_set_config(0) as close all endpoints.

src/device/usbd.c
```diff
@@ -45,14 +45,15 @@ typedef struct {
...
     volatile uint8_t addressed    : 1;
-    volatile uint8_t configured   : 1;
     volatile uint8_t suspended    : 1;
 
     uint8_t remote_wakeup_en      : 1; // enable/disable by host
     uint8_t remote_wakeup_support : 1; // configuration descriptor's attribute
     uint8_t self_powered          : 1; // configuration descriptor's attribute
   };
 
+  volatile uint8_t selected_config;
```
hathach, line 55: can we rename it to have config_xxx, maybe config_value or config_num or config_selected . I kind of like the most relevant word appear first scheme.

src/device/usbd.c
majbthrd: It follows the USB spec.  "If the specified configuration value matches the configuration value from a configuration descriptor, then that configuration is selected and the device remains in the Configured state. Otherwise, the device responds with a Request Error."  A "Request Error" is a stall, and we ostensibly get this by TU_VERIFY returning a false (same as the default 'Unknown/Unsupported request' case).

src/device/usbd.c
majbthrd: Oh, the irony.  I had "config_num" in my local code, but then changed it to be the same as #167.

majbthrd:
> I think we only need to change the handling of SET_CONFIGURE when host switching from non-zero to another non-zero e.g 1 -> 2.

The implementation would not be compliant if it did this.  Hosts must go 1 -> 0 -> 2 when switching from 1 to 2.  If the host tries 1 -> 2, we must balk at the request.

I've added the "selected_config" to "config_num" name change.

src/device/usbd.c
hathach: I think the specs mean if the config num == the one in the descriptors, then use that config descriptor.  `cfg_num == _usbd_dev.selected_config` means host SET the same config value as I did before.

hathach:
> > I think we only need to change the handling of SET_CONFIGURE when host switching from non-zero to another non-zero e.g 1 -> 2.
> 
> The implementation would not be compliant if it did this. Hosts must go 1 -> 0 -> 2 when switching from 1 to 2. If the host tries 1 -> 2, we must balk at the request.
> 
> I've added the "selected_config" to "config_num" name change.

Oh, I didn't know host cannot go from 1 to 2. I haven't read specs for sometime. Let's me double check that. If it is then it is less work for us :)

pigrew:
My reading of the USB spec is that a transition from one configurated state to another IS permissible.

USB 1.1, 9.4.7 `SET_CONFIGURATION` specifies while in a configured state:

>If the specified configuration value is zero, then the device enters the Address state.
If the specified configuration value matches the configuration value from a
configuration descriptor, then that configuration is selected and the device remains in
the Configured state. Otherwise, the device responds with aRequest Error.

Here's my reasoning:

The key is "remains in a Configured state", so it should be able to transition from one to another configuration (or even to the same configuration), maintaining its configured state. During the transition (even from one config to itself), all the endpoints need to be reset to DATA0.

majbthrd:
Sigh... I just can't see to escape getting egg on my face each time I post here.

My erroneous interpretation of "matches the configuration value from a configuration descriptor" was matching the current configuration, and the OS behavior I observed was 1 -> 0 -> 2, which reinforced this misinterpretation.

majbthrd:
Before I submit another commit, does anyone has strong opinions on when dcd_set_config() should be called?  It is only documented in ./device/dcd.h as "Receive Set Configure request".

Should it be called every time a SET_CONFIGURATION request comes in, or only when the Section 9.4.7 logic deems it appropriate to transition configurations?

The legacy code calls it every time a SET_CONFIGURATION request comes in, but that might be code inertia.

pigrew:
> Sigh... I just can't see to escape getting egg on my face each time I post here.

Me too. :)

> Before I submit another commit, does anyone has strong opinions on when dcd_set_config() should be called? It is only documented in ./device/dcd.h as "Receive Set Configure request".
> 
> Should it be called every time a SET_CONFIGURATION request comes in, or only when the Section 9.4.7 logic deems it appropriate to transition configurations?
> 
> The legacy code calls it every time a SET_CONFIGURATION request comes in, but that might be code inertia.

For the STM32FSDEV, its implementation is a NO-OP.

dcd_nuc505 uses it, and so does dcd_lpc17_40. 

On LPC, the function enables all non-zero endpoints (even if the config number parameter is 0). EP0 is always enabled for this MCU.

On NUC505, it schedules clearing the DATA toggle for all enabled EP, once the next setup transaction interrupt happens.

I don't think it's used for any other MCU. I don't know why `dcd_set_config` exists, seems like it's unneeded.

And with not knowing what it should do, I have no clue what it should be called. :/

majbthrd:
Thanks @pigrew.  What you said about all the (non-control?) endpoints resetting to DATA0 rings true to me.

The NUC505 driver does just that (control endpoints have special registers on this peripheral).

For the LPC17xx, the datasheet defines the command as indicating "that the device is configured and all the enabled non-control endpoints will respond".  Note that process_set_config() sets the endpoints *after* the dcd_set_config() call (which is somewhat incongruent with the LPC17xx description).

Not a single driver uses the "config_num" parameter; it seems wholly superfluous.

I'm testing code at the moment with the premise that the dcd_set_config() implementation should always reset the context of non-EP0 endpoints (including DATA0 reset), regardless of "config_num" parameter.  This is a bit different from what @hathach said:

> we will purposely invoke dcd_set_config(rhport, 0) first for dcd to close all opened endpoints

By doing it this way, the dcd_set_config() call at the beginning of handling SET_CONFIGURATION 'cleans the slate' so that if process_set_config() gets called, it will always have full resources to open endpoints with.  If the new configuration is zero or an invalid request is made, it still results in the right action of shutting things down.

Note that all these PR proposed changes only become important when the implementation both has multiple configurations *AND* the host tries changing back and forth between them.

pigrew:
@majbthrd,

Yeah, it seems like there isn't a consistent usage of it.

On STM32, there is a bit for each EP ("DISABLED"), which means the EP will not respond at all to any packet directed to that EP. (This is different than STALL.)

So, do we redefine this function to have the behavior:

>Must be called by stack during `SET_CONFIGURATION` requests.
>* Disable and deallocate all non-zero EP. (Optionally) Set global non-zero EP enable bit. 
 Non-zero EP are to be individually enabled later during an `edpt_open` call which must also set the DTOG to 0.

Or get rid of this function and make an `edpt_close` that will deallocate & DISABLE a particular EP?

For the alternate interface support, there must be a way to disable an EP (since it's different than STALL), e.g. in the case of CDC transitioning from enabled to disabled. It can't just globally disable all of them in the case that there are other interfaces which shouldn't be effected (composite devices).

majbthrd:
> I'd like to see a clear DCD interface for closing endpoints, before thinking about the higher-level implementations of setting alternate settings or configurations.

@pigrew, I think we are on the same page; it is only a question of semantics and timing.

If there are classes that have: a) alternate settings, and b) use different endpoints numbers in different alternate settings, then absolutely I agree that there is also a need for an DCD API call that can surgically close individual EPs.

However, does such a theoretical class exist?  It feels like a paper tiger to me.  Perhaps it is a lack of imagination on my part, but if there was composite device with two or more classes (each with alternate settings), I can't see how endpoint numbers could overlap between the two classes, since all possible permutations of endpoint numbers must not conflict.  As a result, each class effectively "owns" those endpoint numbers (whether it uses them or not).  There is no class driver nor device example (now or under development) that I'm aware of that poses such a threat, so why fret over the prospect?

Going back to this PR, for multiple configurations (whether all configurations use the same or different classes and endpoints), I think we have a straightforward path to achieve a solution by (re-)defining dcd_set_config() as you said:

> Disable and deallocate all non-zero EP. (Optionally) Set global non-zero EP enable bit.
Non-zero EP are to be individually enabled later during an edpt_open call which must also set the DTOG to 0.

To that end, I've updated the PR to show the usbd.c change what I did to implement the revised definition of dcd_set_config() with the Nuvoton NUC120, NUC121/NUC125/NUC126, and NUC505.  I tested with NUC12x, NUC505, and SAMD21 targets.

hathach:
man, that is lots of English text to read. `dcd_set_config()` is added since I started the stack with LPC mcu first. On lpc17xx, the device must enter the configured state first before non-control endpoint is enabled (or maybe it can be any time afterwards, my memory fades). So I thought other mcu may have some similar behavior as well --> turn out not.

For edpt_close() API, it is good to have as well if alternate interface is needed. it actually does exists when I first started the stack but dropped since not used at all.

PS: I am trying to wrap up the eps32s2 port for now, I will catch up with you guys on this later on. I will try to see if `dcd_set_config()` can indeed by dropped.

hathach:
> If there are classes that have: a) alternate settings, and b) use different endpoints numbers in different alternate settings, then absolutely I agree that there is also a need for an DCD API call that can surgically close individual EPs.
> 
> However, does such a theoretical class exist? It feels like a paper tiger to me. Perhaps it is a lack of imagination on my part, but if there was composite device with two or more classes (each with alternate settings), I can't see how endpoint numbers could overlap between the two classes, since all possible permutations of endpoint numbers must not conflict. As a result, each class effectively "owns" those endpoint numbers (whether it uses them or not). There is no class driver nor device example (now or under development) that I'm aware of that poses such a threat, so why fret over the prospect?

Actually there is, especially those with ISO endpoints (audio & video), they always have an ALT with zero endpoints. Pay attention to the number of endpoints between the 2 Alt Interface. So yes, we need to individually disable that endpoints when switching. 

image

> Going back to this PR, for multiple configurations (whether all configurations use the same or different classes and endpoints), I think we have a straightforward path to achieve a solution by (re-)defining dcd_set_config() as you said:
> 
> > Disable and deallocate all non-zero EP. (Optionally) Set global non-zero EP enable bit.
> > Non-zero EP are to be individually enabled later during an edpt_open call which must also set the DTOG to 0.
> 
> To that end, I've updated the PR to show the usbd.c change what I did to implement the revised definition of dcd_set_config() with the Nuvoton NUC120, NUC121/NUC125/NUC126, and NUC505. I tested with NUC12x, NUC505, and SAMD21 targets.

It is a bit surprised to me that mcu vendor doesn't care much about setConfig(), Since both of you vote to remove `dcd_set_config()`, and in fact, it is not needed except for lpc17xx. I think we can safely remove this API(). For API that close all endpoints for switching to new Configuration, it should really be named `dcd_edpt_close_all()`, although technically we can do it with `dcd_set_config(0)` since that put device into Addressed state per spec. However, since the name doesn't ring any bell. Also It will be much quicker for dcd and easier for usbd than closing individual endpoints in this case. This API is of courser optional.

hathach:
pr to remove dcd_set_config() on-the-way #338

src/class/net/net_device.c
```diff
@@ -273,24 +273,27 @@ static void handle_incoming_packet(uint32_t len)
...
       memcpy(p->payload, pnt, size);
       p->len = size;
       accepted = tud_network_recv_cb(p);
-    }
 
-    if (!p || !accepted)
-    {
-      /* if a buffer couldn't be allocated or accepted by the callback, we must discard this packet */
-      tud_network_recv_renew();
+      if (!accepted) pbuf_free(p);
```
hathach, line 288: hmm, are you dynamically malloc-ing :roll_eyes: :roll_eyes: . I have no clue to review this yet, will leave it as it is for now :wink:

src/portable/nuvoton/nuc505/dcd_nuc505.c
```diff
@@ -300,7 +297,15 @@ void dcd_set_config(uint8_t rhport, uint8_t config_num)
 {
   (void) rhport;
   (void) config_num;
-  configuration_changed = true;
+
+  /* reset context of all non-control endpoints */
+
+  for (enum ep_enum ep_index = PERIPH_EPA; ep_index < PERIPH_MAX_EP; ep_index++)
```
hathach, line 303: since we are removing `dcd_set_config()`, we should not implement it here. Don't worry about resetting data toggle, it is enough to do that in dcd_edpt_open(). Since SetConfig() will follows by an series of endpoints open (non-zero) or close_all (zero).

src/portable/nuvoton/nuc505/dcd_nuc505.c
```diff
@@ -574,15 +579,6 @@ void USBD_IRQHandler(void)
       /* given ACK from host has happened, we can now set the address (if not already done) */
       if((USBD->FADDR != assigned_address) && (USBD->FADDR == 0)) USBD->FADDR = assigned_address;
```
hathach, line 580: you could use the optional API `dcd_edpt0_status_complete()` to complete the address, although it causes the function to be invoked every control transfer. It is not a big of a deal since the number of control transfer is much smaller than normal endpoints (mostly in the enumeration).
https://github.com/hathach/tinyusb/blob/master/src/portable/microchip/samd/dcd_samd.c#L189

hathach:
require
- [ ] #338 
- [ ] #339

majbthrd:
> Actually there is, especially those with ISO endpoints (audio & video), they always have an ALT with zero endpoints. Pay attention to the number of endpoints between the 2 Alt Interface. So yes, we need to individually disable that endpoints when switching.

My words appear to have been misinterpreted.  I said "use different endpoints numbers", not "use different quantities (or numbers) of endpoints", i.e. EP 4 in one alternate setting and EP5 in another.

hathach:
> > Actually there is, especially those with ISO endpoints (audio & video), they always have an ALT with zero endpoints. Pay attention to the number of endpoints between the 2 Alt Interface. So yes, we need to individually disable that endpoints when switching.
> 
> My words appear to have been misinterpreted. I said "use different endpoints numbers", not "use different quantities (or numbers) of endpoints", i.e. EP 4 in one alternate setting and EP5 in another.

Ah sorry for my English reading, never be my pros. Yes, I don't see the reason the EP num will change if they are used between Alt Interface.

pigrew:
> > Actually there is, especially those with ISO endpoints (audio & video), they always have an ALT with zero endpoints. Pay attention to the number of endpoints between the 2 Alt Interface. So yes, we need to individually disable that endpoints when switching.
> 
> My words appear to have been misinterpreted. I said "use different endpoints numbers", not "use different quantities (or numbers) of endpoints", i.e. EP 4 in one alternate setting and EP5 in another.

This would be sub-optimal for my stm32fsdev port (and perhaps others). On STM32F103, it now only supports EP 0 through 3 (IN and OUT). This would cause the device to more easily run out of EP. I'm statically mapping EP[n] TX/RX to FIFO[n] TX/RX where n<4 on F103.

majbthrd:
> This would be sub-optimal for my stm32fsdev port (and perhaps others).

@pigrew, what is the "this" that you are referring to as "sub-optimal" in this context?

pigrew:
> > This would be sub-optimal for my stm32fsdev port (and perhaps others).
> 
> @pigrew, what is the "this" that you are referring to as "sub-optimal" in this context?

(Perhaps we're all having English issues today).

I interpreted your  "use different endpoints numbers" to mean that, e.g., configuration 0 would use EP {1,2,3} and config 2 would use EP {3,4,5}, such that there were no overlapping EP numbers between configurations.

majbthrd:
@pigrew, yeah, it does feel that way. ;)  Alas, I suspect English isn't a second language for either of us.

Here's the [original context](https://github.com/hathach/tinyusb/pull/326#issuecomment-611218112).

To summarize: I had said that there wasn't a justified reason that I could think of where we would use different EP numbers.  @hathach then [said here](https://github.com/hathach/tinyusb/pull/326#issuecomment-612836069) that there was indeed a valid reason (but cited quantity of EPs instead of different EP numbers).  When I clarified the difference, you [then said](https://github.com/hathach/tinyusb/pull/326#issuecomment-612986317) that this would be sub-optimal.

It is a barrel of laughs! :)

hathach:
@majbthrd @pigrew things are getting confusing, and I understand why most design choose to soft disconnect then change the configuration descriptor instead. Most discussion is steered toward endpoints manipulation, though these scenario may not be practical at all. And I think these aren't not in the context of configuration, it is better discussed in the Alt Interface

@majbthrd I think we could start to wrap up this PR, once the close API is complete, we can just replace the driver reset() with driver's close() `usbd_reset_config`. Please put a comment there `TODO should be driver close` .  Also could you remove the use of `dcd_set_config()`, we are removing that API (I thought you also think that API is not necessary as well)

Then we could merge this one, it will be good for the scenario where host only pick a specific non-zero and never switch to another non-zero. It is still limited support, but good enough for USB NET usage.

src/portable/nuvoton/nuc120/dcd_nuc120.c
```diff
@@ -223,6 +223,15 @@ void dcd_set_config(uint8_t rhport, uint8_t config_num)
 {
   (void) rhport;
   (void) config_num;
+
+  /* reset context of all non-control endpoints */
+
+  for (enum ep_enum ep_index = PERIPH_EP2; ep_index < PERIPH_MAX_EP; ep_index++)
```
hathach, line 229: since we are removing `dcd_set_config()`, please remove the use of this API. Then we are ready to merge

src/portable/nuvoton/nuc121/dcd_nuc121.c
```diff
@@ -229,6 +229,15 @@ void dcd_set_config(uint8_t rhport, uint8_t config_num)
 {
   (void) rhport;
   (void) config_num;
+
+  /* reset context of all non-control endpoints */
+
+  for (enum ep_enum ep_index = PERIPH_EP2; ep_index < PERIPH_MAX_EP; ep_index++)
```
hathach, line 235: since we are removing `dcd_set_config()`, please remove the use of this API. Then we are ready to merge

majbthrd:
The existing code already supports a non-zero configuration, so if that is the only allowed change, this PR becomes largely pointless.  To switch between configurations needs a DCD API to reset the context of all non-control endpoints, which is what dcd_set_config() was doing.  I'm just going to close this.

hathach:
> The existing code already supports a non-zero configuration, so if that is the only allowed change, this PR becomes largely pointless. To switch between configurations needs a DCD API to reset the context of all non-control endpoints, which is what dcd_set_config() was doing. I'm just going to close this.

Thanks for opening the PR, we all learn a bit more of things. I thought you and Nathan were saying `dcd_set_config()` is not needed, do you want to remove the use of it for NUC505. We can then get rid of that API

majbthrd:
> I thought you and Nathan were saying `dcd_set_config()` is not needed, do you want to remove the use of it for NUC505. We can then get rid of that API

This is one of those discussions where each party had different impressions of the outcome.  I thought the outcome of that was that dcd_set_config() was valid as part of a reset of the non-control endpoint context, and perhaps could be more appropriately named.  However, the torrent of discussions between @hathach and @pigrew about alternate interfaces suggest all manner of different approaches to solve a different problem (that indirectly impacts multiple configurations).

The NUC505 driver is actively using that dcd_set_config() callback, so I can't simply remove it [as you requested](https://github.com/hathach/tinyusb/pull/338#issuecomment-612842952).

hathach:
Indeed, this PR is inflate and mixed with Alt Interface. Since both of them need an endpoints close API() to be fully supported. Since both features are implemented currently, we are easy to mix match them together. ok let's move the dcd_set_config() removal to the new issue.  #338

PR #329, opened 2020-04-07
Correct FreeRTOS wrappers to handle 16-bit ticks.

pigrew:
The FreeRTOS wrappers cannot cope with 16-bit ticks. This PR adds a more complicated conversion function to translate to 16-bit ticks as needed. Since delays are normally compile-time constants, the optimizer should be able to optimize-out the conversion function, so I expect no impact to the compiled binary.

@cr1901 

I'd appreciate if you looked over this PR. Perhaps it'll be useful on MSP430. Note that the FreeRTOS config has to be edited to select 16-bit ticks (`configUSE_16_BIT_TICKS`).

@hathach 

Turns out I can create a draft PR! (As of now, this has only been compile-checked, not run.)

If it looks good to everyone (and preferably if it's tested), I'll mark this as a non-draft PR.

### DIFF
--- a/src/osal/osal_freertos.h
+++ b/src/osal/osal_freertos.h
@@ -37,12 +37,39 @@
 extern "C" {
 #endif
 
+/* Translate TinyUSB msec to FreeRTOS ticks. May clip value in the case
+    of 16-bit ticks. */
+static inline TickType_t _tusb_osal_freertos_translate_msec_to_ticks(uint32_t msec)
+{
+#if (configUSE_16_BIT_TICKS == 1)
+  // Will never overflow here, since portMAX_DELAY is 16-bit
+  const uint32_t max_msec = ((uint32_t)(portMAX_DELAY-1))*1000U/configTICK_RATE_HZ - 1;
+#elif (configTICK_RATE_HZ >= 1000)
+  // slightly conservative, actual max is larger due to rounding.
+  const uint32_t max_msec = ((uint32_t)(portMAX_DELAY-1))/((uint32_t)configTICK_RATE_HZ)*1000U - 1; 
+#else
+  const uint32_t max_sec = (((uint32_t)portMAX_DELAY-1)) / configTICK_RATE_HZ;
+  const uint32_t max_msec =  (max_sec > (UINT32_MAX/1000)) ? (UINT32_MAX-1) : (max_sec * 1000 - 1);
+#endif
+
+  // Translate wait forever
+  if(msec >= OSAL_TIMEOUT_WAIT_FOREVER) 
+  {
+    // Note, FreeRTOS only waits forever if (INCLUDE_vTaskSuspend == 1), in RTOS config.
+    return portMAX_DELAY;
+  }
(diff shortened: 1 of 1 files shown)


cr1901:
I don't know enough about FreeRTOS to comment right now, I'm afraid :(. How do you know the value won't overflow if the `CONFIG_TICKS_HZ` value is <= 1000?

pigrew:
> I don't know enough about FreeRTOS to comment right now, I'm afraid :(. How do you know the value won't overflow if the `CONFIG_TICKS_HZ` value is <= 1000?

You're very correct. It will.

This code is very upsetting, so tricky to get right. I almost want to use 64-bit arithmetic to calculate the max_ms variable, since it would avoid overflows.

I updated this PR again, not perfect, hopefully good enough.

FreeRTOS has three major tick-related configuration options:

* 16-bit vs 32-bit tick counts
* Optional ability to wait forever
* max tick count is customizable
* Ticks per second is definable (often 1000 Hz, but lower is suggested with use of 16-bit ticks)

TinyUSB, on the other hand, is only using `uint32_t` milliseconds.

hathach:
Thank you for  the PR, however I think this is a bit too much work for us, `configUSE_16_BIT_TICKS` is probably only used by 8-bit MCU. even MSP430 should use the 32-bit tick type. If 16-bit tick user start to complain, he/she could help to troubleshoot and test the implementation.

PS: so far only host use semaphore and delay() function, since host stack is kind of obsolete for now. These API may be removed in the future (I tried my best to just keep the queue API), when we rework the host stack. So I would suggest to leave it as it is for now :).

pigrew:
@hathach,

How about adding a static assert that it uses 32bit tick, and that the tick rate is <1000, to guarantee no overflow?

hathach:
> @hathach,
> 
> How about adding a static assert that it uses 32bit tick, and that the tick rate is <1000, to guarantee no overflow?

It is really not necessary, we cannot force application to use tick rate < 1000. The osal is not public API and used internally, the stack will only use several known values eg. 0 (for polling), Max INT for forever, something around 5-10 seconds for decent time out. So it won't overflow anyway.

cr1901:
>Thank you for the PR, however I think this is a bit too much work for us, configUSE_16_BIT_TICKS is probably only used by 8-bit MCU.

I would like to see an 8-bit port one day (maybe Cypress or PIC?); I don't think I would expect an 8-bit to use an RTOS. I don't think I would even use an RTOS (unless I wrote my own for fun) on msp430.

hathach:
@pigrew thank you, we could re-visit this later when it becomes an issue. I still think the tick should always be at least 32-bit for most system, to be sufficient enough. 64k of ticks is only several seconds :).  

> I would like to see an 8-bit port one day (maybe Cypress or PIC?); I don't think I would expect an 8-bit to use an RTOS. I don't think I would even use an RTOS (unless I wrote my own for fun) on msp430.

RTOS can still be very handy in some cases :)

Issue #327, opened 2020-04-07
add dcd_disconnect() dcd_connect()


hathach:
these should be soft disconect/connect via registers. Some MCUs may not support this features.

- [x] Espressif ESP32-S2
- [x] Microchip SAMD
- [x] Microchip SAMG 
- [x] Nordic nRF5x 
- [x] Nuvoton NUC120
- [x] Nuvoton NUC121/125/126
- [x] Nuvoton NUC505
- [x] NXP LPC3511
- [x] NXP LPC17/40
- [x] NXP Transdimension (18/ iMX RT)
- [x] Sony CXD56 <s>(not implemented atm, since I have no idea how).</s>
- [x] STM32 FSDev
- [x] STM32 Synopsys
- [x] TI MSP430
- [x] Fomu

pigrew:
@hathach 
Could you better describe what these functions should do?

Do they enable/disable the pull-up resistors on D+/D-? Is there anything else they would do?

On some STM32, there is no internal pull-up, so for them, there would need to be BSP support (perhaps implemented by using weak functions?).

hathach:
> @hathach
> Could you better describe what these functions should do?
> 
> Do they enable/disable the pull-up resistors on D+/D-? Is there anything else they would do?

@pigrew  yeah, these disables the internal pullup resistor on D+/D-. That's it. It is enough to cause the disconnection on both host & device side.

> On some STM32, there is no internal pull-up, so for them, there would need to be BSP support (perhaps implemented by using weak functions?).

Yeah, several old MCUs doesn't have internal pull-up and replied on the external GPIO. For these case, I think it is better to have dcd_disconnect()/dcd_connect() as weak function. The public version tud_disconnect/tud_connect will have warning attribute to let user know that this functionality is not supported on this mcu. 

We don't need to provide bsp hook in this case, since the stack doesn't use these at all. It is provided for user convenience. Application should know their own way (GPIO) to disconnect/connect when working on those mcu.

pigrew:
Should the _init() function enable the PU, if it exists?

Should the weak dcd_connect() give a TU_LOG1 if it is called on a board that doesn't support it?
Or do we just not define the function on these, forcing the BSP to implement it?

(STM32 F1/F2/F3 discovery boards do not have software control of the PU.)

> The public version tud_disconnect/tud_connect will have warning attribute to let user know that this functionality is not supported on this mcu.

How would the `tud_connect` know if the functionality is supported by the MCU? I don't see any method.

[...]

PR #332, opened 2020-04-09
stm32fsdev: Implement dcd_connect.

pigrew:
This is a followup for #327.

This patch proposes adding dcd_connect/dcd_disconnect as weak functions in the DCD source. Also, changes the intention of dcd_init to NOT enable internal pull-ups by default, so BSP can more easily override behavior.

This allows the BSP to optionally override these functions, so that an internal pull up does not have to be used (in the case, e.g., that external ESD protection already has its own pull-up).

In the case that there is no internal pull-up, by default nothing will change, unless the BSP defines a `dcd_connect` function, at which time this function will be called as part of `dcd_init`.

This patch also adds documentation for the meanings of these functions.

(Hmm, fails the board_test CI. I'll fix it if this patch is viable.)

### DIFF
--- a/docs/porting.md
+++ b/docs/porting.md
@@ -62,6 +62,8 @@ All of the code for the low-level device API is in `src/portable/<vendor>/<chip 
 ##### dcd_init
 Initializes the USB peripheral for device mode and enables it.
 
+If the MCU has an internal pull-up, this function should leave it disabled.
+
 #### dcd_int_enable / dcd_int_disable
 
 Enables or disables the USB device interrupt(s). May be used to prevent concurrency issues when mutating data structures shared between main code and the interrupt handler.
@@ -76,6 +78,10 @@ Called when the device received SET_CONFIG request, you can leave this empty if 
 
 ##### dcd_remote_wakeup
 Called to remote wake up host when suspended (e.g hid keyboard)
+
+##### dcd_connect / dcd_disconnect
+
+Connect or disconnect the data-line pull-up resistor. Define as a weak function if the MCU has an internal pull-up, otherwise do not define.
 
 #### Special events
 You must let TinyUSB know when certain events occur so that it can continue its work. There are a few methods you can call to queue events for TinyUSB to process.
--- a/docs/porting.md
+++ b/docs/porting.md
@@ -62,7 +62,7 @@ All of the code for the low-level device API is in `src/portable/<vendor>/<chip 
 ##### dcd_init
 Initializes the USB peripheral for device mode and enables it.
 
-If the MCU has an internal pull-up, this function should leave it disabled.
+This function should leave an internal D+/D- pull-up in its default power-on state. `dcd_connect` will be called by the USBD core following `dcd_init`.
 
 #### dcd_int_enable / dcd_int_disable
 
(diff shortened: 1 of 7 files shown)


pigrew:
(Rebased this on the connect/disconnect branch)

docs/porting.md
```diff
@@ -62,6 +62,8 @@ All of the code for the low-level device API is in `src/portable/<vendor>/<chip
 ##### dcd_init
 Initializes the USB peripheral for device mode and enables it.
 
+If the MCU has an internal pull-up, this function should leave it disabled.
```
hathach, line 65: as note: dcd doesn't need to explicitly disable PU if it is enabled by default on power/reset

src/portable/st/stm32_fsdev/dcd_stm32_fsdev.c
```diff
@@ -239,17 +239,29 @@ void dcd_init (uint8_t rhport)
...
-  // And finally enable pull-up, which may trigger the RESET IRQ if the host is connected.
-  // (if this MCU has an internal pullup)
+// Define only on MCU with internal pull-up so BSP can override (needed on MCU without internal pull-up)
 #if defined(USB_BCDR_DPPU)
-  USB->BCDR |= USB_BCDR_DPPU;
-#else
-  // FIXME: callback to the user to ask them to twiddle a GPIO to disable/enable D+???
-#endif
 
+TU_ATTR_WEAK
```
hathach, line 249: I think we don't need to hae weak attribute here, only in the dcd.h header. When it is implemented, it should be the one that is used.

docs/porting.md
```diff
@@ -77,6 +79,10 @@ Called when the device received SET_CONFIG request, you can leave this empty if
 ##### dcd_remote_wakeup
 Called to remote wake up host when suspended (e.g hid keyboard)
 
+##### dcd_connect / dcd_disconnect
+
+Connect or disconnect the data-line pull-up resistor. Define as a weak function if the MCU has an internal pull-up, otherwise do not define.
```
hathach, line 84: if PU is available, it will be the strong one, BSP/application has no purpose to use external PU, even if it does, it is not a concern of TinyUSB stack.

pigrew: I was trying to find a use case, but I remembered somewhat wrong. The STM32H742 has an errata stating its internal pull-up is too small and will wear out after 10 years, so someone could conceivably want to use the external one instead.

Regardless, will modify this PR based on feedback.

src/portable/st/stm32_fsdev/dcd_stm32_fsdev.c
pigrew: Modified.

docs/porting.md
pigrew: Modified.

pigrew:
> Thanks, it looks great. Though let's make the supported MCU has the strong define instead of weak.

I've updated this patch based on your feedback. It should be ready for a final review.

docs/porting.md
hathach: wow, that errata for 10 years wear-out is insane though

PR #333, opened 2020-04-11
fix incorrect setup packet

hathach:
also increase usbd stack in example when debug is enabled. We should really merge esp32s2 and stm32 together later on maybe once it is more accessible for everyone and the esp32 usb specs is out. So that problems like this doesn't happen again.

@hathach Change [line 334](https://github.com/hathach/tinyusb/pull/319/files#diff-371e8160b188a80c1a3821c75d194adfR334) from:

```
USB0.out_ep_reg[epnum].doeptsiz = USB_PKTCNT0_M |
        ((xfer->max_size & USB_XFERSIZE0_V) << USB_XFERSIZE0_S);
```

to

```
USB0.out_ep_reg[epnum].doeptsiz |= USB_PKTCNT0_M |
        ((xfer->max_size & USB_XFERSIZE0_V) << USB_XFERSIZE0_S);
```

Do the same for [line 606](https://github.com/hathach/tinyusb/commit/d4302dacc584c4488cf1905caa822f96bd64930d).

Note the ORing. See if the bug goes away and you can save all 3 setup packets again. This seems quite similar to a [bug](https://github.com/hathach/tinyusb/commit/d4302dacc584c4488cf1905caa822f96bd64930d) I fixed from my original STM32 port.

SETUPCNT would become `3` again after receiving a new SETUP packet because it's a 2-bit counter that wraps around.

_Originally posted by @cr1901 in https://github.com/_render_node/MDI0OlB1bGxSZXF1ZXN0UmV2aWV3Q29tbWVudDQwNjg0NDE2OQ==/comments/review_comment_

### DIFF
--- a/examples/device/cdc_msc_freertos/src/main.c
+++ b/examples/device/cdc_msc_freertos/src/main.c
@@ -56,7 +56,13 @@ StaticTimer_t blinky_tmdef;
 TimerHandle_t blinky_tm;
 
 // static task for usbd
-#define USBD_STACK_SIZE     (3*configMINIMAL_STACK_SIZE/2)
+// Increase stack size when debug log is enabled
+#if CFG_TUSB_DEBUG
+  #define USBD_STACK_SIZE     (3*configMINIMAL_STACK_SIZE)
+#else
+  #define USBD_STACK_SIZE     (3*configMINIMAL_STACK_SIZE/2)
+#endif
+
 StackType_t  usb_device_stack[USBD_STACK_SIZE];
 StaticTask_t usb_device_taskdef;
 
--- a/examples/device/cdc_msc_freertos/src/main.c
+++ b/examples/device/cdc_msc_freertos/src/main.c
@@ -200,7 +200,7 @@ void tud_cdc_line_state_cb(uint8_t itf, bool dtr, bool rts)
   if ( dtr && rts )
   {
     // print initial message when connected
-    tud_cdc_write_str("\r\nTinyUSB CDC MSC HID device with FreeRTOS example\r\n");
+    tud_cdc_write_str("\r\nTinyUSB CDC MSC device with FreeRTOS example\r\n");
   }
 }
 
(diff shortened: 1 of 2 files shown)


cr1901:
>We should really merge esp32s2 and stm32 together later on maybe once it is more accessible for everyone and the esp32 usb specs is out. So that problems like this doesn't happen again.

Agreed, with the caveat that the register names don't appear to be the exact same between the two ports :(. I'm not sure the best way to handle this.

hathach:
> Change [these lines](https://github.com/hathach/tinyusb/pull/333/files#diff-371e8160b188a80c1a3821c75d194adfL564-L567) back to match [`dcd_synopsys`](https://github.com/hathach/tinyusb/blob/master/src/portable/st/synopsys/dcd_synopsys.c#L584-L589):
> 
> From:
> 
> ```
>       // We can receive up to three setup packets in succession, but	      // We can receive up to three setup packets in succession, but
>       // only the last one is valid. Therefore we just overwrite it	      // only the last one is valid. Therefore we just overwrite it
>       _setup_packet[0] = (*rx_fifo);	      _setup_packet[0] = (*rx_fifo);
>       _setup_packet[1] = (*rx_fifo);
> ```
> 
> To something similar to:
> 
> ```
>         uint8_t setup_left = ((out_ep[epnum].DOEPTSIZ & USB_OTG_DOEPTSIZ_STUPCNT_Msk) >> USB_OTG_DOEPTSIZ_STUPCNT_Pos);
> 
>         // We can receive up to three setup packets in succession, but
>         // only the last one is valid.
>         _setup_packet[4 - 2*setup_left] = (* rx_fifo);
>         _setup_packet[5 - 2*setup_left] = (* rx_fifo);
> ```
> 
> Also increase `_setup_packet` size back to 6 entries: `static TU_ATTR_ALIGNED(4) uint32_t _setup_packet[6];`
> 
> See if that helps.

The current changes already fixed it, I think we don't really need to store 3 setup packets, just overwrite it to setup[0-1] since we will only examine the last setup anyway :)

hathach:
> > We should really merge esp32s2 and stm32 together later on maybe once it is more accessible for everyone and the esp32 usb specs is out. So that problems like this doesn't happen again.
> 
> Agreed, with the caveat that the register names don't appear to be the exact same between the two ports :(. I'm not sure the best way to handle this.

That is not at problem, I have done it with nxp lpc & mx iRT, basically we will have to write our own register names. Register  mapping won't change anyway, so it is easy enough. Though we could do it later when the esp32s2 is proved to be identical enough to current stm32. 

https://github.com/hathach/tinyusb/blob/master/src/portable/nxp/transdimension/dcd_transdimension.c#L168

hathach:
@cr1901 thanks for reviewing and approving :)

PR #334, opened 2020-04-12
usbnet: OS-agnostic (Windows/Linux/macOS) network driver

majbthrd:
The existing usbnet RNDIS (Windows/Linux) implementation is enhanced to alternately act as CDC-ECM (Linux/macOS).  Windows and Linux users get driver-less RNDIS USB networking; macOS would get driver-less CDC-ECM USB network.

I think this has value because it allows a generic implementation to be agnostic to the user’s OS.

Windows and Linux will default to RNDIS and work the same as they previously did.

In a pinch, macOS users can immediately take advantage of this new code by simply commenting out this line in ./src/device/usb.c:

            TU_ASSERT(alternate == 0);

Otherwise, they must wait until PR #317 or PR #320 or alternative is approved.

@hathach had expressed interest in being able to test different configurations.  Until PR #326 or alternative is approved, a Linux user will not be able to switch from the default RNDIS configuration to the CDC-ECM configuration.  A normal user would not be bothered by the absence of this.

### DIFF
--- a/examples/device/net_lwip_webserver/src/main.c
+++ b/examples/device/net_lwip_webserver/src/main.c
@@ -26,13 +26,14 @@
  */
 
 /*
-depending on the value of CFG_TUD_NET (tusb_config.h), this can be a CDC-ECM, RNDIS, or CDC-EEM USB virtual network adapter
+depending on the value of CFG_TUD_NET (tusb_config.h), this can be a RNDIS+CDC-ECM or CDC-EEM USB virtual network adapter
 
-CDC-ECM should be valid on Linux and MacOS hosts
-RNDIS   should be valid on Linux and Windows hosts
-CDC-EEM should be valid on Linux hosts
+OPT_NET_RNDIS_ECM : RNDIS should be valid on Linux and Windows hosts, and CDC-ECM should be valid on Linux and MacOS hosts
+OPT_NET_EEM       : CDC-EEM should be valid on Linux hosts
 
-You *must* customize tusb_config.h to set the CFG_TUD_NET definition to the type of these network adapters to emulate.
+OPT_NET_RNDIS_ECM should be the best choice, as it makes for a hopefully universal solution.
+
+You *must* customize tusb_config.h to set the CFG_TUD_NET definition to the type of these network option.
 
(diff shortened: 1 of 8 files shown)


majbthrd:
> In the long run, if possible I think we can drop the OPT_NET_* selection, and default to enable all 3.

I am inclined to just remove CDC-EEM, and that would simplify the code.  CDC-EEM has no real-world value; RNDIS and CDC-ECM prevailed.  Are you OK with that?

hathach:
> > In the long run, if possible I think we can drop the OPT_NET_* selection, and default to enable all 3.
> 
> I am inclined to just remove CDC-EEM, and that would simplify the code. CDC-EEM has no real-world value; RNDIS and CDC-ECM prevailed. Are you OK with that?

Yeah, sure. I initially only know of RNDIS, even Android use that for tethering as well. CDC-ECM is new to me (since I rarely use macOS).

pigrew:
> I am inclined to just remove CDC-EEM, and that would simplify the code. CDC-EEM has no real-world value; RNDIS and CDC-ECM prevailed. Are you OK with that?

I also have no clue, never used it.

It looks like it's supposed to be more lightweight/faster than ECM, and Linux has supported for for 10 years. Except for speed/firmware size reasons, I don't see a reason to keep it. TinyUSB isn't performance optimized at the moment, so speed reasons are probably moot.

It looks like CDC-NCM is another network interface that could be supported.

Why are there so many of them?

hathach:
RNFIS alone has a lot of RNDIS interface variant as well check out the Misc class (EF) https://www.usb.org/defined-class-codes#anchor_BaseClassEFh

Also, Microsoft seems to change the RNDIS mapping to USB to CDC-ACM with vendor protocol (0xFF) as well https://docs.microsoft.com/en-us/windows-hardware/drivers/network/interface-descriptor-for-communication-class-interface 

I am glad we got it worked since it is rather confusing to look at the big picture :D

hathach:
> It looks like CDC-NCM is another network interface that could be supported.

I just head to usb site to find CDC-ECM but look like the class is obsolete and replaced by this CDC-NCM. Maybe later on, we will check to see how much is the difference and current support state of OS.

image

PR #335, opened 2020-04-12
net_lwip_webserver: add support for lwip timeouts

majbthrd:
lwip has network timeout routines (for things like packet re-transmission), but these will not fire unless sys_check_timeouts() is periodically called.

### DIFF
--- a/examples/device/net_lwip_webserver/src/main.c
+++ b/examples/device/net_lwip_webserver/src/main.c
@@ -43,6 +43,7 @@ The MCU appears to the host as IP address 192.168.7.1, and provides a DHCP serve
 #include "dhserver.h"
 #include "dnserver.h"
 #include "lwip/init.h"
+#include "lwip/timeouts.h"
 #include "httpd.h"
 
 /* lwip context */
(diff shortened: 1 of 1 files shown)


pigrew:
Looks good.

I think we can assume there is enough traffic to trigger this, otherwise RTOS timers, or SOF callbacks would work.

caobug:
If we frequently call `sys_check_timeouts`, does it affect performance?

pigrew:
> If we frequently call `sys_check_timeouts`, does it affect performance?

Yes, but probably not much?

It looks like LWIP keeps things in a queue sorted by timeout, so the function returns fairly quickly if no timeout has elapsed:

```
void
sys_check_timeouts(void)
{
  u32_t now;

  LWIP_ASSERT_CORE_LOCKED();

  /* Process only timers expired at the start of the function. */
  now = sys_now();

  do {
    struct sys_timeo *tmptimeout;
    sys_timeout_handler handler;
    void *arg;

    PBUF_CHECK_FREE_OOSEQ();

    tmptimeout = next_timeout;
    if (tmptimeout == NULL) {
      return;
    }

    if (TIME_LESS_THAN(now, tmptimeout->time)) {
      return;
    }
```

PR #336, opened 2020-04-13
tud_edpt_close

pigrew:
@majbthrd & @hathach,

I've concluded that alternate endpoint support needs a way to close endpoints (or at minimum a way to reset the DTOG of an endpoint if an endpoint never has to be closed).

Here is a basic implementation of the concept, with a partial implementation for stm32fsdev.

Nothing is tested, and the stm32 implementation leaks packet buffer memory, but I'd like feedback before going forward. I believe the USBD core code is complete (though perhaps buggy).

The event queue has to be filtered to remove pending transfers while the endpoint is being closed.

(Also note that this PR is meant to complement PR #317 or #320.)

TODO:

- [x] ~~Implement dynamic allocation of stm32 packet buffer memory~~
- [x] Test implementation
- [x] Check how this is handled in other USB device libraries
- [x] ~~Add edpt_close_all()~~ (have class drivers close individual ep)

Comparisons:

- libopencm3: Able to disable all endpoints, but not a particular endpoint
- stm32fsdev: Has USBD_LL_CloseEP; PMA address space seems to be manually allocated. Example classes only close EP when class is deinitialized. Driver doesn't seem to reset DTOG when alternate settings are changed (for either audio or CDC). No efforts are made to flush buffers when changing alternate setting.


hathach:
Yeah, I agree supporting switching ALT Interface will indeed need to implement the dcd_edpt_close() API.

src/device/usbd.c
```diff
@@ -1036,4 +1041,60 @@ bool usbd_edpt_stalled(uint8_t rhport, uint8_t ep_addr)
...
+}
+
+/**
+ * tud_edpt_close will disable an endpoint, and clear all pending transfers
+ * through the particular endpoint.
+ * 
+ * It must be called from the usb task (i.e. from the control request
+ * handler while handling SET_ALTERNATE).
+ */
+void tud_edpt_close(uint8_t rhport, uint8_t ep_addr)
```
hathach, line 1086: edpt_close should not be exported to application, it will be handled by the usbd and dcd only.

src/device/usbd.h
```diff
@@ -79,6 +79,8 @@ static inline bool tud_connect(void)
   return true;
 }
 
+void tud_edpt_close(uint8_t rhport, uint8_t ep_addr);
```
hathach, line 82: please remove this, edpt_close should not be exported to application, it will be handled by the usbd and dcd only.

src/device/usbd.c
pigrew: That means it should use the usbd prefix? Or a different header file?

src/device/usbd.c
```diff
@@ -1036,4 +1041,60 @@ bool usbd_edpt_stalled(uint8_t rhport, uint8_t ep_addr)
   return _usbd_dev.ep_status[epnum][dir].stalled;
 }
 
+/**
+ * Remove queued xfer complete messages from event queue,
+ * for a particular ep.
+ */
+static void usbd_abort_transfers(uint8_t rhport, uint8_t ep_addr)
```
hathach, line 1048: I don't think we need this function for now, let's focus on the dcd_edpt_close() for now, we can address this later on.

src/portable/st/stm32_fsdev/dcd_stm32_fsdev.c
```diff
@@ -642,6 +643,40 @@ bool dcd_edpt_open (uint8_t rhport, tusb_desc_endpoint_t const * p_endpoint_desc
   return true;
 }
 
+/**
+ * Close an endpoint.
+ * 
+ * This function should be called with interrupts enabled, though
```
hathach, line 649: there is no need to force the interrupt disabled or eanbled. If the dcd need an interrupt disabled, it could call dcd_init_disable() within this function.

src/portable/st/stm32_fsdev/dcd_stm32_fsdev.c
```diff
@@ -623,13 +622,15 @@ bool dcd_edpt_open (uint8_t rhport, tusb_desc_endpoint_t const * p_endpoint_desc
 
   if(dir == TUSB_DIR_IN)
   {
+    // FIXME: use pma_alloc to allocate memory dynamically
```
hathach, line 625: I am not sure if it is possible to not use alloc memory for stm32fsdev. I absolutely don't want the dcd to have dynamic memory alloc even if that means we have to impose some limitation on the port.

src/device/usbd.c
hathach: yes, it should be usbd_ prefix and in the `usbd_pvt.h` header file for the class usage only.

src/portable/st/stm32_fsdev/dcd_stm32_fsdev.c
pigrew: Actually, for the case that CAN is not being used and no double-bufferig, I think we can statically allocate 64 bytes per endpoint. This will be simpler for now. 

Note I wasn't going to use the system malloc. PMA is not normal RAM, it's a weird double-ported thing. I was going to write a custom allocator.

src/device/usbd.c
pigrew: I think this is needed to flush pending transfers when the edpt is closed. Otherwise, the class driver will receive edpt complete callbacks after the endpoint is closed (and perhaps reopened). The class driver can't know if the transfers happened before or after the endpoint was closed.

src/portable/st/stm32_fsdev/dcd_stm32_fsdev.c
hathach: @pigrew please use that static 64 bytes per endpoints. My will to resist any form of dynamic allocation is very strong :D. we can define an address that is the start of PMA used by USB, since it is static, application could figure out where is the end of PMA used by USB and use the rest for CAN. If needed user can modify it, or we can export the configuration via an OPT_ later on. We can always revisit this later if it is an issue.

src/portable/st/stm32_fsdev/dcd_stm32_fsdev.c
pigrew: dcd_int_enable doesn't keep track of how many times it's called. The queue manipulation code needs interrupts disabled to run properly. If the interrupt were disabled when the function was called, then it would be reenabled prematurely.

FreeRTOS uses a counter to keep track of the nesting depth of disable calls (like a stack) so nested calls can be OK.

hathach: I mean the dcd API doesn't impose any requirement of interrupt state, if need, dcd just get the current interrupt state via NVIC, disable, and enable it if needed. The stack won't impose that requirement, unless we run into issue that need it.

src/device/usbd.c
hathach: Yes, you are right, that we need to have some kind of clean up/abort with the currently pending transfer. But I don't really like the way we loop through the task queue. We will need to find another way to address this.

src/portable/st/stm32_fsdev/dcd_stm32_fsdev.c
hathach: I could be wrong, thing are not really clear from my point of view now, maybe we could figure this out in the process.

pigrew: Yeah, but NVIC is ARM-specific (i think?).  There needs to be more BSP or DCD code to do refcounts. It was just easier to impose a limitation on calling the function, for now.

pigrew: Maybe create usbd_int_enterCritical or something to do refcount, and that calls the dcd functions. But it then creates a whole series of headaches since you don't want to reenable interrupts, if already in an interrupt handler.

hathach: it is inside dcd, it is specific to stm32 already. We could impose it when it is needed across ports. Since stm32 fsdev work with interrupt enabled or disabled, I don't see a reason to impose any limit just now. But it may just be me not seeing the whole picture for now. We can of course address it in the process.

src/portable/st/stm32_fsdev/dcd_stm32_fsdev.c
cr1901: I'm going to second this... no dynamic allocation, I can just _see_ how this will cause all sorts of fun problems down the road.

src/portable/st/stm32_fsdev/dcd_stm32_fsdev.c
pigrew: Oh, it's way too late here. I was thinking we were talking about usbd.c. My cell phone screen is too small.

You're right that it doesn't matter in this case. Apologies.

I'll fix up this and the other issues in the morning.

hathach: > Maybe create usbd_int_enterCritical or something to do refcount, and that calls the dcd functions. But it then creates a whole series of headaches since you don't want to reenable interrupts, if already in an interrupt handler.

Then just implement dcd_edpt_close() as it is, I still don't get why you need to eanble/disable interrupt or mutex access. If needed it will be perform by usbd, dcd just implement its part to close the endpoint.

pigrew: It is needed for the queue filtering. If interrupts were enabled, new events could end up in the middle of the queue.

hathach: @pigrew yeah, it makes sense now, but dcd should not assume usbd will perform any filtering on its queue. We may not end up use that filtering at all. Let's just have dcd close its opened endpoints. The closed endpoints should be best in disabled state. There is Alt interface that has more endpoints than others, and or with complete new configure (Interrupt --> Bulk) etc ..

hathach: > Oh, it's way too late here. I was thinking we were talking about usbd.c. My cell phone screen is too small.
> 
> You're right that it doesn't matter in this case. Apologies.
> 
> I'll fix up this and the other issues in the morning.

Yeah, I also missed this reply. Sorry for being picky, I only mean to remove the comment on top of the `dcd_edpt_close()` :). Yeah, please take some sleep, it must be late on your time (mine is only midnoon for now :) )

src/portable/st/stm32_fsdev/dcd_stm32_fsdev.c
pigrew: Removed in latest patch. Still untested. The F103 will now only support 4 bi-directional endpoints (4 IN and 4 OUT) by default, instead of the 8 it previously did.

src/portable/st/stm32_fsdev/dcd_stm32_fsdev.c
pigrew: I believe this is resolved now?

src/device/usbd.h
pigrew: Moved.

src/device/usbd.c
pigrew: Changed.

pigrew:
@cr1901, @hathach,

I think that your concerns have been resolved.

I will later add a `static void usbd_edpt_close_all()` function and mark this as ready for merging (assuming other issues don't come up).

docs/porting.md
```diff
@@ -119,6 +119,14 @@ Opening an endpoint is done for all non-control endpoints once the host picks a
 
 Also make sure to enable endpoint specific interrupts.
 
+##### dcd_edpt_close
+
+Close an endpoint. his function is used for implementing alternate settings.
+
+After calling this, the device should not respond to any packets directed towards this endpoint. When called, this function must abort any transfers in progress through this endpoint, before returning.
+
+Implementation is optional. Must be called from the USB task. Interrupts could be disabled or enabled during the call.
```
hathach, line 128: We actually don't have to tell where it is called or the context it is called. The only layer could access dcd API is usbd. (OK I know there is a direct class-to-dcd somewhere, it is a hack, will fix it later on :D )

src/device/usbd.c
hathach: please remove this as well.

docs/porting.md
pigrew: I think the edpt_open calls break the separation, as they are called directly from the class drivers.

But, yeah, we should just assume only the USB task can call the dcd_* functions. On the other hand, maybe a comment like this will prevent a bug in the future?

hathach: > I think the edpt_open calls break the separation, as they are called directly from the class drivers.

That is the one, I am too lazy for that changes.

> But, yeah, we should just assume only the USB task can call the dcd_* functions. On the other hand, maybe a comment like this will prevent a bug in the future?

`Interrupts could be disabled or enabled during the call` I am not so sure about this, I didn't see the need for this just now. When there is actually need, we could assert this once.

pigrew:
> To be honest, I would prefer this PR to only implement the dcd_edpt_close() on stm32 fsdev. And a thin `usbd_edpt_close()`. Let's leave the aborting transfer later until we could actually test it with an example.

I'm uneasy to do this. This would introduce a known bug in the case of pending transfers.

Yes, it's very hard to test since the close_ep() must be called at the exact time that data is being received. I'd rather introduce code that should work, than code that I know is buggy.

We could add unit tests to verify the filtering operation works, but it's hard to test with actual hardware.

---

Also, I'm wondering if the `close_all` is actually needed. An alternative would be a `class_driver->close()` call, as the class driver should know exactly which EP need closing.

hathach:
> I'm uneasy to do this. This would introduce a known bug in the case of pending transfers.
> 
> Yes, it's very hard to test since the close_ep() must be called at the exact time that data is being received. I'd rather introduce code that should work, than code that I know is buggy.
> 
> We could add unit tests to verify the filtering operation works, but it's hard to test with actual hardware.

Yes, I also don't like to have code that is not run as well, we could use unit test for this. Until that we shouldn't have code that call dcd_edpt_close(). An working example is still required to verify it works well on actual hardware.
 
> Also, I'm wondering if the `close_all` is actually needed. An alternative would be a `class_driver->close()` call, as the class driver should know exactly which EP need closing.

Right, it is my thought as well, again, I used to have driver close but dropped it due to lack of usage.

src/device/dcd.h
```diff
@@ -66,7 +66,7 @@ typedef struct TU_ATTR_ALIGNED(4)
 
     // USBD_EVT_XFER_COMPLETE
     struct {
-      uint8_t  ep_addr;
+      uint8_t  ep_addr; ///< 0xFF signifies that the transfer was aborted.
```
majbthrd, line 69: I'm not seeing the utility in creating a special ep_addr value to indicate that a non-unknown EP number has had an aborted transfer.  Why not, say, set the len to zero (and keep the EP number as-is)?

hathach: yeah, I think we need to spend a bit more time to evaluate how to address the on-going transfer. I plan to try it out with nrf52 to see what is the best approach.

pigrew: length zero can indicate a ZLP.

We could create a separate event type, EVENT_INVALID?

pigrew: Oh. I'm wrong. We can just get rid of the event entirely, no need to requeue it. The modification was based on the earlier revision of the code where it simplified life. :(

pigrew: (patch pushed to branch, so marking this as resolved.)

cr1901:
@pigrew Is it possible to eventually work back up to 8 bidir endpoints without dynamic allocation (maybe in a future PR)?

Or to allow 8 bidir endpoints without the ability to choose alternate interfaces?

Currently I have no plans of implementing this support in my MSP430 code, but I'll have to take a closer look at what you've done to make this work...

pigrew:
> @pigrew Is it possible to eventually work back up to 8 bidir endpoints without dynamic allocation (maybe in a future PR)?
> 
> Or to allow 8 bidir endpoints without the ability to choose alternate interfaces?
> 
> Currently I have no plans of implementing this support in my MSP430 code, but I'll have to take a closer look at what you've done to make this work...

This is a stm32fsdev limitation, not a TinyUSB limitation. 8 bidir EP is already supported on the STM32FSDEV MCU that have the 1024 byte buffer (e.g. STM32F0). The issue I mentioned shouldn't apply to MSP430 or the ST-Synopsys USB hardware.

For the 512 byte buffer stm32fsdev MCUs (e.g. STM32F1), I think they need dynamic buffer allocator in order to make it easy for the programmer if 8 EP are to be supported.

Previously, I did use dynamic allocation for the packet buffer memory for stm32fsdev, starting from offset 0 upon USB reset. I assumed that deallocation never happen, which greatly simplified the allocator.

Workarounds for 512-byte buffer MCU to get more EP:

* Set the packet size to 32 byte max and change a `#define` for it in `dcd_stm32fsdev`, it still can have the extra endpoints, as another option. (32*2*8 == 512 byte)
* The library user could override `dcd_stm32_pma_alloc` with a different static allocation that fits in the available buffer memory. It isn't currently set as weak, so would require changes to the library code, but isn't so bad. I don't expect anyone to do this. This could allow, e.g., some EP to use 64 byte, but others to use 16 byte packets. Perhaps the allocation has to change with configuration, and the DCD doesn't currently know  which configuration is active.
* Write a dynamic packet memory allocator.

The first workaround is already implemented. The other two could be, but might not be worth doing at the moment.

pigrew:
Ugh, on stm32fsdev, the EP limits are even lower than I thought. (I just fixed a bug in this PR found through testing...).

Each EP needs 8 bytes of control data in the packet buffer, so with 64 byte packets, there can only be 3 EP for the 512 byte models, or 7 for the 1024 byte models.

``` requiredBufferSpace = NUM_BIDIRECTIONAL_EP * (8 + packetSizeTX + packetSizeRX)```

hathach:
> > @pigrew Is it possible to eventually work back up to 8 bidir endpoints without dynamic allocation (maybe in a future PR)?
> > Or to allow 8 bidir endpoints without the ability to choose alternate interfaces?
> > Currently I have no plans of implementing this support in my MSP430 code, but I'll have to take a closer look at what you've done to make this work...
> 
> This is a stm32fsdev limitation, not a TinyUSB limitation. 8 bidir EP is already supported on the STM32FSDEV MCU that have the 1024 byte buffer (e.g. STM32F0). The issue I mentioned shouldn't apply to MSP430 or the ST-Synopsys USB hardware.
> 
> For the 512 byte buffer stm32fsdev MCUs (e.g. STM32F1), I think they need dynamic buffer allocator in order to make it easy for the programmer if 8 EP are to be supported.
> 
> Previously, I did use dynamic allocation for the packet buffer memory for stm32fsdev, starting from offset 0 upon USB reset. I assumed that deallocation never happen, which greatly simplified the allocator.
> 
> Workarounds for 512-byte buffer MCU to get more EP:
> 
> * Set the packet size to 32 byte max and change a `#define` for it in `dcd_stm32fsdev`, it still can have the extra endpoints, as another option. (32 * 2 * 8 == 512 byte)
> * The library user could override `dcd_stm32_pma_alloc` with a different static allocation that fits in the available buffer memory. It isn't currently set as weak, so would require changes to the library code, but isn't so bad. I don't expect anyone to do this. This could allow, e.g., some EP to use 64 byte, but others to use 16 byte packets. Perhaps the allocation has to change with configuration, and the DCD doesn't currently know  which configuration is active.
> * Write a dynamic packet memory allocator. (perhaps use a very simple implementation which doesn't support deallocation, for the case with no alternate configuration/interface)
> 
> The first workaround is already implemented. The other two could be, but might not be worth doing at the moment.

Ah, I see the problem now, 512 bytes for 8 pairs of endpoints is a bit short. Control and Bulk are all 64, Interrupt endpoints is only type can be smaller. The number of endpoints limit is actually worth the trouble we get into though. 

From my observation for Alt Interface, it is either use an EP or not, it is rarely both use an endpoints with different packet size (maybe a real de-allocator is not needed). EXCEPT well for the ISOCHRONOUS endpoint, these comes usually comes with series of Alt Interface with different packet size for different bandwidth requirement. However, those ISO is not supported yet, and it seems to need a special buffer allocation with double buffer as well. Since we need to update dcd to support ISO later, maybe we can postpone this until that ?

TLDR: for non-iso, close endpoint then re-open it will use the same buffer location.

pigrew:
@hathach,

I'm tempted to go back to dynamic allocation, with no deallocation (like it had been). Stipulating that all instances of a particular EP number have equal max packet lengths removes the need for one-by-one deallocations. Do classes ever need to change the packet length of an endpoint? (My guess is no)

Special code can be added to clear allocations when all EP are closed (such as during reset and set_config).

hathach:
> @hathach,
> 
> I'm tempted to go back to dynamic allocation, with no deallocation (like it had been). Stipulating that all instances of a particular EP number have equal max packet lengths removes the need for one-by-one deallocations. Do classes ever need to change the packet length of an endpoint? (My guess is no)
> 
> Special code can be added to clear allocations when all EP are closed (such as during reset and set_config).

Yeah, let's go back to it, simple dynamic without deallocating is OK though. I don't it is count as dynamic alloc, since we only increase the head of buffer address. totally OK with that. Yes, the packet size won't change, the only scenario is with ALT ITF + ISO endpoints. However, since we didn't support ISO yet, we can keep it as it is for now. 

ITF 1 ALT 1 has iso endpoint size = 0x00A0 while ITF 1 ALT 2 has size = 0x00D0, this is normally involved to bandwidth feature such as webcam resolution = 360p, 480p (each resolution is an ALT interface). 
image

PS: Sorry for over-reacting with the term dynamic alloc :D

pigrew:
@hathach 

I've reimplemented the PMA allocator for stm32fsdev. Basic testing has been done, and I believe that this could be ready for merging.

src/device/usbd.c
hathach: I still couldn't find this event queue filtering is a good approach. It potentially introduce other problems into the stack. In case it does indeed remove an few events. Those are valid and successful transfer, application/class driver e.g MSC could be blocking wait for those. The host close the pipe after those transfer, e.g HOST send eject command to MSC, then close the interface. But since we drop eject transfer, it makes no sense to application.

pigrew:
@hathach 

The issue is the following (imagine with streaming video, though the same could happen with CDC):

1. Host sends bulk OUT data (480p)
2. Host decides to change interface to (240p), sends SET_INTERFACE
3. Device receives both transfers; both received, and transfers are put into event queue (with SET_INTERFACE first, since EP0 may have higher priority than other EP.)
4. Device handles SET_INTERFACE, closes  & reopens other EP, now expects 240p data
5. Event queue processes 480p bulk OUT xfer, but this confuses the class driver since it is now expecting 240p.

Similar to that, for CDC:
1. Host sends bulk OUT data (480p)
2. Host decides to close interface (alt setting 0), sends SET_INTERFACE
3. Device receives both transfers; both received, and transfers are put into event queue (with SET_INTERFACE first, since EP0 may have higher priority than other EP.)
4. Device handles SET_INTERFACE, closes  EP.
5. Event queue processes BULK OUT xfer. Class driver gets confused since it isn't expecting any more data since alt setting 0 is active.

You're right that this may take some more thinking, since the BULK OUT data had been ACKed by the device, so the host would think that it had been delivered. I think the key is that when SET_INTERFACE is sent, the host should be expecting the device to mostly reset itself, so losing data is expected. And also that the USB ACK doesn't mean that the packet was processed by the device, it only means that it was received by the device.

hathach:
> I think the key is that when SET_INTERFACE is sent, the host should be expecting the device to mostly reset itself, so losing data is expected. And also that the USB ACK doesn't mean that the packet was processed by the device, it only means that it was received by the device.

This is not quite correct, a decent Host will never try to communicate to an endpoint that it knows is closed, not even try for NAK, it will simply remove the endpoint from its active queue list. However, it will expect device honors any transfer complete before the SET_INTERFACE. Normally host will stop communicate with that endpoint in advance and send out the request. It won't suddenly close endpoint in the middle of transferring. I mean it won't assume any corruption transfer happened with ACKed transfer.

pigrew:
@hathach,

I've removed the queue filtering, (so hopefully this PR can be merged).

I'm still of the opinion that it should be there, but I'll give it some more time to think about.

If you notice my chain of events above, the bulk transfer was started BEFORE the SET_INTERFACE call. The USB device hardware swaps the order of them being delivered. On STM32, it gives priority to the lower-numbered EP index.

hathach:
> @hathach,
> 
> I've removed the queue filtering, (so hopefully this PR can be merged).
> 
> I'm still of the opinion that it should be there, but I'll give it some more time to think about.

Yeah, It give us more time to think, especially finding an actual example that we can test on a hardware. 

> If you notice my chain of events above, the bulk transfer was started BEFORE the SET_INTERFACE call. The USB device hardware swaps the order of them being delivered. On STM32, it gives priority to the lower-numbered EP index.

It shouldn't be an issue, control is 2,3-stage. Host will stop communicating with about-to-close endpoint before sending out the setup packet.

PR #337, opened 2020-04-13
introducing usbd_interface_t

hathach:
it is an enhancement, there will be an additional class_get_interface() which return the pointer to currently active (alt) interface from class driver. 

Note: currently there is not much benefit to use this, we may not use this at all. I am sorting my thought on how to implement alt interface.

### DIFF
--- a/src/class/hid/hid_device.c
+++ b/src/class/hid/hid_device.c
@@ -40,9 +40,15 @@
 //--------------------------------------------------------------------+
 typedef struct
 {
-  uint8_t itf_num;
-  uint8_t ep_in;
-  uint8_t ep_out;        // optional Out endpoint
+//  uint8_t itf_num;
+//  uint8_t itf_alt;
+//  uint8_t ep_in;
+//  uint8_t ep_out;        // optional Out endpoint
+
+  // Interface with max 2 endpoints
+  // must be the first
+  usbd_interface_t(2);
+
   uint8_t boot_protocol; // Boot mouse or keyboard
   bool    boot_mode;     // default = false (Report)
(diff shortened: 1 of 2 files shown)


pigrew:
I'm also not sure if this is helpful. It seems to make the memory footprint increase with little benefit. My feeling is that the class driver should handle SET_INTERFACE, or USBD should handle it (by deinitializing and reinitializing the class driver during SET_INTERFACE).

Designing an API is hard.

One related thing I was pondering was to let the application pass a class-specific configuration structure to the class driver initializer (instead of having the class driver parse the configuration descriptor. Also, maybe pass the IAD descriptor to the class driver's open function, instead of  the first interface.

Lots of ideas, not sure if there is much benefit to any of them.

hathach:
> I'm also not sure if this is helpful. It seems to make the memory footprint increase with little benefit. My feeling is that the class driver should handle SET_INTERFACE, or USBD should handle it (by deinitializing and reinitializing the class driver during SET_INTERFACE).
> 
> Designing an API is hard.
> 
> One related thing I was pondering was to let the application pass a class-specific configuration structure to the class driver initializer (instead of having the class driver parse the configuration descriptor. Also, maybe pass the IAD descriptor to the class driver's open function, instead of the first interface.
> 
> Lots of ideas, not sure if there is much benefit to any of them.

Yeah, there isn't much benefit for this yet. I will try to focus on the ALT ITF descriptor side instead. Though I still want a consistent interface data structure for usbd to manage later on, much like base  interface in C++. We will keep this open for now until it has more clear benefit.

hathach:
close now, back later if ever needed.

PR #338, opened 2020-04-13
lpc17xx move set configdevice into set address for removing dcd_set_config()

hathach:
This is technically not right, but it will have lpc17 behaves the same as other mcu vendor. Tested and works well on lpc4088 qs

### DIFF
--- a/docs/porting.md
+++ b/docs/porting.md
@@ -78,10 +78,6 @@ Called when the device is given a new bus address.
 
 If your peripheral automatically changes address during enumeration (like the nrf52) you may leave this empty and also no queue an event for the corresponding SETUP packet.
 
-##### dcd_set_config
-
-Called when the device received SET_CONFIG request, you can leave this empty if your peripheral does not require any specific action.
-
 ##### dcd_remote_wakeup
 
 Called to remote wake up host when suspended (e.g hid keyboard)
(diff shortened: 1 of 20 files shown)


hathach:
@majbthrd if you could remove the use of `dcd_set_config()` in nuc505, I will remove the API altogether in this PR. We can then sync your PR #326 with this to continue our work there.

hathach:
@majbthrd currently nuc505 only use set_config() to reset DToggle for endpoints. Maybe it can be achieved somewhere. Normally after setConfig is endpoint open, which already set DToggle to DATA0 already !!

majbthrd:
> 
> 
> @majbthrd currently nuc505 only use set_config() to reset DToggle for endpoints. Maybe it can be achieved somewhere. Normally after setConfig is endpoint open, which already set DToggle to DATA0 already !!

I've added PR #347, which follows the same sort of approach that you did for lpc17xx.

hathach:
> The existing dcd_set_config() call in ./device/usbd.c may disappear, but something eventually needs to replace that call in the future once SET_CONFIGURATION supports multiple configurations. IMHO, the replacement should disable all non-control endpoints (whether that is implemented as: a) a direct DCD API to perform this action, or b) some middleman/shim routine that uses another DCD API to exhaustively close all endpoints).
> 
> I'm personally partial to (a), but I can also see the upside of (b), since the same DCD API called in that situation might also benefit future classes with more elaborate alternate interfaces.

Yeah, the (b) dedicated close_all() is also very useful for some ports that are require a bit of buffer allocating based on packetsize like stm32 fsdev. It is essentially an bus_reset() without the actual hardware signal. We will see though, no rush into that, since there is no benefit yet :smiley:

PR #339, opened 2020-04-13
implment dcd_edpt_close_all() API

hathach:
Although the API will only be used (for now) when switching the configuration. But it is trivial enough for DCD to implement, mostly is a loop to disable all non-control endpoints enabled bit or so. It should be mandatory instead of optional.

Port checklikst
- [ ] Espressif ESP32-S2
- [ ] Microchip SAMD
- [ ] Microchip SAMG 
- [ ] Nordic nRF5x 
- [ ] Nuvoton NUC120
- [ ] Nuvoton NUC121/125/126
- [ ] Nuvoton NUC505
- [ ] NXP LPC3511
- [ ] NXP LPC17/40
- [ ] NXP Transdimension (18/ iMX RT)
- [ ] Sony CXD56
- [ ] STM32 FSDev
- [ ] STM32 Synopsys
- [ ] TI MSP430
- [ ] ValentyUSB eptri (fomu)

### DIFF
--- a/src/device/dcd.h
+++ b/src/device/dcd.h
@@ -132,6 +132,9 @@ void dcd_edpt_stall       (uint8_t rhport, uint8_t ep_addr);
 // clear stall, data toggle is also reset to DATA0
 void dcd_edpt_clear_stall (uint8_t rhport, uint8_t ep_addr);
 
+// Close/Disable all non-control endpoints
+void dcd_edpt_close_all(uint8_t rhport);
+
 //--------------------------------------------------------------------+
 // Event API (Implemented by device stack)
 //--------------------------------------------------------------------+


pigrew:
I think that this doesn't need to be part of the dcd API. I think it'll only be called internally from `usbd.c` when changing configuration?

Perhaps I could add something to my `edpt_close` patch, such as:

```
static void usbd_edpt_close_all(rhport) {
  for i=0:END-1:
    dcd_edpt_close(i);
    dcd_edpt_close(i | 0x80);
  queue_filter_remove_pending_transfer(ALL);
}
```

hathach:
> I think that this doesn't need to be part of the dcd API. I think it'll only be called internally from `usbd.c` when changing configuration?
> 
> Perhaps I could add something to my `edpt_close` patch, such as:
> 
> ```
> static void usbd_edpt_close_all(rhport) {
>   for i=0:END-1:
>     dcd_edpt_close(i);
>     dcd_edpt_close(i | 0x80);
>   queue_filter_remove_pending_transfer(ALL);
> }
> ```

Yeah, maybe you are right, I thought it is a good idea to have a close_all(). After having dinner full of rice, now I start to see it is not really a good one.

hathach:
I created an new issue here to keep check of dcd_edpt_close() for all ports here #340

PR #342, opened 2020-04-14
add subclass and protocol for matching driver id

### LINKED ISSUE #205: Class code driver conflict application specific 0xfe between USBTMC and Runtime DFU

hathach:
fix #205 , this help to resolve conflict of currently class driver. Tested with msc_cdc, hid_composite, and net_lwip_webserver with nrf52840. All seems to work

### DIFF
--- a/src/device/usbd.c
+++ b/src/device/usbd.c
@@ -190,11 +190,6 @@ static usbd_class_driver_t const _usbd_driver[] =
   #if CFG_TUD_NET
   /* RNDIS management interface */
   {
-//      .class_code       = TUD_RNDIS_ITF_CLASS,
-//      .subclass         = TUD_RNDIS_ITF_SUBCLASS,
-//      .protocol         = TUD_RNDIS_ITF_PROTOCOL,
-//      .all_subclass     = 0,
-//      .all_protocol     = 0,
       DRIVER_NAME("RNDIS")
       .init             = netd_init,
       .reset            = netd_reset,
@@ -204,14 +199,9 @@ static usbd_class_driver_t const _usbd_driver[] =
       .xfer_cb          = netd_xfer_cb,
       .sof              = NULL,
   },
+
   /* CDC-ECM management interface */
   {
-//      .class_code       = TUSB_CLASS_CDC,
-//      .subclass         = CDC_COMM_SUBCLASS_ETHERNET_NETWORKING_CONTROL_MODEL,
-//      .protocol         = 0x00,
-//      .all_subclass     = 0,
-//      .all_protocol     = 0,
-
       DRIVER_NAME("CDC-ECM")
       .init             = netd_init,
       .reset            = netd_reset,
@@ -221,13 +211,9 @@ static usbd_class_driver_t const _usbd_driver[] =
       .xfer_cb          = netd_xfer_cb,
       .sof              = NULL,
   },
+
   /* RNDIS/CDC-ECM data interface */
   {
-//      .class_code       = TUSB_CLASS_CDC_DATA,
-//      .subclass         = 0x00,
-//      .protocol         = 0x00,
-//      .all_subclass     = 0,
-//      .all_protocol     = 0,
       DRIVER_NAME("CDC-DATA")
       .init             = netd_init_data,
       .reset            = NULL,
--- a/src/device/usbd.c
+++ b/src/device/usbd.c
@@ -76,8 +76,8 @@ enum { DRVID_INVALID = 0xFFu };
 typedef struct
 {
   uint8_t class_code;
-  uint8_t subclass; // 0xFF support all values of subclass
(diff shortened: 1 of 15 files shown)


src/device/usbd.c
```diff
@@ -79,6 +79,12 @@ typedef struct
   uint8_t subclass; // 0xFF support all values of subclass
   uint8_t protocol; // 0xFF support all values of protocol
```
pigrew: Comments should be removed?

src/device/usbd.c
```diff
@@ -209,6 +229,8 @@ static usbd_class_driver_t const _usbd_driver[] =
       .class_code       = TUD_RNDIS_ITF_CLASS,
       .subclass         = TUD_RNDIS_ITF_SUBCLASS,
       .protocol         = TUD_RNDIS_ITF_PROTOCOL,
+      .all_subclass     = 0,
+      .all_protocol     = 0,
```
pigrew, line 233: Coding style question: Add these lines, or just leave them out (the compiler will default to setting them to 0)?

src/device/usbd.c
hathach: yeah right

src/device/usbd.c
hathach: better to explicitly assign them, else I think we can got into C++ compiler issue with non-trivial initialization error.

hathach:
> However, in the bigger picture, I'd like to think about this PR vs a class callback:
> 
> `bool class_driver->supports_intf(usb_desc_t* desc)`
> 
> where desc is an IAD descriptor or an interface descriptor.
> 
> I think the total firmware size would be nearly unchanged (might go up or down by a few bytes). The callback would provide more flexibility than this solution.

Yeah, I think your suggestion is much better than the current one. Actually we can bundle this into open() API, upon return false, the usbd will try to open with next driver. I will try to do that in this PR as well.

pigrew:
> Yeah, I think your suggestion is much better than the current one. Actually we can bundle this into open() API, upon return false, the usbd will try to open with next driver. I will try to do that in this PR as well.

Sounds good. I can't think of a reason why usbd needs to know if a driver supports an interface, other than when it is being opened.

So, ```open```  will have three results:

* EP opened (returns true)
* Interface is incorrect (returns false).
* Bad things happened (TU_ASSERT enters infinite loop)

Something about returning ```false``` not being an error bothers me, but I think it's fine in this case.

hathach:
> > Yeah, I think your suggestion is much better than the current one. Actually we can bundle this into open() API, upon return false, the usbd will try to open with next driver. I will try to do that in this PR as well.
> 
> Sounds good. I can't think of a reason why usbd needs to know if a driver supports an interface, other than when it is being opened.
> 
> So, `open` will have three results:
> 
> * EP opened (returns true)
> * Interface is incorrect (returns false).
> * Bad things happened (TU_ASSERT enters infinite loop)
> 
> Something about returning `false` not being an error bothers me, but I think it's fine in this case.

The original version is much different than current one, it is kind of the coding-while-figuring out code :D

src/device/usbd.h
```diff
@@ -377,7 +377,7 @@ TU_ATTR_WEAK bool tud_vendor_control_complete_cb(uint8_t rhport, tusb_control_re
   #define TUD_RNDIS_ITF_PROTOCOL 0xFF /* CDC_COMM_PROTOCOL_MICROSOFT_RNDIS */
 #else
   /* Windows 7+ */
-  #define TUD_RNDIS_ITF_CLASS    0xE0
+  #define TUD_RNDIS_ITF_CLASS    TUSB_CLASS_WIRELESS_CONTROLLER
```
hathach, line 380: @majbthrd  First it is not related to this 380 line, but on line 350 `#define TUD_CDC_ECM_DESCRIPTOR()` , github doesn't allow to comment on unmodified line !!!. I just notice ECM doesn't have Interface Assoc Descriptor. It is a bit awkward since it is an CDC subclass. It could work without IAD, but maybe since our example has only 1 usbnet, it could run into issue with a composite device. I will try to add IAD and test. Though just want to check with you first

hathach: update: hehe, I couldn't wait until you wake up, give it a try by adding IAD and it works fine on Linux. I will include this in my next push with IAD enhance, then ALT ITF. Let me know if there is anything preventing us to use IAD in this case. 

```
  /* Interface Association */\
  8, TUSB_DESC_INTERFACE_ASSOCIATION, _itfnum, 2, TUSB_CLASS_CDC, CDC_COMM_SUBCLASS_ETHERNET_NETWORKING_CONTROL_MODEL, 0, 0,\
```

hathach:
> Wow, so many more changes than I had envisioned. :(
> 
> I think it's improved. Firmware even gets smaller!

Yeah, it is better and yet smaller. That is the power of collaborating :tada:

PR #343, opened 2020-04-15
usbnet: remove CDC-EEM

majbthrd:
CDC-EEM is too obscure a protocol to justify [the request to include it as a run-time option](https://github.com/hathach/tinyusb/pull/334#pullrequestreview-392807281).

It seems a more prudent choice is to just cull the source code associated with CDC-EEM.

### DIFF
--- a/src/device/usbd.c
+++ b/src/device/usbd.c
@@ -184,7 +184,6 @@ static usbd_class_driver_t const _usbd_driver[] =
   #endif
 
   #if CFG_TUD_NET
-#if CFG_TUD_NET != OPT_NET_EEM
   /* RNDIS management interface */
   {
       .class_code       = TUD_RNDIS_ITF_CLASS,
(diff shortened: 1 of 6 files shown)


src/device/usbd.c
```diff
@@ -196,24 +195,18 @@ static usbd_class_driver_t const _usbd_driver[] =
...
-#endif
       .control_request  = netd_control_request,
       .control_complete = netd_control_complete,
       .xfer_cb          = netd_xfer_cb,
       .sof              = NULL,
   },
   /* RNDIS/CDC-ECM data interface */
-#if CFG_TUD_NET != OPT_NET_EEM
   {
       .class_code       = TUSB_CLASS_CDC_DATA,
```
hathach, line 211: if possible, please consider to merge them together as `cdc_device.c`, it is much cleaner and easier for class driver that way.

hathach:
Ah I see, due to the setInterface() you need to seperate the data to its own class. Nevermind that, I will merge them together with my next PR for alt itf

PR #344, opened 2020-04-15
Enhance Interface Associate descriptor handling

hathach:
Follow up to #342 (inherit all its changes for now, should be updated when it is merged)
- assign driver ID for all associated interface
- cdc template protocol to None (previously ATCommand)
- add IAD for USB NET CDC-ECM descriptor template
- merge net_data back into net driver

### DIFF
--- a/src/device/usbd.c
+++ b/src/device/usbd.c
@@ -188,38 +188,13 @@ static usbd_class_driver_t const _usbd_driver[] =
   #endif
 
   #if CFG_TUD_NET
-  /* RNDIS management interface */
   {
-      DRIVER_NAME("RNDIS")
+      DRIVER_NAME("NET")
       .init             = netd_init,
       .reset            = netd_reset,
       .open             = netd_open,
       .control_request  = netd_control_request,
       .control_complete = netd_control_complete,
-      .xfer_cb          = netd_xfer_cb,
-      .sof              = NULL,
-  },
-
-  /* CDC-ECM management interface */
-  {
-      DRIVER_NAME("CDC-ECM")
-      .init             = netd_init,
-      .reset            = netd_reset,
-      .open             = netd_open,
-      .control_request  = netd_control_request,
-      .control_complete = netd_control_complete,
-      .xfer_cb          = netd_xfer_cb,
-      .sof              = NULL,
-  },
-
-  /* RNDIS/CDC-ECM data interface */
-  {
-      DRIVER_NAME("CDC-DATA")
-      .init             = netd_init_data,
-      .reset            = NULL,
-      .open             = netd_open_data,
-      .control_request  = NULL,
-      .control_complete = NULL,
       .xfer_cb          = netd_xfer_cb,
       .sof              = NULL,
   },
@@ -611,9 +586,10 @@ static bool process_control_request(uint8_t rhport, tusb_control_request_t const
           case TUSB_REQ_SET_INTERFACE:
           {
             uint8_t const alternate = (uint8_t) p_request->wValue;
+            (void) alternate;
 
             // TODO not support alternate interface yet
-            TU_ASSERT(alternate == 0);
+//            TU_ASSERT(alternate == 0);
             tud_control_status(rhport, p_request);
           }
           break;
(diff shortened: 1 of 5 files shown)


src/device/usbd.c
```diff
@@ -727,41 +703,59 @@ static bool process_set_config(uint8_t rhport, uint8_t cfg_num)
...
+    uint8_t drv_id;
+    uint16_t drv_len;
 
-      for (drv_id = 0; drv_id < USBD_CLASS_DRIVER_COUNT; drv_id++)
+    for (drv_id = 0; drv_id < USBD_CLASS_DRIVER_COUNT; drv_id++)
+    {
+      usbd_class_driver_t const *driver = &_usbd_driver[drv_id];
+
+      drv_len = 0;
+      if ( driver->open(rhport, desc_itf, &drv_len) )
```
pigrew, line 726: Should the IAD be passed to ```open```, or the first interface?

I don't think it matters so much since USB classes are not defined with a variable number of interfaces?

src/device/usbd.c
```diff
@@ -727,41 +703,59 @@ static bool process_set_config(uint8_t rhport, uint8_t cfg_num)
...
-          TU_ASSERT( DRVID_INVALID == _usbd_dev.itf2drv[desc_itf->bInterfaceNumber] );
-          TU_LOG2("  %s open\r\n", _usbd_driver[drv_id].name);
-          _usbd_dev.itf2drv[desc_itf->bInterfaceNumber] = drv_id;
-          break;
+          // IAD's first interface number and class/subclass/protocol should match with opened interface
+          TU_ASSERT(desc_itf_assoc->bFirstInterface   == desc_itf->bInterfaceNumber   &&
+                    desc_itf_assoc->bFunctionClass    == desc_itf->bInterfaceClass    &&
+                    desc_itf_assoc->bFunctionSubClass == desc_itf->bInterfaceSubClass &&
+                    desc_itf_assoc->bFunctionProtocol == desc_itf->bInterfaceProtocol);
+
```
pigrew, line 742: Should there be checks that the device descriptor's class==0xEF, subclass=0x02, protocol=0x01?

This seems to be said in the ```iadclasscode_r10.pdf``` spec. But, we probably can't check that here, as there isn't a pointer to the device descriptor?

However, in the USB 2.0 IAD ECN, it says this is "strongly recommended", but not required.

src/device/usbd.c
```diff
@@ -727,41 +703,59 @@ static bool process_set_config(uint8_t rhport, uint8_t cfg_num)
...
+        // If IAD exist, assign all interfaces to the same driver
+        if (desc_itf_assoc)
         {
-          // Interface number must not be used already TODO alternate interface
-          TU_ASSERT( DRVID_INVALID == _usbd_dev.itf2drv[desc_itf->bInterfaceNumber] );
-          TU_LOG2("  %s open\r\n", _usbd_driver[drv_id].name);
-          _usbd_dev.itf2drv[desc_itf->bInterfaceNumber] = drv_id;
-          break;
+          // IAD's first interface number and class/subclass/protocol should match with opened interface
```
pigrew, line 737: The spec says:

```The values in these fields can be the same as the interface class, subclass and protocol values from any one of the associated interfaces```

not that it is the first one.

So, maybe change the comment to say that this is a TinyUSB requirement, not a spec requirement.

File: src/device/usbd.c [hunk 4]
```diff
@@ -611,9 +586,10 @@ static bool process_control_request(uint8_t rhport, tusb_control_request_t const
           case TUSB_REQ_SET_INTERFACE:
           {
             uint8_t const alternate = (uint8_t) p_request->wValue;
+            (void) alternate;
 
             // TODO not support alternate interface yet
-            TU_ASSERT(alternate == 0);
+//            TU_ASSERT(alternate == 0);
```
pigrew, line 592: This isn't technically correct, but should be fixed soon with other PR.

hathach:
gh tip: have to change base to other, then back to master to get the file diffs updated :D

src/device/usbd.h
```diff
@@ -343,11 +343,13 @@ TU_ATTR_WEAK bool tud_vendor_control_complete_cb(uint8_t rhport, tusb_control_re
...
 
 // Length of template descriptor: 71 bytes
-#define TUD_CDC_ECM_DESC_LEN  (9+5+5+13+7+9+9+7+7)
+#define TUD_CDC_ECM_DESC_LEN  (8+9+5+5+13+7+9+9+7+7)
 
 // CDC-ECM Descriptor Template
 // Interface number, description string index, MAC address string index, EP notification address and size, EP data address (out, in), and size, max segment size.
 #define TUD_CDC_ECM_DESCRIPTOR(_itfnum, _desc_stridx, _mac_stridx, _ep_notif, _ep_notif_size, _epout, _epin, _epsize, _maxsegmentsize) \
+  /* Interface Association */\
+  8, TUSB_DESC_INTERFACE_ASSOCIATION, _itfnum, 2, TUSB_CLASS_CDC, CDC_COMM_SUBCLASS_ETHERNET_NETWORKING_CONTROL_MODEL, 0, 0,\
```
pigrew, line 352: For IAD, do we need to change the device descriptor's class,subclass,protocol to be IAD device?

pigrew:
> gh tip: have to change base to other, then back to master to get the file diffs updated :D

Yep.

Looks like I'm 20 seconds too late with my review!

It's mostly comment changes, though.

hathach:
> > gh tip: have to change base to other, then back to master to get the file diffs updated :D
> 
> Yep.
> 
> Looks like I'm 20 seconds too late with my review!
> 
> It's mostly comment changes, though.

oops, sorry, I am too eager to merge, please just leave the comment, I will fix it in the final PR #345.

src/device/usbd.c
hathach: It doesn't make much difference for class driver since they already know what to do. It is mostly for host for managing and dynamic driver binding.  not all class support IAD, since we use open() as driver check now, best to keep it simple for all. We can always change later on if needed.

src/device/usbd.c
hathach: It is actually host responsibility to check for that, and not send setConfig() if device violate the IAD specs. Though I doubt if Linux would do that. As device, we don't really sanity check everything to keep it simple :).

src/device/usbd.c
hathach: You are right on the specs, but it does also say its prefer as well :) . I will keep this as it is, until there is a actual class that does differently. There is TU_ASSSERT(), we can always catch this error.
image

src/device/usbd.c
hathach: yeah, it is tempt hack to test out the request. I forgot to revert the change.

src/device/usbd.h
hathach: the IAD specs says so, I did it in most of my example. That is duty of host to validate, if it get to this point setConfig(), it is all settled.

pigrew: I've created a PR for this change, so setting this comment as resolved.

PR #345, opened 2020-04-15
Implement setInterface(alt) for usb net driver

hathach:
- mostly forward these request (recipient = interface) to class driver.
- also slightly update the lwip example descriptor

follow up to #342  and #344 (merging these first make bisecting easier)
related to #320 #317


pigrew:
I believe that GET_INTERFACE must be always implemented?

I think that either it needs to be handled in usbd (as in#317 and #320), or a handler must be added to all classes.

examples/device/net_lwip_webserver/src/usb_descriptors.c
```diff
@@ -106,30 +105,41 @@ enum
...
+  [CONFIG_NUM_ECM-1  ] = ecm_configuration
+};
+
 // Invoked when received GET CONFIGURATION DESCRIPTOR
 // Application return pointer to descriptor
 // Descriptor contents must exist long enough for transfer to complete
 uint8_t const * tud_descriptor_configuration_cb(uint8_t index)
 {
-  return (0 == index) ? main_configuration : alt_configuration;
+  return configuration_arr[index];
```
majbthrd, line 142: Should there be bound checking?  The advantage of the old approach was a valid pointer configuration descriptor would always be returned.

hathach:
> I believe that GET_INTERFACE must be always implemented?
> 
> I think that either it needs to be handled in usbd (as in #317 and #320), or a handler must be added to all classes' control request callback

Yeah, you are right, macOS does sometime send out that request. I think it will just send that request toward CDC class. Just fixed in the new commit.

examples/device/net_lwip_webserver/src/usb_descriptors.c
hathach: yeah right, I am too focused on the usbd. Just fixed it in the new commit. Also slightly changed the CONFIG_NUM to CONFIG_ID. Let's me know if you want to keep the NUM convention more.

hathach:
> The revised tud_descriptor_configuration_cb() in examples/device/net_lwip_webserver/src/usb_descriptors.c causes me some heartburn, but it isn't that big a deal.

Why is that, I am open to all suggestion. 

> One of the tests I did was to hack up usbd.c to support multiple configurations and use a DCD API that can reset non-control endpoints. That allows repeatedly switching between RNDIS and ECM, which is a good way to torture test operation.
> 
> Thanks, @hathach!

Yeah, dynamic switching configuration require an close all endpoints. We will get there with dcd_edpt_close(), but I am not rush to do that. Since we solve all the problems for usb-net (multiple configuration + setInterface()). Although we don't close the Data's endpoints. Host won't actually try to read/write to those endpoints anyway (since we trick it to think so :D ) . So it is rather safe in this class driver. 

When adding Audio or UVC (video) class driver, a proper edpt_close is needed, hopefully this buys us enough of time.

majbthrd:
> > The revised tud_descriptor_configuration_cb() in examples/device/net_lwip_webserver/src/usb_descriptors.c causes me some heartburn, but it isn't that big a deal.
> 
> Why is that, I am open to all suggestion.

Sorry, I was referring to the older implementation without bounds-checking.  I submitted the review about the same time you pushed the latest commits.

hathach:
> > > The revised tud_descriptor_configuration_cb() in examples/device/net_lwip_webserver/src/usb_descriptors.c causes me some heartburn, but it isn't that big a deal.
> > 
> > 
> > Why is that, I am open to all suggestion.
> 
> Sorry, I was referring to the older implementation without bounds-checking. I submitted the review about the same time you pushed the latest commits.

yeah, It is hard to get that though :D since host won't attempt to set an invalid config number. But right, better to be safe.

pigrew:
@hathach,

I'm finding it hard to go through all these changes at once. 

Could you change this PR to be against the 'enhance-itf-assoc' branch or merge 'class-driver-id' so it's more clear exactly what this PR is?

I'm trying to go through the PR in order of application to 'master' right now.

hathach:
> @hathach,
> 
> I'm finding it hard to go through all these changes at once.
> 
> Could you change this PR to be against the 'enhance-itf-assoc' branch or merge 'class-driver-id' so it's more clear exactly what this PR is?
> 
> I'm trying to go through the PR in order of application to 'master' right now.

Yeah, it is a bit confusing, Basically they share same commit path, I made an PR #342 then branch from that to make PR #344  then make a branch and to this PR. I those PR is merged in order, it will make it easier to follow.

hathach:
gh tips: change target base to another, then back to master to get file changes updated :D

src/class/net/net_device.c
```diff
@@ -226,33 +249,89 @@ static void ecm_report(bool nc)
...
+
+          // Only valid for Data Interface
+          TU_ASSERT(_netd_itf.itf_num+1 == req_itfnum);
+
+          _netd_itf.itf_data_alt = (uint8_t) request->wValue;
+
+          if ( _netd_itf.itf_data_alt )
+          {
+            // TODO since we don't actually close endpoint
+            // hack here to not re-open it
```
pigrew, line 276: Should we delay this PR until the close_endpoint PR is merged? That way we're not introducing more hacky code?

src/class/net/net_device.c
```diff
@@ -226,33 +249,89 @@ static void ecm_report(bool nc)
...
+
+        case TUSB_REQ_SET_INTERFACE:
+        {
+          uint8_t const req_itfnum = (uint8_t) request->wIndex;
+
+          // Request to enable/disable network activities on ACM-ECM only
+          TU_ASSERT(_netd_itf.ecm_mode);
+
+          // Only valid for Data Interface
+          TU_ASSERT(_netd_itf.itf_num+1 == req_itfnum);
```
pigrew, line 269: Not a TU_ASSERT. This should be a TU_VERIFY.

src/class/net/net_device.c
```diff
@@ -226,33 +249,89 @@ static void ecm_report(bool nc)
...
+        case TUSB_REQ_GET_INTERFACE:
+          tud_control_xfer(rhport, request, &_netd_itf.itf_data_alt, 1);
+        break;
+
+        case TUSB_REQ_SET_INTERFACE:
+        {
+          uint8_t const req_itfnum = (uint8_t) request->wIndex;
+
+          // Request to enable/disable network activities on ACM-ECM only
+          TU_ASSERT(_netd_itf.ecm_mode);
```
pigrew, line 266: Maybe should be TU_VERIFY?

src/class/net/net_device.c
```diff
@@ -226,33 +249,89 @@ static void ecm_report(bool nc)
...
+
+          // Only valid for Data Interface
+          TU_ASSERT(_netd_itf.itf_num+1 == req_itfnum);
+
+          _netd_itf.itf_data_alt = (uint8_t) request->wValue;
+
```
pigrew, line 272: Should return false (error) if the requested altSetting is >1.

src/device/usbd.c
```diff
@@ -571,39 +571,17 @@ static bool process_control_request(uint8_t rhport, tusb_control_request_t const
...
-            tud_control_xfer(rhport, p_request, &alternate, 1);
-          }
-          break;
-
-          case TUSB_REQ_SET_INTERFACE:
-          {
-            uint8_t const alternate = (uint8_t) p_request->wValue;
-            (void) alternate;
+        // For STD GET_INTERFACE even if class driver doesn't support alternate setting
+        // It is still mandatory to response with value of zero
```
pigrew, line 579: Maybe improve the English here?

```
// For GET_INTERFACE, it is mandatory to respond even if the class
// driver doesn't use alternate settings.
```

src/class/net/net_device.c
hathach: No I think we should merge it, the dcd edpt close for all ports still has a long way to go. We also need an example to justify the change for edpt close as well. The currently is better than current master. There is many hacks already in place, I won't worry too much about those.

src/class/net/net_device.c
hathach: right, even me got mixed between those two. One is a check, one will print out error on log. Hmm, will try to remember that.

src/device/usbd.c
hathach: my English is getting worse by the time :disappointed:

src/class/net/net_device.c
```diff
@@ -226,33 +249,95 @@ static void ecm_report(bool nc)
...
+            {
+              TU_ASSERT(_netd_itf.ecm_desc_epdata);
+              TU_ASSERT( usbd_open_edpt_pair(rhport, _netd_itf.ecm_desc_epdata, 2, TUSB_XFER_BULK, &_netd_itf.ep_out, &_netd_itf.ep_in) );
+
+              // TODO should be merge with RNDIS's after endpoint opened
+              // Also should have opposite callback for application to disable network !!
+              tud_network_init_cb();
+              can_xmit = true; // we are ready to transmit a packet
+              tud_network_recv_renew(); // prepare for incoming packets
+            }
```
pigrew, line 293: Perhaps another hack could be to do a "stall" and then an "unstall" on the two bulk EP, in order to reset their DTOGs, in the case that the EP are already open. But, not necessary, since it's hacky anyway.

hathach: that is a good way, though let's just leave it this way, to make sure we got an actual error and have motivation to fix it :)

PR #347, opened 2020-04-15
nuc505: change dcd_set_config() behavior

majbthrd:
This addresses @hathach's request in PR #338.  It is technically not right, but it will cause the NUC505 to behave the same as LPC17xx would in PR #338.

### DIFF
--- a/src/portable/nuvoton/nuc505/dcd_nuc505.c
+++ b/src/portable/nuvoton/nuc505/dcd_nuc505.c
@@ -112,8 +112,6 @@ static struct
   uint16_t total_bytes;
 } ctrl_in_xfer;
 
-static volatile bool configuration_changed;
-
 static volatile struct xfer_ctl_t *current_dma_xfer;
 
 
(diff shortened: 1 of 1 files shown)


src/portable/nuvoton/nuc505/dcd_nuc505.c
```diff
@@ -579,15 +575,14 @@ void dcd_irq_handler(uint8_t rhport)
     else if (cep_state & USBD_CEPINTSTS_STSDONEIF_Msk)
     {
       /* given ACK from host has happened, we can now set the address (if not already done) */
-      if((USBD->FADDR != assigned_address) && (USBD->FADDR == 0)) USBD->FADDR = assigned_address;
-
-      if (configuration_changed)
+      if((USBD->FADDR != assigned_address) && (USBD->FADDR == 0))
```
hathach, line 578: how about using the optional `dcd_edpt0_status_complete()` check for the complete of set_address and change it there. It is function called at the end of every Control transfer. Don't worry if it slows down the mcu, control transfer only happens in a very brief time of enumeration and rarely in the whole device activities.

https://github.com/hathach/tinyusb/blob/master/src/portable/microchip/samd/dcd_samd.c#L180

PR #348, opened 2020-04-15
Set Net example to use IAD device class

pigrew:
Since the net example now has IAD descriptors, I believe that the device class should be set to be a IAD device.


pigrew:
That, I don't know.

Personally, I've been using Windows, and only had to manually reset drivers when I've used Zadig.

hathach:
> Thanks, @pigrew, for looking out for such issues. I checked operation with Linux, Win 7 & 10, and macOS.
> 
> Is it worthwhile or overkill to also revise .bcdDevice to 0x0102 to delineate the change? If someone has a formal guideline somewhere for what Windows presumes or tolerates when descriptors change, I'm not aware of it. I've just come to depend on usbdeview during development to erase existing information.

You will need bcd = 2.0 for the support of IAD (when it is introduced). I think best to keep it as 2.0.

PR #350, opened 2020-04-16
Synopsys msp430 dcd disconnect connect

hathach:
implement pullup enable/disable for synopsys and msp430. #327

### DIFF
--- a/src/portable/ti/msp430x5xx/dcd_msp430x5xx.c
+++ b/src/portable/ti/msp430x5xx/dcd_msp430x5xx.c
@@ -206,20 +206,24 @@ void dcd_remote_wakeup(uint8_t rhport)
 
 void dcd_connect(uint8_t rhport)
 {
-  (void) rhport;
+  dcd_int_disable(rhport);
 
   USBKEYPID = USBKEY;
   USBCNF |= PUR_EN; // Enable pullup.
   USBKEYPID = 0;
+
+  dcd_int_enable(rhport);
 }
 
 void dcd_disconnect(uint8_t rhport)
 {
-  (void) rhport;
+  dcd_int_disable(rhport);
 
   USBKEYPID = USBKEY;
   USBCNF &= ~PUR_EN; // Disable pullup.
   USBKEYPID = 0;
+
+  dcd_int_enable(rhport);
 }
 
 /*------------------------------------------------------------------*/
--- a/src/portable/ti/msp430x5xx/dcd_msp430x5xx.c
+++ b/src/portable/ti/msp430x5xx/dcd_msp430x5xx.c
@@ -134,9 +134,6 @@ void dcd_init (uint8_t rhport)
   // Enable reset and wait for it before continuing.
   USBIE |= RSTRIE;
 
-  // Enable pullup.
-  USBCNF |= PUR_EN;
-
   USBKEYPID = 0;
 }
 
@@ -205,6 +202,24 @@ void dcd_set_config (uint8_t rhport, uint8_t config_num)
 void dcd_remote_wakeup(uint8_t rhport)
 {
   (void) rhport;
+}
+
+void dcd_connect(uint8_t rhport)
+{
+  (void) rhport;
+
+  USBKEYPID = USBKEY;
+  USBCNF |= PUR_EN; // Enable pullup.
+  USBKEYPID = 0;
+}
+
+void dcd_disconnect(uint8_t rhport)
+{
+  (void) rhport;
+
+  USBKEYPID = USBKEY;
+  USBCNF &= ~PUR_EN; // Disable pullup.
+  USBKEYPID = 0;
 }
 
 /*------------------------------------------------------------------*/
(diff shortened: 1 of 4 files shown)


src/portable/ti/msp430x5xx/dcd_msp430x5xx.c
```diff
@@ -207,6 +204,24 @@ void dcd_remote_wakeup(uint8_t rhport)
   (void) rhport;
 }
 
+void dcd_connect(uint8_t rhport)
+{
+  (void) rhport;
+
+  USBKEYPID = USBKEY;
+  USBCNF |= PUR_EN; // Enable pullup.
+  USBKEYPID = 0;
```
pigrew, line 213: Is there a concurrency issue here?

USBKEYPID is also used in ```bus_reset```, which looks to be called from the ISR?

hathach: ah yeah, we better to be safe, just take a look at dcd_init_enable()/disable() look like there is quite a lot of stuff. We can simply call it, unless @cr1901 know a better way :)

cr1901:
>ah yeah, we better to be safe, just take a look at dcd_init_enable()/disable() look like there is quite a lot of stuff. We can simply call it, unless @cr1901 know a better way :)

There is so much code there so that other interrupts can continue to process data while the USB peripheral interrupts are disabled; this is the only way I could think of to disable _only_ the USB peripheral interrupt.


Re: USBKEYPID: Strictly speaking, every register write in every port should probably be done with interrupts disabled, unless we can prove that "a read/write to a register from the main thread won't be interrupted in the middle of said read/write". That might be a decent follow-up PR, but skimming I don't see any dangerous reads/writes that pop out.

However, since `USBKEYPID`'s side effects are longer than a single register write, and potentially _chunks_ of register reads/writes can be invalidated by the interrupt thread changing `USBKEYPID`, I agree that each use should be gated. _If you modify `USBKEYPID` from another interrupt besides the USB peripheral, you're on your own :)._

Maybe replace all accesses to `USBKEYPID` with something like this with an appropriate comment on the macro definitions?

```
#define USBKEY_UNLOCK() dcd_int_disable(); USBKEYPID = USBKEY
#define USBKEY_LOCK() USBKEYPID = 0; dcd_int_enable()
```

pigrew:
> There is so much code there so that other interrupts can continue to process data while the USB peripheral interrupts are disabled; this is the only way I could think of to disable _only_ the USB peripheral interrupt.

That's so much longer than I would have guessed. I don't understand how ```in_isr``` is set properly to true. ```in_isr``` seems to only be set when dcd_int_disable() is called.... I wonder if there is danger if the disable function is called from inside the USB ISR?

> Re: USBKEYPID: Strictly speaking, every register write in every port should probably be done with interrupts disabled, unless we can prove that "a read/write to a register from the main thread won't be interrupted in the middle of said read/write". That might be a decent follow-up PR, but skimming I don't see any dangerous reads/writes that pop out.

I think there is an issue with ```dcd_edpt_open```, which also uses USBKEYPID. Interrupts probably need to be disabled for it.

> However, since `USBKEYPID`'s side effects are longer than a single register write, and potentially _chunks_ of register reads/writes can be invalidated by the interrupt thread changing `USBKEYPID`, I agree that each use should be gated. _If you modify `USBKEYPID` from another interrupt besides the USB peripheral, you're on your own :)._

Yep, but we can declare that it'll never happen. :)

In stm32fsdev, I also assume that the ```dcd_*``` functions are only called from the USB task, and only with interrupts enabled. If they are disabled, then some counting needs to happen for nested calls to disable/enable interrupts (like an enterCritical and leaveCritical function pair).

> Maybe replace all accesses to `USBKEYPID` with something like this with an appropriate comment on the macro definitions?
> 
> ```
> #define USBKEY_UNLOCK() dcd_int_disable(); USBKEYPID = USBKEY
> #define USBKEY_LOCK() USBKEYPID = 0; dcd_int_enable()
> ```

Seems good, except perhaps the usage while in the USB ISR.

cr1901:
>That's so much longer than I would have guessed. I don't understand how in_isr is set properly to true. in_isr seems to only be set when dcd_int_disable() is called.... 

`in_isr` is/was a bad name. `in_isr` represents "were USB peripheral interrupts already disabled when this function was called?". Maybe `ints_already_disabled` is a better name.

>I wonder if there is danger if the disable function is called from inside the USB ISR?

The register contents restored when interrupts are re-enabled reflect their contents when interrupts were _most recently_ disabled. I think in that sense, we should be okay because interrupts will be reenabled before we leave the USB ISR. So register contents will be as up-to-date as possible. 

I think this also solves the hypothetical (?) issue of restoring registers to their most recent values in the "application disables interrupts, ISR disables interrupts but doesn't reenable them, application reenables interrupts" case.

>I think there is an issue with dcd_edpt_open, which also uses USBKEYPID. Interrupts probably need to be disabled for it.

Yes, my mistake. I meant to include that w/ USBKEYPID.

hathach:
thanks @cr1901 @pigrew for a very helpful discussion as usual. We can leave it as it is now when it become an issue or when we have more time to work with it in the future.

PR #351, opened 2020-04-16
Valentyusb dcd disconnect connect

hathach:
implement dcd disconnect and connect, part of #327

### DIFF
--- a/hw/bsp/fomu/dfu.py
+++ b/hw/bsp/fomu/dfu.py
@@ -8,7 +8,7 @@
 import sys,struct,zlib,os
 from optparse import OptionParser
 
-DEFAULT_DEVICE="0x0483:0xdf11"
+DEFAULT_DEVICE="0x1209:0x5bf0"
 
 def named(tuple,names):
   return dict(zip(names.split(),tuple))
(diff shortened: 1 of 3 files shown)


hw/bsp/fomu/dfu.py
```diff
@@ -0,0 +1,95 @@
...
+
+# Written by Antonio Galea - 2010/11/18
+# Updated for DFU 1.1 by Sean Cross - 2020/03/31
+# Distributed under Gnu LGPL 3.0
+# see http://www.gnu.org/licenses/lgpl-3.0.txt
+
+import sys,struct,zlib,os
+from optparse import OptionParser
+
+DEFAULT_DEVICE="0x0483:0xdf11"
```
xobs, line 11: This is the VID and PID for ArduPilot. Since this is in the `Fomu` directory, maybe this should be updated to `0x1209:0x5bf0`

xobs: Although leaving it as-is does give an idea of some of the lineage of this code!

hathach: updated, I just copy these from the circuitpython https://github.com/adafruit/circuitpython/blob/master/tools/dfu.py for an way to covert the bin to dfu file

PR #352, opened 2020-04-16
samg implement dcd connect/disconnect

hathach:
implement part of #327

### DIFF
--- a/src/portable/microchip/samg/dcd_samg.c
+++ b/src/portable/microchip/samg/dcd_samg.c
@@ -138,9 +138,6 @@ void dcd_init (uint8_t rhport)
   (void) rhport;
 
   tu_memclr(_dcd_xfer, sizeof(_dcd_xfer));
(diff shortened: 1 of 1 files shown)


pigrew:
@hathach 

Looks like this has the same concurrency issue.....   ```UDP_TXVC```  is also used by bus_reset().

hathach:
Don't worry for this port. There won't be a bus reset while we doing disconnect/connect. There is also no usbkey guard like the msp430 to worry about

pigrew:
Wow, that's tricky. Since the R-M-W cycle is in the ISR, I think it's safe.

hathach:
> Wow, that's tricky. Since the R-M-W cycle is in the ISR, I think it's safe.

Not what I mean though, I mean there shouldn't a bus reset occurs anywhere near the execution of these function. But yeah, that helps as well.

PR #355, opened 2020-04-17
follow up to #206

hathach:
- improve cdc write flush on completion
- add a bit of document for cdc device API

### DIFF
--- a/src/class/cdc/cdc_device.h
+++ b/src/class/cdc/cdc_device.h
@@ -78,7 +78,7 @@ int32_t  tud_cdc_n_read_char       (uint8_t itf);
 // Clear the received FIFO
 void     tud_cdc_n_read_flush      (uint8_t itf);
 
-// Get a byte from FIFO at the specified position without revmoing it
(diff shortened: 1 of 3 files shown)


src/class/cdc/cdc_device.h
```diff
@@ -48,28 +48,58 @@
...
 uint32_t tud_cdc_n_read            (uint8_t itf, void* buffer, uint32_t bufsize);
+
+// Read a byte, return -1 if there is none
+static inline
+int32_t  tud_cdc_n_read_char       (uint8_t itf);
+
+// Clear the received FIFO
 void     tud_cdc_n_read_flush      (uint8_t itf);
+
+// Get a byte from FIFO at the specified position without revmoing it
```
majbthrd, line 81: "removing" rather than "revmoing"

hathach: man, I can never escape from English typos :sweat_smile:

hathach:
> works for me!

Thanks for reviewing :+1:

PR #356, opened 2020-04-18
Use docker for esp32s2 ci

hathach:
docker build save 3-4 minutes comparing to normal build process. pretty good result

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -13,6 +13,7 @@ jobs:
     - name: failed
       run: |
         docker --version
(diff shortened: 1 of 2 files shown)


hathach:
@ladyada thanks :smiley:  we can use this for circuitpython later on. Espressif has CI to public their IDF master  to docker on every commit. Much faster than downloading and setup toolchain. IDF docker guide can be found here https://docs.espressif.com/projects/esp-idf/en/latest/esp32/api-guides/tools/idf-docker-image.html

ladyada:
@tannewt

PR #357, opened 2020-04-18
add disconnection detection for stm32 synopsys

### LINKED ISSUE #209: STM32 Synopsis doesn't seems to detect disconnection event
**Describe the bug**

**Set up (please complete the following information):**
 - OS: Ubuntu 18.04
 - Board: stm32f407disco, stm32f411disco, stm32f767nucleo
 - Firmware Code: examples/device/cdc_msc

**To Reproduce**
1. Run example without USB: led blink fast indicate no connection
2. Plug USB cable, device enumerate, also led blink slow
3. Unplug USB cable (keep stlink for power), led still blink slow --> no disconnect event detected

**Expected behavior**
LED should blink slow in the above example indicating there is no connections. maybe related to #179

hathach:
Fix #209
implement disconnection detection for stm32 synopsys
- disconnection is OTG INT session end bit
- add USE_SOF to disable 1ms interrupt on mcu which isn't used now by
the stack
- add suspend detection
- add resume detection

tested on stm32f411 disco board. However, esp32s2 saola doens't seem to detect disconnection with the same changes. Probably the vbus sensing is used for detecting the event, which saola is lacking (I have to jump wire the D+ D- to break out usb connector). We will revise this later with another S2 board later on. 

image

### DIFF
--- a/src/portable/espressif/esp32s2/dcd_esp32s2.c
+++ b/src/portable/espressif/esp32s2/dcd_esp32s2.c
@@ -179,7 +179,8 @@ void dcd_init(uint8_t rhport)
   USB0.gintmsk = 0;   //mask all
   USB0.gotgint = ~0U; //clear OTG ints
   USB0.gintsts = ~0U; //clear pending ints
-  USB0.gintmsk = USB_MODEMISMSK_M  |
+  USB0.gintmsk = USB_OTGINTMSK_M   |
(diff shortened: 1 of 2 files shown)


pigrew:
Is it permitted to call printf/ESP_EARLY_LOGV from an interrupt handler?

hathach:
thanks @cr1901 for reviewing.

Issue #358, opened 2020-04-19
MIDI: missing flow-control drops packets


kaysievers:
I'm running this patch to make the tinyusb MIDI code behave as expected:
  https://github.com/versioduo/Adafruit_TinyUSB_ArduinoCore/commit/4e8e4a3d6dcad2313169461bb4b60f8025c21c4e

The git commit of the patch message should explain what is going on. The current tinyusb  MIDI code does not work correctly for anything non-trivial, it just drops random packets. Large sysex transfers (16kb) drop almost half of the packets.

It is more a general USB problem, nothing really MIDI specific. The patch does not apply directly anymore though, there are too  many changes in the meantime. Could please let me know how we can get this fixed?

Should I try to forward-port it, or can you add the proper flow control to the core code?

cr1901:
>The patch does not apply directly anymore though, there are too many changes in the meantime.

What happens when you apply that patch as-is to the current HEAD? That was Feb 28? Most activity on tinyusb since then has been dedicated to:

* The ESP32 port
* Getting msp430 merged after it stagnated
* CDC-EEM/ECM support
  * Discussing multiple configuration support
  * Discussing alternate interface support
* Various bug fixes (incl ZLP handling)

Of these, only the bug fix code could really affect MIDI, and I don't think anything really fundamental changed about tinyusb's event-handling code in the past month that would make your patch obsolete.

hathach:
> I'm running this patch to make the tinyusb MIDI code behave as expected:
> [versioduo/Adafruit_TinyUSB_ArduinoCore@4e8e4a3](https://github.com/versioduo/Adafruit_TinyUSB_ArduinoCore/commit/4e8e4a3d6dcad2313169461bb4b60f8025c21c4e)
> 
> The git commit of the patch message should explain what is going on. The current tinyusb MIDI code does not work correctly for anything non-trivial, it just drops random packets. Large sysex transfers (16kb) drop almost half of the packets.
> 
> It is more a general USB problem, nothing really MIDI specific. The patch does not apply directly anymore though, there are too many changes in the meantime. Could please let me know how we can get this fixed?
> 
> Should I try to forward-port it, or can you add the proper flow control to the core code?

Yeah, It seems to take some inspiration from cdc. The config fifo for not allow overwriting would help. @kaysievers Can you tell us steps to reproduce this issue as detail as possible (I don't even know how to send sysEx message. ). I will try to analyze on my side.

kaysievers:
[...]

PR #359, opened 2020-04-19
MIDI - Add flow control to incoming packet stream

kaysievers:
Here is the current patch for https://github.com/hathach/tinyusb/issues/358
It is only compile-tested so far.

Larger SysEx transfers get corrupted by incoming packets.

This changes the FIFOs not to overwrite their data. MIDI should not be
a transport that drops packets. A potentially blocking device is easier
to detect and handle than a device that silently corrupts the packet
stream at random overflows, especially when SysEx messages are involved.

### DIFF
--- a/src/class/midi/midi_device.c
+++ b/src/class/midi/midi_device.c
@@ -85,6 +85,18 @@ bool tud_midi_n_mounted (uint8_t itf)
   return midi->ep_in && midi->ep_out;
 }
 
+static void _prep_out_transaction (midid_interface_t* p_midi)
+{
+  // skip if previous transfer not complete
+  if ( usbd_edpt_busy(TUD_OPT_RHPORT, p_midi->ep_out) )
+    return;
+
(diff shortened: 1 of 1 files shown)


src/class/midi/midi_device.c
```diff
@@ -267,8 +285,8 @@ void midid_init(void)
     midid_interface_t* midi = &_midid_itf[i];
 
     // config fifo
-    tu_fifo_config(&midi->rx_ff, midi->rx_ff_buf, CFG_TUD_MIDI_RX_BUFSIZE, 1, true);
-    tu_fifo_config(&midi->tx_ff, midi->tx_ff_buf, CFG_TUD_MIDI_TX_BUFSIZE, 1, true);
+    tu_fifo_config(&midi->rx_ff, midi->rx_ff_buf, CFG_TUD_MIDI_RX_BUFSIZE, 1, false);
```
hathach, line 288: I thought MIDI is music and it is better to drop data if we run out of space, but do that with sysex is a bad idea indeed.

PR #365, opened 2020-04-20
tu_verify for getting descriptors

pigrew:
TU_ASSERT was being used for getting config and string descriptors.

This could be triggered by the host (and doesn't mean that TinyUSB is buggy) AND represents an abnormal condition, so it should be TU_VERIFY.

(Found while examining code while thinking about #361).

### DIFF
--- a/src/device/usbd.c
+++ b/src/device/usbd.c
@@ -450,7 +450,7 @@ void tud_task (void)
 // Helper to invoke class driver control request handler
 static bool invoke_class_control(uint8_t rhport, uint8_t drvid, tusb_control_request_t const * request)
 {
-  TU_ASSERT(_usbd_driver[drvid].control_request);
+  TU_VERIFY(_usbd_driver[drvid].control_request);
 
   usbd_control_set_complete_callback(_usbd_driver[drvid].control_complete);
   TU_LOG2("  %s control request\r\n", _usbd_driver[drvid].name);
@@ -463,7 +463,7 @@ static bool process_control_request(uint8_t rhport, tusb_control_request_t const
 {
   usbd_control_set_complete_callback(NULL);
 
-  TU_ASSERT(p_request->bmRequestType_bit.type < TUSB_REQ_TYPE_INVALID);
+  TU_VERIFY(p_request->bmRequestType_bit.type < TUSB_REQ_TYPE_INVALID);
 
   // Vendor request
   if ( p_request->bmRequestType_bit.type == TUSB_REQ_TYPE_VENDOR )
@@ -801,7 +801,7 @@ static bool process_get_descriptor(uint8_t rhport, tusb_control_request_t const 
     case TUSB_DESC_CONFIGURATION:
     {
       tusb_desc_configuration_t const* desc_config = (tusb_desc_configuration_t const*) tud_descriptor_configuration_cb(desc_index);
-      TU_ASSERT(desc_config);
+      TU_VERIFY(desc_config);
 
       uint16_t total_len;
       memcpy(&total_len, &desc_config->wTotalLength, 2); // possibly mis-aligned memory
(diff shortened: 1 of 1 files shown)


src/device/usbd.c
```diff
@@ -463,7 +463,7 @@ static bool process_control_request(uint8_t rhport, tusb_control_request_t const
 {
   usbd_control_set_complete_callback(NULL);
 
-  TU_ASSERT(p_request->bmRequestType_bit.type < TUSB_REQ_TYPE_INVALID);
+  TU_VERIFY(p_request->bmRequestType_bit.type < TUSB_REQ_TYPE_INVALID);
```
hathach, line 466: this is illegal request, host won't ever send this. It normally signals an issue with dcd getting incorrect or corrupted setup packet.

src/device/usbd.c
```diff
@@ -450,7 +450,7 @@ void tud_task (void)
 // Helper to invoke class driver control request handler
 static bool invoke_class_control(uint8_t rhport, uint8_t drvid, tusb_control_request_t const * request)
 {
-  TU_ASSERT(_usbd_driver[drvid].control_request);
+  TU_VERIFY(_usbd_driver[drvid].control_request);
```
hathach, line 453: we could actually remove this, the control request is mandatory for all driver, drvid is already valid to this point.

src/device/usbd.c
```diff
@@ -801,7 +801,7 @@ static bool process_get_descriptor(uint8_t rhport, tusb_control_request_t const
     case TUSB_DESC_CONFIGURATION:
     {
       tusb_desc_configuration_t const* desc_config = (tusb_desc_configuration_t const*) tud_descriptor_configuration_cb(desc_index);
-      TU_ASSERT(desc_config);
+      TU_VERIFY(desc_config);
```
hathach, line 804: This should be assert to tell application in the log that they are returning NULL config descriptors. Again it is unlikely host request the out-of-bound config descriptor. It is likely the application logic issue.

src/device/usbd.c
```diff
@@ -817,10 +817,11 @@ static bool process_get_descriptor(uint8_t rhport, tusb_control_request_t const
         // The 0xEE index string is a Microsoft OS Descriptors.
         // https://docs.microsoft.com/en-us/windows-hardware/drivers/usbcon/microsoft-defined-usb-descriptors
         return false;
-      }else
+      }
+      else
       {
         uint8_t const* desc_str = (uint8_t const*) tud_descriptor_string_cb(desc_index, p_request->wIndex);
-        TU_ASSERT(desc_str);
+        TU_VERIFY(desc_str);
```
hathach, line 824: like above, host is well behaved, if it request out of bound descriptors, it is usually device  report it has that one by mistake. This help troubleshooting using assert to print out the error in the LOG

src/device/usbd.c
pigrew: I didn't realize it was mandatory. Removing the check seems reasonable, but maybe also adding a comment to state that the sof() function is optional in the struct definition.

Or maybe we can add a TU_ASSERT for all the function members in tud_init where they would be optimised out by the compiler...

hathach: @pigrew yeah, you are right, since the driver array is const, even this will surely be optimized by compiler. Maybe just leave it this for now. I am still not sure what to do with SOF. It is probably required for ISO transfer, but yes, it is optional for most class driver.

hathach: But again, I am not sure if this would change :D. Maybe there is a class that won't bother to do the control request.

pigrew: Currently the MIDI class doesn't bother, it has empty function definitions. But, the MIDI spec does define a pair of requests.

hathach: It is still better to have empty function than leaving it NULL I think. Yeah, let's TU_ASSERT() all of them (except for the SOF) when init the stack :+1: :+1:

pigrew:
@hathach,

I'll close this PR tomorrow (if the unneeded assert is actually optimised out).

My understanding was that TU_VERIFY should be used to detect misbehaving hosts (as we put in tusb_verify.h), so that's why I created this PR.... It sounds like the criteria to choose is a bit more nuanced than that.

hathach:
yeah, I would normally use VERIFY for normal logic check, which the failed is normal and happens from time to time. The ASSERT is on the other hand will print out to LOG=1, and or halt the MCU with breakpoint command if debugger is attached. This helps greatly in initial phase of developing the stack since there are too many thing got wrong, and the breakpoint help to understand the reason (variable.memory, register) at that exact moment.

pigrew:
@hathach 

Huh, it doesn't optimize out where it is right now.

However, it does optimize out if it's put in tud_init(), but if put there, probably there should be a check for all of the member functions existing.

So, my thought is to remove the check, and add a comment that all members are required, except `sof()`.


(making this PR pretty pointless, but I updated it anyway)

PR #372, opened 2020-04-22
Add rtt as logger option

### LINKED ISSUE #367: use SWO & Segger RTT for logging.
**Feature Description**
Current logging with uart is only viable within the main task. SWO (on supported board) and RTT is much faster and even can be used within ISR. Which is handy to debug/add new mcu. 

Note: both requires swd debugger. RTT is specific to jlink, swo is standard ARM but not all board/mcu support SWO.

- https://www.segger.com/products/debug-probes/j-link/technology/about-real-time-transfer/

hathach:
copy from readme 

> By default log message is printed via on-board UART which is slow and take lots of CPU time comparing to USB speed. If your board support on-board/external JLink debugger, it would be more efficient to use it with [RTT protocol](https://www.segger.com/products/debug-probes/j-link/technology/about-real-time-transfer/) for logging. To do that add option `LOGGER=rtt` to build command e.g

image

implement the RTT part of #367


examples/make.mk
```diff
@@ -75,7 +75,17 @@ else
...
 endif
 
-# TUSB Logging option
+# Log level is mapped to TUSB DEBUG option
 ifneq ($(LOG),)
   CFLAGS += -DCFG_TUSB_DEBUG=$(LOG)
 endif
+
+# Logger: default is UART, can be set to RTT
+ifeq ($(LOGGER),rtt)
```
henrygab, line 84: Existing boards and sketches may already be using/enabled RTT.

Given a sketch that is already configured to use RTT, if the BSP starts to use RTT from TinyUSB, what result?  It seems that there would need to be some way to ensure versions of RTT headers were synchronized between TinyUSB and the sketch (or BSP, as appropriate).

henrygab: actually, this shows the need to consider whether TinyUSB is the right location to enable RTT / SWO ... because neither of those requires USB at all.  Instead, it seems much cleaner to provide a separate extensible logging library, which allows simple compile-time selection of RTT, Serial, or SWO.  TinyUSB could use that library, for example.

pigrew: My feeling is that this PR is to enable RTT as a part of the provided makefiles/examples. There are no direct references to RTT or SWO within the TinyUsb library code (src directory). Presently all the logging is done through printf, which the goes through libc to the overrides in the BSP. Users of TinyUSB will have their own BSP. The BSP can use whichever RTT version they need. Maybe there could be a hook to use something other than printf, but I don't think that's an immediate concern.

henrygab: AWESOME!  I definitely misunderstood the context. That's the risk when I jump in without looking closely enough.

Given these are examples, then I have to say:

## Well done!

hathach: yup, tinyusb only call tu_printf and doesn't know about the retarget :)

PR #373, opened 2020-04-22
added swo as logger

hathach:
tested with feather nrf52840 + jlink. Implement swo part of #367 . Text from examples readme

````
By default log message is printed via on-board UART which is slow and take lots of CPU time comparing to USB speed. If your board support on-board/external debugger, it would be more efficient to use it for logging. There are 2 protocols: 

- `LOGGER=rtt`: use [Segger RTT protocol](https://www.segger.com/products/debug-probes/j-link/technology/about-real-time-transfer/)   
  - Cons: requires jlink as the debugger.
  - Pros: work with most if not all MCUs
  - Software viewer is JLink RTT Viewer/Client/Logger which is bundled with JLink driver package.
- `LOGGER=swo`: Use dedicated SWO pin of ARM Cortex SWD debug header.
  - Cons: only work with ARM Cortex MCUs minus M0
  - Pros: even faster than RTT, and should be compatible with hardware and software debugger that support SWO.
  - Software viewer is JLink SWO Viewer which is also bundled with JLink driver package.

$ make LOG=2 LOGGER=rtt BOARD=feather_nrf52840_express all
$ make LOG=2 LOGGER=swo BOARD=feather_nrf52840_express all
````

image

### DIFF
--- a/src/tusb_option.h
+++ b/src/tusb_option.h
@@ -75,17 +75,21 @@
 #define OPT_MCU_CXD56             400 ///< SONY CXD56
 
 // TI MSP430
-#define OPT_MCU_MSP430x5xx    500 ///< TI MSP430x5xx
+#define OPT_MCU_MSP430x5xx        500 ///< TI MSP430x5xx
 
+// ValentyUSB eptri
 #define OPT_MCU_VALENTYUSB_EPTRI  600 ///< Fomu eptri config
 
+// NXP iMX RT
 #define OPT_MCU_MIMXRT10XX        700 ///< NXP iMX RT10xx
 
+// Nuvoton
 #define OPT_MCU_NUC121            800
 #define OPT_MCU_NUC126            801
 #define OPT_MCU_NUC120            802
 #define OPT_MCU_NUC505            803
 
+// Espressif
 #define OPT_MCU_ESP32S2           900 ///< Espressif ESP32-S2
 
 /** @} */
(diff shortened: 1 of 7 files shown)

### CODE REVIEW

File: src/tusb_option.h [hunk 1]
```diff
@@ -75,17 +75,21 @@
 #define OPT_MCU_CXD56             400 ///< SONY CXD56
 
 // TI MSP430
-#define OPT_MCU_MSP430x5xx    500 ///< TI MSP430x5xx
+#define OPT_MCU_MSP430x5xx        500 ///< TI MSP430x5xx
```
henrygab, line 78: Just a thought ...

You might consider changing this to a union, unless there are benefits I am not aware of to leaving as `#define`.

e.g.:
```C
enum {
    OPT_MCU_NONE = 0,
    OPT_MCU_LPC11UXX = 1,
    /* etc. */
} /* OPT_MCU_Enum */ ;

Why?
* You get the same compile-time values, but you've now added type information -- so debuggers can print the friendly name when it's used.
* Values can still be used in C++11 (and later) `constexpr`, just like `#define`
* If you give the enum a name, function / template parameters can specify the type, improving usage and finding more errors at compile-time 

Obviously, this is optional, but does result in a "friendlier" library.

hathach: it is complicated, trust me I tried to use enum wherever I could. just leave that one for now.

PR #374, opened 2020-04-22
Added support for STM32F1 Connectivity Line MCU

duempel:
There is currently no support for STM32F105 and STM32F107 MCUs. These microcontrollers are connectivity line devices and are using the Synopsys IP instead of other STM32F1 devices.

I made some changes to get the F1 connectivity line into dcd. Successfully tested the cdc_msc example with my hardware. But I do not have an evaluation board to provide a full board example.

### DIFF
--- a/src/portable/st/synopsys/dcd_synopsys.c
+++ b/src/portable/st/synopsys/dcd_synopsys.c
@@ -27,6 +27,11 @@
 
 #include "tusb_option.h"
 
+#if defined (STM32F105x8) || defined (STM32F105xB) || defined (STM32F105xC) || \
+    defined (STM32F107xB) || defined (STM32F107xC)
+#define STM32F1_SYNOPSYS
+#endif
+
 #if defined (STM32L475xx) || defined (STM32L476xx) ||                          \
     defined (STM32L485xx) || defined (STM32L486xx) || defined (STM32L496xx) || \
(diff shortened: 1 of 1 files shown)


ladyada:
thanks :)

hathach:
there is F103 support, I thought F105/107 are the same as F103, no ?

duempel:
> I thought F105/107 are the same as F103, no ?

I also thought they were the same and tried to get it running with only the STM32F1 MCU option. Was a bad idea and took me some time to realize the real problem.

The reference manual actually includes two different types of USB peripherals, since it is valid for the whole F1 family. These are **Universal serial bus full-speed device interface (USB)** for STM32F103xx performance line and STM32F102xx USB access line families and **USB on-the-go full-speed (OTG_FS)** used by STM32F105xx and STM32F107xx connectivity line devices.

Here's a snip out of the reference manual showing STM32F105 and STM32F107 are using the OTG_FS (Synopsys) peripheral.

stm32f1_usb_otg

pigrew:
https://www.st.com/resource/en/application_note/dm00296349-usb-hardware-and-pcb-guidelines-using-stm32-mcus-stmicroelectronics.pdf

The ST USB hardware design guide is the best summary of the USB features of the various models.

Issue #289, opened 2020-03-03
check RNDIS with all ports


hathach:
**Describe the bug**
RNDIS example PR #287  may not work with all ports, following is port that work
- [x] SAMD
- [x] NUC126

Ports that tested and not work
- [ ] STM32F4 STM32F0
- [ ] nRF52840

**Set up (please complete the following information):**
 - OS: Ubuntu 18.04
 - Board: stm32f411disco, stm32f072, feather_nrf52840_express 
 - Firmware Code: examples/device/lwip_webserver

**To Reproduce**
flash and run the example

**Expected behavior**
failed to complete RNDIS message 

**Screenshots**
nrf52840 capture
![image](https://user-images.githubusercontent.com/249515/75797899-634b7580-5da8-11ea-8969-d00cdd5fb488.png)

majbthrd:
No, not exactly 128 bytes.  The intent was to transfer request->wLength, but no more than 128.  tu_control_xfer() calls tu_min() to cap all transfers to request->wLength.

pigrew:
@majbthrd : Based hathach's log, the SETUP packet has a wLength is 1025, so it wouldn't be truncated. My guess is that the host always sets wLength to 1025, so it'll always send 128 bytes (`tu_min(1025,128)`).

Regardless, we should figure out why the ZLP is not being sent.

majbthrd:
I've submitted a PR that should address the STM32 RNDIS issue, and I've regression tested it with SAMD21 and NUC126.  It may be worth re-trying on the other targets mentioned above.

hathach:
@pigrew @majbthrd great analysis, I think the host expect the usbd to return ZLP packet to indicate the end of transfer. I think this is a great chance for us to fix this ZLP behavior of usbd. It is not easy to get an reproducible issue for us to work on. This will prevent similar issue down the road.  @pigrew already did  ZLP response previously for MSP430 port, maybe it is merged there, I will try to look it up.

hathach:
This commin https://github.com/hathach/tinyusb/commit/c8247f0907040eb86c301e8bff8bb322fc9cb566 fix ZLP issue for nrf52840. For STM32F072, it is little bit difference, the device did response with ZLP correctly, however STALL all request afterwards.

![image](https://user-images.githubusercontent.com/249515/76159525-1d99fe80-6154-11ea-84ea-d79e4b17cb56.png)

hathach:
[...]

PR #380, opened 2020-04-26
synopsys turn off TX FIFO Empty for EPIN if all bytes are written

hathach:
**Describe the PR**
TXFE is currently cleared after all data is transferred to host. However host can be blocking wait for another EP (e.g control)  and doesn't issue IN token to receive/complete the transfer. This cause the TXFE constantly fired repeatedly and block/reduce processing other tasks. 

Fixed by turning of TXFE when the last data is written to FIFO (nothing else to write). Related to  #378 

This also fix the issue with usbnet mentioned in #289 

**Additional context**
image
screenshot when running lwip_webserver example with stm32f4 
- Host send RNDIS control message trigger Control DATA OUT 24 bytes (status phase not complete yet) 
- usbd forward data to the usbnet driver 
- usbnet parse rndis message and queue an 8 bytes transfer on notification endpoint 0x81
- TXFE occurred internally, stm32 write 8 bytes to FIFO of 0x81
- However since the control status is not yet complete, Host continue to send IN token to EP0, and **never** attempt to send IN token to 0x81
- the above TXFE keeps firing and block stm32 mcu to response to the control status response

Host waits for Control Status before sending IN to 0x81, but it is blocked by constatnt TXFE on 0x81 internally prevent stm32 to response to Control which is resulted to deadblock.

Update: also update esp32s2 as well, we should really merge esp32 and stm32 synopsys sometime in the future.

### DIFF
--- a/src/device/usbd_control.c
+++ b/src/device/usbd_control.c
@@ -82,7 +82,7 @@ bool tud_control_status(uint8_t rhport, tusb_control_request_t const * request)
   return _status_stage_xact(rhport, request);
 }
 
-// Queue an transaction in Data Stage
+// Queue a transaction in Data Stage
 // Each transaction has up to Endpoint0's max packet size.
 // This function can also transfer an zero-length packet
 static bool _data_stage_xact(uint8_t rhport)
--- a/src/device/usbd_control.c
+++ b/src/device/usbd_control.c
@@ -58,6 +58,7 @@ static uint8_t _usbd_ctrl_buf[CFG_TUD_ENDPOINT0_SIZE];
 // Application API
 //--------------------------------------------------------------------+
 
+// Queue ZLP status transaction
 static inline bool _status_stage_xact(uint8_t rhport, tusb_control_request_t const * request)
 {
   // Opposite to endpoint in Data Phase
@@ -81,7 +82,7 @@ bool tud_control_status(uint8_t rhport, tusb_control_request_t const * request)
   return _status_stage_xact(rhport, request);
 }
 
-// Transfer an transaction in Data Stage
+// Queue an transaction in Data Stage
 // Each transaction has up to Endpoint0's max packet size.
 // This function can also transfer an zero-length packet
 static bool _data_stage_xact(uint8_t rhport)
@@ -102,7 +103,6 @@ static bool _data_stage_xact(uint8_t rhport)
 }
 
 // Transmit data to/from the control endpoint.
-// 
 // If the request's wLength is zero, a status packet is sent instead.
 bool tud_control_xfer(uint8_t rhport, tusb_control_request_t const * request, void* buffer, uint16_t len)
 {
@@ -182,7 +182,7 @@ bool usbd_control_xfer_cb (uint8_t rhport, uint8_t ep_addr, xfer_result_t result
 
   // Data Stage is complete when all request's length are transferred or
   // a short packet is sent including zero-length packet.
-  if ( (_ctrl_xfer.request.wLength == _ctrl_xfer.total_xferred) || xferred_bytes < CFG_TUD_ENDPOINT0_SIZE )
+  if ( (_ctrl_xfer.request.wLength == _ctrl_xfer.total_xferred) || (xferred_bytes < CFG_TUD_ENDPOINT0_SIZE) )
   {
     // DATA stage is complete
     bool is_ok = true;
@@ -191,6 +191,11 @@ bool usbd_control_xfer_cb (uint8_t rhport, uint8_t ep_addr, xfer_result_t result
     // callback can still stall control in status phase e.g out data does not make sense
     if ( _ctrl_xfer.complete_cb )
     {
+      #if CFG_TUSB_DEBUG >= 2
+      extern void usbd_driver_print_control_complete_name(bool (*control_complete) (uint8_t, tusb_control_request_t const *));
+      usbd_driver_print_control_complete_name(_ctrl_xfer.complete_cb);
+      #endif
+
       is_ok = _ctrl_xfer.complete_cb(rhport, &_ctrl_xfer.request);
     }
 
--- a/src/portable/st/synopsys/dcd_synopsys.c
+++ b/src/portable/st/synopsys/dcd_synopsys.c
@@ -679,7 +679,8 @@ static void handle_epin_ints(USB_OTG_DeviceTypeDef * dev, USB_OTG_INEndpointType
       // XFER FIFO empty
       if ( in_ep[n].DIEPINT & USB_OTG_DIEPINT_TXFE )
       {
-        in_ep[n].DIEPINT = USB_OTG_DIEPINT_TXFE;
+        // DIEPINT's TXFE bit is read-only -> no need to clear
+
         transmit_packet(xfer, &in_ep[n], n);
 
         // Turn off TXFE if all bytes are written.
--- a/src/portable/st/synopsys/dcd_synopsys.c
+++ b/src/portable/st/synopsys/dcd_synopsys.c
@@ -291,9 +291,10 @@ bool dcd_edpt_open (uint8_t rhport, tusb_desc_endpoint_t const * desc_edpt)
 
   if(dir == TUSB_DIR_OUT)
   {
-    out_ep[epnum].DOEPCTL |= (1 << USB_OTG_DOEPCTL_USBAEP_Pos) | \
-      desc_edpt->bmAttributes.xfer << USB_OTG_DOEPCTL_EPTYP_Pos | \
-      desc_edpt->wMaxPacketSize.size << USB_OTG_DOEPCTL_MPSIZ_Pos;
+    out_ep[epnum].DOEPCTL |= (1 << USB_OTG_DOEPCTL_USBAEP_Pos) |
+                             (desc_edpt->bmAttributes.xfer << USB_OTG_DOEPCTL_EPTYP_Pos) |
+                             (desc_edpt->wMaxPacketSize.size << USB_OTG_DOEPCTL_MPSIZ_Pos);
+
     dev->DAINTMSK |= (1 << (USB_OTG_DAINTMSK_OEPM_Pos + epnum));
   }
   else
@@ -321,11 +322,12 @@ bool dcd_edpt_open (uint8_t rhport, tusb_desc_endpoint_t const * desc_edpt)
     // - Offset: GRXFSIZ + 16 + Size*(epnum-1)
     // - IN EP 1 gets FIFO 1, IN EP "n" gets FIFO "n".
 
-    in_ep[epnum].DIEPCTL |= (1 << USB_OTG_DIEPCTL_USBAEP_Pos) | \
-      epnum << USB_OTG_DIEPCTL_TXFNUM_Pos | \
-      desc_edpt->bmAttributes.xfer << USB_OTG_DIEPCTL_EPTYP_Pos | \
-      (desc_edpt->bmAttributes.xfer != TUSB_XFER_ISOCHRONOUS ? USB_OTG_DOEPCTL_SD0PID_SEVNFRM : 0) | \
-      desc_edpt->wMaxPacketSize.size << USB_OTG_DIEPCTL_MPSIZ_Pos;
+    in_ep[epnum].DIEPCTL |= (1 << USB_OTG_DIEPCTL_USBAEP_Pos) |
+                            (epnum << USB_OTG_DIEPCTL_TXFNUM_Pos) |
+                            (desc_edpt->bmAttributes.xfer << USB_OTG_DIEPCTL_EPTYP_Pos) |
+                            (desc_edpt->bmAttributes.xfer != TUSB_XFER_ISOCHRONOUS ? USB_OTG_DOEPCTL_SD0PID_SEVNFRM : 0) |
+                            (desc_edpt->wMaxPacketSize.size << USB_OTG_DIEPCTL_MPSIZ_Pos);
+
     dev->DAINTMSK |= (1 << (USB_OTG_DAINTMSK_IEPM_Pos + epnum));
 
     // Both TXFD and TXSA are in unit of 32-bit words.
@@ -352,9 +354,9 @@ bool dcd_edpt_xfer (uint8_t rhport, uint8_t ep_addr, uint8_t * buffer, uint16_t 
   uint8_t const dir   = tu_edpt_dir(ep_addr);
 
   xfer_ctl_t * xfer = XFER_CTL_BASE(epnum, dir);
-  xfer->buffer = buffer;
-  xfer->total_len = total_bytes;
-  xfer->queued_len = 0;
+  xfer->buffer       = buffer;
+  xfer->total_len    = total_bytes;
+  xfer->queued_len   = 0;
   xfer->short_packet = false;
 
   uint16_t num_packets = (total_bytes / xfer->max_size);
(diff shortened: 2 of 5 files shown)


src/portable/st/synopsys/dcd_synopsys.c
```diff
@@ -658,21 +663,30 @@ static void handle_epout_ints(USB_OTG_DeviceTypeDef * dev, USB_OTG_OUTEndpointTy
...
-      if(in_ep[n].DIEPINT & USB_OTG_DIEPINT_TXFE) {
+      if ( in_ep[n].DIEPINT & USB_OTG_DIEPINT_TXFE )
+      {
         in_ep[n].DIEPINT = USB_OTG_DIEPINT_TXFE;
         transmit_packet(xfer, &in_ep[n], n);
+
+        // Turn off TXFE if all bytes are written.
+        if (xfer->queued_len == xfer->total_len)
+        {
+          dev->DIEPEMPMSK &= ~(1 << n);
```
hathach, line 688: please skip other changes, those are clean up mostly. Only this change is important, moving clearing TXFE from complete to after no data left.

pigrew, line 688: Is there a concurrency problem here?

```DIEPEMPMSK``` is modified using read-modify-write on this line (while NOT in an interrupt).

It's also read-modify-written in ```handle_epin_ints``` (which I believe would be called inside the ISR).

**Comment by majbthrd** [comment]
> There's likely a concurrency issue in the updated xfer function.

Could there also be a concurrency issue in the preexisting code?  handle_epin_ints() was already doing another read-modify-write to DIEPEMPMSK, putting it in potential conflict with the one in dcd_edpt_xfer().

File: src/portable/st/synopsys/dcd_synopsys.c [hunk 2]
```diff
@@ -365,21 +367,23 @@ bool dcd_edpt_xfer (uint8_t rhport, uint8_t ep_addr, uint8_t * buffer, uint16_t
...
-    in_ep[epnum].DIEPTSIZ = (num_packets << USB_OTG_DIEPTSIZ_PKTCNT_Pos) | \
-        ((total_bytes & USB_OTG_DIEPTSIZ_XFRSIZ_Msk) << USB_OTG_DIEPTSIZ_XFRSIZ_Pos);
+    in_ep[epnum].DIEPTSIZ = (num_packets << USB_OTG_DIEPTSIZ_PKTCNT_Pos) |
+                            ((total_bytes & USB_OTG_DIEPTSIZ_XFRSIZ_Msk) << USB_OTG_DIEPTSIZ_XFRSIZ_Pos);
+
     in_ep[epnum].DIEPCTL |= USB_OTG_DIEPCTL_EPENA | USB_OTG_DIEPCTL_CNAK;
+
     // Enable fifo empty interrupt only if there are something to put in the fifo.
     if(total_bytes != 0) {
       dev->DIEPEMPMSK |= (1 << epnum);
```
pigrew, line 380: Also a (preexisting) race condition here.

src/portable/st/synopsys/dcd_synopsys.c
```diff
@@ -658,21 +663,30 @@ static void handle_epout_ints(USB_OTG_DeviceTypeDef * dev, USB_OTG_OUTEndpointTy
...
+      if ( in_ep[n].DIEPINT & USB_OTG_DIEPINT_XFRC )
+      {
         in_ep[n].DIEPINT = USB_OTG_DIEPINT_XFRC;
-        dev->DIEPEMPMSK &= ~(1 << n); // Turn off TXFE b/c xfer inactive.
         dcd_event_xfer_complete(0, n | TUSB_DIR_IN_MASK, xfer->total_len, XFER_RESULT_SUCCESS, true);
       }
 
       // XFER FIFO empty
-      if(in_ep[n].DIEPINT & USB_OTG_DIEPINT_TXFE) {
+      if ( in_ep[n].DIEPINT & USB_OTG_DIEPINT_TXFE )
```
pigrew, line 680: Also a likely race condition on these in_ep/out_ep usages, concurrent with edpt_stall() being called.

pigrew:
> > 
> 
> Could there also be a concurrency issue in the preexisting code? handle_epin_ints() was already doing another read-modify-write to DIEPEMPMSK, putting it in potential conflict with the one in dcd_edpt_xfer().

Yes, I agree (and made another comment). I'm not sure if @hathach or @cr1901 want to fix these issues in this PR, or create a second.

There appears to also be issues with ```in_ep[n]``` and ```out_ep[n]```, relating to an attempt to stall an endpoint while a transfer is completed.

I think that (USB) interrupts will need to be disabled in the xfer, stall, (open) and perhaps unstall  and close function calls. Reference counting (or the interrupt handler returning the old interrupt state) may need to be added for enabling/disabling interrupts (good idea for all the drivers, actually), for the case that the interrupts are disabled when the dcd calls are made.

cr1901:
@pigrew @majbthrd Let's keep this PR focused- make another issue to discuss potential race conditions.

@hathach The change looks fine to me. Is this specifically for ZLPs? I can't think of any other situation where the TX FIFO will be empty without the host having issued an IN packet first (in the Synopsys driver, TI driver at least, ZLPs have to be scheduled as a second xfer).

Looks like a genuine oversight on my part- I think I designed `transmit_packet` to be [a noop](https://github.com/hathach/tinyusb/blob/master/src/portable/st/synopsys/dcd_synopsys.c#L541-L544) if the length left to xfer was 0. Also, how does this PR interact w/ #378?

Are there cases where #378 won't fix this bug, and thus this PR is required?

pigrew:
@cr1202,

Fine by me to fix race conditions later. This PR was touching them, so I wanted to bring it up.

If there is an active IN transfer, and an EP is halted/stalled, does that clear the FIFO and trigger the EP empty interrupt? (I'm not sure if this case matters in this context, though)

cr1901:
Yes, you have to flush the FIFOs on STALL; don't remember if you have to do the same on disable, I do not remember offhand if/how the TXFE interrupt interacts with STALL (and it's not convenient to check right now).

hathach:
> @pigrew @majbthrd Let's keep this PR focused- make another issue to discuss potential race conditions.
> 
> @hathach The change looks fine to me. Is this specifically for ZLPs? I can't think of any other situation where the TX FIFO will be empty without the host having issued an IN packet first (in the Synopsys driver, TI driver at least, ZLPs have to be scheduled as a second xfer).

@cr1901 It is not specifically to ZLP, the deadblock is describeb as 1st post in screeshot, maybe I didn't make it clear enough. I updated it. please check it out again. To sum up, usbnet driver response to Control DATA OUT is to call edpt_xfer on notification enpoint 0x81 with 8 bytes. But since control is not complete yet, host will try to send IN to EP0 and won't send IN to EP 0x81. 

> Looks like a genuine oversight on my part- I think I designed `transmit_packet` to be [a noop](https://github.com/hathach/tinyusb/blob/master/src/portable/st/synopsys/dcd_synopsys.c#L541-L544) if the length left to xfer was 0. Also, how does this PR interact w/ #378?
> 
> Are there cases where #378 won't fix this bug, and thus this PR is required?

#378 give me an hint to troubleshoot this, I found this issue when troubleshooting the lwip_webserver issue running on stm32f4. 378 only fix ZLP, this one fix the other case when data is all send out while host does not send IN token to receive it.

src/portable/st/synopsys/dcd_synopsys.c
hathach: Thanks @pigrew for an eagle eye on this matter, since there are probably more registers like this, and I probably need to re-read the specs to do this properly. I have opened an separated issue for this https://github.com/hathach/tinyusb/issues/381 as reminder.

I like to get this PR merged first to fix issue with usbnet first. And it is actually better to have PR focus on fixing one bug at a time.

cr1901:
@hathach I'm still confused I'm afraid. Why is TXFE being sent to EP 0x81 repeatedly? The FIFO is still full of 8 bytes that have yet to be sent- that doesn't sound like an empty FIFO condition to me.

hathach:
> @hathach I'm still confused I'm afraid. Why is TXFE being sent to EP 0x81 repeatedly? The FIFO is still full of 8 bytes that have yet to be sent- that doesn't sound like an empty FIFO condition to me.

Yeah, but somehow it is triggered continuously. Now you mentioned it, I wonder why as well it occurs as well.

I have an RTT debug log which is handy for printf in isr.  line 385 is printf just before setting `dev->DIEPEMPMSK |= (1 << epnum)` of 0x81 in edpt xfer. and it blocks there almost forever. 

```
USBD Xfer Complete on EP 00 with 24 bytes
  NET control complete
  000:  02 00 00 00 18 00 00 00 01 00 00 00 01 00 00 00 | ................
  010:  00 00 00 00 40 06 00 00                         | ....@...
  Queue EP 81 with 8 bytes ... dcd_edpt_xfer: 370:
dcd_edpt_xfer: 385:

USB_OTG_FS->GINTMSK = 800C3816
USB_OTG_FS->GINTSTS = 4848428
dev->DAINT = 2
in_ep[n].DIEPINT = 80
USB_OTG_FS->GINTMSK = 800C3816
USB_OTG_FS->GINTSTS = 4848428
dev->DAINT = 2
in_ep[n].DIEPINT = 80
USB_OTG_FS->GINTMSK = 800C3816
USB_OTG_FS->GINTSTS = 4848428
dev->DAINT = 2
in_ep[n].DIEPINT = 80
```

hathach:
Looking deeper, it is configurable half/empty fifo. However we already configure 
the GAHBCFG as empty https://github.com/hathach/tinyusb/blob/master/src/portable/st/synopsys/dcd_synopsys.c#L198 

BTW, I just realized DIEPINT is readonly, which means we cannot clear it, and it will keep firing until HW think it is enough, which could be the case here, maybe half of FIFO is what it consider enough !!!

image

image

hathach:
@cr1901 I think it is due to the way we are currently allocate fifo
https://github.com/hathach/tinyusb/blob/master/src/portable/st/synopsys/dcd_synopsys.c#L334

we equally divided fifo for all EPIN without taking EPSize into account, since the average divided FIFO is more than 64 bytes which is simple and sufficient enough without DMA. I think hw need at least half of fifo size written to actually clear the TXFE bit. Since it is read-only from sw side, we couldn't truly clear it causing this huge confusion so far :D .

cr1901:
>BTW, I just realized DIEPINT is readonly, which means we cannot clear it.

Oh damn... that oversight has been around [since the beginning](https://github.com/hathach/tinyusb/pull/38/files#diff-a8b1af2d37038f5533918f40b5a5df97R607) ._.

>and it will keep firing until HW think it is enough, which could be the case here, maybe half of FIFO is what it consider enough !!!

Is there any easy way to test this behavior on your end,, by e.g. forcing the FIFO to fill up to over half the size when `condition_is_met`, and then removing that code and seeing how the `TXFE` bit reacts?

hathach:
> Oh damn... that oversight has been around [since the beginning](https://github.com/hathach/tinyusb/pull/38/files#diff-a8b1af2d37038f5533918f40b5a5df97R607) ._.

Yeah, that is thing I made all the time. Naturally we write to clear the bit in INTSTS
 
> > and it will keep firing until HW think it is enough, which could be the case here, maybe half of FIFO is what it consider enough !!!
> 
> Is there any easy way to test this behavior on your end,, by e.g. forcing the FIFO to fill up to over half the size when `condition_is_met`, and then removing that code and seeing how the `TXFE` bit reacts?

Yes, you just simply compile current master with net_lwip_webserver example using LOGGER=RTT (since uart won't work well within ISR). Which means you need to convert on-board stlink to jlink. But that is is rather simple and straightfoward https://www.segger.com/products/debug-probes/j-link/models/other-j-links/st-link-on-board/

Here is the file that I added the log previsouly, you can just replace the one in master. (github only allow to attach .txt). `_debug_en=1` when using notification 0x81 only to prevent unnecessary printf.
[dcd_synopsys.c.txt](https://github.com/hathach/tinyusb/files/4537651/dcd_synopsys.c.txt)

```
make BOARD=stm32f407disco LOG=2 LOGGER=rtt clean all flash-jlink
```
**flash-jlink** since we already converted it. Then using the JLink RTT VIewer (bundled with jlink driver) to connect and view the rtt log. 

image

cr1901:
Ahhh I see, I'm not really in a position to test right now :P.

I was asking if you had an example b/c I was wondering if you had gone as far to confirm the TXFE behavior, or if it was just a theory for now :).

Anyways, I'll approve, but I would appreciate a comment clarifying that "TXFE doesn't reset itself until the FIFO is over half full" in the source, for future reference.

hathach:
> Ahhh I see, I'm not really in a position to test right now :P.
> 
> I was asking if you had an example b/c I was wondering if you had gone as far to confirm the TXFE behavior, or if it was just a theory for now :).
> 
> Anyways, I'll approve, but I would appreciate a comment clarifying that "TXFE doesn't reset itself until the FIFO is over half full" in the source, for future reference.

haha, I am too lazy to write code based on theory, I am even too lazy to fix a more obvious bug unless there is a example that show the bug is real :D. Though I am try to increase more bytes than 8 to see if how many bytes to clear the event.

hathach:
@cr1901 I have just done some test. 
- current equally allocated fifo size is 84 words
- anything less than 64 bytes won't be enough to clear the TXFE event, therefore we will like to hit this issue with short package as well if host does not send IN token

I did a quick hack to change fifo_size of 0x81 and this is result
- fifo_size = 4 words ( 16 bytes) --> write 8 bytes does not enough to clear TXFE
- fifo_size =3 or 2 words ---> write 8 bytes clears the TXFE right away

Therefore the sumup would be, even though the TXFE is configured to trigger when it is completely empty. Hardware will only clear it if **MORE** than half of fifo is written or number of bytes is at least 64 ( 64 is still less than half of 84 words, but I guess there is a special if for that within hardware), I will put a bit more comment to elaborate this. However moving to clear this when there is nothing to write is also a good approach without any cons, since we have nothing left, there is no need to wait for TXFE anymore.

Thanks for commenting, this helps me to understand how hw functioning now with TXFE :)

cr1901:
@hathach Works for me. Thank you for testing :)! What an... involved... fix :P!

hathach:
> @hathach Works for me. Thank you for testing :)! What an... involved... fix :P!

Added note and push, great discussion so far on the TXFE :). Maybe specs mentioned it somewhere, but it is great finding.

duempel:
Since I'm new here and not very experienced with TinyUSB yet, I wonder if there is any reason why we set TXFE interrupt to trigger at **completely empty** (TXFELVL in GAHBCFG) instead of **half empty**? It should not change the way the interrupt handler is called, since before this PR it fired as long as dedicated Tx-FIFO is less (or equal) than half full (very great testing @hathach . This is a important information I can not find anywhere else! 👍  ) but should reduce the time _transfer_packet()_ starts to push data at the buffer again. Maybe we would need to read the DTXFSTS register in _transfer_packet()_ to check the free space available at FIFO.
I did not check this kind of implementation yet. Just wanted to think about if there are any cons for this.

src/device/usbd_control.c
```diff
@@ -81,7 +82,7 @@ bool tud_control_status(uint8_t rhport, tusb_control_request_t const * request)
   return _status_stage_xact(rhport, request);
 }
 
-// Transfer an transaction in Data Stage
+// Queue an transaction in Data Stage
```
pigrew, line 85: s/an/a/

hathach: damn, why English uses different words for the same thing !!! :sweat:

hathach:
> Since I'm new here and not very experienced with TinyUSB yet, I wonder if there is any reason why we set TXFE interrupt to trigger at **completely empty** (TXFELVL in GAHBCFG) instead of **half empty**? It should not change the way the interrupt handler is called, since before this PR it fired as long as dedicated Tx-FIFO is less (or equal) than half full (very great testing @hathach . This is a important information I can not find anywhere else!  ) but should reduce the time _transfer_packet()_ starts to push data at the buffer again. Maybe we would need to read the DTXFSTS register in _transfer_packet()_ to check the free space available at FIFO.
> I did not check this kind of implementation yet. Just wanted to think about if there are any cons for this.

Yes, I see your point, the fact that HW doesn't clear the TXFE until half of allocated FIFO is occupied is new finding ( I am not sure that if we could ever confirm this with actual specs). Before that, it does not make any differences either. Moving from empty to half empty make more sense, though it is not urgent and may need a bit of testing. Doing initial dcd port is difficult, the stm32 document is not really well written and organized (at least to me), so it does help picking the fastest config to implement. Currently we equally allocated dedicated FIFO for all IN endpoints, we can actually write more than a packet to FIFO. There is room for improving. But I guess it has to be its own PR with a decent testing. Thanks for your feedback.

cr1901:
>I wonder if there is any reason why we set TXFE interrupt to trigger at completely empty (TXFELVL in GAHBCFG) instead of half empty?

I think my thought process at the time was "half empty can fire while the USB core is in the middle of xferring data", whereas an empty FIFO definitely signals the end of a packet being sent (IIRC it's on the programmer to ensure that the FIFO's last entry before an xfer begins is a packet boundary).

But I don't remember offhand if the half-empty interrupt fires _after_ a full packet is sent, or if it fires while the FIFO is being drained.

duempel:
@hathach @cr1901 thanks for your feedback. I've done a new PR #387 to discuss about further improvements to the transmission.

PR #385, opened 2020-04-29
support class drivers implemented by application

### LINKED ISSUE #384: Allow overriding class drivers table
At the moment, is is only possible to define one custom device class driver through custom_device.h/.c. When tinyusb is used as part of a larger application or framework, it becomes inconvenient since the specific driver to be used might only be known at run time, not at compile time.

It would be great if `static usbd_class_driver_t const usbd_class_drivers[]` would be made overridable by the application, without having to modify TinyUSB source code.

There are a few ways to achieve this, here is a relatively simple modification which should not affect the existing apps. It involves 3 changes:

1. Define each standard class driver structure as a separate const variable, expose this variable in the public header file.
2. Compose the default array of class drivers from the pointers to the class driver structures.
3. Allow the array of class drivers to be overridden.

Here is how it can look like:

```c++
  /* Define each class driver as a separate structure */
  #if CFG_TUD_CDC
    usbd_class_driver_t const tusb_cdc_class_driver = 
    {
        .class_code      = TUSB_CLASS_CDC,
        .init            = cdcd_init,
        .open            = cdcd_open,
        .control_request = cdcd_control_request,
        .control_request_complete = cdcd_control_request_complete,
        .xfer_cb         = cdcd_xfer_cb,
        .sof             = NULL,
        .reset           = cdcd_reset
    },
  #endif

  /* Same for the rest of the default class drivers */

  /* The default drivers table. Note that these are pointers to usbd_class_driver_t */
  static usbd_class_driver_t const* usbd_class_drivers_default[] = {
  #if CFG_TUD_CDC
    tusb_cdc_class_driver,
  #endif
  /* The rest of the default class drivers */
  };

  const usbd_class_driver_t **usbd_class_drivers = usbd_class_drivers_default;
  size_t usbd_class_drivers_count = sizeof(usbd_class_drivers_default)/sizeof(usbd_class_driver_t*);
```

In some public header file, expose the class driver structures, and the `usbd_class_drivers` variable:

```c++
    usbd_class_driver_t const tusb_cdc_class_driver;
    /* the rest of the class driver structures */

   const usbd_class_driver_t **usbd_class_drivers usbd_class_drivers;
   size_t usbd_class_drivers_count;
```

With this, the application can define a custom list of class drivers:

```c++
const usbd_class_driver_t** my_class_drivers = {
  tusb_cdc_class_driver,
  my_class_driver_1,
  my_class_driver_2
};

usbd_class_drivers = my_class_drivers;
usbd_class_drivers_count = sizeof(my_class_drivers) / sizeof(my_class_drivers[0]);
```

Alternatively, instead of exposing the array + size, it can be required to make the array terminated by a NULL pointer.

Instead of exposing global `usbd_class_drivers` and `usbd_class_drivers_count`, it is possible to expose a setter function.

hathach:
**Describe the PR**
fix #384 
Allow application to extend class driver that is not currently support by the stack. This is a draft PR, most of work is complete, though need feedback and a bit of polish for naming.

@pigrew @igrr


src/device/usbd_pvt.h
```diff
@@ -33,6 +33,29 @@
...
+  bool (* open             ) (uint8_t rhport, tusb_desc_interface_t const * desc_intf, uint16_t* p_length);
+  bool (* control_request  ) (uint8_t rhport, tusb_control_request_t const * request);
+  bool (* control_complete ) (uint8_t rhport, tusb_control_request_t const * request);
+  bool (* xfer_cb          ) (uint8_t rhport, uint8_t ep_addr, xfer_result_t event, uint32_t xferred_bytes);
+  void (* sof              ) (uint8_t rhport); /* optional */
+} usbd_class_driver_t;
+
+// Invoked when initializing device stack to get additional class drivers.
+// Can optionally implemented by application to extend/overwrite class driver support.
+usbd_class_driver_t const* usbd_app_driver_get_cb(uint8_t* driver_count) TU_ATTR_WEAK;
```
hathach, line 57: @igrr callback implemented by application to add more drivers. Callback remove the need for application to know when to call driver register.

src/device/usbd.c
```diff
@@ -203,6 +189,29 @@ static usbd_class_driver_t const _usbd_driver[] =
 
 enum { USBD_CLASS_DRIVER_COUNT = TU_ARRAY_SIZE(_usbd_driver) };
 
+// Additional class drivers implemented by application
+static usbd_class_driver_t const * _app_driver = NULL;
+static uint8_t _app_driver_count = 0;
+
+// virtually joins built-in and application drivers together
+//   All drivers = built-in + application
+static inline usbd_class_driver_t const * get_driver(uint8_t drvid)
```
hathach, line 198: join 2 drivers

src/device/usbd.c
```diff
@@ -203,6 +189,29 @@ static usbd_class_driver_t const _usbd_driver[] =
...
+{
+  // Built-in drivers
+  if (drvid < USBD_CLASS_DRIVER_COUNT) return &_usbd_driver[drvid];
+
+  // App drivers
+  if ( usbd_app_driver_get_cb )
```
hathach, line 204: when not implemented, this will be easily optimized by compiler when linking.

src/device/usbd.c
```diff
@@ -309,11 +319,18 @@ bool tud_init (void)
   _usbd_q = osal_queue_create(&_usbd_qdef);
   TU_ASSERT(_usbd_q != NULL);
 
+  // Get application driver if available
+  if ( usbd_app_driver_get_cb )
+  {
+    _app_driver = usbd_app_driver_get_cb(&_app_driver_count);
```
hathach, line 325: app driver callback when init.

src/device/usbd.c
```diff
@@ -708,17 +722,17 @@ static bool process_set_config(uint8_t rhport, uint8_t cfg_num)
     uint8_t drv_id;
     uint16_t drv_len;
 
-    for (drv_id = 0; drv_id < USBD_CLASS_DRIVER_COUNT; drv_id++)
+    for (drv_id = 0; drv_id < TOTAL_DRIVER_COUNT; drv_id++)
```
hathach, line 725: reverse order to run from TOTAL_DRIVER_COUNT-1 to 0 will effectively allow application to overwrite built-in class driver. However, I cannot think of a good reason why application would do such a thing. Like providing its own msc/cdc instead of bultin one. If there is a bug within stack driver, it should be fixed rather. So we could just leave it as it is here until there is a scenario to justify the change.

igrr: I think an alternative when we need to override a driver is to exclude one of tinyusb driver files during the build process, and provide a custom driver implementation with the same named functions. So I don't mind if this stays as is.

src/device/usbd.c
igrr: Does this callback apply both to host and device drivers? (If not, suggest indicating which ones in the name)

hathach: yeah, host if supported will have its own callbacks :). Host and device stack are independent to each other. Only sharing common usb definitions and types

src/device/usbd_pvt.h
```diff
@@ -33,6 +33,29 @@
...
+  void (* reset            ) (uint8_t rhport);
+  bool (* open             ) (uint8_t rhport, tusb_desc_interface_t const * desc_intf, uint16_t* p_length);
+  bool (* control_request  ) (uint8_t rhport, tusb_control_request_t const * request);
+  bool (* control_complete ) (uint8_t rhport, tusb_control_request_t const * request);
+  bool (* xfer_cb          ) (uint8_t rhport, uint8_t ep_addr, xfer_result_t event, uint32_t xferred_bytes);
+  void (* sof              ) (uint8_t rhport); /* optional */
+} usbd_class_driver_t;
+
+// Invoked when initializing device stack to get additional class drivers.
+// Can optionally implemented by application to extend/overwrite class driver support.
```
pigrew, line 56: I think we need a comment that the array must remain in RAM indefinitely?

hathach: > Everything looks OK (except for adding a bit more documentation).

Yeah, I am always behind in documentation. I will try to do a better job with docs.
 
> I'd still lean towards only maintaining the driver list in the app (instead of dynamically modifying it with the config header file):

I want application to only focus on its extended driver, and not have to construct other class drivers, and I don't like to expose those built-in classes either. App could miss-calculated the driver struct and all the good old built-in stop working, nah not a chance. 
 
> Without the static list in usbd, the usbd.c code would be slightly shorter, and it also allows the linker to optimize out drivers not explicitly listed by the app without the need to modify the tusb_config header (though maybe that's a bad idea, since all the class drivers would need to be built even if only one was used).
> 
> On the other hand, it would break existing code.

It is not that much, especially when the callback is not implemented, the code is still pretty much the same after optimization. I would care less about the code size. One unique array and 2 virtual ones are the same to me. 

Yeah, the app driver should be long-lived, most of the time it is on ROM, I will add it on the next push

HiFiPhile:
Hi, have you made any progress ?

hathach:
> Hi, have you made any progress ?

I am too busy with other works for now. Will look at this again in a couple of weeks.

me-no-dev:
Re: Not loading all compiled drivers if they are not needed:

One downside that I found in my particular use case (ESP32-S2 Arduino with all drivers precompiled) is that even if a driver is not to be used (descriptor is not loaded), all required callbacks for that driver need to exist at link time. HID and MSC being tested offenders, HCI and NET not tested, but if they have callbacks that are not weak, case will be the same.

src/device/usbd.c
```diff
@@ -217,7 +203,30 @@ static usbd_class_driver_t const _usbd_driver[] =
...
+// Additional class drivers implemented by application
+static usbd_class_driver_t const * _app_driver = NULL;
+static uint8_t _app_driver_count = 0;
+
+// virtually joins built-in and application drivers together.
+// Application is positioned first to allow overwriting built-in ones.
+static inline usbd_class_driver_t const * get_driver(uint8_t drvid)
+{
+  // Application drivers
+  if ( usbd_app_driver_get_cb )
```
hathach, line 217: Application is positioned first before built-in this allows overwriting built-in driver.

hathach:
Application driver open() is invoked first before built-in one to allow overwriting behavior. 

@me-no-dev @igrr @HiFiPhile the PR is ready for review, it may take a couple more PRs to complete, but let's me know what you think.

src/device/usbd.c
HiFiPhile, line 217: I prefer to test `_app_driver_count`, since only the linker knows the existence of `usbd_app_driver_get_cb`, this `if{}` won't be optimized out if `usbd_app_driver_get_cb` not exist.

src/device/usbd_pvt.h
```diff
@@ -33,6 +33,30 @@
...
+  bool     (* control_request  ) (uint8_t rhport, tusb_control_request_t const * request);
+  bool     (* control_complete ) (uint8_t rhport, tusb_control_request_t const * request);
+  bool     (* xfer_cb          ) (uint8_t rhport, uint8_t ep_addr, xfer_result_t event, uint32_t xferred_bytes);
+  void     (* sof              ) (uint8_t rhport); /* optional */
+} usbd_class_driver_t;
+
+// Invoked when initializing device stack to get additional class drivers.
+// Can optionally implemented by application to extend/overwrite class driver support.
+// Note: The drivers array must be accessible at all time when stack is active
+usbd_class_driver_t const* usbd_app_driver_get_cb(uint8_t* driver_count) TU_ATTR_WEAK;
```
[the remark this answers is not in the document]
hathach: that would be `uint8_t usbd_app_driver_get_cb(usbd_class_driver_t const** app_driver)`, which I find pointer to pointer is a bit annoying.

src/device/usbd.c
hathach: `usbd_app_driver_get_cb` allow LTO to optimize out the code, though it is much the same with _app_driver_count, gcc is smart enough to figure it out with either.

HiFiPhile: > gcc is smart enough to figure it out with either.

Seems gcc is smarter than IAR :)

src/device/usbd_pvt.h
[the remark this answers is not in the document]
HiFiPhile: app_driver itself is an array, you don't need pointer to pointer of an array.

src/device/usbd.c
hathach: can you try to enable LTO option on IAR, I think IAR can only be smarter than gcc, they charge premium for that

src/device/usbd_pvt.h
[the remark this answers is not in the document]
hathach: Yeah, but you still need the pointer-to-pointer in the signature, I just won't tell you why :smiling_imp:

src/device/usbd.c
HiFiPhile: > I think IAR can only be smarter than gcc

I've already enabled every options. Frankly their optimization is not so impressive, many times with `const` `restrict` keywords it re-read everything from the stack. 

Their advantages are functional safety, MISRA and easy to use, for example to write a ram routine instead of fiddle with linker options and copy the routine with `memcpy`, I can use simply the keyword `__ramfunc`.

hathach: yeah, MISRA is a super nice thing to have.

hathach:
thank you everyone for participating to the PR

Issue #126, opened 2019-09-10
Port STM32 L4


hathach:
**Description**
Port STM32 L4

hathach:
tried but couldn't get stm32L476 discovery board running with USB in crystal less mode. Further study is needed.

pigrew:
I see a note in the manual about having to enable power gating for USB? Post your code (though I don't have a L476...)?

> After reset, the USB features supplied by VDDUSB are logically and electrically isolated and
therefore are not available. The isolation must be removed before using the USB OTG
peripheral, by setting the USV bit in the PWR_CR2 register, once the VDDUSB supply is
present.

hathach:
> I see a note in the manual about having to enable power gating for USB? Post your code (though I don't have a L476...)?
> 
> > After reset, the USB features supplied by VDDUSB are logically and electrically isolated and
> > therefore are not available. The isolation must be removed before using the USB OTG
> > peripheral, by setting the USV bit in the PWR_CR2 register, once the VDDUSB supply is
> > present.

Thanks for the note, I have struggled with L4 for sometime. Got to move on with other issues (too many already :D) . Will come back to test this later on :)

hathach:
maybe fixed by #391

kasjer:
> maybe fixed by #391

Unfortunately it will not help.

Here is what helped me to get started with enumeration:
```c
    USB_OTG_FS->GCCFG |= USB_OTG_GCCFG_PWRDWN;

    USB_OTG_FS->GUSBCFG &= ~USB_OTG_GUSBCFG_FHMOD;
    USB_OTG_FS->GUSBCFG |= USB_OTG_GUSBCFG_FDMOD;

    USB_OTG_FS->GOTGCTL |= USB_OTG_GOTGCTL_BVALOEN;
    USB_OTG_FS->GOTGCTL |= USB_OTG_GOTGCTL_BVALOVAL;
```

But that is not enough synopsys differs more from what is in F4 chips.
Funny thing is that with just this enumeration succeeded for cdc_msc but only msc worked with some huge delay.
While CDC did not work.
It is caused (maybe among other reasons) by EP0 transfer end interrupt triggered at the end of setup packed, before setup interrupt shows up. With this extra interrupt synopsys driver tries to handle end of transfer interrupt when it did not schedule any transfer yet using transfer data from previous transfer.

hathach:
[...]

PR #391, opened 2020-05-06
stm32l476disco clock configuration

kasjer:
**Describe the PR**
There were two problems with **STM32L476** System clock setup.
One is that uninitialized variable could lead to reboot loop due to possible accidental assert.

Second problem resulted in System clock being driven by MSI **48MHz** clock and NOT what code
suggested **80MHz**, due to PLL not being initialized.
PLL was not initialized by accident since Tick interrupt priority was never specified and default
ST value was intentionally invalid:
```c
uint32_t uwTickPrio   = (1UL << __NVIC_PRIO_BITS); /* Invalid PRIO */
```

**Additional context**
Code also initializes PC11 pin as VBUS detection pin which is suggested by board documentation while is seems that it's not needed since like other ST chips PA9 is used for this and not PC11.

### DIFF
--- a/hw/bsp/stm32l476disco/stm32l476disco.c
+++ b/hw/bsp/stm32l476disco/stm32l476disco.c
@@ -95,6 +95,10 @@ static void SystemClock_Config(void)
   /* Enable the CSS interrupt in case LSE signal is corrupted or not present */
   HAL_RCCEx_DisableLSECSS();
 
+  /* Set tick interrupt priority, default HAL value is intentionally invalid
+     and that prevents PLL initialization in HAL_RCC_OscConfig() */
+  HAL_InitTick((1UL << __NVIC_PRIO_BITS) - 1UL);
+
   /* Enable MSI Oscillator and activate PLL with MSI as source */
   RCC_OscInitStruct.OscillatorType      = RCC_OSCILLATORTYPE_MSI;
   RCC_OscInitStruct.MSIState            = RCC_MSI_ON;
--- a/hw/bsp/stm32l476disco/stm32l476disco.c
+++ b/hw/bsp/stm32l476disco/stm32l476disco.c
@@ -86,9 +86,10 @@ static void SystemClock_Config(void)
   RCC_OscInitTypeDef RCC_OscInitStruct;
   RCC_PeriphCLKInitTypeDef PeriphClkInitStruct;
 
-  /* Enable the LSE Oscilator */
+  /* Enable the LSE Oscillator */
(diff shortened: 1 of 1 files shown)


hathach:
Thank you for your PR, I have been struggling with this specific board STM32L476 Disco, however, I just pull out the board and run `cdc_msc`, it doesn't seem to enumerate. Maybe I miss something hmm.

kasjer:
Hi, this one is just to correct obvious clocking issues.
It will not help with USB working (USB clock was correct so it is not reason why USB is not working).
This is whole different story and I'm glad that there is issue opened for this one. I'll add comments in #126.

hw/bsp/stm32l476disco/stm32l476disco.c
```diff
@@ -86,14 +86,19 @@ static void SystemClock_Config(void)
...
   RCC_OscInitStruct.LSEState = RCC_LSE_ON;
+  RCC_OscInitStruct.PLL.PLLState = RCC_PLL_NONE;
   HAL_RCC_OscConfig(&RCC_OscInitStruct);
 
   /* Enable the CSS interrupt in case LSE signal is corrupted or not present */
   HAL_RCCEx_DisableLSECSS();
 
+  /* Set tick interrupt priority, default HAL value is intentionally invalid
+     and that prevents PLL initialization in HAL_RCC_OscConfig() */
+  HAL_InitTick((1UL << __NVIC_PRIO_BITS) - 1UL);
```
hathach, line 100: this is too tricky for me

salkinium:
Just FYI for the next unlucky dev:
I've also unsuccessfully tried to get USB working on this board, apparently the board needs to be modified first. (From the User Manual DocID027676 Rev 3):

hathach:
> Just FYI for the next unlucky dev:
> I've also unsuccessfully tried to get USB working on this board, apparently the board needs to be modified first. (From the User Manual DocID027676 Rev 3):

I remembered seeing something like this, hmm, I think I tried it out one way or another. Not entirely sure though, my memory isn't good anymore. Have you finally got it working .

salkinium:
No, it doesn't work for me either with or without these changes. I'll poke around more with a debugger, and get back to you if I have something.

hathach:
> No, it doesn't work for me either with or without these changes. I'll poke around more with a debugger, and get back to you if I have something.

I think st sdk usb example works with the board, so I guess the hw is fine. Though again my memory is not good, but it is worth for you to check it out.

PR #395, opened 2020-05-08
add CFG_TUSB_DEBUG_PRINTF() for log retargeting

### LINKED ISSUE #394: Support using SEGGER_RTT_printf
The tinyusb core now has partial support for SEGGER RTT.  However, support is incomplete, and it is not possible to use it with something such as circuit python.

This is caused by the definition of `tu_printf`:

```cpp
#ifndef tu_printf
    #define tu_printf           printf
#endif
```

The Segger RTT library has the following definition:

```cpp
int SEGGER_RTT_printf(unsigned BufferIndex, const char * sFormat, ...);
```

This is incompatible with printf due to the additional first argument.

It is possible to get this working by applying the following patch to `src/common/tusb_common.h`:

```patch
diff --git a/src/common/tusb_common.h b/src/common/tusb_common.h
index 052caef2..46468e5b 100644
--- a/src/common/tusb_common.h
+++ b/src/common/tusb_common.h
@@ -214,6 +214,8 @@ static inline bool     tu_bit_test (uint32_t value, uint8_t pos) { return (value
 #if CFG_TUSB_DEBUG

 void tu_print_mem(void const *buf, uint16_t count, uint8_t indent);
+#include "SEGGER_RTT.h"
+#define tu_printf(...) SEGGER_RTT_printf(0, __VA_ARGS__)

 #ifndef tu_printf
   #define tu_printf         printf
```

It would be nice if there was a CPP macro that did this.

hathach:
**Describe the PR**
Allow using custom printf function such as SEGERR_RTT_PRINTF() via macro define

**Additional context**
fix #394

### DIFF
--- a/src/common/tusb_common.h
+++ b/src/common/tusb_common.h
@@ -215,8 +215,11 @@ static inline bool     tu_bit_test (uint32_t value, uint8_t pos) { return (value
 
 void tu_print_mem(void const *buf, uint16_t count, uint8_t indent);
 
-#ifndef tu_printf
-  #define tu_printf         printf
+#ifdef CFG_TUSB_DEBUG_PRINTF
+  extern int CFG_TUSB_DEBUG_PRINTF(const char *format, ...);
+  #define tu_printf    CFG_TUSB_DEBUG_PRINTF
+#else
+  #define tu_printf    printf
 #endif
 
 static inline


hathach:
@xobs let's me know if this work for you, since rtt printf has extra channel argument, you will need to implement an proxy function to pass channel=0.

xobs:
The only problem with this is that your printf function also needs to have a variadic version.  I suppose most do, including Segger, which has `SEGGER_RTT_printf()` and `SEGGER_RTT_vprintf()`.

There may be special embedded environments that don't provide the `vprintf` form of the function, in which case this patch won't work for them.  However, that's probably such a niche case that it's not important to accommodate.

hathach:
yeah, that is std printf() signature, tinyusb log depends on it since it does use std format. We can try to fix minimal platforms that doesn't support that when it arises later on.

PR #398, opened 2020-05-08
Add IAR EWARM compiler attribute and endian support.

HiFiPhile:
**Describe the PR**
Add IAR compiler attribute and endian support in `tusb_compiler.h`, which let TinyUSB build with IAR EWARM.

Tested with IAR 8.50.1 and STM32L072KB.


HiFiPhile:
It's indeed the same syntax as gcc 🙂.  There are some compiler warning, I'll try to fix them.

Issue #400, opened 2020-05-09
CDC DTR compatibility on Windows


HiFiPhile:
**Set up**
[Mandatory] Provide details of your setup help us to reproduce the issue as quick as possible  
 - **PC OS**   : Windows 10
 - **Board**   : STM32L072KB
 - **Firmware**: examples/device/cdc_dual_ports

**Describe the bug**
On Windows DTR bit is not set automatically when the port is opened, and some terminal programs don't set this bit. Which makes TinyUSB believe it's in unconnected state.

**To reproduce**
1. Launch HTerm 0.81.
2. Open the port and send some data.
3. See error.

To make it work, I used a ugly workaround, which ignores the DTR bit:
```
case CDC_REQUEST_SET_CONTROL_LINE_STATE:
    {
      // CDC PSTN v1.2 section 6.3.12
      // Bit 0: Indicates if DTE is present or not.
      //        This signal corresponds to V.24 signal 108/2 and RS-232 signal DTR (Data Terminal Ready)
      // Bit 1: Carrier control for half-duplex modems.
      //        This signal corresponds to V.24 signal 105 and RS-232 signal RTS (Request to Send)
      bool const dtr = tu_bit_test(request->wValue, 0);
      bool const rts = tu_bit_test(request->wValue, 1);

      p_cdc->line_state = (uint8_t) request->wValue | 0x01;
```

hathach:
Most terminal did, it is a well known norm. I know no alternatives to detect connection. I would suggest you to change terminal client or file issue to request the dtr implementation for that terminal

hathach:
Also please provide the log specifically when you connect and disconnect with your terminal

HiFiPhile:
> Most terminal did, it is a well known norm.

I agree with you, proper terminal clients handle this correctly. 

But there are some cases, especially some proprietary industrial control softwares doesn't handle it, they are hard to be replaced.

Silab, ST and J-Link's CDC implementation don't rely on DTR neither.

I suggest to add an option to bypass DTR detection.

> PS: when you connect/disconnect. Did the terminal set the rts bit ?

RTS is also not managed in my cases.

HiFiPhile:
> I suggest to add an option to bypass DTR detection.

In fact it's a feature request to make TinyUSB compatible with more terminal clients, rather than a bug.

hathach:
Unfortunately, It doesn't make much sense for me to add that option for now.

hathach:
[...]

PR #401, opened 2020-05-09
Add option to ignore DTR bit in CDC connection check

HiFiPhile:
Some non-standard terminal programs don't set DTR bit while connected, which makes TinyUSB believe that no connection is established.
 
Added an option to ignore DTR bit in connection check.

### DIFF
--- a/src/class/cdc/cdc_device.c
+++ b/src/class/cdc/cdc_device.c
@@ -93,8 +93,12 @@ static void _prep_out_transaction (uint8_t itf)
 //--------------------------------------------------------------------+
 bool tud_cdc_n_connected(uint8_t itf)
 {
+#if CFG_TUD_CDC_IGNORE_DTR
+  return tud_ready();
+#else
   // DTR (bit 0) active  is considered as connected
   return tud_ready() && tu_bit_test(_cdcd_itf[itf].line_state, 0);
+#endif
 }
 
 uint8_t tud_cdc_n_get_line_state (uint8_t itf)
--- a/src/class/cdc/cdc_device.h
+++ b/src/class/cdc/cdc_device.h
@@ -38,6 +38,14 @@
 #define CFG_TUD_CDC_EPSIZE 64
 #endif
 
+/* Some non-standard terminal programs don't set DTR bit while connected, 
+ * which makes TinyUSB believe that no connection is established.
+ * Use this option to ignore DTR bit in connection check.
+ */
+#ifndef CFG_TUD_CDC_IGNORE_DTR
+#define CFG_TUD_CDC_IGNORE_DTR 0
+#endif
+
 #ifdef __cplusplus
  extern "C" {
 #endif


duempel:
I think this PR will work great with the current CDC example application. But I'm not sure, if there is a need for this feature.

TinyUSB CDC class also works without DTR to be set. Just don't check the connection status with `tud_cdc_connected()`. Because we actually want to know the connection state of a data terminal with this function, it would not show sense to me to call this function with CFG_TUD_CDC_IGNORE_DTR to be set.

Also this won't work with all terminal software. Maybe there are some terminals which always ignore DTR. This would be fine then. But if using HTerm (just as an example because you mentioned it at #400 ) or others, which have manual control about the DTR, you will also not receive any data if you do not set DTR.

The main problem in my eyes is, that an unexperienced user might try out CDC example with a terminal that does not set DTR. In this case the example application won't work and the user maybe does not know anything about RS-232 control signals or how the communication interface works. This user also would not know that setting CFG_TUD_CDC_IGNORE_DTR can help him out. If the user would recognize that he's checking for a connection state that was not enabled by his terminal software, he might just turn off the check.

To prevent this case, we maybe should change the CDC exmaple as well. Maybe a comment at `cdc_task()` or at main to explain the user he has to use a terminal software which sets DTR. Or never checking for connection state at the example. (The example echos back received data. If data is received there's always a valid connection. No need to check tud_cdc_connected() for this use case.)

I don't think this PR is bad. Just want to tell my concerns about this feature. If we want to keep the cdc example as it is right now, then a comment to set CFG_TUD_CDC_IGNORE_DTR if it's not working would be the best solution. But since TinyUSB already works fine without DTR it's maybe better to change the example into not using this check.

hathach:
@duempel thank you for your comment, cdc driver does check for the connection state before flushing https://github.com/hathach/tinyusb/blob/master/src/class/cdc/cdc_device.c#L170 resulting the data dropped if not connected.  without the connected() check, when using with normal terminal, it will cause the overflow of the TX buffer, since host driver won't send IN token to receive data from endpoints. And the edpt will keep busying, eventually overflow due to more data written from application.

Of course, windows bus driver can work differently and always send IN token and buffered all data until some driver pull from that or the TinyUSB internal FIFO  is just not full. 

Although I don't like the idea at first, when commenting in #400 , though I don't want to leave people without option either. @HiFiPhile I just checked the HTerm log http://www.der-hammer.info/terminal/changelog.txt it does support DTR/RTS command to set line state. Would you be able to use those for your purpose ?

duempel:
> cdc driver does check for the connection state before flushing

Oh, this was my fault. I didn't check the code and thought I knew how it is working. Thanks for this information. But is it necessary to check the connection before flushing? The comment on this line `// fifo is empty if not connected` is not clear to me. We just get to this point, because the fifo is **not** empty (count != 0). If verification of connection fails and we return from write_flush the data was already dropped off tx_ff but not transmitted. I'm not sure if this is the wished behavior of class driver. If we can't transmit data because the data terminal isn't ready, I want to wait and retry at a later point. If we already discard the tx_ff we would have to write the data to fifo again. Just to mention this, it's better to discuss in a new issue.

hathach:
@duempel it is my intended behavior to drop characters if it is not connected. As mentioned above, I rembmered linux driver doesn't send IN token if there is no terminal connected, if application continue to print, it will overflow the buffer. Maybe we can change it a bit to only drop data when the buffer is full.

HiFiPhile:
I think we can leave these changes for now, once #385 is merged I can 'hack' my own cdc class.
Instead in cdc examples we can add comments about the DTR bit.

hathach:
Ok, we will leave it open for now, I converted it to draft PR.

me-no-dev:
@hathach is it possible that instead of deciding based on DTR, actually to use the fact that data has not been sent/fifo is not empty? Or just leave it up to the application (returning 0 if no space available). I also have a case where DTR/RTS lines are used for different things and we need to communicate with the other end even when lines are low. The application itself can detect that the other end is alive, but bytes can not be sent in any way. (Actually one packet comes, proving connection works, but that is it). Forking the driver will be a way once #385 is merged, but change is rather minimal/generic enough that could maybe make sense to be upstream (optional?).

hathach:
@me-no-dev can you describe your issue in detail, the DTR is standard OS way for terminal to connect, using DTR for other purpose like Arduino autoreset is a actually hack. However, I am open for discussion if you could break it down and there are no other alternatives to address a real issue.

me-no-dev:
If I want to create a transparent USB-UART bridge to use for flashing ESP devices, the app needs to handle particular DTR/RTS changes that will trigger the target to reboot in the proper mode, after which it needs to communicate with the target. Unfortunately the last change leaves both RTS and DTR low and further communication is impossible.

hathach:
> If I want to create a transparent USB-UART bridge to use for flashing ESP devices, the app needs to handle particular DTR/RTS changes that will trigger the target to reboot in the proper mode, after which it needs to communicate with the target. Unfortunately the last change leaves both RTS and DTR low and further communication is impossible.
> 
> All starts as expected with both lines to high and baudrate update, then the series of 4 line changes are issued (while connected).

I still don't understand where is your issue with DTR, have you got into issue with your existing upload process for now ? if yes, please break it down the process into small step and where is the step the issue occurs.

me-no-dev:
Here is the signal flow of such upload process:

- Line State: DTR:1, RTS:1 (Connected)
- Line Coding: baud, etc.
- Reset Target
  1. Line State: DTR:0, RTS:1 (Disconnected)
  2. Line State: DTR:1, RTS:1 (Connected)
  3. Line State: DTR:1, RTS:0 (Connected)
  4. Line State: DTR:0, RTS:0 (Disconnected)

- Send the firmware **(but TinyUSB thinks we are disconnected, so it fails)**
- Line State: DTR:0, RTS:0 (Disconnected)

hathach:
I have a couple of question
1. at state iv, the device is already in ROM loader right ? 
2. why not set DTR to 1 before sending firmware

me-no-dev:
1. No, it is not in bootloader yet. The toggles interface with two fets to toggle RST and IO0 in the proper sequence to reset to bootloader. In reality a few milliseconds later is when the target is in bootloader.
2. If we toggle the lines again, they will reset the target back into the firmware

hathach:
> 1. No, it is not in bootloader yet. The toggles interface with two fets to toggle RST and IO0 in the proper sequence to reset to bootloader. In reality a few milliseconds later is when the target is in bootloader.
> 2. If we toggle the lines again, they will reset the target back into the firmware

Ok, so after a few ms of state iv, the device reset into bootloader. Why setting DTR to 1 again while in bootloader cause esp32s2 reset into application ?  If I manually holding GPIO0 to force bootloader mode, and then connect with another terminal, it will also set the DTR to 1 as well, in that case would it reset to app upon connection as well ?

me-no-dev:
> If I manually holding GPIO0 to force bootloader mode, and then connect with another terminal, it will also set the DTR to 1 as well, in that case would it reset to app upon connection as well ?

My English level did not help me understand what you mean? If you manually hold IO0 and connect from another terminal (just regular terminal, not esptool) on connect the chip will enter bootloader.

> Ok, so after a few ms of state iv, the device reset into bootloader. Why setting DTR to 1 again while in bootloader cause esp32s2 reset into application ?

because of the resulting switches of the fets. Any toggle of DTR at this point will cause the reset. Here is a schematic of such circuit:
Screenshot 2020-07-13 at 11 54 11

hathach:
> because of the resulting switches of the fets. Any toggle of DTR at this point will cause the reset. Here is a schematic of such circuit:

OK, I am not good with hardware, so I am still failed to understand why DTR cannot be set again. It is native USB connection, DTR/RTS are software control line state and  are not necessary hard-writed to anything. In my point of view, DTR is used to indicate active connection, you are using it for different purpose as hardware hack, and it is not really the good reason for the stack to change its behavior !! As I said, I am still open for discussion, but so far I still don't understand the need to change.

me-no-dev:
I get your point, but regular USB UARTs do not prevent the data flow when toggling the lines, therefore we have been using this for the last 6 years (as far as I remember, not sure for longer). I am not sure if this is a hack really, we have APIs to control the lines in all operating systems.

Could this just be made optional? So schemes where lines need to switch and data needs to flow can be handled?

hathach:
> I get your point, but regular USB UARTs do not prevent the data flow when toggling the lines, therefore we have been using this for the last 6 years (as far as I remember, not sure for longer). I am not sure if this is a hack really, we have APIs to control the lines in all operating systems.
> 
> Could this just be made optional? So schemes where lines need to switch and data needs to flow can be handled?

sure, It is totally possible, however the stack will loose the ability to detect active and/or new connection ( since DTR is used differently now). So I would like to understand reason behind that. Could you provide an in-depth with explanation again why it is a must in upload scenario.

me-no-dev:
This is not an S2 upload scenario. This is the scenario where we use an S2 just like Arduino uses ATMega chips for USB-UART and the S2 is just acting as USB-UART and passing data/toggles through to the target ESP (can be any ESP model. they all work the same).

In this case the application is well aware of when it is connected and when not, but we can not tell tinyusb that. Flush is the only issue. If we were just to receive data from the host and not need to talk back to it, it would have worked even now.

> It is totally possible, however the stack will loose the ability to detect active and/or new connection

This surely needs to be optional and not default. In "regular" applications this is not necessary for us either.

hathach:
> This is not an S2 upload scenario. This is the scenario where we use an S2 just like Arduino uses ATMega chips for USB-UART and the S2 is just acting as USB-UART and passing data/toggles through to the target ESP (can be any ESP model. they all work the same).

I see, seem like we are talking about totally different thing from the start. Could you explain it all again starting with the schematics. If you intend to use ATMega, you should probably want to use LUFA since TinyUSB doesn't support it yet.

me-no-dev:
We intend to use ESP32-S2, not ATMega. ATMega was just an example that the Arduino guys use on their boards.

Schematics are irrelevant to the point where they are the reason why esptool sends those particular line changes. In the resulting application where we use the S2, we can handle that in software without transistors on the other end. They are necessary only on boards that use regular USB-UART chips.

The application schematic is one S2 with USB, directly connected with RX/TX/EN/IO0 to the target ESP on the board (the ESP that will run the user applications and have antenna etc.).

Goal is the S2 to act just like any other FTDI, SiLabs, etc. USB-UART chip and pass transparently RX/TX and LineState changes. Everything is working as expected, except tx flush, because the last LineState that esptool sends before starting communication with the bootloader, leaves DTR low.

me-no-dev:
> However if the scenario isn't materialized yet, we can wait until you are working on it, and discuss more later on. Otherwise, we can just add an option for just in case scenario where in fact no body is using it.

Ahh you edit as I write :D We are in fact working on it. It is an active project and we hit this issue.

hathach:
> The application schematic is one S2 with USB, directly connected with RX/TX/EN/IO0 to the target ESP on the board (the ESP that will run the user applications and have antenna etc.).
> 
> Goal is the S2 to act just like any other FTDI, SiLabs, etc. USB-UART chip and pass transparently RX/TX and LineState changes. Everything is working as expected, except tx flush, because the last LineState that esptool sends before starting communication with the bootloader, leaves DTR low.

I see, so S2 is playing the role of FTDI chip, passing through UART data whereas
- DTR is connected to Reset, and 
- RTS is connected to  GPIO0 of target MCU.

I see why you would need the option now. Thank you for your patients in explaining.

me-no-dev:
Thanks for having the conversation :) I am not always the best at explaining.

hathach:
@duempel after a lengthy discussion with @me-no-dev (there is a bit of misunderstanding at firs), you could just read the last 2,3 comments. I think there is actual a scenario where this option is a must. The DTR/CTS are used as proxy singal controlling to another target MCU pins rather than its original meaning of connection detection. in Arduino schematics, DTR is mostly connected to target mcu reset. Let's me know if you have any other opinion, otherwise, I think we can merge this PR.

hathach:
> Thanks for having the conversation :) I am not always the best at explaining.

no problems, English is not my native language as well :)

duempel:
Thanks @me-no-dev and @hathach for bringing up this PR again. I've followed your conversation the last days but was too busy to leave any comments.
I think it is important to get CDC working without DTR being set. The example @me-no-dev gave is just one of the reasons. There are a lot of proprietary software solution which do not set this bit as it should. I just run into this trap again when changing USB middleware of an existing project to TinyUSB for development cases. There will be a lot of small embedded solution which do not make use of flow control to keep things easy. And for those systems controlling software is also developed without thinking about DTR bit. Most of those software will not work with TinyUSB anymore.

But I am not sure if a define is a good way to ignore DTR. I was reviewing CDC implementation and think that the way we are currently checking `tud_cdc_n_connected` within `tud_cdc_n_write_flush` is not very useful.
**Problem:** DTR bit should be used to check if the data terminal is ready and able to receive new data. If DTR is not set and we send out the data, we can lose all of it, because we don't know if the terminal will take the data or not. But with our current implementation the data is pulled out the FIFO before `tud_cdc_n_connected` is checked. This means the data is lost anyway (it's not real lost because still saved at cdc->epin_buf, but we can not transmit this data afterwards with our current API). Of course we know that the data has not been transmitted because we return 0, but it's not at FIFO anymore which means it's lost. So I don't see a reason for not just sending it out.
**My opinion:** We should just remove the check within `tud_cdc_n_write_flush`. If the user wants to make use of DTR he just calls `tud_cdc_n_connected` before flushing. If he wants to ignore DTR just flush without checking. The comment documentation for this API should be improved to explain this to the user.

Also if we would use a define to ignore DTR we can not use DTR for a 2nd CDC interface. I can also imagine there can be projects which can send out important setup data only if terminal is ready but ignore this bit for frequently sent out measurement data.

duempel:
I just remembered we were already talking about this [here](https://github.com/hathach/tinyusb/pull/401#issuecomment-627897830) and [here](https://github.com/hathach/tinyusb/pull/401#issuecomment-627934450).

If no IN token is being sent by host the DTR bit is irrelevant anyway. But as you said the endpoint will be busy and no data can be sent out. In this case maybe it's better to not start a transfer. 🤔 Maybe a API to abort a transfer which can be called if a new connection was started by change of line coding could help here..

hathach:
> But I am not sure if a define is a good way to ignore DTR. I was reviewing CDC implementation and think that the way we are currently checking `tud_cdc_n_connected` within `tud_cdc_n_write_flush` is not very useful.
> **Problem:** DTR bit should be used to check if the data terminal is ready and able to receive new data. If DTR is not set and we send out the data, we can lose all of it, because we don't know if the terminal will take the data or not. But with our current implementation the data is pulled out the FIFO before `tud_cdc_n_connected` is checked. This means the data is lost anyway (it's not real lost because still saved at cdc->epin_buf, but we can not transmit this data afterwards with our current API). Of course we know that the data has not been transmitted because we return 0, but it's not at FIFO anymore which means it's lost. So I don't see a reason for not just sending it out.

Yeah, it is not very smart right now, I was looking for an easy way to drain fifo when host not sending out IN token to avoid overflow FIFO (e.g application doing tons of logging). But we could tweak it to overwrite FIFO when DTR is not connected, and keep holding data in FIFO as long as possible.

> **My opinion:** We should just remove the check within `tud_cdc_n_write_flush`. If the user wants to make use of DTR he just calls `tud_cdc_n_connected` before flushing. If he wants to ignore DTR just flush without checking. The comment documentation for this API should be improved to explain this to the user.

Yes, this makes sense to me, giving it to application to decide make things more flexible. 

> Also if we would use a define to ignore DTR we can not use DTR for a 2nd CDC interface. I can also imagine there can be projects which can send out important setup data only if terminal is ready but ignore this bit for frequently sent out measurement data.

Right, this will disable DTR for all the interfaces.

hathach:
> I just remembered we were already talking about this [here](https://github.com/hathach/tinyusb/pull/401#issuecomment-627897830) and [here](https://github.com/hathach/tinyusb/pull/401#issuecomment-627934450).
> 
> If no IN token is being sent by host the DTR bit is irrelevant anyway. But as you said the endpoint will be busy and no data can be sent out. In this case maybe it's better to not start a transfer.  Maybe a API to abort a transfer which can be called if a new connection was started by change of line coding could help here..

Right, however one may argue to have the data sent to host anyway, to have a full print log even terminal connected a few second late (especially log around the enumeration), provided the FIFO is not full yet, there is no need to drop data then. More importantly, I am too lazy for now to add dcd abort API for all ports. How about doing all above and skip the abort, here is the sum up 

1. Skip connected() check within write_flush() --> up to application, not need for IGNORE_DTR macro
2. TX FIFO become overwritable if DTR is not set (nonoverwritable if DTR is set), this to prevent FIFO overflow without IN token from host. Application should use write_available() and/or connected() to prevent accidental ovewrite.
3. Skip abort API, which could possibly cause missing a chunk of data after EP BUF due to buffer overflow up when DTR is not set (and host not sending IN token). 

 No 3 may sound severe but not really, the abort API if added help to completely remove the data queued in EP BUF, but yeah I guess having it wouldn't hurt much, we could update it later on. I will try to be lazy until there is a strong reason to add dcd abort later on :smile:

HiFiPhile:
> Right, however one may argue to have the data sent to host anyway, to have a full print log even terminal connected a few second late.

I think we could check the behavior of USB-Uart chips.

duempel:
> But we could tweak it to overwrite FIFO when DTR is not connected, and keep holding data in FIFO as long as possible.

This is a pretty good idea. If I understand you correctly we just have to set
```
p_cdc->tx_ff.overwritable = dtr?0:1;
```
within control request (e.g. [here](https://github.com/hathach/tinyusb/blob/master/src/class/cdc/cdc_device.c#L359)). This will enable that we can still write data to the buffer, even if we do not transfer them (in case no IN token are being requested). 👍

> How about doing all above and skip the abort, here is the sum up
> 1. Skip connected() check within write_flush() --> up to application, not need for IGNORE_DTR macro
> 2. TX FIFO become overwritable if DTR is not set (nonoverwritable if DTR is set), this to prevent FIFO overflow without IN token from host. Application should use write_available() and/or connected() to prevent accidental ovewrite.
> 3. Skip abort API, which could possibly cause missing a chunk of data after EP BUF due to buffer overflow up when DTR is not set (and host not sending IN token).

This sounds good! Also skipping the abort API is no problem, since this is just useful if no IN token are sent. Worst case we send out one very old packet stored at ep_in buffer just before a terminal software releases DTR and the driver won't request IN ep anymore. But still the application can check `tud_cdc_n_connected` and manage this problem.

> I think we could check the behavior of USB-Uart chips.

I just tested the behavior with one FTDI chip and also with STLink Virtual COM Port of Nucleo board. The FTDI seemed to never store and send old data after starting a new connection. ST's virtual com port had the same behavior when first connecting to it. But after disconnection and reconnection I instandly received some old data. Both of them also transmit data even I do not set DTR with HTerm!

HiFiPhile:
> I just tested the behavior with one FTDI chip and also with STLink Virtual COM Port of Nucleo board. The FTDI seemed to never store and send old data after starting a new connection.

Personally I prefer the FTDI behavior, maybe it's not a big thing but in some case we may have data leak problem if we store old data.

hathach:
> > I just tested the behavior with one FTDI chip and also with STLink Virtual COM Port of Nucleo board. The FTDI seemed to never store and send old data after starting a new connection.
> 
> Personally I prefer the FTDI behavior, maybe it's not a big thing but in some case we may have data leak problem if we store old data.
> 
> But with the ACM model it could be difficult to detect the connection state without DTR.

FTDI is custom usb2uart, they wrote both PC driver and firmware, they probably have other way to tell if terminal is connected. For ACM, if client doesn't set DTR, we have no way to know if they are ready. I guess we have to live with that. Let's me know if you have anything to add to my above proposed approach. CDC is most used interface, hopefully the changes won't break other project.

duempel:
Yes, as @hathach said the FTDI driver have much more possibilities to detect connection and disconnection events. I see a lot of control packets for connection management here.
But I think the way we want to go now is just fine. Every user can decide on his own if he wants to implement DTR checking for his application or not.

> Let's me know if you have anything to add to my above proposed approach. CDC is most used interface, hopefully the changes won't break other project.

Maybe an option for fifo overwriting would be a nice feature to keep consistency? Something like `CFG_TUD_CDC_DISABLE_UNCONNECTED_BUF_OVERWRITING` with a shorter name 😂. But this is just optional since write_available API can handle it.

hathach:
> Maybe an option for fifo overwriting would be a nice feature to keep consistency? Something like `CFG_TUD_CDC_DISABLE_UNCONNECTED_BUF_OVERWRITING` with a shorter name . But this is just optional since write_available API can handle it.

IMHO, TX fifo must not be overwritten when we know there is client connected to drain data. Since most of the time, application will just call write() and check the written_bytes. Disabling overwrite will definitely cause loss of data for lots of current project. And I don't see CDC interface could want that in an active connection.

duempel:
@hathach I think there is a misunderstanding. Currently overwrite is always disabled. Because current projects are not overwriting the buffer, I asked for an option which just optionally disables 2. of [your comment](https://github.com/hathach/tinyusb/pull/401#issuecomment-658003219) to keep behavior as it was before.

Changes of behavior:

- If DTR bit is set new and old implementation will work the same
- If no DTR is set, flushing of both implementation will remove data from fifo, but only new will schedule a edpt xfer. Also tx fifo will become overwritable.
- If fifo is overwritable `tud_cdc_n_connected` and `tud_cdc_n_write_available` will behave the same. Just `tud_cdc_n_write` can still push data to fifo if the bytes available have not been checked first.

Well it's difficult to think about every possible application. My use cases won't be effected by this change. Not checking for DTR before flushing will cause a bigger change of behavior than enabling buffer overwriting. Since current application may only flush without checking `tud_cdc_n_conntected` they might see old data come up in terminal just after connecting. If we are doing these changes to CDC class old applications have to be changed if they do not wish this behavior. Future applications can only make profit from it. One day dcd abort api can make improvements to this and cancel a scheduled transfer in case DTR bit is changed back to 1.

HiFiPhile:
@duempel would you like to get write access to this branch or create a new PR to show your idea ?

duempel:
Thanks @HiFiPhile but since the new discussed implementation would differ from this PR's description I think it's better to open a new PR for this.

Maybe we can just merge this PR in the first step, as @hathach already wanted to [here](https://github.com/hathach/tinyusb/pull/401#issuecomment-657483151). Since this is a very easy concept and will not bring any other issues up to existing applications.

I could then create a new issue to discuss about other possibilities. Let's wait what @hathach thinks about this. Maybe he does not like to merge short term interim solution.

hathach:
> Thanks @HiFiPhile but since the new discussed implementation would differ from this PR's description I think it's better to open a new PR for this.

Yeah, it makes more sense to implement this on a separated PR. There should be a reference to this PR along with a sum up as well.

> Maybe we can just merge this PR in the first step, as @hathach already wanted to [here](https://github.com/hathach/tinyusb/pull/401#issuecomment-657483151). Since this is a very easy concept and will not bring any other issues up to existing applications.

Open an issue is with sum up is good as well, I am a bit lazy to implement the change anyway ;)

> I could then create a new issue to discuss about other possibilities. Let's wait what @hathach thinks about this. Maybe he does not like to merge short term interim solution.

Yeah, I don't think we should merge this IGNORE_DTR pr, it is much easier to postpone the feature than adding a interim and removed it later. It will cause user frustration, and push a pressure to implement the feature asap on us.

me-no-dev:
> If we are doing these changes to CDC class old applications have to be changed if they do not wish this behavior. Future applications can only make profit from it

Those applications will need to specifically enable the "do not check DTR" option. If they do, they should be aware of the consequences, as that is an option that you usually first need to learn the protocol in order to figure out that you need. Such applications then can/should detect on their own when they are connected or not. I have also gotten old data in the UART before... that is not something strange.

How do you actually know that you are connected? Do you not get any setup at all? You should still get a LineState even if DTR is 0.

hathach:
> @hathach I think there is a misunderstanding. Currently overwrite is always disabled. Because current projects are not overwriting the buffer, I asked for an option which just optionally disables 2. of [your comment](https://github.com/hathach/tinyusb/pull/401#issuecomment-658003219) to keep behavior as it was before.
> 
> Changes of behavior:
> 
> * If DTR bit is set new and old implementation will work the same
> * If no DTR is set, flushing of both implementation will remove data from fifo, but only new will schedule a edpt xfer. Also tx fifo will become overwritable.
> * If fifo is overwritable `tud_cdc_n_connected` and `tud_cdc_n_write_available` will behave the same. Just `tud_cdc_n_write` can still push data to fifo if the bytes available have not been checked first.
> 
> Well it's difficult to think about every possible application. My use cases won't be effected by this change. Not checking for DTR before flushing will cause a bigger change of behavior than enabling buffer overwriting. Since current application may only flush without checking `tud_cdc_n_conntected` they might see old data come up in terminal just after connecting. If we are doing these changes to CDC class old applications have to be changed if they do not wish this behavior. Future applications can only make profit from it. One day dcd abort api can make improvements to this and cancel a scheduled transfer in case DTR bit is changed back to 1.

Thanks for the sum up, yeah, I kind of mixing up thing. Maybe we can add a new API() to enable/disable the overwritable of the TX FIFO rather than an compiler switch. Sound like we are trying to cover too many scenarios for now. But it is good to pre-consider these before writing actual code.

hathach:
> Those applications will need to specifically enable the "do not check DTR" option. If they do, they should be aware of the consequences, as that is an option that you usually first need to learn the protocol in order to figure out that you need. Such applications then can/should detect on their own when they are connected or not. I have also gotten old data in the UART before... that is not something strange.

We will use a more dynamic approach the compiler switch proposed in this PR.
 
> How do you actually know that you are connected? Do you not get any setup at all? You should still get a LineState even if DTR is 0.

I really don't understand these questions of yours. Seem missing a context I think.

me-no-dev:
> I really don't understand these questions of yours. Seem missing a context I think.

Sorry, they were meant for @duempel

duempel:
@hathach alright I will be opening a new issue with reference to this PR and writing a little summary of everything we've already discussed about. Also I can show up a draft PR with our favorite changes for now. But today I will most likely not be able to do that anymore.

@me-no-dev yes you are right about the "do not check DTR" option. But as @hathach already said we were already discussing about an other solution which is different from this PR's approach.

> How do you actually know that you are connected? Do you not get any setup at all? You should still get a LineState even if DTR is 0.

This is one of the main problem we are actually facing. TinyUSB's CDC implementation uses the communications class subclass _Abstract Control Model_. The request _SetControlLineState_ is actually used to tell the DCE device that the DTE device is present. But this is exactly our DTR bit and not an information about the physical connection. Also it's just optional to be used by ACM. I'm sure the requests _SetLineCoding_ and _GetLineCoding_ will be used everytime a connection is set up. But also this can be handled different by different systems. I'm sure we could use this as well to detect a new connection is set up. But I don't see a chance to detect disconnection events with this.

me-no-dev:
> But I don't see a chance to detect disconnection events with this.

Clear. Thanks!

hathach:
closed in favor of #506

PR #402, opened 2020-05-11
Add tud_cdc_flushed function

kasjer:
**Describe the PR**

tud_cdc_write_flush() returns 0 when ednpoint is busy (function was called
to soon) or endpoint fifo is empty.
To make it easier to decide if flush should be called later, new function
tud_cdc_flushed() returns true if output fifo is empty.

Without this it is possible it accomplish same task with following
condition

tud_cdc_write_available() < USBD_CDC_DATA_EP_SIZE

...but code is more awkward.

**Additional context**
Naming of the function could be change if it does not match TinyUSB conventions (like fifo_empty, is_fifo_empty or such)

### DIFF
--- a/src/class/cdc/cdc_device.c
+++ b/src/class/cdc/cdc_device.c
@@ -179,6 +179,10 @@ uint32_t tud_cdc_n_write_available (uint8_t itf)
   return tu_fifo_remaining(&_cdcd_itf[itf].tx_ff);
 }
 
+bool tud_cdc_n_flushed (uint8_t itf)
+{
+  return tu_fifo_empty(&_cdcd_itf[itf].tx_ff);
+}
 
 //--------------------------------------------------------------------+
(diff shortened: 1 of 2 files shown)


hathach:
why don't we just call write_flush() anyway ? it does the same check and return without queuing any usb data.

kasjer:
> why don't we just call write_flush() anyway ? it does the same check and return without queuing any usb data.

I did not noticed that commit 4571ce0d29398089cf6615515190a70ff9be2361 fixes the issue since once transfer is started there is no need to issue flush any more it will be automatically executed.

In that case there is no need for this PR.

hathach:
yeah, I wasn't really sure to do that at first, since it will mess up the baudrate setting in the LINE_CODING however people seem to prefer sending ASAP :D . I will let the application do the timing with the baudrate, that would be the best.

Issue #403, opened 2020-05-12
Performance improvements in tusb_fifo


HiFiPhile:
**Is your feature request related to a problem? Please describe.**
Currently I'm making a USB-RS485 bridge, I noticed that at high baud rate(1M), data transmitted from RS485 is discontinued, seems like data receiving of  USB is slower than UART transmit, so I decided to do some performance investigation.

|Configuration ||
---|---
MUC|STM32L072KB
Frequency|32MHz
Compiler| IAR 8.50.1, High optimization

From the `cdc_dual_ports` example, I toggle a GPIO to show the USB receiving speed of 140 bytes:
```
while (1)
{
    tud_task();
    for (int itf = 0; itf < CFG_TUD_CDC; itf++)
    {
        if ( tud_cdc_n_available(itf) )
        {
            uint8_t buf[64];
            BSP_L_DE_GPIO->BSRR = 1 << BSP_L_DE_PIN;
            uint32_t count = tud_cdc_n_read(itf, buf, sizeof(buf));
            BSP_L_DE_GPIO->BRR = 1 << BSP_L_DE_PIN;
        }
    }
}
```
Between 2 packet receptions I got 650us, and I was a little surprised that `tud_cdc_n_read` execute for 240us !

Then I looked inside and found `tusb_fifo` is over-complicated in order to take care of different item sizes, but in general we fill only bytes, only exception is `osal_queue_t`.

**Describe the solution you'd like**

- Simply tsub_fifo to take care only bytes, and make changes in `osal`.

- Add functions to read/write bytes efficiently.

Here I tested the 2nd way, I created a function `tu_fifo_read_n_bytes`:
```
uint16_t tu_fifo_read_n_bytes (tu_fifo_t* f, void * buffer, uint16_t count)
{
  if( tu_fifo_empty(f) ) return 0;

  tu_fifo_lock(f);

  /* Limit up to fifo's count */
  if ( count > f->count ) count = f->count;

  if(count + f->rd_idx < f->depth )
  {
      memcpy(buffer, f->buffer + f->rd_idx, count);
      f->rd_idx += count;
  }
  else
  {
      memcpy(buffer, f->buffer + f->rd_idx, f->depth - f->rd_idx);
      memcpy((uint8_t*)buffer + f->depth - f->rd_idx, f->buffer, count - (f->depth - f->rd_idx));
      f->rd_idx = (f->rd_idx + count) % f->depth;
  }
  f->count -= count;
  
[...]

PR #404, opened 2020-05-12
Optimize tusb_fifo transfer speed.

HiFiPhile:
Signed-off-by: Mengsk <admin@hifiphile.com>

**Describe the PR**
Optimize FIFO for byte transfer following issue #403. 

In my case USB OUT packet interval decreased from 650us to 208us, which is 3x performance gain.

### DIFF
--- a/src/common/tusb_fifo.c
+++ b/src/common/tusb_fifo.c
@@ -103,9 +103,9 @@ static void _tu_ff_push(tu_fifo_t* f, void const * data)
 
 /******************************************************************************/
 /*!
-    @brief Read one byte out of the RX buffer.
+    @brief Read one element out of the RX buffer.
 
-    This function will return the byte located at the array index of the
+    This function will return the element located at the array index of the
     read pointer, and then increment the read pointer index.  If the read
     pointer exceeds the maximum buffer size, it will roll over to zero.
 
@@ -132,8 +132,8 @@ bool tu_fifo_read(tu_fifo_t* f, void * buffer)
 
 /******************************************************************************/
 /*!
-    @brief This function will read n elements into the array index specified by
-    the write pointer and increment the write index. If the write index
+    @brief This function will read n elements from the array index specified by
+    the read pointer and increment the read index. If the read index
     exceeds the max buffer size, then it will roll over to zero.
 
     @param[in]  f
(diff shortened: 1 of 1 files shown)


kasjer:
I think that culprit for slow transfer in case that this PR tries to fix is lack of **DIV** instruction on **Cortex-M0**.
Code for handling FIFO (and other places as well) tend to use **%** (modulo) operator which is fast if CPU has instruction for it.

If _tu_fifo_pull() instead of having
```c
  f->rd_idx = (f->rd_idx + 1) % f->depth;
```
had this:
```c
  if (++(f->rd_idx) >= f->depth)
    f->rd_idx = 0;
```
...division that is done in software in a loop would not be used and code would be faster.

This PR reduced number of times division is used and that I guess makes a difference.

If, however, optimization did further reduced FIFO handling time I don't see why it should not be applied to non-byte size elements. Same schema with one or two memcpy could be used with no need to introduce new functions that basically has same arguments and do same thing.

hathach:
> If, however, optimization did further reduced FIFO handling time I don't see why it should not be applied to non-byte size elements. Same schema with one or two memcpy could be used with no need to introduce new functions that basically has same arguments and do same thing.

Yeah, This is a great idea, and is better to do it with current write_n/read_n , I actually was thinking to do this previously as well.

src/common/tusb_fifo.c
```diff
@@ -132,8 +132,8 @@ bool tu_fifo_read(tu_fifo_t* f, void * buffer)
 
 /******************************************************************************/
 /*!
-    @brief This function will read n elements into the array index specified by
-    the write pointer and increment the write index. If the write index
+    @brief This function will read n elements from the array index specified by
```
kasjer, line 135: Since this comment is updated anyway, it could be changed, **n** in description can easily be confused with **n** in function name tu_fifo_read_**n** and they are not related. It it was **count** instead of n confusion would be gone.

hathach: right, this happens when I try to write the function descriptions before its implementation. :D

HiFiPhile: Could you take a look also other commets ? English not my native language ;)

src/common/tusb_fifo.c
```diff
@@ -148,27 +148,34 @@ bool tu_fifo_read(tu_fifo_t* f, void * buffer)
 /******************************************************************************/
 uint16_t tu_fifo_read_n (tu_fifo_t* f, void * buffer, uint16_t count)
 {
-  if( tu_fifo_empty(f) ) return 0;
+  if(tu_fifo_empty(f))
```
kasjer, line 151: I'm not that familiar with TinyUSB style guidelines, but one-liners are pretty common so I guess that they are preferred, and in this case this line and next change should not be here at all.

HiFiPhile:
Wait,
```
    f->rd_idx = f->rd_idx + count;
    if (f->rd_idx >= f->depth)
      f->rd_idx = 0;
```
should be 
```
    f->rd_idx = f->rd_idx + count;
    if (f->rd_idx >= f->depth)
      f->rd_idx -= f->depth;
```

kasjer:
> Wait,
> 
> ```
>     f->rd_idx = f->rd_idx + count;
>     if (f->rd_idx >= f->depth)
>       f->rd_idx = 0;
> ```
> 
> should be
> 
> ```
>     f->rd_idx = f->rd_idx + count;
>     if (f->rd_idx >= f->depth)
>       f->rd_idx -= f->depth;
> ```
Yes, original comment was for pull that increments by 1 hence = 0 was safe

src/common/tusb_fifo.c
```diff
@@ -148,27 +148,34 @@ bool tu_fifo_read(tu_fifo_t* f, void * buffer)
...
   /* Limit up to fifo's count */
-  if ( count > f->count ) count = f->count;
-
-  uint8_t* buf8 = (uint8_t*) buffer;
-  uint16_t len = 0;
+  if(count > f->count)
+    count = f->count;
 
-  while (len < count)
+  if(count + f->rd_idx < f->depth)
```
duempel, line 160: Shouldn't this one better be `if(count + f->rd_idx <= f->depth)`?

Then we have to take care that f->rd_idx can get out of valid buffer area. We can increment the `f->rd_idx += count;` here. Same at Line 170 and then write

```
if (f->rd_idx >= f->depth){
  f->rd_idx -= f->depth;
}
```
Just after the if condition to be used with both memcpy cases.

HiFiPhile: Yeap

src/common/tusb_fifo.c
```diff
@@ -148,32 +148,39 @@ bool tu_fifo_read(tu_fifo_t* f, void * buffer)
...
+  if(tu_fifo_empty(f)) return 0;
 
   tu_fifo_lock(f);
 
   /* Limit up to fifo's count */
-  if ( count > f->count ) count = f->count;
-
-  uint8_t* buf8 = (uint8_t*) buffer;
-  uint16_t len = 0;
+  if(count > f->count)
```
hathach, line 156: I am either single line if or newline with full bracket.  Though this is not important, I would do a follow up pr for unit test and clean up anywayl.

PR #412, opened 2020-05-18
suppress cast-align warnings for net device driver

hathach:
**Describe the PR**
follow up to #411 use memcpy and/or `(void*)`  to suppress cast-align warning. It is indeed legit for gcc to warn when we increase the alignment requirement when casting pointer, even though M3/M4 has not problems with mis-aligned memory access, M0 and other MCUs may not

### DIFF
--- a/examples/device/net_lwip_webserver/Makefile
+++ b/examples/device/net_lwip_webserver/Makefile
@@ -6,9 +6,6 @@ CFLAGS += \
   -DTCP_WND=2*TCP_MSS \
   -DHTTPD_USE_CUSTOM_FSDATA=0
 
-# TODO rndis_reports.c and net_device cause cast algin warnings
-CFLAGS += -Wno-error=cast-align
-
 INC += \
   src \
   $(TOP)/hw \
--- a/lib/networking/rndis_reports.c
+++ b/lib/networking/rndis_reports.c
@@ -74,7 +74,7 @@ static const uint32_t OIDSupportedList[] =
 #define OID_LIST_LENGTH TU_ARRAY_SIZE(OIDSupportedList)
 #define ENC_BUF_SIZE    (OID_LIST_LENGTH * 4 + 32)
 
-static uint8_t *encapsulated_buffer;
+static void *encapsulated_buffer;
 
 static void rndis_report(void)
 {
@@ -152,7 +152,7 @@ static void rndis_query(void)
   }
 }
 
-#define INFBUF ((uint32_t *)((uint8_t *)&(m->RequestId) + m->InformationBufferOffset))
+#define INFBUF  ((uint8_t *)&(m->RequestId) + m->InformationBufferOffset)
 
 static void rndis_handle_config_parm(const char *data, int keyoffset, int valoffset, int keylen, int vallen)
 {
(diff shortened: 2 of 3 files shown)


hathach:
> Do you have a specific example of M0 failing, or did you arbitrarily lump in M0 to your disclaimer of "M0 and other MCUs may not"?
> I'll look at this PR more closely when I have a spare moment, but I don't have any strong objections.
> 
> The author of the RNDIS driver was fond of using pointers without regard to alignment; if you compare the code to the original, you'll see that I had already added some memcpy calls for similar reasons.

In general M0 cannot access `*((uint32_t*) addr)` if addr is not multiple of 4,else  it will cause hardfault. M3/M4 has no problems with that. In this example, there should be no issues since  **CFG_TUSB_MEM_ALIGN is default to aligned 4** --> encapsulated_buffer is aligned 4. Regardless of MCU, GCC `-Wcast-align` will warns whenever there is a cast that increase the alignment e.g from uint8_t* --> uint32_t* . To silence it, we either memcpy or cast uint8_t* -> void* -> uint32_t* (kind of annoying). Feel free to review it at anytime you like. I only enable cast-align with #411 , in general, this warning option is good to enable.

majbthrd:
> In this example, there should be no issues since **CFG_TUSB_MEM_ALIGN is default to aligned 4** --> encapsulated_buffer is aligned 4.

OK, so you are not trying to fix a specific problem; you are just quelling compiler warnings.  I agree there should be no existing issue; the compiler just isn't clever enough to know the pointer is being used judiciously.

hathach:
Yeah, it is exactly what I am trying to do :D . It could help to solve a potential problem in the future.

hathach:
@majbthrd the PR is relatively simple, it doesn't change anything, only do some casting to suppress gcc warnings. I did run the tests, it is good enough.

hathach:
> The changes to src/class/net/net_device.c seem unnecessary. The file could remain unchanged and the project still compiles without issue.

Which board and gcc version you are building. revert those changes still cause issue on multiple target e.g SAMD51/SAMD21/stm32f4 etc .. Hmm although we did declare notify with `CFG_TUSB_MEM_ALIGN`,  I even go further with 

```
CFG_TUSB_MEM_SECTION CFG_TUSB_MEM_ALIGN static union
{
  uint8_t rndis_buf[120] __attribute__ ((aligned(4))) ;
  struct ecm_notify_struct ecm_buf;
} notify;
```

Gcc still emit the warning anyway

```
src/class/net/net_device.c:329:44: error: cast increases required alignment of target type [-Werror=cast-align]
  329 |           rndis_generic_msg_t *rndis_msg = (rndis_generic_msg_t *) notify.rndis_buf;
```
Here is my gcc version

```
arm-none-eabi-gcc (xPack GNU Arm Embedded GCC, 64-bit) 9.3.1 20200408 (release)
Copyright (C) 2019 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
```

majbthrd:
OK, the difference is that I'm using:

````
gcc version 9.2.1 20191025 (release) [ARM/arm-9-branch revision 277599] (GNU Tools for Arm Embedded Processors 9-2019-q4-major) 
````
There doesn't seem to be much choice; it feels inelegant to workaround the misguided compiler, but I don't have a good alternative that isn't convoluted.  Thanks for your patience.

hathach:
> OK, the difference is that I'm using:
> 
> ```
> gcc version 9.2.1 20191025 (release) [ARM/arm-9-branch revision 277599] (GNU Tools for Arm Embedded Processors 9-2019-q4-major) 
> ```
> 
> There doesn't seem to be much choice; it feels inelegant to workaround the misguided compiler, but I don't have a good alternative that isn't convoluted. Thanks for your patience.

I admit the changes is a bit odd to look at, maybe we could do it better later on by changing the declaration of notify buffer or something. For now, I tried to just suppress the warnings :). Thanks for the review.

PR #413, opened 2020-05-19
Bluetooth HCI transport over USB

kasjer:
**Describe the PR**
Bluetooth specification in Vol. 4 part B describes USB transport for HCI (Host Controller Interface).
This PR provides necessary API so Bluetooth controller can use USB as transport layer.

Only Primary Controller is supported.
Isochronous endpoints are not handled (as they are not use in Bluetooth LE for now).
Isochronous endpoints can be specified in descriptor though as without them Windows
Bluetooth stack will not recognize device as Bleutooth controller.
Linux on the other hand can handle BLE only controller without extra ISO endpoints.

**Additional context**
Code was tested with **Apache Mynewt NimBLE** which is open source BLE only (not basic rate) host and controller stack. Additional information will be added with cross referencing PR in NimBLE project to include USB transport there.

There are some points that could be brought to attention:
- To handle Bluetooth transport as described in Bluetooth core specification requests with type *class* as and recipient *device* must be handled.
- Isochronous endpoints that are included in implementation (to allow Windows enumeration) can have numerous interface alternatives. For Windows to work correctly at least one is needed with endpoint size other then 0. So for this there is bunch of macros to create up to 6 interface alternatives.
- Nordic nRF52840-DK was used as testing platform.


hathach:
wow this is a great PR, and will be very useful for nRF mcus. Could you mind adding an example to demonstrate the usage of class. If the example requires mynet build system, just give me a link/build instruction to your project. I could will include it in my mynewt-example https://github.com/hathach/mynewt-tinyusb-example (just fixed its build issue recently). It will be easier for people for testing. It doesn't need to be complex, just simple enough to show how it works. 

Such as great class.

kasjer:
Let me put something together...

hathach:
> Let me put something together...

great, I am looking forward to test this out.

kasjer:
Bluetooth HCI transport allows communication between host and controller.
In this case host resides in PC (it can be bluez on Linux, or Microsoft stack), so there is no
application that user will use and modify.
NimBLE repository has application **blehci** that is meant to be used if only controller is flashed to MCU, and with this PR bluetooth dongle can be build and used.

This work requires some coordination of hithub repositories to test.

It also requires newest **newt** tool from github (released ones can have
problems building due to **-I** not being added for TinyUSB paths).

First mynewt project with NimBLE is needed.

```sh
newt new mynewt-example
cd mynewt-example/
```
This sets up basic project with blinky example.
File **project.yml** generated by **newt** tool needs to have some
repositories added for USB and BLE, following modifications should be enough to start.

**project.yml**
```yml
project.name: "my_project"

project.repositories:
    - apache-mynewt-core
    - apache-mynewt-nimble
    - tinyusb

# Use github's distribution mechanism for core ASF libraries.
# This provides mirroring automatically for us.
#
repository.apache-mynewt-core:
    type: github
    vers: 0.0.0
    user: apache
    repo: mynewt-core

repository.apache-mynewt-nimble:
    type: github
    vers: 0.0.0
    user: apache
    repo: mynewt-nimble

repository.tinyusb:
    type: github
    vers: 0.0.0
    user: hathach
    repo: tinyusb
```

Download repositories
```sh
newt upgrade
```

Two mynewt targets are needed: bootloader and application.
Bootloader is provided by mcuboot repository.
Application (blehci) is provided by NimBLE.
In this example nordic nRF52840-DK will be used.
Lets create target for bootloader **pca10056-boot**.


```sh
newt target create pca10056-boot
newt target set pca10056-boot bsp=@apache-mynewt-core/hw/bsp/nordic_pca10056
newt target set pca10056-boot app=@mcuboot/boot/mynewt
newt build pca10056-boot
newt load pca10056-boot
```

Before application can be made, get code from pull requests
https://github.com/hathach/tinyusb/pull/413
https://github.com/apache/mynewt-nimble/pull/819
https://github.com/apache/mynewt-core/pull/2292

This step is needed till all are merged into respective repositories.
```shell script
git -C repos/tinyusb/ fetch origin pull/413/head
git -C repos/tinyusb/ checkout FETCH_HEAD
git -C repos/apache-mynewt-nimble/ fetch origin pull/819/head
git -C repos/apache-mynewt-nimble/ checkout FETCH_HEAD
git -C repos/apache-mynewt-core/ fetch origin pull/2292/head
git -C repos/apache-mynewt-core/ checkout FETCH_HEAD
```
Create application target for **blehci** with some additional **syscfg** values
```sh
newt target create pca10056-blehci
newt target set pca10056-blehci bsp=@apache-mynewt-core/hw/bsp/nordic_pca10056
newt target set pca10056-blehci app=@apache-mynewt-nimble/apps/blehci
newt target set pca10056-blehci build_profile=debug
newt target set pca10056-blehci "syscfg=BLE_EXT_ADV=1:BLE_HCI_TRANSPORT=usb:BLE_PUBLIC_DEV_ADDR=(uint8_t[6]){0x20, 0x20, 0x10, 0x10, 0xBB, 0xCC}:USBD_BTH=1:USBD_PID=0x1234:USBD_VID=0x5678:BLE_MAX_CONNECTIONS=20"

```

**newt** toold does not have option to add target package dependencies,
so single dependency must be added maually to the target **pkg.yml**

**targtes/pca10056-blehci/pkg.yml**
```yml
pkg.deps:
    - "@apache-mynewt-core/hw/usb/tinyusb/std_descriptors"
```

or copy this to script along with all other commands in this description
```shell script
echo "pkg.deps:" >> targets/pca10056-blehci/pkg.yml
echo '    - "@apache-mynewt-core/hw/usb/tinyusb/std_descriptors"' >> targets/pca10056-blehci/pkg.yml
```
Now build and download to device
```sh
newt build pca10056-blehci
newt create-image pca10056-blehci timestamp
newt load pca10056-blehci 0
```

Now baord can be connected and it should be visible in Linux as bluetooth controller.
It will show up in Windows as not working device.
Reason for not working with Windows right away is that NimBLE is BLE only controller
while Windows expects also Bluetooth classic to be supported.
Some BLE experts I work with created modification of NimBLE that can convince
Windows that it is indeed valid controller and then BLE devices will also
work on Windows. It this fake Bluetooth classic support is added to NimBLE in
the future USB part will not need to be updated.

Here are some Linux command to use:
Show available controllers (NimBLE should show up along with preexisting ones)
```shell script
hciconfig
```
To scan device around command line tool can be used
```
bluetoothctl

[bluetooth]# list
[bluetooth]# select CC:BB:10:10:20:20 
[bluetooth]# scan on
[bluetooth]# scan off
[bluetooth]# devices
[bluetooth]# connect <some address from devices> 

```
To see packets (that should match what USB transfers)
```shell script
btmon
```

hathach:
wow, it is a lengthy process :+1: :+1: , I will try it out and feedback asap.

src/class/bth/bth_device.c
```diff
@@ -0,0 +1,238 @@
...
+    TU_VERIFY(request->bRequest == 0 && request->wValue == 0 && request->wIndex == 0);
+  }
+  else if (request->bmRequestType_bit.recipient == TUSB_REQ_RCPT_INTERFACE)
+  {
+    if (request->bRequest == TUSB_REQ_SET_INTERFACE && _btd_itf.itf_num + 1 == request->wIndex)
+    {
+      // TODO: Set interface it would involve changing size of endpoint size
+    }
+    else // HCI command packet for Primary Controller function in a composite device
+        TU_VERIFY(request->bRequest == 0 && request->wValue == 0 && request->wIndex == _btd_itf.itf_num);
```
hathach, line 207: we should open an `{ }` here, since TU_VERIFY() can screw up by expanding to multiple statements (although I have no intention to do that). We are still better to be safe.

src/class/bth/bth_device.c
```diff
@@ -0,0 +1,238 @@
...
+{
+  (void)result;
+
+  // received new data from host
+  if (ep_addr == _btd_itf.ep_acl_out)
+  {
+    if (tud_bt_acl_data_received_cb) tud_bt_acl_data_received_cb(_btd_itf.epout_buf, xferred_bytes);
+
+    // prepare for next data
+    TU_ASSERT(usbd_edpt_xfer(rhport, _btd_itf.ep_acl_out, _btd_itf.epout_buf, CFG_TUD_BTH_DATA_EPSIZE), false);
```
hathach, line 224: you could drop the `, false` part, it returns false by default 

```
TU_ASSERT(usbd_edpt_xfer(rhport, _btd_itf.ep_acl_out, _btd_itf.epout_buf, CFG_TUD_BTH_DATA_EPSIZE));
```

src/class/bth/bth_device.h
```diff
@@ -0,0 +1,86 @@
...
+ extern "C" {
+#endif
+
+//--------------------------------------------------------------------+
+// Application Callback API (weak is optional)
+//--------------------------------------------------------------------+
+
+// Invoked when HCI command was received
+TU_ATTR_WEAK void tud_bt_hci_cmd_cb(void *hci_cmd, size_t cmd_len);
+// Invoked when ACL data was received
```
hathach, line 60: maybe added newline for separation, it makes it easier to read :)

src/class/bth/bth_device.h
```diff
@@ -0,0 +1,86 @@
...
+TU_ATTR_WEAK void tud_bt_acl_data_received_cb(void *acl_data, uint16_t data_len);
+// Called when event was sent to BT stack
+TU_ATTR_WEAK void tud_bt_event_sent_cb(uint16_t sent_bytes);
+// Called when ACL data was sent to BT stack
+TU_ATTR_WEAK void tud_bt_acl_data_sent_cb(uint16_t sent_bytes);
+
+bool tud_bt_event_send(void *event, uint16_t event_len);
```
hathach, line 67: maybe adding a short description for these two send() API as well to help others.

src/device/usbd.c
```diff
@@ -499,6 +512,18 @@ static bool process_control_request(uint8_t rhport, tusb_control_request_t const
   {
     //------------- Device Requests e.g in enumeration -------------//
     case TUSB_REQ_RCPT_DEVICE:
+      if ( TUSB_REQ_TYPE_CLASS == p_request->bmRequestType_bit.type )
```
hathach, line 515: great !! So far there is no **Class requests toward Device** just yet.

src/device/usbd.h
```diff
@@ -435,6 +435,63 @@ TU_ATTR_WEAK bool tud_vendor_control_complete_cb(uint8_t rhport, tusb_control_re
...
+  7, TUSB_DESC_ENDPOINT, _ep_out, TUSB_XFER_BULK, U16_TO_U8S_LE(_ep_size), 1
+
+#define TUD_BTH_ISO_ITF(_itfnum, _alt, _ep_in, _ep_out, _n) ,\
+  /* Interface with 2 endpoints */ \
+  9, TUSB_DESC_INTERFACE, _itfnum, _alt, 2, TUD_BT_APP_CLASS, TUD_BT_APP_SUBCLASS, TUD_BT_PROTOCOL_PRIMARY_CONTROLLER, 0, \
+  /* Isochronous endpoints */ \
+  7, TUSB_DESC_ENDPOINT, _ep_in, TUSB_XFER_ISOCHRONOUS, U16_TO_U8S_LE(_n), 1, \
+  7, TUSB_DESC_ENDPOINT, _ep_out, TUSB_XFER_ISOCHRONOUS, U16_TO_U8S_LE(_n), 1
+
+#define TU_CONCAT_(a, b) a ## b
```
hathach, line 468: Great macro expanding for ISO interface, it is getting big now. I may find another place to move them, probably to the driver header later on :).

There is string cat and expaned and cat `TU_STRCAT/TU_XSTRCAT` already here.
https://github.com/hathach/tinyusb/blob/master/src/common/tusb_compiler.h#L37

src/class/bth/bth_device.c
```diff
@@ -0,0 +1,238 @@
...
+
+      desc_ep = (tusb_desc_endpoint_t const *)tu_desc_next(desc_ep);
+      dir = tu_edpt_dir(desc_ep->bEndpointAddress);
+      // Verify that alternative endpoint are same as first ones
+      TU_ASSERT(desc_ep->bDescriptorType == TUSB_DESC_ENDPOINT &&
+                _btd_itf.ep_voice[dir] == desc_ep->bEndpointAddress);
+      _btd_itf.ep_voice_size[dir][i] = (uint8_t) desc_ep->wMaxPacketSize.size;
+    }
+    // Consume all alternatives descriptors
+    (*p_length) += CFG_TUD_BTH_ISO_ALT_COUNT * (sizeof(tusb_desc_interface_t) + 2 * sizeof(tusb_desc_endpoint_t));
```
hathach, line 172: although most users will use descriptor template, but it is totally up to application to define their own configuration descriptors, the `CFG_TUD_BTH_ISO_ALT_COUNT` should be treated as **maximum** number of ALT instead of **exact** number, primarily for storing alternate EPSize. We should still count the actual dynamic alternate interface. **It is probably better to increase `p_length` within the above loop()**

hathach:
@kasjer thanks to your details instruction, I have able to tested with blehci, and got it working on the first try with my Linux. Windows Bluetooth stack is known (at least to me) lacking behind Linux/macOS support so I am not too worried about it for now. We will sort it out later on :) 

I have also added blehci to my newt example repo as well, to save time for people wanting to test it out https://github.com/hathach/mynewt-tinyusb-example/pull/8

kasjer:
@hathach thanks for checking it out.
requested changes on the way...

src/class/bth/bth_device.c
kasjer: I understand your concern.
There is one problem though.
If ```CFG_TUD_BTH_ISO_ALT_COUNT``` was a maximum ```btd_open``` would need to check during runtime if descriptor is valid and should be used to increase ```p_length```.
In this case however function does not know how much memory to check since there is not upper limit provided. So it may end up checking memory that does not contain configuration descriptor.
If open function had limit it would be enough.
I guess that other interfaces do not have this dynamic size nature.

On the other hand if alternate interfaces were handled in function that calls open it could be also covered.

src/class/bth/bth_device.c
```diff
@@ -0,0 +1,238 @@
...
+    TU_ASSERT(desc_ep->bDescriptorType == TUSB_DESC_ENDPOINT);
+    dir = tu_edpt_dir(desc_ep->bEndpointAddress);
+    _btd_itf.ep_voice[dir] = desc_ep->bEndpointAddress;
+    // Store endpoint size for alternative 0
+    _btd_itf.ep_voice_size[dir][0] = (uint8_t) desc_ep->wMaxPacketSize.size;
+
+    for (int i = 1; i < CFG_TUD_BTH_ISO_ALT_COUNT; ++i) {
+      // Make sure rest of alternatives matches
+      itf_desc = (tusb_desc_interface_t const *)tu_desc_next(desc_ep);
+      TU_ASSERT(itf_desc->bDescriptorType == TUSB_DESC_INTERFACE);
```
hathach, line 152: @kasjer I the ATL interface will be back to back, we could break whenever the descriptor type is not a Interface type, or alternate number is zero (next driver interface).

src/class/bth/bth_device.c
hathach: @kasjer ah, I see, the driver open() should pass the remaining length left from config descriptor to the open(). Don't worry about that for now, I will create an separated issue for it.

src/class/bth/bth_device.h
kasjer: done

src/class/bth/bth_device.c
kasjer: done

src/class/bth/bth_device.c
kasjer: done

src/class/bth/bth_device.h
kasjer: done

src/device/usbd.h
kasjer: TU_CONCAT removed

src/class/bth/bth_device.c
kasjer: Done, waits to fix for out of bounds access

src/class/bth/bth_device.h
hathach: wow, this is perfect,and  much more than I expected, only want a couple more of description lines.

src/class/bth/bth_device.c
hathach: implemented by pr #418 and merge, please pull from master to update

kasjer: Update to reflect changes in #418

PR #414, opened 2020-05-20
added tud_task_event_ready() 

hathach:
**Describe the PR**
added tud_task_event_ready() to check if there is pending events in the tud task without executing
it. Useful to check before entering low power mode with WFI/WFE. This helps to resovle https://github.com/adafruit/circuitpython/pull/2868 

This requires an additional `osal_queue_empty()` to be implemented.

**Additional context**

TockOS, like a lot of other operating systems divides interrupt processing in two halves. The top-half runs in interrupt context, which is mapped to ARMv7-M's handler mode. The bottom-half processing happens in kernel context, which is mapped to ARMv7-M's privileged thread mode. 

The main USB interrupt processing happens in the bottom half, and within the bottom half we add to TinyUSB task queue. Therefore we can call `wfi`, _only when_ both TinyUSB task queue is empty **and** no more bottom half interrupt processing is left to do. 

Hence the need for `tud_task_is_queue_empty()` function.

_Originally posted by @rajivr in https://github.com/hathach/tinyusb/issues/279#issuecomment-583957143_

### DIFF
--- a/src/osal/osal_none.h
+++ b/src/osal/osal_none.h
@@ -142,7 +142,7 @@ typedef osal_queue_def_t* osal_queue_t;
     }\
   }
 
-// lock queue by disable usb isr
+// lock queue by disable USB interrupt
 static inline void _osal_q_lock(osal_queue_t qhdl)
 {
   (void) qhdl;
@@ -176,7 +176,6 @@ static inline osal_queue_t osal_queue_create(osal_queue_def_t* qdef)
   return (osal_queue_t) qdef;
 }
 
-// non blocking
 static inline bool osal_queue_receive(osal_queue_t const qhdl, void* data)
 {
   _osal_q_lock(qhdl);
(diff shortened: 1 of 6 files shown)


src/osal/osal_none.h
```diff
@@ -203,6 +202,15 @@ static inline bool osal_queue_send(osal_queue_t const qhdl, void const * data, b
   return success;
 }
 
+static inline bool osal_queue_empty(osal_queue_t qhdl)
+{
+  _osal_q_lock(qhdl);
```
xobs, line 207: This approach won't work.

The user needs to call `tud_task_event_ready()` with interrupts already disabled in order to check whether the queue is empty. If the queue is empty, then it can issue `WFI`. If not, it will re-enable interrupts and not issue `WFI`.

I'm uncertain what happens if you issue `cpsid i` when interrupts are already disabled, but this function will re-enable them when it exits. If a USB event comes in between the time this function exits and when the calling function issues `WFI`, then that interrupt will be handled and the queue will no longer be empty, but the calling function will still issue `WFI`.

This function needs to be called with interrupts already disabled, and shouldn't lock/unlock the system on its own.

hathach: @xobs Yeah, you are right, I was thinking to have this function being used in normal/general mode as well. However, there is not much benefit in normal mode, also locking the queue is not crucial as well in that mode.
  
NVIC enable probably shouldn't have any impact on the global CPSID, but I could be wrong. I   will remove the queue lock/unlock anyway since it is better to do.

hathach:
thanks @xobs for reviewing

PR #427, opened 2020-06-09
Support for DA1469x MCU from Dialog Semiconductor

kasjer:
**Describe the PR**
Code adds DCD for Bluetooth Low Energy MCU **DA1469x** that has USB full speed device controller.
https://www.dialog-semiconductor.com/products/da14695-development-kit-pro
Code was tested to support control interrupt and bulk endpoints.
ISO endpoints may work but were not tested yet.

**Additional context**
It may be that code is compatible with previous version of Dialog chips **DA1468x** but that was not tested.
Dialog provides SDK that has support for USB (with Segger stack).
Most of files that are available in SDK have license that prevents their inclusion in outside repositories.
Header file with register description has Apache license and was added here.
Driver uses only this public header.

Since SDK files can't be included here and SDK is available as download only (not git access) it makes it hard for board package to be made (startup files, clocks gpio would need to be written from scrtach).
Additionally DA1469x requires special way of flashing board (product header binary file, documented in datasheet is needed apart of binary file with application). Dialog SDK has command line tool to flash images but they are also proprietary.

Support for Dialog's development kit (with DA1469x MCU) is available in Apache Mynewt with full source code and all tools ready to build and flash application.

PR with Mynewt side of support for this code ~~will be available soon~~ https://github.com/apache/mynewt-core/pull/2310


hathach:
Thank you @kasjer for another great PR. I have just go through the registration and download dialog sdk. It is indeed tons of "Confidential, Proprietary Information" files. This is actually quite similar to Nordic when they started nrf5 series with nrf51, Nordic even go further to require enter Serial Number of the devkit to be able to download the sdk. Eventually when the MCU get more popular, they will change the license term so I don't see any issues with it.

> Since SDK files can't be included here and SDK is available as download only (not git access) it makes it hard for board package to be made (startup files, clocks gpio would need to be written from scrtach).
Additionally DA1469x requires special way of flashing board (product header binary file, documented in datasheet is needed apart of binary file with application). Dialog SDK has command line tool to flash images but they are also proprietary.
>
> Support for Dialog's development kit (with DA1469x MCU) is available in Apache Mynewt with full source code and all tools ready to build and flash application.

Since MyNewt already has an open source friendly low level driver and tool, we have no reasons to use the DA sdk files at all. I still want to see a BSP for bare metal build and includes it in the CI, which is needed for regression test and maintain in the future. I will order an DA devkit probably the [budget friendly one](https://www.dialog-semiconductor.com/products/da14695-development-kit-usb) to test it out later on as well.

Therefore, Is it possible that you could include some of the `da1469x_abc.c` from MyNewt to the hw/mcu folder with the same structure. Those files seems to be independent enough, we could hardcode `MYNEWT_VAL()` and stub out `os_trace` in a separated files. The ultimate goal is getting the clock setup and the GPIO for LED / Button, UART is optional, TinyUSB doesn't need other fancy peripherals.

kasjer:
> Therefore, Is it possible that you could include some of the `da1469x_abc.c` from MyNewt to the hw/mcu folder with the same structure. Those files seems to be independent enough, we could hardcode `MYNEWT_VAL()` and stub out `os_trace` in a separated files or in `board.mk`. The ultimate goal is getting the clock setup and the GPIO for LED / Button, UART is optional, TinyUSB doesn't need other fancy peripherals.

I'll try to put something together...

hathach:
that would be perfect, best to keep the same file orgs to make syncing with mynewt easier later on. You could just left out all the hal_ API, but you know mynewt much better than me :)

hathach:
the usb devkit https://www.dialog-semiconductor.com/products/da14695-development-kit-usb just arrived on my desk so I could help testing as well. You don't have to get it all working, just make an WIP, I could help with testing/adding bsp as well :)

kasjer:
I was busy with something else for a last couple of days. I may start this bsp tomorrow if nothing comes up.

hathach:
No hurry, takes your time, I only want to say I am able to help with testing as well :)

kasjer:
Code was tested with cdc_msc example project only.

There is one Tiny problem that requires one tool from **Dialog SDK**.
In short one can not just build elf file convert it to bin or hex and flash it due to security feature that is used in MCU.
Tool `mkimage` from SDK is needed. It produces image file that could be written to flash at address 0x2000, but for some reason on board jlink when instructed to write this image file, also erases flash at address 0x0000 which contains another header needed for booting.
This PR adds ``product_header.dump`` file that needs to be concatenated with image produced by mkimage. Make file covers usage of mkimage provided it is in PATH.

Before trying to flash board I advise to dump beginning of flash from new board in jlink.
```
savebin product_header.dump 16000000 2000
```
Address 0x16000000 is real flash address that contains product header.

In case of accident Smart Snippets studio can be used to flash things to board even if there is no valid product header (which it will rebuild).

kasjer:
I tried also on USB dongle version and it is not that good.
For some reason jlink fails to flash image.
This MCU uses external SPI flash and dongle has different chip that was used in pro-dk.
So this maybe the factor.

Image produced by ``make`` is correct and can be flashed to board with cli_programmer from Dialog SDK.
```
make BOARD=dialog_da1469x-dk-pro flash
```
So if above line fails, image can be written to flash using UART.
```
cli_programmer /dev/ttyACM0 write_qspi 0 _build/build-dialog_da1469x-dk-pro/dialog_da1469x-dk-pro-image.bin
```
 **K2** button should be pressed when console prompts
```
Connecting to device...
Press RESET.
```

hathach:
@kasjer wow look like there is a bit of learning curve for this MCU. Give me a bit of time, I will try to catch up in the next week.

hathach:
> I tried also on USB dongle version and it is not that good.
> For some reason jlink fails to flash image.
> This MCU uses external SPI flash and dongle has different chip that was used in pro-dk.
> So this maybe the factor.

Indeed, I gave it a try with USB dongle version, and it couldn't flash the board, maybe the flash layout is different. I should have just order the dk pro. I will try again later to see if I could manage to get it flash 

```
****** Error: Error while determining flash info (Bank @ 0x16000000)
Unspecified error -1
```

hw/bsp/dialog_da1469x-dk-pro/board.mk
```diff
@@ -0,0 +1,61 @@
...
+    $(TOP)/hw/bsp/$(BOARD) \
+	$(TOP)/$(MCU_DIR)/include \
+	$(TOP)/$(MCU_FAMILY_DIR)/include \
+	$(TOP)/$(MCU_FAMILY_DIR)/SDK_10.0.8.105/sdk/bsp/include \
+
+# For TinyUSB port source
+VENDOR = dialog
+# While this is for da1469x chip, there is chance that da1468x chip family will also work
+CHIP_FAMILY = da146xx
+
```
hathach, line 40: please also add following line for freeRTOS ports, I have tested, and it seems that cdc_msc_freertos works well enough

```
# For freeRTOS port source
FREERTOS_PORT = ARM_CM33_NTZ/non_secure
```

kasjer:
give me a minute...

hathach:
> give me a minute...

All the time you need :smiley:

kasjer:
I think that problem with flashing board first is really JLink problem.
Board comes from Dialog with some software, this code runs with 96MHz PLL and this clock is used for accessing QSPI flash.
Problem is that some of the commands that JLink tries to execute are not allowed to run at such high clock.
Most obvious command is standard read command (03h) that usually has some clock restrictions and for flash that is on board it should be set to < 54MHz and Jlink does not change QSPI clock so most likely it can't read correctly flash content.

In example BSP, PLL is started (as it is needed for USB) but system clock and QSPI clock remain at 32MHz which is safe for flash.

hathach:
> I think that problem with flashing board first is really JLink problem.
> Board comes from Dialog with some software, this code runs with 96MHz PLL and this clock is used for accessing QSPI flash.
> Problem is that some of the commands that JLink tries to execute are not allowed to run at such high clock.
> Most obvious command is standard read command (03h) that usually has some clock restrictions and for flash that is on board it should be set to < 54MHz and Jlink does not change QSPI clock so most likely it can't read correctly flash content.
> 
> In example BSP, PLL is started (as it is needed for USB) but system clock and QSPI clock remain at 32MHz which is safe for flash.

Yeah, you know this better than me, maybe the qspi config in product header is for jlink to know which flash device, and/or the read mode quad/bi/single data etc ... To be honest, I will take anything that you tell me about this. I am glad that I could finally test it out on the usb devkit :smile:

PR #431, opened 2020-06-15
Fix IAR compiler warnings.

HiFiPhile:
**Describe the PR**
Some workarounds are used to prevent unaligned memory access of struct member, but they are not needed since both GCC and IAR generate correct instructions for packed struct.

For example`usbd.c` line 784, instead of
```
uint16_t total_len;
memcpy(&total_len, &desc_bos->wTotalLength, 2); // possibly mis-aligned memory

return tud_control_xfer(rhport, p_request, (void*) desc_bos, total_len);
```
We can simplely do:
```
return tud_control_xfer(rhport, p_request, (void*) desc_bos, desc_bos->wTotalLength);
```
Generated code:
```
return tud_control_xfer(rhport, p_request, (void*) desc_bos, desc_bos->wTotalLength);
    80027fa:   7881        ldrb    r1, [r0, #2]
    80027fc:   78c3        ldrb    r3, [r0, #3]
    80027fe:   021b        lsls    r3, r3, #8
    8002800:   430b        orrs    r3, r1
    8002802:   0021        movs    r1, r4
    8002804:   0028        movs    r0, r5
    8002806:   f7ff fb1d   bl  8001e44 <tud_control_xfer>
```

In this way we can  make a little optimization and prevent  a IAR warning `Pa039 : use of address of unaligned structure member`

### DIFF
--- a/src/class/hid/hid_device.c
+++ b/src/class/hid/hid_device.c
@@ -194,7 +194,9 @@ uint16_t hidd_open(uint8_t rhport, tusb_desc_interface_t const * desc_itf, uint1
 
   p_hid->boot_mode = false; // default mode is REPORT
   p_hid->itf_num   = desc_itf->bInterfaceNumber;
-  memcpy(&p_hid->report_desc_len, &(p_hid->hid_descriptor->wReportLength), 2);
+  
+  // Use offsetof to avoid pointer to the odd/misaligned address
+  memcpy(&p_hid->report_desc_len, (uint8_t*) p_hid->hid_descriptor + offsetof(tusb_hid_descriptor_hid_t, wReportLength), 2);
 
   // Prepare for output endpoint
   if (p_hid->ep_out)
--- a/src/device/usbd.c
+++ b/src/device/usbd.c
@@ -437,7 +437,7 @@ void tud_task (void)
 
         if ( 0 == epnum )
         {
-          usbd_control_xfer_cb(event.rhport, ep_addr, event.xfer_complete.result, event.xfer_complete.len);
+          usbd_control_xfer_cb(event.rhport, ep_addr, (xfer_result_t)event.xfer_complete.result, event.xfer_complete.len);
         }
         else
         {
@@ -445,7 +445,7 @@ void tud_task (void)
           TU_ASSERT(drv_id < USBD_CLASS_DRIVER_COUNT,);
 
           TU_LOG2("  %s xfer callback\r\n", _usbd_driver[drv_id].name);
-          _usbd_driver[drv_id].xfer_cb(event.rhport, ep_addr, event.xfer_complete.result, event.xfer_complete.len);
+          _usbd_driver[drv_id].xfer_cb(event.rhport, ep_addr, (xfer_result_t)event.xfer_complete.result, event.xfer_complete.len);
         }
       }
       break;
@@ -846,8 +846,10 @@ static bool process_get_descriptor(uint8_t rhport, tusb_control_request_t const 
       if (!tud_descriptor_bos_cb) return false;
 
       tusb_desc_bos_t const* desc_bos = (tusb_desc_bos_t const*) tud_descriptor_bos_cb();
+
       uint16_t total_len;
-      memcpy(&total_len, &desc_bos->wTotalLength, 2); // possibly mis-aligned memory
+      // Use offsetof to avoid pointer to the odd/misaligned address
+      memcpy(&total_len, (uint8_t*) desc_bos + offsetof(tusb_desc_bos_t, wTotalLength), 2);
 
       return tud_control_xfer(rhport, p_request, (void*) desc_bos, total_len);
     }
@@ -861,7 +863,8 @@ static bool process_get_descriptor(uint8_t rhport, tusb_control_request_t const 
       TU_ASSERT(desc_config);
 
       uint16_t total_len;
-      memcpy(&total_len, &desc_config->wTotalLength, 2); // possibly mis-aligned memory
+      // Use offsetof to avoid pointer to the odd/misaligned address
+      memcpy(&total_len, (uint8_t*) desc_config + offsetof(tusb_desc_configuration_t, wTotalLength), 2);
 
       return tud_control_xfer(rhport, p_request, (void*) desc_config, total_len);
     }
@@ -915,8 +918,6 @@ static bool process_get_descriptor(uint8_t rhport, tusb_control_request_t const 
 
     default: return false;
   }
-
-  return true;
 }
 
 //--------------------------------------------------------------------+
--- a/src/class/msc/msc_device.c
+++ b/src/class/msc/msc_device.c
@@ -80,7 +80,8 @@ static inline uint32_t rdwr10_get_lba(uint8_t const command[])
 
   // copy first to prevent mis-aligned access
   uint32_t lba;
-  memcpy(&lba, &p_rdwr10->lba, 4);
+  // use offsetof to avoid pointer to the odd/misaligned address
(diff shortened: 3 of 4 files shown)


HiFiPhile:
Need test of TI Compiler.

HiFiPhile:
Hi,

IMO the only problem is when we use the unaligned address as a pointer.

Could you give an example how it failed ?

HiFiPhile:
Or what about add a macro in `tusb_compiler.h`, which looks like
```
#define tu_unaligned_read(dst, src, len) (memcpy(dst, src, len))
```
So we can customize the behavior for different compilers / cores ?

hathach:
> Hi,
> 
> IMO the only problem is when we use the unaligned address as a pointer.
> 
> Could you give an example how it failed ?

SCSI command is odd align depending on command. Like I said there is no promise from compiler, and even it does, I don't trust it.

hathach:
> Or what about add a macro in `tusb_compiler.h`, which looks like
> 
> ```
> #define tu_unaligned_read(dst, src, len) (memcpy(dst, src, len))
> ```
> 
> So we can customize the behavior for different compilers / cores ?

I don't see any benefit to do that, I prefer one less thing to worry about.

HiFiPhile:
> I don't see any benefit to do that.

Personally it's for IAR compiler, since the behavior has been tested for years. 

In order to suppress the warning `Pa039 : use of address of unaligned structure member` **within** the scope of tinyusb, curretly the only way is put precompiler macros in the source file, which is ugly.

hathach:
I see it now, you should stay clearly that the main purpose of this PR is to suppress the IAR warning. hmm, I am not familiar with IAR, why did it raise that warning, and on which line ?

HiFiPhile:
IAR emit a warning when we take the address of an unaligned  struct member.

For example : 
```
memcpy(&p_hid->report_desc_len, &(p_hid->hid_descriptor->wReportLength), 2);
```
The compiler is not smart enough to know that memcpy() can handle pointer to unaligned object.

hathach:
can you try to explicitly cast `(void*) &(p_hid->hid_descriptor->wReportLength)` to see if that fixes the problem.

HiFiPhile:
> can you try to explicitly cast (void*) &(p_hid->hid_descriptor->wReportLength) to see if that fixes the problem.

Unfortunatelly this doesn't work. At the same time I can ask IAR why memcpy is not whitelisted.

HiFiPhile:
A workaround is using the `offsetof` function:
`  memcpy(&p_hid->report_desc_len, (uint8_t*)p_hid->hid_descriptor + offsetof(tusb_hid_descriptor_hid_t, wReportLength), 2);
`

hathach:
> A workaround is using the `offsetof` function:
> ` memcpy(&p_hid->report_desc_len, (uint8_t*)p_hid->hid_descriptor + offsetof(tusb_hid_descriptor_hid_t, wReportLength), 2);`

I won't make this change over IAR warning, that is a surprise that casting to void* doesn't help, maybe trying to cast it to (uint8_t*) would help, since it is byte access.

HiFiPhile:
> maybe trying to cast it to (uint8_t*) would help

It doesn't help either.

hathach:
how about declare an separated pointer `uint8_t* temp = ...` if this doesn't work, then I don't really understand how IAR expect us to reference to the packed struct. There is lots of packed struct with possible odd address ref throughout usb specs in class driver.

HiFiPhile:
> how about declare an separated pointer uint8_t* temp = ...

No chance...

Here is a similar situation in LwIP, they fixed it by using the value instead of the pointer.
http://savannah.nongnu.org/bugs/?54006
http://git.savannah.nongnu.org/cgit/lwip.git/commit/?id=288d3c2802e53ab42464a917e9fd8712f0d58b29

Since LwIP is much widely used than tinyusb, and they didn't have problem in 2 years, I believe this change is safe.

hathach:
I think for GCC casting to (void*) does seem to suppress the warning. memcpy does just that as side effect. 

> Here is a similar situation in LwIP, they fixed it by using the value instead of the pointer.
> http://savannah.nongnu.org/bugs/?54006
> http://git.savannah.nongnu.org/cgit/lwip.git/commit/?id=288d3c2802e53ab42464a917e9fd8712f0d58b29
> 
> Since LwIP is much widely used than tinyusb, and they didn't have problem in 2 years, I believe this change is safe.

That is interesting that lwip choose to do that, but It won't change my opinion. Since this could cause very headache hard-fault with mis-algined memory access on particular platform which is annoying and very hard to troubleshoot. `offsetof()` is much better workaround, yeah, if there is no other way to suppress this, then I am ok with using `offsetof()` at least everything is the same, the generated code should be no different from current one. If you don't mind, please update the PR.

HiFiPhile:
I've updated the PR using `offsetof` workaround.

And cast `event.xfer_complete.result` explicitly to xfer_result_t to get rid of `Warning[Pe188]: enumerated type mixed with another type`

HiFiPhile:
There are still 2 warnings of `statement is unreachable`:
In dcd_stm32_fsdev.c line 701, execution is blacked by TU_ASSERT:
```
  default:
    TU_ASSERT(false);
    return false;
```
In usbd.c line 920, prior code has already returned.
```

    default: return false;
  }

  return true;
```
I don't know if we can do something ?

pigrew:
That's my bad habit in stm32 fsdev. I wanted the assert to get the debug/breakpoint, but the compiler wants a forced return. I think we could do:

```TU_ASSERT(false, /**/ );
return true;
```

HiFiPhile:
> TU_ASSERT(false, /**/ );
return true;

Doesn't work, with error:
`Error[Pe117]: non-void function "dcd_edpt_open" should return a value `

src/class/hid/hid_device.c
```diff
@@ -194,7 +194,9 @@ uint16_t hidd_open(uint8_t rhport, tusb_desc_interface_t const * desc_itf, uint1
 
   p_hid->boot_mode = false; // default mode is REPORT
   p_hid->itf_num   = desc_itf->bInterfaceNumber;
-  memcpy(&p_hid->report_desc_len, &(p_hid->hid_descriptor->wReportLength), 2);
+  
+  // Suppress IAR warning Pa039
```
hathach, line 198: it can be tricky for people that doesn't use IAR. Maybe saying "Use offsetof to avoid pointer to the odd/misaligned address" would be better. I am not too picky, but we should have a space between castting and its operand e.g ` (uint8_t*) p_hid->hid_descriptor`

src/device/usbd.c
```diff
@@ -859,9 +860,9 @@ static bool process_get_descriptor(uint8_t rhport, tusb_control_request_t const
 
       tusb_desc_configuration_t const* desc_config = (tusb_desc_configuration_t const*) tud_descriptor_configuration_cb(desc_index);
       TU_ASSERT(desc_config);
-
+
```
hathach, line 863: please remove unnecessary indent spaces

hathach:
> There are still 2 warnings of `statement is unreachable`:
> In dcd_stm32_fsdev.c line 701, execution is blacked by TU_ASSERT:
> 
> ```
>   default:
>     TU_ASSERT(false);
>     return false;
> ```

`TU_ASSERT(false)` already return false, we can safely remove the `return false` after it.

> In usbd.c line 920, prior code has already returned.
> 
> ```
> 
>     default: return false;
>   }
> 
>   return true;
> ```

every cases of the switch have return on its every if/else. The last `return true;` after the switch can be safely remove.

HiFiPhile:
I've updated the PR.

PR #433, opened 2020-06-17
Disable SOF in dcd_stm32_fsdev

HiFiPhile:
**Describe the PR**
Disable SOF in dcd_stm32_fsdev
Since TinyUSB doesn't use SOF for now, and this interrupt too often (1ms interval)

### DIFF
--- a/src/portable/st/stm32_fsdev/dcd_stm32_fsdev.c
+++ b/src/portable/st/stm32_fsdev/dcd_stm32_fsdev.c
@@ -144,6 +144,12 @@
 #  define DCD_STM32_BTABLE_LENGTH (PMA_LENGTH - DCD_STM32_BTABLE_BASE)
 #endif
 
+// Since TinyUSB doesn't use SOF for now, and this interrupt too often (1ms interval)
+// We disable SOF for now until needed later on
+#ifndef USE_SOF
+#  define USE_SOF     0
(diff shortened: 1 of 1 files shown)


hathach:
Look good to me, but I will leave the review for @pigrew 😄

src/portable/st/stm32_fsdev/dcd_stm32_fsdev.c
```diff
@@ -144,6 +144,10 @@
 #  define DCD_STM32_BTABLE_LENGTH (PMA_LENGTH - DCD_STM32_BTABLE_BASE)
 #endif
 
+// Since TinyUSB doesn't use SOF for now, and this interrupt too often (1ms interval)
+// We disable SOF for now until needed later on
+#define USE_SOF     0
```
gh2o, line 149: Like the `#define`s above, I'd recommend putting a `#ifndef` around this.

hathach: actually, we don't plan to allow application to override this value. If there is it would follow the `CFG_TUD_*` naming, so it is good as it is.

HiFiPhile: done.

PR #434, opened 2020-06-18
Add TX callback to CDC device

gh2o:
Useful for continuous transmission of data, which is difficult currently because there is no notification of tx completion.

**Describe the PR**
Transmitting data continuously via CDC is difficult because there is no way to know when the TX FIFO has room for more data. This PR adds a callback to allow the TX FIFO to be refilled by the user.

### DIFF
--- a/src/class/cdc/cdc_device.c
+++ b/src/class/cdc/cdc_device.c
@@ -416,7 +416,12 @@ bool cdcd_xfer_cb(uint8_t rhport, uint8_t ep_addr, xfer_result_t result, uint32_
   //       Though maybe the baudrate is not really important !!!
   if ( ep_addr == p_cdc->ep_in )
   {
-    if ( 0 == tud_cdc_n_write_flush(itf) )
+    uint32_t flushed = tud_cdc_n_write_flush(itf);
+
+    // invoke transmit callback to possibly refill tx fifo
+    if ( tud_cdc_tx_complete_cb ) tud_cdc_tx_complete_cb(itf);
+
+    if ( 0 == flushed && tu_fifo_empty(&p_cdc->tx_ff) )
     {
       // There is no data left, a ZLP should be sent if
       // xferred_bytes is multiple of EP size and not zero
@@ -425,9 +430,6 @@ bool cdcd_xfer_cb(uint8_t rhport, uint8_t ep_addr, xfer_result_t result, uint32_
         usbd_edpt_xfer(TUD_OPT_RHPORT, p_cdc->ep_in, NULL, 0);
       }
     }
-
-    // invoke transmit callback to possibly refill tx fifo
-    if ( tud_cdc_tx_cb && !tu_fifo_full(&p_cdc->tx_ff) ) tud_cdc_tx_cb(itf);
   }
 
   // nothing to do with notif endpoint for now
--- a/src/class/cdc/cdc_device.c
+++ b/src/class/cdc/cdc_device.c
@@ -416,9 +416,6 @@ bool cdcd_xfer_cb(uint8_t rhport, uint8_t ep_addr, xfer_result_t result, uint32_
   //       Though maybe the baudrate is not really important !!!
   if ( ep_addr == p_cdc->ep_in )
   {
-    // invoke transmit callback to possibly refill tx fifo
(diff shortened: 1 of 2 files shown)


src/class/cdc/cdc_device.c
```diff
@@ -416,6 +416,9 @@ bool cdcd_xfer_cb(uint8_t rhport, uint8_t ep_addr, xfer_result_t result, uint32_
   //       Though maybe the baudrate is not really important !!!
   if ( ep_addr == p_cdc->ep_in )
   {
+    // invoke transmit callback to possibly refill tx fifo
+    if ( tud_cdc_tx_cb && !tu_fifo_full(&p_cdc->tx_ff) ) tud_cdc_tx_cb(itf);
```
hathach, line 420: This is good idea, though I think we shouldn't invoke every transfer complete. Maybe invoke it when the TX FIFO is empty, which is just inside the below if() for Zero length packet. We should also rename it to `tud_cdc_tx_empty_cb()`

gh2o:
I was thinking of calling it whenever the TX FIFO has space because that way we could maximize bandwidth by ensuring that the TX FIFO is never empty. Though for that to work I guess I would have to put the `tud_cdc_tx_cb` call after `tud_cdc_n_write_flush`. What do you think?

gh2o:
I've updated the PR to call `tud_cdc_tx_cb` after `tud_cdc_n_write_flush`, as it makes more sense for it to run after there is more space in the tx fifo. Please let me know what you think!

src/class/cdc/cdc_device.c
```diff
@@ -425,6 +425,9 @@ bool cdcd_xfer_cb(uint8_t rhport, uint8_t ep_addr, xfer_result_t result, uint32_
         usbd_edpt_xfer(TUD_OPT_RHPORT, p_cdc->ep_in, NULL, 0);
       }
     }
+
+    // invoke transmit callback to possibly refill tx fifo
+    if ( tud_cdc_tx_cb && !tu_fifo_full(&p_cdc->tx_ff) ) tud_cdc_tx_cb(itf);
```
hathach, line 430: 1. This should be called after the above flush, else we could send ZLP while this fill up the fifo which is not ideal.
2. To make it clear, we should rename the callback to `tud_cdc_tx_complete_cb()`

me-no-dev:
Looking forward to this!

hathach:
@gh2o Let's me know if you have time to wrap up the PR, else I could help with the changes. This callback maybe handy for many projects.

gh2o:
@hathach I've made the requested changes. Looking forward to getting them incorporated!

src/class/cdc/cdc_device.c
```diff
@@ -416,7 +416,12 @@ bool cdcd_xfer_cb(uint8_t rhport, uint8_t ep_addr, xfer_result_t result, uint32_
   //       Though maybe the baudrate is not really important !!!
   if ( ep_addr == p_cdc->ep_in )
   {
-    if ( 0 == tud_cdc_n_write_flush(itf) )
+    uint32_t flushed = tud_cdc_n_write_flush(itf);
+
+    // invoke transmit callback to possibly refill tx fifo
+    if ( tud_cdc_tx_complete_cb ) tud_cdc_tx_complete_cb(itf);
+
+    if ( 0 == flushed && tu_fifo_empty(&p_cdc->tx_ff) )
```
hathach, line 424: This is still not ideal since we may not transfer all available bytes e.g flush send out 5 bytes, but the complete_cb() fill more 200 bytes just a few statement later. Ideally it should be 

```
  if ( ep_addr == p_cdc->ep_in )
  {
    // invoke transmit callback to possibly refill tx fifo
    if ( tud_cdc_tx_complete_cb ) tud_cdc_tx_complete_cb(itf);

    if ( 0 == tud_cdc_n_write_flush(itf) )
    {
      // There is no data left, a ZLP should be sent if
      // xferred_bytes is multiple of EP size and not zero
      if ( xferred_bytes && (0 == (xferred_bytes % CFG_TUD_CDC_EP_BUFSIZE)) )
      {
        usbd_edpt_xfer(TUD_OPT_RHPORT, p_cdc->ep_in, NULL, 0);
      }
    }
  }
```

hathach (changes requested):
Thanks for updating the PR, there is still one request left though to optimize the USB transfer.

PR #442, opened 2020-06-24
Update NRFX to v2.2.0

czeslawmakarski:
Updates the NRFX from Nordic Semiconductor to the version v2.2.0


hathach:
thanks, it looks good, I quickly check nrfx changelog, it shoudn't break anything. Though just to be safe, have you tested the nrfx 2.2 on actual board ?

czeslawmakarski:
Yes, I did test it with pca10056 (nRF52840) board. Also, the reason for this PR is that it shall enable to use the nRF52820 chip (nRF52833 copy with less memory) with tinyusb.

PR #439, opened 2020-06-21
Add stm hs

UweBonnes:
**Describe the PR**
Add STM32F7xxdisco related BSP as a start

**Additional context**
If applicable, add any other context about the PR and/or screenshots here.

PR #447, opened 2020-06-29
Add support for STM32 OTG HS core

### LINKED ISSUE #198: Port STM32 Highspeed core
**Description**
It is more likely a test than port, probably adjusting base address etc to get the OTG HS core work with the stack. Need to find an dev kit that support OTG HS connection.

board with OTG HS connection 
- [x] STM32F723E Discovery with builtin HS PHY
- [x] STM32F769I Discovery require external PHY for OTG HS
- [x] STM32H743Eval with external PHY for OTG HS

hathach:
**Describe the PR**
This PR add support for stm32 otg hs core, fix #198  board with OTG HS connection for testing.
 
- [x] STM32F723E Discovery with builtin HS PHY
- [x] STM32F769I Discovery require external PHY for OTG HS
- [x] STM32H743Eval with external PHY for OTG HS

**Additional context**
~~#439 update this add-stm-hs branch to support more boards and core with internal HS PHY.  It is best to merge 439 first to update this PR before merging.~~ merged already

@UweBonnes @cr1901 @duempel


hathach:
PR is ready for review and merged, tested and work with all the OTG HS board I have above. Please let me know if you have any feedback on the changes @duempel @UweBonnes @cr1901

src/device/usbd.c
```diff
@@ -53,6 +54,8 @@ typedef struct {
     uint8_t self_powered          : 1; // configuration descriptor's attribute
   };
 
+  uint8_t speed;
```
duempel, line 57: Maybe using `tusb_speed_t` type would fit a little better here?

src/portable/st/synopsys/dcd_synopsys.c
```diff
@@ -149,134 +203,282 @@ static void bus_reset(void) {
...
+  dev->DCFG |= (bitvalue << USB_OTG_DCFG_DSPD_Pos);
 }
 
+#if defined(USB_HS_PHYC) && TUD_OPT_HIGH_SPEED
+static bool USB_HS_PHYCInit(void)
+{
+  USB_HS_PHYC_GlobalTypeDef *usb_hs_phyc = (USB_HS_PHYC_GlobalTypeDef*) USB_HS_PHYC_CONTROLLER_BASE;
+
+  // Enable LDO
+  usb_hs_phyc->USB_HS_PHYC_LDO |= USB_HS_PHYC_LDO_ENABLE;
```
duempel, line 321: Which documentation have you been used for referencing USBPHYC LDO register? I was checking with ST's [RM0431](https://www.st.com/resource/en/reference_manual/dm00305990-stm32f72xxx-and-stm32f73xxx-advanced-armbased-32bit-mcus-stmicroelectronics.pdf) and can only see a `LDO_DISABLE` bit.

src/portable/st/synopsys/dcd_synopsys.c
```diff
@@ -149,134 +203,282 @@ static void bus_reset(void) {
...
+    case 25000000: phyc_pll = USB_HS_PHYC_PLL1_PLLSEL_25MHZ   ; break;
+    case 32000000: phyc_pll = USB_HS_PHYC_PLL1_PLLSEL_Msk     ; break; // Value not defined in header
+    default:
+      TU_ASSERT(0);
+  }
+  usb_hs_phyc->USB_HS_PHYC_PLL = phyc_pll;
+
+  // Control the tuning interface of the High Speed PHY
+  // Use magic value from ST driver
+  usb_hs_phyc->USB_HS_PHYC_TUNE |= 0x00000F13U;
```
duempel, line 344: This seems to be very hardware specific. Maybe we can think about an option to give control about this to the user. But should be fine for this PR.

src/portable/st/synopsys/dcd_synopsys.c
```diff
@@ -149,134 +203,282 @@ static void bus_reset(void) {
...
-  (void) rhport;
-  NVIC_EnableIRQ(OTG_FS_IRQn);
+  NVIC_EnableIRQ(_dcd_rhport[rhport].irqnum);
 }
 
 void dcd_int_disable (uint8_t rhport)
 {
-  (void) rhport;
-  NVIC_DisableIRQ(OTG_FS_IRQn);
+  NVIC_EnableIRQ(_dcd_rhport[rhport].irqnum);
```
duempel, line 476: Copy mistake: Use `NVIC_DisableIRQ` instead.

src/portable/st/synopsys/dcd_synopsys.c
UweBonnes: The header have some inconsistant definition. Seee
https://community.st.com/s/question/0D53W00000BMA8HSAX/stm32f7xxh-strange-usbhsphycldoenable-definition

duempel: Thanks @UweBonnes forwarding this to ST community. If I trust in TRM this bit will disable the LDO. But if it's disabled instead of enabled why don't we get stuck in while loop when waiting for it to be ready? I hope ST can clearify this.

UweBonnes: I have not many question answered and hints noticed on the ST community forum. So my hope is not high.

src/portable/st/synopsys/dcd_synopsys.c
hathach: :facepalm: cannot believe I still made this silly mistake. Better to get more sleeps.

src/device/usbd.c
hathach: I am bit afraid the enum will cause the size of variable to 4 unnecessary

src/portable/st/synopsys/dcd_synopsys.c
hathach: I agree, I like to move any board specific such as VBUS sense into BSP layer. But I am new to this, and there isn't much document for how these bits affect the USB operation as well. So it is fine as it is until there is issue.

src/portable/st/synopsys/dcd_synopsys.c
```diff
@@ -49,40 +49,85 @@
...
+
 #else
   #error "Unsupported MCUs"
+
 #endif
 
 #include "device/dcd.h"
 
+//--------------------------------------------------------------------+
+//
```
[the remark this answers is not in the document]
hathach: Yeah right, I forgot this

src/portable/st/synopsys/dcd_synopsys.c
```diff
@@ -555,61 +767,62 @@ static void write_fifo_packet(uint8_t fifo_num, uint8_t * src, uint16_t len){
...
-      break;
+    break;
+
     case 0x06: // Setup packet recvd
-      {
-        uint8_t setup_left = ((out_ep[epnum].DOEPTSIZ & USB_OTG_DOEPTSIZ_STUPCNT_Msk) >> USB_OTG_DOEPTSIZ_STUPCNT_Pos);
+      // We can receive up to three setup packets in succession, but
+      // only the last one is valid.
+      _setup_packet[0] = (* rx_fifo);
+      _setup_packet[1] = (* rx_fifo);
```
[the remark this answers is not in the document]
hathach: Yes, your implementation is spot-on with the manual. However, since we only use the last setup packet and ignore the first two setup. I think we don't really need to store them, just overwrite it with each time we pull the setup packet from FIFO. Anything else still keeps the same, the STUPCNT is still 3. The semantic should be the same, it just we don't store the first 2 setup. Let's me know if I misunderstood the workflow, if yes then I am happy to revert this :)

hathach:
> @hathach I'm a bit confused- what is the purpose of `rhport` in the device API? I thought `rhport` was mostly a host API thing; is it also relevant when a board has multiple USB controllers on board? 

> I notice that you use a compile-time define to set which port is in use on STM32 boards. _Is it possible to use both ports at once with tinyusb, using the same driver for both?_
> 
> If it's _not_ possible to use both ports at once, why is `rhport` needed when the correct address and IRQ to use could be a compile-time constant*, and `rhport` is ignored as it is currently?
> 
> * This won't matter if LTO is enabled, but I'm interested in the reasoning anyway.

Ah you are right, It is more relevant to host stack than device yes, since one host stack can run on multiple ports (roothub or externa hub). However, there is still a use case where user want to have 2 separated devices running on 2 separated ports. e.g on STM32
- Port 0 running device streaming video/audio class driver and
- Port 1 running device with hid/cdc

For smaller project, just using a composite class in a single port makes more sense. To be honest, I am not sure if there will be any projects that needs to run 2 devices on 2 ports like that. Though if `rhport` is not part of the top-down API from beginning, it will be impossible later on to support this configuration without breaking the API.

That is the reason I am kind of always prefer using `rhport` than the macro `TUD_OPT_RPHORT` in general. Back to your point, yes, IRQ alone in the array doesn't make sense, we will need to also group EP_MAX/EP_FIFO_SIZE together as well. So yeah I will make the change to have IRQ as macro like EP_MAX/EP_FIFO_SIZE and we could worry about muliple ports on devices later on, as long as the API has rhport, any changes is transparent to application code.

src/device/usbd.c
duempel: Oh yeah, I always forget short enum are not gcc standard. Better keep it as uint8_t then.

src/portable/st/synopsys/dcd_synopsys.c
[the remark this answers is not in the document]
cr1901: >The semantic should be the same, it just we don't store the first 2 setup.

This sounds reasonable to me. If it breaks something later, or it turns out we did actually need the other 2 setup packets, we can revert.

cr1901:
@hathach Looks good to me after your changes and explanations. Approving.

src/portable/st/synopsys/dcd_synopsys.c
duempel: Yeah, seems they are not going to answer this question. I've checked the F723 discovery board schematic and could not find external power supply for HS peripheral. I don't have this hardware to test the LDO but if HS is working with this code the LDO is defenitely enabled since there is no exernal source provided by this board. We should keep it like this for the moment.

duempel:
@hathach maybe you missed the edit I wrote. How about **cdc_dual_ports** and **webusb_serial** exmaples? I'm not used to webusb but at least CDC should be changed to 512 bytes endpoint size in `usb_descriptos.c` for high speed I think.

hathach:
> @hathach maybe you missed the edit I wrote. How about **cdc_dual_ports** and **webusb_serial** exmaples? I'm not used to webusb but at least CDC should be changed to 512 bytes endpoint size in `usb_descriptos.c` for high speed I think.

ah yeah, sorry about that, I am totally missed that

duempel:
> ah yeah, sorry about that, I am totally missed that

No problem. Looks good now! 😊

hathach:
finally we could be able to merge this, thank you every for participating in the PR. Thanks @UweBonnes for adding internal HS PHY init and bunches of other BSPs, @duempel & @cr1901 as usual for inspecting & reviewing the chagnes.

Issue #448, opened 2020-06-29
Synopsys driver out endpoint handling broken by PR #399


kasjer:
**Set up**
 - **PC OS**   : Ubuntu 18.04 / Windows 10 (does not matter)
 - **Board**   : stm32f401blackpill (anything with synopsys)
 - **Firmware**: examples/device/cdc_msc

**Describe the bug**
#399 broke handling of OUT endpoints.
Code ignores data length in ***handle_epout_ints()*** and uses xfer->total_len even though packet was short,
later FIFO is filled with garbage based on this false total_len.
```c
          dcd_event_xfer_complete(0, n, xfer->total_len, XFER_RESULT_SUCCESS, true);
```

**To reproduce**
1. Build any CDC application
2. Hit enter to see `TinyUSB CDC MSC device example`
3. **Paste** something from clipboard or selection (on Linux)
4. What was copied should be visible fine
5. **Type** just one character, at this point previously copied data shows up again

hathach:
hmm, that is weird, I couldn't reproduce the issue with stm32f411disco board. Everything still work fine. I will try to find my 401 blackpill to test again.

hathach:
update: I didn't reproduce the exact behavior but found out another related bug that is pasting a long text will result in text got truncated in echo. E.g paste following text into the terminal (minicom on ubuntu 18.04) is fine pre PR #399 

```
// Stop transmitting packets and NAK IN xfers.
in_ep[epnum].DIEPCTL |= USB_OTG_DIEPCTL_SNAK;
while((in_ep[epnum].DIEPINT & USB_OTG_DIEPINT_INEPNE) == 0);

// Disable the endpoint.
in_ep[epnum].DIEPCTL |= (USB_OTG_DIEPCTL_STALL | USB_OTG_DIEPCTL_EPDIS);
while((in_ep[epnum].DIEPINT & USB_OTG_DIEPINT_EPDISD_Msk) == 0);
in_ep[epnum].DIEPINT = USB_OTG_DIEPINT_EPDISD;
```

After #399 the terminal can only echo

```
// Stop transmitting packets and NAK IN xfers.                                  
in_ep[epnum].DIEPCTL |= USB_OTG_DIEPCTDIEPINT_EPDISD;
```

I am doing more troubleshooting

kasjer:
I just checked STM32F407 target stm32f407disco, fresh master same behavior.
Tested with putty terminal, (minicom seems to send one byte at a time even when something is pasted into it).
[...]

PR #451, opened 2020-06-30
Fix receiving of short packet data (Synopsys)

duempel:
**Describe the PR**
Described in #448 there was an error that DCD did not inform USBD about the real number of received bytes. This PR will fix the problem and shorten the transfer length when receiving a short packet that indicates end of transfer.

**Additional context**
Only tested with CDC_MSC example yet.

### DIFF
--- a/src/portable/st/synopsys/dcd_synopsys.c
+++ b/src/portable/st/synopsys/dcd_synopsys.c
@@ -576,6 +576,12 @@ static void handle_rxflvl_ints(USB_OTG_OUTEndpointTypeDef * out_ep) {
 
         // Increment pointer to xfer data
         xfer->buffer += bcnt;
+
+        // Truncate transfer length in case of short packet
+        if(bcnt < xfer->max_size){
+          xfer->total_len -= (out_ep[epnum].DOEPTSIZ & USB_OTG_DOEPTSIZ_XFRSIZ_Msk) >> USB_OTG_DOEPTSIZ_XFRSIZ_Pos;
+          if(epnum == 0) xfer->total_len -= ep0_pending[TUSB_DIR_OUT];
+        }
       }
       break;
     case 0x03: // Out packet done (Interrupt)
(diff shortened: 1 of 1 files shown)


kasjer:
Tested with success on STM32F407, with CDC_MSC and Mynewt.

src/portable/st/synopsys/dcd_synopsys.c
```diff
@@ -576,6 +576,12 @@ static void handle_rxflvl_ints(USB_OTG_OUTEndpointTypeDef * out_ep) {
 
         // Increment pointer to xfer data
         xfer->buffer += bcnt;
+
+        // Truncate transfer length in case of short packet
+        if(bcnt < xfer->max_size){
+          xfer->total_len -= (out_ep[epnum].DOEPTSIZ & USB_OTG_DOEPTSIZ_XFRSIZ_Msk) >> USB_OTG_DOEPTSIZ_XFRSIZ_Pos;
+          if(epnum == 0) xfer->total_len -= ep0_pending[TUSB_DIR_OUT];
```
hathach, line 583: ep0_pending[TUSB_DIR_OUT] should be zero as well I think to end the whole transfer

duempel: I think `ep0_pending` in this case can have a different value than zero. It contains the bytes that do not fit into `DOEPTSIZ0`. Therefore `XFRSIZ` of `DOEPTSIZn` of different epnums should be equal to `XFRSIZ` of `DOEPTSIZ0` of EP0 **+** `ep0_pending`.

e.g. we request 512 bytes to be received at EP0. We schedule 64 bytes and set ep0_pending to 448. If the first scheduled packet has short length 10 the transfer will be already finished. Then `DOEPTSIZ` will contain 54 bytes only and `ep0_pending` has to be used to get the real remaining bytes of 502.

duempel: Ahh sorry I missunderstood you. Yes we should **set** ep0_pending to zero here. 🙄

PR #454, opened 2020-07-01
ESP32-S2: Handle the fact that available EP IN FIFOs are less than the number of available EP INs

me-no-dev:
ESP32-S2 has only 5 available endpoint-in FIFOs (including EP0) but 7 available EP IN numbers. This change decouples the fifo number from the endpoint number, providing FIFO numbers until they reach the limit, at which point it will return false and assert an error that too many endpoints were allocated.

### DIFF
--- a/src/portable/espressif/esp32s2/dcd_esp32s2.c
+++ b/src/portable/espressif/esp32s2/dcd_esp32s2.c
@@ -71,6 +71,20 @@ static uint32_t _setup_packet[2];
 #define XFER_CTL_BASE(_ep, _dir) &xfer_status[_ep][_dir]
 static xfer_ctl_t xfer_status[EP_MAX][2];
 
+// Max number of IN EP FIFOs
+#define USB_FIFO_NUM 5
+
+// Keep count of how many FIFOs are in use
+static uint8_t dcd_allocated_fifos = 1; //FIFO0 is always in use
+
+// Will either return an unused FIFO number, or 0 if all are used.
+static uint8_t dcd_get_free_fifo(){
+  if (dcd_allocated_fifos < USB_FIFO_NUM) {
+    return dcd_allocated_fifos++;
+  }
+  return 0;
+}
+
 // Setup the control endpoint 0.
 static void bus_reset(void)
 {
@@ -271,8 +285,12 @@ bool dcd_edpt_open(uint8_t rhport, tusb_desc_endpoint_t const *desc_edpt)
     // - Offset: GRXFSIZ + 16 + Size*(epnum-1)
     // - IN EP 1 gets FIFO 1, IN EP "n" gets FIFO "n".
 
+    uint8_t fifo_num = dcd_get_free_fifo();
+    TU_ASSERT(fifo_num != 0);
+
+    in_ep[epnum].diepctl &= ~(USB_D_TXFNUM1_M | USB_D_EPTYPE1_M | USB_DI_SETD0PID1 | USB_D_MPS1_M);
     in_ep[epnum].diepctl |= USB_D_USBACTEP1_M |
-                            epnum << USB_D_TXFNUM1_S |
+                            fifo_num << USB_D_TXFNUM1_S |
                             desc_edpt->bmAttributes.xfer << USB_D_EPTYPE1_S |
                             (desc_edpt->bmAttributes.xfer != TUSB_XFER_ISOCHRONOUS ? (1 << USB_DI_SETD0PID1_S) : 0) |
                             desc_edpt->wMaxPacketSize.size << 0;
@@ -282,8 +300,8 @@ bool dcd_edpt_open(uint8_t rhport, tusb_desc_endpoint_t const *desc_edpt)
     // Both TXFD and TXSA are in unit of 32-bit words.
     // IN FIFO 0 was configured during enumeration, hence the "+ 16".
     uint16_t const allocated_size = (USB0.grxfsiz & 0x0000ffff) + 16;
-    uint16_t const fifo_size = (EP_FIFO_SIZE/4 - allocated_size) / (EP_MAX-1);
-    uint32_t const fifo_offset = allocated_size + fifo_size*(epnum-1);
+    uint16_t const fifo_size = (EP_FIFO_SIZE/4 - allocated_size) / (USB_FIFO_NUM-1);
+    uint32_t const fifo_offset = allocated_size + fifo_size*(fifo_num-1);
 
     // DIEPTXF starts at FIFO #1.
     USB0.dieptxf[epnum - 1] = (fifo_size << USB_NPTXFDEP_S) | fifo_offset;
@@ -361,7 +379,8 @@ void dcd_edpt_stall(uint8_t rhport, uint8_t ep_addr)
     }
 
     // Flush the FIFO, and wait until we have confirmed it cleared.
-    USB0.grstctl |= ((epnum - 1) << USB_TXFNUM_S);
+    uint8_t const fifo_num = ((in_ep[epnum].diepctl >> USB_D_TXFNUM1_S) & USB_D_TXFNUM1_V);
+    USB0.grstctl |= (fifo_num << USB_TXFNUM_S);
     USB0.grstctl |= USB_TXFFLSH_M;
     while ((USB0.grstctl & USB_TXFFLSH_M) != 0) ;
   } else {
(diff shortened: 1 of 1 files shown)


hathach:
It is really odd to have 7 EP IN but only 5 FIFO, to be honest I don't see why people wouldn't use the  EP IN from 1-4 instead and jump to EP5,6 IN. How about us forcing application to only us EP IN from 1-4 instead, since I have a plan to merge dcd_esp32s2 with dcd_synopsys of stm32, that will make the code more neutral.

me-no-dev:
How will you force something like that only on one platform? Surely both ways would work. And yes, I totally agree that it's really weird that endpoints and fifos do not match, especially since you can split the whole fifo memory to more than 5... but :) next silicon I guess. I will be looking into DMA as well, as that is something that can remove the limitation.
Any way you go is fine with me. TBH I have even added a hack locally (you know I like hacks :D) that assigns a fake fifo to EP5 because it's used for the CDC control EP, that is never called and thus saving a fifo.

I am not sure why this approach would not work on the other chip? it's same IP so, should be fine? The approach itself is not bad or flawed in any way.

p.s. since you want to merge the dcd for both chips, do you plan on host support for the STM?

me-no-dev:
One case where one could use endpoints 5 and 6 is if they have predefined descriptors for different interfaces, but load only what is required on boot. Currently I handle this by not having the descriptors static and requesting endpoint numbers to use.

hathach:
> How will you force something like that only on one platform? Surely both ways would work. And yes, I totally agree that it's really weird that endpoints and fifos do not match, especially since you can split the whole fifo memory to more than 5... but :) next silicon I guess. I will be looking into DMA as well, as that is something that can remove the limitation.

We can bring up this PR when using DMA, though that may take long time since current contributors seems to be happy with FIFO approach.

> I am not sure why this approach would not work on the other chip? it's same IP so, should be fine? The approach itself is not bad or flawed in any way.

Like always, we need the reason to add it. I don't like to add code that didn't run at all e.g the situation when user skip EP 2 In and go with EP 5 IN is unlikely, and we can simply tell user to use the lower number EP. Therefore this code is redundant and not needed.

> Any way you go is fine with me. TBH I have even added a hack locally (you know I like hacks :D) that assigns a fake fifo to EP5 because it's used for the CDC control EP, that is never called and thus saving a fifo.

Hacky is totally fine, though we could most do that to optimize for a specific project/platform. But It is really hard to make it run all on MCUs from small to big. You are right about the notification endpoints, which is totally optional for Virtual Com. **This is actually a legit reason to merge this PR**, however it is too specific to ESP32s2. USBD API doesn't have way to tell DCD which endpoint to skip FIFO. The DCD doesn't know which class the EP belongs to. Maybe we could try to figure out how to tell DCD to skip FIFO assignment in portable fashion, then we could merge this PR.

> p.s. since you want to merge the dcd for both chips, do you plan on host support for the STM?

I had a plan to rework host stack from the LST year. But couldn't manage the time to do so. Current host only support EHCI/OHCI which are actually much more complicated than STM/ESP  (those are pc-graded host controller). In short, I will try when I could find time.

me-no-dev:
> Maybe we could try to figure out how to tell DCD to skip FIFO assignment in portable fashion, then we could merge this PR.

I could not come up with one :(
Endpoints are opened as they are found in the descriptors (as far as I understand the flow), so in order for endpoint to not be opened, it have to be somehow designated by another api call or something. Or... can we define the EP transfer size as 0 in the descriptor and have the dcd not open it? or not assign fifo? Would that still work and enumerate fine? Will try!

me-no-dev:
One more point for this PR (S2 specific): FIFOs will be larger, because the total space is divided by 4 instead of 6 (excluding EP0)

me-no-dev:
> I don't like to add code that didn't run at all

Do you mean that you do not expect this to be ever needed? Because this code will run the same on STM (no reason fifo number to be the same a EP number. Designation is so internal that 99% of the dcd code do not care for that fifo number) and will run in more cases on ESP32S2. This is for the case where descriptors for multiple drivers are predefined by an upper layer library and loaded when necessary. You can fit into 6 endpoints constant descriptors for MSC, CDC, MIDI, HID and VENDOR. While you can not have them all running at the same time on ESP32S2, with the current case you will need to implement some sort of EP pool and descriptor altering in order to not have the EPs constant.

Again, we can deal with it either way, this one will not require to explain anything to the users.

hathach:
> Endpoints are opened as they are found in the descriptors (as far as I understand the flow), so in order for endpoint to not be opened, it have to be somehow designated by another api call or something. Or... can we define the EP transfer size as 0 in the descriptor and have the dcd not open it? or not assign fifo? Would that still work and enumerate fine? Will try!

Define EP size = 0 shouldn't be done, since it will potentially mess up with PC driver, and never guarantee to run. An windows update that tighten the driver can cause all the production board **failed to work**. I am not sure if we could get away from having no EP in the CDC Management interface. The CDC specs say notification endpoint is optional, but I had issue with windowser `usbser` in the pass without the notification EP. Maybe things have change, I am not sure.

> Do you mean that you do not expect this to be ever needed? Because this code will run the same on STM (no reason fifo number to be the same a EP number. Designation is so internal that 99% of the dcd code do not care for that fifo number) and will run in more cases on ESP32S2. This is for the case where descriptors for multiple drivers are predefined by an upper layer library and loaded when necessary. You can fit into 6 endpoints constant descriptors for MSC, CDC, MIDI, HID and VENDOR. While you can not have them all running at the same time on ESP32S2, with the current case you will need to implement some sort of EP pool and descriptor altering in order to not have the EPs constant.

OK, I got your point now, and it makes sense. I will review the PR in a couple of days. In your constant EP assignment, you seem to miss the notification EP of CDC. Although it is not used, the EP still need to response with NAK.

me-no-dev:
>  In your constant EP assignment, you seem to miss the notification EP of CDC. Although it is not used, the EP still need to response with NAK.

It's up there :) 0x84 and 0x03 are for DATA, 0x85 is the notification EP

hathach:
> > In your constant EP assignment, you seem to miss the notification EP of CDC. Although it is not used, the EP still need to response with NAK.
> 
> It's up there :) 0x84 and 0x03 are for DATA, 0x85 is the notification EP (weird numbers because of S2 ROM CDC)

ah right, there is also an extra 0x05 as well !!!, I was overlook.

src/portable/espressif/esp32s2/dcd_esp32s2.c
```diff
@@ -71,6 +71,20 @@ static uint32_t _setup_packet[2];
...
 
+// Max number of IN EP FIFOs
+#define USB_FIFO_NUM 5
+
+// Keep count of how many FIFOs are in use
+static uint8_t dcd_allocated_fifos = 1; //FIFO0 is always in use
+
+// Will either return an unused FIFO number, or 0 if all are used.
+static uint8_t dcd_get_free_fifo(){
+  if (dcd_allocated_fifos < USB_FIFO_NUM) {
```
hathach, line 82: simple if should stay in the same line, it is just my preference. This is optional

src/portable/espressif/esp32s2/dcd_esp32s2.c
```diff
@@ -681,6 +702,8 @@ static void _dcd_int_handler(void* arg)
     if (otg_int & USB_SESENDDET_M)
     {
       dcd_event_bus_signal(0, DCD_EVENT_UNPLUGGED, true);
+      // FIFOs will be reassigned when the endpoints are reopen
+      dcd_allocated_fifos = 1;
```
hathach, line 706: this one is redundant, reset on bus_reset() is good enough already. Please remove this.

src/portable/espressif/esp32s2/dcd_esp32s2.c
```diff
@@ -71,6 +71,20 @@ static uint32_t _setup_packet[2];
 #define XFER_CTL_BASE(_ep, _dir) &xfer_status[_ep][_dir]
 static xfer_ctl_t xfer_status[EP_MAX][2];
 
+// Max number of IN EP FIFOs
+#define USB_FIFO_NUM 5
```
hathach, line 75: To be consistent, this should be name `EP_FIFO_NUM` and should be placed next to `EP_FIFO_SIZE`

me-no-dev:
I think I covered all requests :)

me-no-dev:
with this PR merged, persistence is coming next :) one of the lines here was related more or less (would have failed without it)

me-no-dev:
Maybe should clarify :)
This line here was missing: https://github.com/hathach/tinyusb/pull/454/files#diff-371e8160b188a80c1a3821c75d194adfR289

Code without it is assuming a clean peripheral registers and just setting the new values without clearing their masks before, which with persistence was causing previous FIFO to be 2, new FIFO to be 4 and resulting in 6, which first does not exists and second is not the correct value :)

PR #455, opened 2020-07-01
Add stm hs

UweBonnes:
**Describe the PR**
Remove PYC PHY PLL stabilization delay. Add stlinkv3mini board BSP.

**Additional context**

### DIFF
--- a/src/portable/st/synopsys/dcd_synopsys.c
+++ b/src/portable/st/synopsys/dcd_synopsys.c
@@ -27,7 +27,6 @@
  */
 
 #include "tusb_option.h"
-#include "../../../../hw/bsp//board.h"
 
 #if defined (STM32F105x8) || defined (STM32F105xB) || defined (STM32F105xC) || \
     defined (STM32F107xB) || defined (STM32F107xC)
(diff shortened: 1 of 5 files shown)


src/portable/st/synopsys/dcd_synopsys.c
```diff
@@ -349,7 +348,12 @@ static bool USB_HS_PHYCInit(void)
     // Enable PLL internal PHY
     usb_hs_phyc->USB_HS_PHYC_PLL = phyc_pll | USB_HS_PHYC_PLL_PLLEN;
 
-    board_delay(2);
+    /* Original ST code has 2 ms delay for PLL stabilization.
+     * Primitive test shows that more than 10 USB un/replug cycle
+     * showed no error with enumeration
+     * //#include "../../../../hw/bsp//board.h"
```
hathach, line 354: This is called withtin dcd_init() which only happens one when powering on or reset. Plug and unplug shouldn't be affected at all. If it does, it possibly cause the initial failure of the 1st enumeration or something like that. Though the 2ms delay isn't there for no reasons. Since it works for now, we could worry about this later on when there is problem with it :)

PR #456, opened 2020-07-02
ESP32-S2: Detect EP IN Xfer Timeout

me-no-dev:
In some rare ocasions (bad cable, noise, etc.) data transfer might timeout and hang the endpoint, unless the interrupt flag is cleared.
This pull request targets to solve that case.


me-no-dev:
oh... I got more weird stuff, just nothing concrete or a fix for the situation. Things like getting transfer complete, but fifo is not empty, so nothing else can be sent. Flipping an undocumented endpoint interrupt bit that hangs the whole bus (digital was not able to help). And so on. To be fair, they are really rare and I usually abuse the bus when they happen.

PR #468, opened 2020-07-19
FreeRTOS: Yield from ISR to notify the USB task

me-no-dev:
If we do not yeld in ISR when we write to queue/give semaphore, the scheduler will not know of the change and will not check the queue untill the next OS tick. This change causes the task to be called immediately and makes communication many times faster.

### Example reading the device descriptor

In this example a simple exchange with two data packets is shown. In the first case, the whole exchange takes 3ms and in the second, less than 240us. OS Tick is running at 1000Hz. If lower than that, gain will be even more.

#### Before the change

Screenshot 2020-07-19 at 1 22 29

#### After the change

Screenshot 2020-07-19 at 23 47 53

@hathach this is also the progress with my USB decoder and direct result of it :)

### DIFF
--- a/src/osal/osal_freertos.h
+++ b/src/osal/osal_freertos.h
@@ -32,6 +32,7 @@
 #include "semphr.h"
 #include "queue.h"
 #include "task.h"
+#include "tusb_option.h"
 
 #ifdef __cplusplus
 extern "C" {
@@ -58,7 +59,19 @@ static inline osal_semaphore_t osal_semaphore_create(osal_semaphore_def_t* semde
 
 static inline bool osal_semaphore_post(osal_semaphore_t sem_hdl, bool in_isr)
 {
-  return in_isr ?  xSemaphoreGiveFromISR(sem_hdl, NULL) : xSemaphoreGive(sem_hdl);
+  if(!in_isr){
+    return xSemaphoreGive(sem_hdl) != 0;
+  }
+  BaseType_t xHigherPriorityTaskWoken;
+  BaseType_t res = xSemaphoreGiveFromISR(sem_hdl, &xHigherPriorityTaskWoken);
+#if CFG_TUSB_MCU == OPT_MCU_ESP32S2
+  if (xHigherPriorityTaskWoken) {
+    portYIELD_FROM_ISR();
+  }
+#else
+  portYIELD_FROM_ISR(xHigherPriorityTaskWoken);
+#endif
+  return res != 0;
 }
 
 static inline bool osal_semaphore_wait (osal_semaphore_t sem_hdl, uint32_t msec)
@@ -125,7 +138,19 @@ static inline bool osal_queue_receive(osal_queue_t qhdl, void* data)
 
 static inline bool osal_queue_send(osal_queue_t qhdl, void const * data, bool in_isr)
 {
-  return in_isr ? xQueueSendToBackFromISR(qhdl, data, NULL) : xQueueSendToBack(qhdl, data, OSAL_TIMEOUT_WAIT_FOREVER);
+  if(!in_isr){
+    return xQueueSendToBack(qhdl, data, OSAL_TIMEOUT_WAIT_FOREVER) != 0;
+  }
+  BaseType_t xHigherPriorityTaskWoken;
+  BaseType_t res = xQueueSendToBackFromISR(qhdl, data, &xHigherPriorityTaskWoken);
+#if CFG_TUSB_MCU == OPT_MCU_ESP32S2
+  if (xHigherPriorityTaskWoken) {
+    portYIELD_FROM_ISR();
+  }
+#else
+  portYIELD_FROM_ISR(xHigherPriorityTaskWoken);
+#endif
+  return res != 0;
 }
 
 static inline bool osal_queue_empty(osal_queue_t qhdl)


src/osal/osal_freertos.h
```diff
@@ -32,6 +32,7 @@
 #include "semphr.h"
 #include "queue.h"
 #include "task.h"
+#include "tusb_option.h"
```
hathach, line 35: this shouldn't be needed, since osal_freertos.h is only meant to be included by osal.h which already included tusb_common.h.

me-no-dev: agreed, but I had a couple of failed builds and opted for "better safe than sorry" approach :)

hathach:
> @hathach this is also the progress with my USB decoder and direct result of it :)

Are you writing the USB decoder script with an OSC/Logic analyzer, that looks fantastic. Super work there

me-no-dev:
> Are you writing the USB decoder script with an OSC/Logic analyzer

Yes :) The LA that I have is not widely used or support everything, so some coding had to be done :) I added since SETUP decoding and I am currently adding CRC check and calculation. It's limited to 12Mbps USB and far from your LA, but it works great and shows everything the I might need to debug the bus up to driver level.

Screenshot 2020-07-21 at 17 23 11

hathach:
> > Are you writing the USB decoder script with an OSC/Logic analyzer
> 
> Yes :) The LA that I have is not widely used or support everything, so some coding had to be done :) I added since SETUP decoding and I am currently adding CRC check and calculation. It's limited to 12Mbps USB and far from your LA, but it works great and shows everything the I might need to debug the bus up to driver level.

Super cool, I am sure lots of people will find your decoding script useful. I am too lazy for write script. Usually when pulling the lecroy, I am in some kind of a trouble with the code :)

me-no-dev:
> I am sure lots of people will find your decoding script useful.

I will definitely publish it. I am not sure if others will find it useful if they do not have an LA from the same company. The good thing is that it's all in JavaScript, so in reality a web layer can be written to display the data and also to accept files from different LAs (I em pretty sure that the APP that my LA uses is just a web app with QT backend), but requires someone with the free time and will at hand :D

hathach:
Note to everyone:  while working with nRF52 + freeRTOS, the USB IRQ can be triggered (BUS_RESET) before task scheduler is started, which will cause portYIELD_FROM_ISR -> SVCHandler run into hardfault. I could resolve this by moving the tud_init() next to `vTaskStartScheduler()`. 

https://github.com/adafruit/Adafruit_nRF52_Arduino/pull/542/files

This can be resolved completely by adding.

```
if(xTaskGetSchedulerState() != taskSCHEDULER_NOT_STARTED) portYIELD_FROM_ISR(xHigherPriorityTaskWoken)
```

Though I am a bit tentative since it will add unnecessary overhead for a check that 99.9 is true. Maybe we should put an note somewhere to tell user about this instead. @me-no-dev what do you think about this.

me-no-dev:
This is never the case for ESP32 either. Maybe worth checking only until scheduler is running?

hathach:
> This is never the case for ESP32 either. Maybe worth checking only until scheduler is running?

there is no way to do just that, the only way is adding it into the osal_freertos.h . I will try to put up some docs, if anyone run into the issue. Then google this out, they may raise the issue for further action.

me-no-dev:
```cpp
static bool _scheduler_started = false;
static bool check_if_scheduler_was_started(){
  _scheduler_started = (xTaskGetSchedulerState() != taskSCHEDULER_NOT_STARTED);
  return _scheduler_started;
}
...
if(_scheduler_started || check_if_scheduler_was_started()) portYIELD_FROM_ISR(xHigherPriorityTaskWoken);
```

`check_if_scheduler_was_started()` should not be called once start was detected

hathach:
> ```c++
> static bool _scheduler_started = false;
> static bool check_if_scheduler_was_started(){
>   _scheduler_started = (xTaskGetSchedulerState() != taskSCHEDULER_NOT_STARTED);
>   return _scheduler_started;
> }
> ...
> if(_scheduler_started || check_if_scheduler_was_started()) portYIELD_FROM_ISR(xHigherPriorityTaskWoken);
> ```
> 
> `check_if_scheduler_was_started()` should not be called once start was detected

it is an great idea, however, since osal_freertos is all static inline, therefore every callers has a copy of the function and also the `_scheduler_started` as well. For simplicity, I didn't want to introduce a `osal.c`, the whole inline is to prevent myself to overwrite the osal code than it needs to be.

hathach:
I think it is best to just move the tud_init() into the app device task in the example with a note to user is good enough for now.
https://github.com/hathach/tinyusb/blob/master/examples/device/cdc_msc_freertos/src/main.c#L118

me-no-dev:
> I think it is best to just move the tud_init() into the app device task

this is the valid RTOS approach IMHO :) You start a task, init whatever it needs inside, then loop about it.

hathach:
Yeah, I feel much better now. Since most people will use stock example as template, it is better to fix it asap :)

me-no-dev:
decoder is now online :) https://github.com/ikalogic/ScanaStudio-scripts-v3/pull/15

hathach:
superb !!

PR #476, opened 2020-07-31
Fix synopsys odd/even frame bit for IN ISO endpoints

kasjer:
**Describe the PR**
For ISO endpoint driver has to specify when data
is to be transmitted (odd or even frame).
Currently code was not updating this bit resulting in
data being sent every other frame.
If interval was 1ms full data packed was sent every 2ms, and
ZLP was sent in between.

**Additional context**
~~I'm not sure if second part (handle_epin_ints) is really needed.~~

### DIFF
--- a/src/portable/st/synopsys/dcd_synopsys.c
+++ b/src/portable/st/synopsys/dcd_synopsys.c
@@ -365,6 +365,13 @@ static void edpt_schedule_packets(uint8_t rhport, uint8_t const epnum, uint8_t c
         ((total_bytes << USB_OTG_DIEPTSIZ_XFRSIZ_Pos) & USB_OTG_DIEPTSIZ_XFRSIZ_Msk);
 
     in_ep[epnum].DIEPCTL |= USB_OTG_DIEPCTL_EPENA | USB_OTG_DIEPCTL_CNAK;
+    // For ISO endpoint set correct odd/even bit for next frame.
+    if ((in_ep[epnum].DIEPCTL & USB_OTG_DIEPCTL_EPTYP) == USB_OTG_DIEPCTL_EPTYP_0)
+    {
+      // Take odd/even bit from frame counter.
+      uint32_t const odd_frame_now = (dev->DSTS & (1u << USB_OTG_DSTS_FNSOF_Pos));
+      in_ep[epnum].DIEPCTL |= (odd_frame_now ? USB_OTG_DIEPCTL_SD0PID_SEVNFRM_Msk : USB_OTG_DIEPCTL_SODDFRM_Msk);
+    }
     // Enable fifo empty interrupt only if there are something to put in the fifo.
     if(total_bytes != 0) {
       dev->DIEPEMPMSK |= (1 << epnum);


src/portable/st/synopsys/dcd_synopsys.c
```diff
@@ -365,6 +365,14 @@ static void edpt_schedule_packets(uint8_t rhport, uint8_t const epnum, uint8_t c
         ((total_bytes << USB_OTG_DIEPTSIZ_XFRSIZ_Pos) & USB_OTG_DIEPTSIZ_XFRSIZ_Msk);
 
     in_ep[epnum].DIEPCTL |= USB_OTG_DIEPCTL_EPENA | USB_OTG_DIEPCTL_CNAK;
+    // For ISO endpoint set correct odd/even bit for next frame.
+    if ((in_ep[epnum].DIEPCTL & USB_OTG_DIEPCTL_EPTYP) == USB_OTG_DIEPCTL_EPTYP_0)
+    {
+        const uint32_t odd_frame_now = (dev->DSTS & (1u << USB_OTG_DSTS_FNSOF_Pos));
+        // Now is odd, next should be even.
+        if (odd_frame_now) in_ep[epnum].DIEPCTL |= USB_OTG_DIEPCTL_SD0PID_SEVNFRM_Msk;
```
hathach, line 373: tenary operator is more preferred, but it is only a minor suggestion. 

```
in_ep[epnum].DIEPCTL |= (odd_frame_now ? USB_OTG_DIEPCTL_SD0PID_SEVNFRM_Msk : USB_OTG_DIEPCTL_SODDFRM_Msk);
```

kasjer: Code adjusted according to suggestion.
**const** moved to match coding style.

duempel:
I'm glad you're working on isochronous endpoints. I think there are a few more changes to do to get a full iso support. But these little change seems to get them working already. 👍

To focus on this PR's content: As you already mentioned I do not think that we need to set the EONUM bit within TXFE interrupt. Since a full iso transfer has to be sent within one frame, it's not useful to change this bit again. Did you try to run it without the changes to `handle_ep_ints`? I can not test it myself at the moment. 

@hathach how does the stack react if you do not get `dcd_event_xfer_complete`? Will you schedule a new transfer or has the previous to be finished first? This is very important to iso endpoints because it can happen that we do not get the full transfer transmitted within one frame and in this case device driver won't invoke `dcd_event_xfer_complete`. We should introduce incomplete transfer interrupts `IISOIXFR` and `IISOOXFR` to isochronous endpoints maybe.

@kasjer OUT ISO endpoints seem to use the same way of handling odd and even frame numbers. Please feel free to set up a PR for OUT as well. 😊

kasjer:
@duempel my USB knowledge is very limited so I'm not sure what are the constraints on ISO. I simply added this code in two places just in case. My application did not use second path as whole data frame is sent in one packet.
I can drop it and if later some finds out that ISO packets needs this frame bit set in interrupt maybe some of you will remember and add it accordingly.
As to OUT ISO, I'm not there yet.

hathach:
> @hathach how does the stack react if you do not get `dcd_event_xfer_complete`? Will you schedule a new transfer or has the previous to be finished first? This is very important to iso endpoints because it can happen that we do not get the full transfer transmitted within one frame and in this case device driver won't invoke `dcd_event_xfer_complete`. We should introduce incomplete transfer interrupts `IISOIXFR` and `IISOOXFR` to isochronous endpoints maybe.

The stack should still wait for the complete event to prevent race condition (multiple xfer calls from preempted rtos). In case the data is not sent fast enough, dcd should detect this and let the stack know so that it could mark the endpoint as ready for next transfer. Though I don't think we need an extra INCOMPLETE event, ISO data is time-sensitive (audio/video), if it is sent late, it is just better to not send at all. 

However I am not entirely sure as well, since I haven't done much work with ISO. I probably need to educate myself and learn it from some handon with existing consumer device as well. That says, anything to get it running as proof of concept is good enough. We could add TODO anywhere we think we could do/improve later on.

PanRe:
I also did a bit of research concerning the ISO EP even/odd frame bit. This topic was heavly discussed [here](http://www.chibios.com/forum/viewtopic.php?f=16&t=926), where also a flowchart of how to handle it was presented. This might be of interest to you if you haven't seen it already!

duempel:
> In case the data is not sent fast enough, dcd should detect this and let the stack know so that it could mark the endpoint as ready for next transfer.

Alright, our current ISO handling is not safe in this point. It can happen that we do not send out data and stack never gets a complete event. I'll write an issue to keep this in mind.

> Though I don't think we need an extra INCOMPLETE event, ISO data is time-sensitive (audio/video), if it is sent late, it is just better to not send at all.

I agree. Usually it's no problem if a iso transfer fails. Maybe there are some special cases in which the application needs to get notification about it. But since I never used iso endpoints before I don't know how rare these cases are. I did not want to introduce a new event to stack. Only mentioned these two interrupts to be handled within dcd to invoke `dcd_event_xfer_complete` at failure to set the ep free for a new transfer.

@kasjer thanks for removing second part. Since I never read about it in ST's programming model it should work fine without. I just wondered why ST's define for iso type is `USB_OTG_DIEPCTL_EPTYP_0`. Very confusing since the EPTYP field has the value 01 for it. Well but it seems to work and since it does not effect endpoints other than isochronous it's good to merge.

hathach:
Thank you everyone of participating in the helpful discussion,hopefully we could got ISO support soon enough.

PR #479, opened 2020-08-02
add app driver as suggested by #467

### LINKED ISSUE #384: Allow overriding class drivers table

### LINKED ISSUE #467: Custom addition usbd driver
In _usbd_driver array (in file usbd.c), could add a include like
```c
#ifdef USBD_DRIVER_INCLUDE_H
#include USBD_DRIVER_INCLUDE_H
#endif

#ifdef USBD_DRIVER_INCLUDE
#include USBD_DRIVER_INCLUDE
#endif
```
to support user custom usbd drivers, avoiding modifying files.

The file will be like:
```c
#include "tusb.h"
#include "device/usbd.h"
#include "device/usbd_pvt.h"
#include "device/dcd.h"

#ifdef USBD_DRIVER_INCLUDE_H
#include USBD_DRIVER_INCLUDE_H
#endif

// ...

static usbd_class_driver_t const _usbd_driver[] =
{
// Original content
  #if CFG_TUD_CDC
  {
      DRIVER_NAME("CDC")
      .init             = cdcd_init,
      .reset            = cdcd_reset,
      .open             = cdcd_open,
      .control_request  = cdcd_control_request,
      .control_complete = cdcd_control_complete,
      .xfer_cb          = cdcd_xfer_cb,
      .sof              = NULL
  },
  #endif

// More ...

// Add this
#ifdef USBD_DRIVER_INCLUDE
#include USBD_DRIVER_INCLUDE
#endif

};

// ...
```

User can define USBD_DRIVER_INCLUDE, e.g. `<usbd_driver.include>`
The file `usbd_driver.include` content could be like:
```c
{
	DRIVER_NAME("CCID")
	.init = ccid_init,
	.reset = ccid_reset,
	.open = ccid_open,
	.control_request = ccid_control_request,
	.control_complete = ccid_control_complete,
	.xfer_cb = ccid_xfer_cb,
	.sof = NULL
},
// More ...
```

hathach:
**Describe the PR**
- Add app driver as suggested by #467, more preferable than #385 
- Fix #384 , Fix #467 

**Additional context**

@igrr @me-no-dev

### DIFF
--- a/src/device/usbd.c
+++ b/src/device/usbd.c
@@ -33,6 +33,11 @@
 #include "device/usbd_pvt.h"
 #include "dcd.h"
 
+// Include application driver header if defined
+#ifdef CFG_TUD_APP_DRIVER_HEADER
+#include CFG_TUD_APP_DRIVER_HEADER
+#endif
+
 #ifndef CFG_TUD_TASK_QUEUE_SZ
 #define CFG_TUD_TASK_QUEUE_SZ   16
 #endif
@@ -99,6 +104,10 @@ typedef struct
 
 static usbd_class_driver_t const _usbd_driver[] =
 {
+  #ifdef CFG_TUD_APP_DRIVER_LIST
+  CFG_TUD_APP_DRIVER_LIST
+  #endif
+
   #if CFG_TUD_CDC
   {
       DRIVER_NAME("CDC")


me-no-dev:
We can definitely work with that 👍

igrr:
@hathach Am I right that this approach, compared to #385, does not allow the list of drivers to be defined dynamically, and we still need to know the list of drivers at compile time?

This option is certainly better than not having any way to customize the driver list, so 👍 on that. However the callback option still seems to be preferable since it allows both compile-time defied list of custom drivers, or run-time defined list of custom drivers.

HiFiPhile:
I think it's enough for most user cases.

hathach:
> @hathach Am I right that this approach, compared to #385, does not allow the list of drivers to be defined dynamically, and we still need to know the list of drivers at compile time?
> 
> This option is certainly better than not having any way to customize the driver list, so  on that. However the callback option still seems to be preferable since it allows both compile-time defied list of custom drivers, or run-time defined list of custom drivers.

@igrr yeah, this approach is very static, much like current implementation. Which is why I really like it. Could you tell me in which scenario you need to have an run-time driver list. If it is lightly used, this approach can actually be implemented as run-time defined as well, application only need to
- define the all the driver it could possibly enable with the macro
- for driver that is disabled in run-time, just use an local variable to return 0/false upon called e.g `if (driver_disabled) return 0;` in open().

I know esp32 has a great heap management, but I still prefer to having everything static as much as possible within the stack. As always, I am open for any suggestion/discussion as well.

me-no-dev:
@hathach I can try to give an example: 
- The ESP32 camera driver is a separate component that can benefit from having it's own webcam driver support that can be loaded if the component is in the project and enabled.
- ESP32-Arduino will also be able to load it's own custom drivers (or the above camera driver) independent of what is precompiled from ESP-IDF.

I am sure @igrr could fill in more cases :)

igrr:
Yes, pretty much what @me-no-dev said — the main reason for dynamic registration is to not centralize the knowledge of all the custom drivers in one place.
Which is not easy from the build system perspective — how to allow 3rd party components to contribute entries to CFG_TUD_APP_DRIVER_HEADER?
As an example, we have a virtual filesystem (VFS) component in IDF, which allows dynamic registration of filesystem drivers. The filesystem driver can be provided in a third party component, and VFS layer doesn't need to know about the specific driver at compile time.

I think the callback approach is compatible both with compile-time driver registration (no heap allocation required), as well as run-time registration.

As to the approach you suggest with reserving some "slots" in the driver table, i think it can work as well, although it does feel a bit hacky compared to having a callback.

hathach:
thanks @igrr , it is a valid point to support 3rd party drivers. IDF is much more comprehensive that what I have been done so far. I will get back to the #385 . Thanks for explanation.

PR #485, opened 2020-08-12
Fix synopsys fifo flush during stall

kasjer:
**Describe the PR**
Wrong FIFO was flushed in dcd_edpt_stall().
(epnum - 1) should only be used when accessing DIEPTXF registers.

For DIEPCTL and GRSTCTL epnum is correct index.

### DIFF
--- a/src/portable/st/synopsys/dcd_synopsys.c
+++ b/src/portable/st/synopsys/dcd_synopsys.c
@@ -667,7 +667,7 @@ void dcd_edpt_stall (uint8_t rhport, uint8_t ep_addr)
     }
 
     // Flush the FIFO, and wait until we have confirmed it cleared.
-    usb_otg->GRSTCTL |= ((epnum - 1) << USB_OTG_GRSTCTL_TXFNUM_Pos);
+    usb_otg->GRSTCTL |= (epnum << USB_OTG_GRSTCTL_TXFNUM_Pos);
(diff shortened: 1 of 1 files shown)


kasjer:
As I mentioned enum -1 is correct for DIEPTXF registers.
For other it is not, see few lines below in same function.
https://github.com/hathach/tinyusb/blob/b0f617ba209110a4d5f144f46ccfc9f2cbfa25e7/src/portable/st/synopsys/dcd_synopsys.c#L598

hathach:
> As I mentioned enum -1 is correct for DIEPTXF registers.
> For other it is not, see few lines below in same function.
> https://github.com/hathach/tinyusb/blob/b0f617ba209110a4d5f144f46ccfc9f2cbfa25e7/src/portable/st/synopsys/dcd_synopsys.c#L598

Sorry, I missed the context, and jump right to the diff. Will pull out the manual soon enough :)

cr1901:
Looks like this is the "dual" of [another](https://github.com/hathach/tinyusb/pull/177) bug in my initial implementation.

hathach:
> Looks like this is the "[dual](https://en.wikipedia.org/wiki/Duality_(mathematics))" of [another](https://github.com/hathach/tinyusb/pull/177) bug in my initial implementation.

:+1:  This is still way better than my initial port of anything else :smile: . Thank you for confirmation.

PR #486, opened 2020-08-14
Add dcd_edpt_close() to synopsys

kasjer:
**Describe the PR**
Endpoint close was implemented only in one driver so far.
This function is needed for interfaces with several alternate settings.

The way FIFO is allocated in `dcd_edpt_open()` allows to correctly close
only one IN endpoint (the one that was opened last).

Stall and close share code as was suggested by TODO: comment.

**Additional context**
FIFO allocation schema could be changed to allow closing endpoints.
It could be that FIFO allocation is not done during open but based on all possible
interfaces. This would required some major API changes.

This functionality is needed for audio devices where **SET_INTERFACE** request
can be called several times requesting different endpoint sizes.

Work was tested only on IN endpoint so far (microphone), it may not be enough for OUT.

There is one point for discussion.
When **SET_INTERFACE** request is processed (resulting in endpoint close) for ISO IN endpoint
it is very likely that transfer is already started (FIFO is full).
But there will not be any IN token for this endpoint since host just set interface to 0 bandwidth alternate with no endpoints.
Close function will empty endpoint FIFO but **busy** flag will not be changed.
Transfer should be terminated.
I'm dealing with this problem by calling `usbd_edpt_clear_stall()` just after opening endpoint anew that clear busy flag.
Clearing busy flag only when endpoint is opened again has advantage that it can be check in application
code and when busy flash is true writes to IN endpoint would not be carried over.

### DIFF
--- a/src/portable/st/synopsys/dcd_synopsys.c
+++ b/src/portable/st/synopsys/dcd_synopsys.c
@@ -637,9 +637,7 @@ bool dcd_edpt_xfer (uint8_t rhport, uint8_t ep_addr, uint8_t * buffer, uint16_t 
   return true;
 }
 
-// TODO: The logic for STALLing and disabling an endpoint is very similar
-// (send STALL versus NAK handshakes back). Refactor into resuable function.
-void dcd_edpt_stall (uint8_t rhport, uint8_t ep_addr)
+static void dcd_edpt_disable (uint8_t rhport, uint8_t ep_addr, bool stall)
 {
   (void) rhport;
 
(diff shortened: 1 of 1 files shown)


hathach:
Hi Hi, sorry for late response, I have been busy doing other work for Adafruit.

> **Describe the PR**
> Endpoint close was implemented only in one driver so far.
> This function is needed for interfaces with several alternate settings.
> 
> The way FIFO is allocated in `dcd_edpt_open()` allows to correctly close
> only one IN endpoint (the one that was opened last).

This is good start, we can focus on closing random endpoint later on when other thing work.

> Stall and close share code as was suggested by TODO: comment.

I think we could do better with this, but it is not a issue at all. We could always refactor.

> **Additional context**
> FIFO allocation schema could be changed to allow closing endpoints.
> It could be that FIFO allocation is not done during open but based on all possible
> interfaces. This would required some major API changes.

Yeah, the fifo allocation scheme now is just enough to get the stack running with all the fifo size/ep number in all STM32 mcu. But I would prefer not changing the API in a significant way. Since the only endpoint that got closed and re-open is ISO ( I don't see any reason why hosts would do this with other types). We can allocate ISO buffer backward from the last byte (like stack & heap) and manage as minimal malloc (alhough I don't like dynamic memory, I don't know any way better). 

Actually I think we don't have to do this on purpose, by USB std, after `SET_CONFIGURE` all other endpoints got opened, while ISO won't, since there is always a default alt interface without endpoint for ISO. That makes ISO endpoints are always the one got opened last after `SET_INTERFACE`. We can run into issue like fragmented & wasted spaces when having multiple ISOs and reopening the fist ISO with larger size. If necessary (short of space) we could disable all the ISOs, and reopen them with adjusted index.  

> This functionality is needed for audio devices where **SET_INTERFACE** request
> can be called several times requesting different endpoint sizes.
> 
> Work was tested only on IN endpoint so far (microphone), it may not be enough for OUT.

OUT shared fifo, we only need to allocated the largest ISO size possible for it i.e 1023 / 1024 bytes. However, these value are huge for smaller MCU (largest has 4K, smallest only has 512-1K in total). Maybe we can add an additional macro for TUD_CFG_ISO_MAX_SIZE or similar to help with determining the OUT ISO.

> There is one point for discussion.
> When **SET_INTERFACE** request is processed (resulting in endpoint close) for ISO IN endpoint
> it is very likely that transfer is already started (FIFO is full).
> But there will not be any IN token for this endpoint since host just set interface to 0 bandwidth alternate with no endpoints.
> Close function will empty endpoint FIFO but **busy** flag will not be changed.
> Transfer should be terminated.
> I'm dealing with this problem by calling `usbd_edpt_clear_stall()` just after opening endpoint anew that clear busy flag.
> Clearing busy flag only when endpoint is opened again has advantage that it can be check in application
> code and when busy flash is true writes to IN endpoint would not be carried over.

Yeah, before closing the endpoint, the usbd close API should mark ep as busy ( to prevent application schedule new transfer), then dcd should abort all going transfer (there is no need to notify the stack/application, ISO transfer can be allowed to drop) and proceeding with closing and configuring new fifo size.

hathach (approved):
Thanks for the PR, it looks good to me. ISO is troublesome (hence it is last to got supported), there is no surprise that it will take several more PRs to get it running. Let's leave this open for a more few days to see if anyone else have any other idea.

PR #490, opened 2020-08-21
add support for SAMD11 devices / add samd11_xplained board

ktemkin:
This pull request adds support for USB device mode on SAMD11 family of devices. The SAMD11 devices use a USB peripheral which is binary compatible with the SAMD21, so changes are minimal, and the SAMD11 family use the SAMD21 TUSB driver.

This requires https://github.com/hathach/microchip_driver/pull/1, which adds the SAMD11 _Atmel Software Framework_ files to bring up the SAMD21.


hw/bsp/samd11_xplained/board.mk
```diff
@@ -0,0 +1,53 @@
...
+  -fdata-sections \
+  -Wl,--gc-sections \
+  -mthumb \
+  -mabi=aapcs-linux \
+  -mcpu=cortex-m0plus \
+  -nostdlib -nostartfiles \
+  -D__SAMD11D14AM__ \
+  -DCONF_DFLL_OVERWRITE_CALIBRATION=0 \
+  -DOSC32K_OVERWRITE_CALIBRATION=0 \
+  -DCFG_TUSB_MCU=OPT_MCU_SAMD21 \
```
hathach, line 12: you could add `-DCFG_EXAMPLE_MSC_READONLY` to allow using ROM instead of RAM for read-only mass storage device. It can still be enumerated and show up the readme though PC won't be able to change the disk and/or readme contents. Here is how other small chip define this.  
https://github.com/hathach/tinyusb/blob/master/hw/bsp/lpcxpresso1347/board.mk#L9

The macros is kind of required as well to pass CI build test.

hathach, line 12: Ideally we should add a new `OPT_MCU_SAMD11`, but it is not an issue. I will do a follow up later on. There is not much differences.

PR #514, opened 2020-09-15
nrf52: Fix edpt_dma_start() wrong condition check

kasjer:
**Describe the PR**
Operator < used in while condition was obviously incorrect.
Loop starts with checking if unsigned variable is less then 0.
This condition is always false.

This reverses condition to follow intention of of the code.

**Additional context**


hathach (approved):
Oh my, you really got an eagle eye, it seems like I didn't test this code or the tests I used is irrelevant. Thank you for spotting the issue

PR #520, opened 2020-09-26
STM32F3 IRQ remap option and some minor improvements

salkinium:
Hi, thanks for this awesome library! I'm currently integrating TinyUSB into modm (https://github.com/modm-io/modm/pull/478) and I found some minor things to improve:

1. I added a simple `STM32F3_IRQ_REMAP`, which can be defined in the config to remap the IRQs. The user must remap the IRQs in their hardware setup (it's just `SYSCFG->CFGR1 |= SYSCFG_CFGR1_USB_IT_RMP;`).
2. We're using TinyUSB in C++, which doesn't define the `__STDC_VERSION__` macro, thus throwing a warning. Just a simple fix.
3. The `_MESS_ERR()` and `_MESS_FAILED()` macros use `printf` instead of `tu_printf`, which is unfortunate for debugging and also links the newlib version of printf which requires heap, thus pulling a whole lot of unrelated code.
4. To improve debugability I'm overwriting the `TU_ASSERT` macros with our own fault reporting implementation, and to allow this I just wrapped them in `#ifndef`. Unfortunately OpenOCD does not by default reset the debugger connected flag on disconnect, thus `TU_BREAKPOINT()` causes a HardFault, which isn't great to debug.
5. We're using SCons to compile, which has issues resolving `#include CFG_TUSB_CONFIG_FILE`, thus not recompiling TinyUSB sources when the config changes. So I changed the mechanism to "fallback" to a plain `#include "tusb_config.h"`, which solves this issue.

### DIFF
--- a/hw/bsp/stm32f303disco/stm32f303disco.c
+++ b/hw/bsp/stm32f303disco/stm32f303disco.c
@@ -31,27 +31,29 @@
 // Forward USB interrupt events to TinyUSB IRQ Handler
 //--------------------------------------------------------------------+
 
-// USB defaults to using interrupts 19, 20, and 42 (based on SYSCFG_CFGR1.USB_IT_RMP)
+// USB defaults to using interrupts 19, 20 and 42, however, this BSP sets the
+// SYSCFG_CFGR1.USB_IT_RMP bit remapping interrupts to 74, 75 and 76.
+
 // FIXME: Do all three need to be handled, or just the LP one?
-// USB high-priority interrupt (Channel 19): Triggered only by a correct
+// USB high-priority interrupt (Channel 74): Triggered only by a correct
 // transfer event for isochronous and double-buffer bulk transfer to reach
 // the highest possible transfer rate.
-void USB_HP_CAN_TX_IRQHandler(void)
+void USB_HP_IRQHandler(void)
 {
   tud_int_handler(0);
 }
 
-// USB low-priority interrupt (Channel 20): Triggered by all USB events
+// USB low-priority interrupt (Channel 75): Triggered by all USB events
 // (Correct transfer, USB reset, etc.). The firmware has to check the
 // interrupt source before serving the interrupt.
-void USB_LP_CAN_RX0_IRQHandler(void)
+void USB_LP_IRQHandler(void)
 {
   tud_int_handler(0);
 }
 
-// USB wakeup interrupt (Channel 42): Triggered by the wakeup event from the USB
+// USB wakeup interrupt (Channel 76): Triggered by the wakeup event from the USB
 // Suspend mode.
-void USBWakeUp_IRQHandler(void)
+void USBWakeUp_RMP_IRQHandler(void)
 {
   tud_int_handler(0);
 }
@@ -126,6 +128,10 @@ void board_init(void)
   // 1ms tick timer
   SysTick_Config(SystemCoreClock / 1000);
   #endif
+
+  // Remap the USB interrupts
+  __HAL_RCC_SYSCFG_CLK_ENABLE();
+  __HAL_REMAPINTERRUPT_USB_ENABLE();
 
   // LED
   __HAL_RCC_GPIOE_CLK_ENABLE();
--- a/src/common/tusb_compiler.h
+++ b/src/common/tusb_compiler.h
@@ -44,8 +44,10 @@
 #endif
 
 // Compile-time Assert
-#if __STDC_VERSION__ >= 201112L
+#if defined (__STDC_VERSION__) && __STDC_VERSION__ >= 201112L
   #define TU_VERIFY_STATIC   _Static_assert
+#elif defined (__cplusplus) && __cplusplus >= 201103L
+  #define TU_VERIFY_STATIC   static_assert
 #else
   #define TU_VERIFY_STATIC(const_expr, _mess) enum { TU_XSTRCAT(_verify_static_, _TU_COUNTER_) = 1/(!!(const_expr)) }
 #endif
--- a/src/common/tusb_verify.h
+++ b/src/common/tusb_verify.h
@@ -142,7 +142,9 @@
 #define ASSERT_1ARGS(_cond)            TU_VERIFY_DEFINE(_cond, _MESS_FAILED(); TU_BREAKPOINT(), false)
 #define ASSERT_2ARGS(_cond, _ret)      TU_VERIFY_DEFINE(_cond, _MESS_FAILED(); TU_BREAKPOINT(), _ret)
 
+#ifndef TU_ASSERT
 #define TU_ASSERT(...)             GET_3RD_ARG(__VA_ARGS__, ASSERT_2ARGS, ASSERT_1ARGS,UNUSED)(__VA_ARGS__)
+#endif
 
 // TODO remove TU_ASSERT_ERR() later
 
@@ -163,10 +165,12 @@
 /* ASSERT Error
  * basically TU_VERIFY Error with TU_BREAKPOINT() as handler
  *------------------------------------------------------------------*/
-#define ASERT_ERR_1ARGS(_error)         TU_VERIFY_ERR_DEF2(_error, TU_BREAKPOINT())
-#define ASERT_ERR_2ARGS(_error, _ret)   TU_VERIFY_ERR_DEF3(_error, TU_BREAKPOINT(), _ret)
+#define ASSERT_ERR_1ARGS(_error)         TU_VERIFY_ERR_DEF2(_error, TU_BREAKPOINT())
+#define ASSERT_ERR_2ARGS(_error, _ret)   TU_VERIFY_ERR_DEF3(_error, TU_BREAKPOINT(), _ret)
 
-#define TU_ASSERT_ERR(...)         GET_3RD_ARG(__VA_ARGS__, ASERT_ERR_2ARGS, ASERT_ERR_1ARGS,UNUSED)(__VA_ARGS__)
+#ifndef TU_ASSERT_ERR
+#define TU_ASSERT_ERR(...)         GET_3RD_ARG(__VA_ARGS__, ASSERT_ERR_2ARGS, ASSERT_ERR_1ARGS,UNUSED)(__VA_ARGS__)
+#endif
 
 /*------------------------------------------------------------------*/
 /* ASSERT HDLR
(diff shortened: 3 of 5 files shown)


src/common/tusb_verify.h
```diff
@@ -120,7 +120,9 @@
 #define TU_VERIFY_1ARGS(_cond)                         TU_VERIFY_DEFINE(_cond, , false)
 #define TU_VERIFY_2ARGS(_cond, _ret)                   TU_VERIFY_DEFINE(_cond, , _ret)
 
+#ifndef TU_VERIFY
```
hathach, line 123: verify doesn't trigger any error, it is considered normal scenario for an VERIFY() to failed. Therefore I think we don't have a need to overwrite it.

src/common/tusb_compiler.h
```diff
@@ -44,7 +44,7 @@
 #endif
 
 // Compile-time Assert
-#if __STDC_VERSION__ >= 201112L
+#if defined (__STDC_VERSION__) && __STDC_VERSION__ >= 201112L
```
hathach, line 47: thanks, cpp has its own version  `static_assert` since C++11, it would be great if you could add ifelf for `__cplusplus` for this as well. It is OK if don't want to do it now, I will do it later on.

https://en.cppreference.com/w/cpp/language/static_assert

```
__cplusplus
This macro is defined when the C++ compiler is in use. You can use __cplusplus to test whether a header is compiled by a C compiler or a C++ compiler. This macro is similar to __STDC_VERSION__, in that it expands to a version number. Depending on the language standard selected, the value of the macro is 199711L for the 1998 C++ standard, 201103L for the 2011 C++ standard, 201402L for the 2014 C++ standard, 201703L for the 2017 C++ standard, or an unspecified value strictly larger than 201703L for the experimental languages enabled by -std=c++2a and -std=gnu++2a.
```

src/common/tusb_verify.h
```diff
@@ -131,7 +133,9 @@
 #define TU_VERIFY_HDLR_2ARGS(_cond, _handler)           TU_VERIFY_DEFINE(_cond, _handler, false)
 #define TU_VERIFY_HDLR_3ARGS(_cond, _handler, _ret)     TU_VERIFY_DEFINE(_cond, _handler, _ret)
 
+#ifndef TU_VERIFY_HDLR
```
hathach, line 136: like above, verify doesn't need to be overwritten.

src/common/tusb_verify.h
```diff
@@ -76,8 +76,8 @@
 
 #if CFG_TUSB_DEBUG
   #include <stdio.h>
-  #define _MESS_ERR(_err)   printf("%s %d: failed, error = %s\r\n", __func__, __LINE__, tusb_strerr[_err])
-  #define _MESS_FAILED()    printf("%s %d: assert failed\r\n", __func__, __LINE__)
+  #define _MESS_ERR(_err)   tu_printf("%s %d: failed, error = %s\r\n", __func__, __LINE__, tusb_strerr[_err])
```
hathach, line 79: thanks, nice catch. Actually I tried to update this once, but some error prevent this in a certain build. I forgot the reason, but look like it is addressed recently :).

src/common/tusb_verify.h
```diff
@@ -142,7 +146,9 @@
 #define ASSERT_1ARGS(_cond)            TU_VERIFY_DEFINE(_cond, _MESS_FAILED(); TU_BREAKPOINT(), false)
 #define ASSERT_2ARGS(_cond, _ret)      TU_VERIFY_DEFINE(_cond, _MESS_FAILED(); TU_BREAKPOINT(), _ret)
 
+#ifndef TU_ASSERT
```
hathach, line 149: I am ok with this, however overwriting the TU_ASSERT() can be a bit of works, since it does have a macro trick to take optional parameter. If you has issue with TU_BREAKPOINT(), maybe you only need to overwrite it

src/portable/st/stm32_fsdev/dcd_stm32_fsdev.c
```diff
@@ -150,6 +150,12 @@
 #  define USE_SOF     0
 #endif
 
+// Some STM32F3 devices allow to remap the USB interrupt vectors from
+// shared USB/CAN IRQs to separate CAN and USB IRQs
+#ifndef STM32F3_IRQ_REMAP
```
hathach, line 155: This is interesting, I didn't know about this, could you tell me more about the specific MCU that support this, and which section in the reference manual for this. I would like to read this up a bit, to see if we could have a better systematic solution. I kind of not wanting to have custom macro for specific DCD if there is a better way.

hathach (commented):
Thank for the PR, sorry for late response, I have been too busy lately. The PR looks great, your modification and explanation all make sense. I only want to know a bit more about the STM32 remap to see if we could implement it without having an extra DCD specific macro.

src/common/tusb_verify.h
salkinium: Agreed.

src/common/tusb_compiler.h
salkinium: Done.

src/common/tusb_verify.h
salkinium: Not that big an issue, since I can reuse the tinyusb variadic helpers for that purpose, and redirect to [our assertion mechanism](https://modm.io/reference/module/modm-tinyusb/#debugging-tinyusb):

```c
// Redirect TinyUSB asserts to use modm_assert
#define MODM_ASSERT_1ARGS(_cond) \
	TU_VERIFY_DEFINE(_cond, modm_assert(0, "tu", \
		__FILE__ ":" MODM_STRINGIFY(__LINE__) " -> \"" #_cond "\""), false)

#define MODM_ASSERT_2ARGS(_cond, _ret) \
	TU_VERIFY_DEFINE(_cond, modm_assert_continue_fail(0, "tu", \
		__FILE__ ":" MODM_STRINGIFY(__LINE__) " -> \"" #_cond "\""), _ret)

#define TU_ASSERT(...) GET_3RD_ARG(__VA_ARGS__, MODM_ASSERT_2ARGS, MODM_ASSERT_1ARGS,UNUSED)(__VA_ARGS__)
```

This prints out:

```
Assertion 'tu' failed!
  modm/ext/tinyusb/portable/st/synopsys/dcd_synopsys.c:524 -> "epnum < 4U"
Abandoning...
```

which gets treated as an URI by my terminal, and when I ctrl-click on it, opens the file at the exact line in my editor. The TinyUSB assert formatter prints the function name and line number, which wasn't helpful for someone unfamiliar with the code. The above makes it relatively clear that you've run out of endpoints.

src/portable/st/stm32_fsdev/dcd_stm32_fsdev.c
salkinium: This only exists for the [STM32F302 and the larger STM32F303 devices](https://github.com/modm-io/modm-devices/search?q=USBWakeUp_RMP).

See [RM0365 Table 40, page 208, footnote 1](https://www.st.com/resource/en/reference_manual/dm00094349-stm32f302xbcde-and-stm32f302x68-advanced-armbased-32bit-mcus-stmicroelectronics.pdf#page=208) (STM32F302) and [RM0318 Table 82, page 288, footnote 1](https://www.st.com/resource/en/reference_manual/dm00043574-stm32f303xb-c-d-e-stm32f303x6-8-stm32f328x8-stm32f358xc-stm32f398xe-advanced-arm-based-mcus-stmicroelectronics.pdf#page=288) (STM32F303xb/c/d/e):

> 1. It is possible to remap the USB interrupts (USB_HP, USB_LP and USB_WKUP) on interrupt lines 74, 75 and 76 respectively by setting the USB_IT_RMP bit in the Section 11.1.1: SYSCFG configuration register 1 (SYSCFG_CFGR1) on page 172.

salkinium: I guess you could read the bit at runtime and select the right interrupts. Wouldn't be too much overhead.

hathach: @salkinium sorry for late response, this is great idea to solve the shared CAN/USB. I like your last suggestion by checking the actual remap bit. The overhead is too small ( a couple of cycle for bit test) and this aren't call at any fast rate either. Would you mind changing it to dynamic check. Also since it only available to selected chips, it should be wrapped with `SYSCFG_CFGR1_USB_IT_RMP` e.g

```
#ifdef SYSCFG_CFGR1_USB_IT_RMP
if ( remapped )
{
 use RMP IRQ
}else
#endif
{
  use normal IRQ 
}
```

hathach (commented):
Thanks for the great idea with remap, would you mind taking a bit more time to make it dynamically check of the remapped bit in the SYSCFG_CFGR1, that would be perfect.

src/portable/st/stm32_fsdev/dcd_stm32_fsdev.c
salkinium: I implemented it like this and remapped the IRQs on the STM32F303 DISCO BSP.

hw/bsp/stm32f303disco/stm32f303disco.c
```diff
@@ -127,6 +129,9 @@ void board_init(void)
   SysTick_Config(SystemCoreClock / 1000);
   #endif
 
+  // Remap the USB interrupts
+  __HAL_REMAPINTERRUPT_USB_ENABLE();
```
hathach, line 133: I pull out my f3 disco board to test with, I believe we need to enable the syscfg clk first for the remap to take effect.

```
__HAL_RCC_SYSCFG_CLK_ENABLE();
__HAL_REMAPINTERRUPT_USB_ENABLE();
``

hathach (commented):
almost perfect, though hand-on testing with f3 disco prove that syscfg clk must be enabled before we could do the remap.

hw/bsp/stm32f303disco/stm32f303disco.c
salkinium: Whoops, I got distracted with work, so I forgot to test this in hardware. Fixed now.

hathach (approved):
perfect now, thank you very much for your PR. I will merge when ci passed.

hw/bsp/stm32f303disco/stm32f303disco.c
hathach: no problem at all, writing code is much harder than testing :)

Issue #504, opened 2020-09-03
Convert EP buffers to ring buffers (at least for ISO EPs)


PanRe:
Hello,

Right now @kasjer and i are working on finalizing a working version for an audio 2.0 driver. During tests a possible bottleneck came up and i wanted your opinion on this issue. In general a FIFO is necessary in order to decouple the sampling process from the USB transmission process. 

So the setup right now (e.g. for a microphone) is: Samples -> FIFO -> EP buffer -> USB FIFO, where -> means "put into".

As how tinyUSB is set up right now, all EP buffers are considered to be linear buffers, however, at least for ISO EPs, ring buffers (or FIFOs if you like) would be far more practical. The benefits would be:

- Less memory space required as the EP ring buffer would not need to be as large as the current FIFO plus the linear EP buffer
- One unnecessary copy process less
- The ring buffer can be filled by use of a DMA in circular mode with little intervention (picture an I2S interface which directly moves its samples to USB, where the only thing to do would be to update the write pointer of the ring buffer adequately)

However, this would mean a rather big amount of work to do since all implementations of dcd_edpt_xfer() must be adapted. A possible compromise would be to setup a new function exclusivly for ISO EPs e.g. dcd_edpt_ISO_xfer() which gets a pointer to a FIFO. Within the device controler drivers, we could check if the EP is an ISO EP and if so read the data from there.

All reasons stated above also hold in case of OUT EPs.

What do you think of this?

Talking about ring buffers:

For the current FIFO implementation (which are ring buffers) i would also suggest to change from relative to absolute pointers (or unmasked indices), which is very nicely described [here](https://www.snellman.net/blog/archive/2016-12-13-ring-buffers/). The benefits are:

- You don't loose one byte since empty and full are distinguishable from the unmasked or absolute pointers
- An overrun can be determined (also nice for DMA use)

In the article, the editor says that ring buffers with unmasked pointers can only work for buffer lengths which are a power of two (and wrap around of unsigned integers must be possible). However, i think i came up with a solution to this problem (making any length possible). So if you are interested please let me know!

Also in your current implementation of the FIFOs i would recommend you not to store a count variable and modify it in both pull() and push() as the count variable might get messed up if two processes modify it (the writing of the count variable is not atomic). Rather i would compute count all time new.

Thanks for your opinions!
Reinhard

hathach:
> Hello,
> 
[...]

PR #521, opened 2020-09-28
Uac2

PanRe:
**Describe the PR**
This PR adds a USB audio 2.0 (UAC2) driver.

First a very BIG thanks to @kasjer for helping finalizing and testing this driver! It took quiet some time to set all of this up and there definitely is room for improvement. But as a first start this sould be fine!

**Additional context**
In general
- multiple audio functions (drivers) are useable in parallel.
- at most one IN EP, one OUT EP, one feed back EP, and one INT EP are supported per audio function. This should be enough for 99% of all applications.
- FIFO structures are supported for TX and/or RX applications. If desired, FIFOs can be set up for individual channels, or for a more performant approach, all channels can be put into one single FIFO. FIFOs for individual channels are intended for quick start and debugging purpose. Currently individual channels are encoded in PCM format only. In case of only one FIFO you are responsible for encoding yourself (which allows for optimization). A third option to avoid FIFOs completely and put data directly into the EP buffer (which is the last software buffer before hardware buffers of MCU follow) is currently under development  #504. This version should also allow for DMA use.
- A big part of audio requests are already supported.
- Only required code is enabled (e.g. if only IN EPs are required no OUT EP code is enabled etc.).
- There is a big bunch of callback functions available to customize your applications.

There were some changes made to dcd_synopsis.
When a USB configuration is set up, within dcd_synopsis the current configuration descriptor is scanned through and a report is set up which EPs (type and size) are used. From this the worst case scenario is determined and the USB FIFO is allocated accordingly. This avoids nasty fragmentation issues when opening and closing EPs since their corresponding USB FIFO sizes are fixed and need to be set up only once. While this approach restricts the FIFO allocation flexibility, it guarantees a smooth operation. Once someone has found a way how to solve defragmentation this can be improved further.
Also ISO EPs in STM32 have a strange even/odd number framing which needs to get handled.

For this to work, within usbd.c the function `dcd_alloc_mem_for_conf()` was implemented triggering the aforementioned configuration scanning process.

Within usbd.h, descriptors were added for mono microphone and mono speaker in a sophistcated way to allow people to setup new descriptors quickly. However, there is no way around reading UAC2 specifications if you are not familiar with it already.

Tested with PCM 16 bits samples on STM32F411RE:
- sending and receiving audio requests
- mono and stereo microphone
- mono and stereo speaker with and without feedback EP


ladyada:
cool!!

hathach:
wow wow wow, this is **astounding**, thank you very much @PanRe and @kasjer   for this awesome work. This is quite a lots of changes, please give me a few days for proper reviewing and testing. I may try to merge the  fifo enhancement #516 first. 

> @hathach There might be the need to adjust the unit tests because it does not know anything of dcd_alloc_mem_for_conf() and it complains that this function is called too often. Can assist here? Thanks a lot!

no problem, don't worry about the unit test, it does its work to catch unknown (to its) DCD API. I will update it accordingly. 

> Also for the SAMD11 device the linking process seems not working. A similar issue was stated here.

Thanks for the head up, the end symbol is mostly for malloc/heap related function when linking with newlib nano. As far as I remember it is only used for static check SRAM overflow (no dynamic heap check).  I will check this out later on,.

src/tusb_option.h
```diff
@@ -171,7 +171,7 @@
 
 // Debug enable to print out error message
 #ifndef CFG_TUSB_DEBUG
-  #define CFG_TUSB_DEBUG 0
+  #define CFG_TUSB_DEBUG 2
```
atanisoft, line 174: I think this update should be reverted so the debug output is off by default when not explicitly enabled.

PanRe: Well spotted!

PanRe:
> 
> 
> wow wow wow, this is **astounding**, thank you very much @PanRe and @kasjer for this awesome work. This is quite a lots of changes, please give me a few days for proper reviewing and testing. I may try to merge the fifo enhancement #516 first.
> 
> > @hathach There might be the need to adjust the unit tests because it does not know anything of dcd_alloc_mem_for_conf() and it complains that this function is called too often. Can assist here? Thanks a lot!
> 
> no problem, don't worry about the unit test, it does its work to catch unknown (to its) DCD API. I will update it accordingly.
> 
> > Also for the SAMD11 device the linking process seems not working. A similar issue was stated here.
> 
> Thanks for the head up, the end symbol is mostly for malloc/heap related function when linking with newlib nano. As far as I remember it is only used for static check SRAM overflow (no dynamic heap check). I will check this out later on,.

The issue with the end symbol seems to be fixed by setting debugging messages disabled by default (as indicated by @atanisoft).

zlittell:
Everytime you try to turn debug or log on you will get the error back. It can be fixed by adding hw/mcu/microchip/samd11/hal/utils/src/utils_syscalls.c \
to the SRC_C += in side board.mk for the samd11

this shouldn't be necessary since there is a -specs=nosys.specs in the rules.mk file, so it is probably correct that the linker is wrong, but I can't find where. It matches quite a few examples.

src/portable/st/synopsys/dcd_synopsys.c
```diff
@@ -1040,6 +1114,210 @@ void dcd_int_handler(uint8_t rhport)
...
+// The idea is to use this information such that the FIFOs need to be configured only once
+// at configuration time and no more later. This makes it easy in case you want to close and open EPs for different
+// purposes at any time (without taking care of other active EPs). You also avoid the nasty need of defragmenting
+// the TX buffers, which is likely to happen.
+// Certainly, this function does not allow for the highest possible flexibility as it only works in the worst case
+// (all biggest EPs can be active at the same time). However, this should not be a problem for almost all applications.
+
+// Spare space is assigned equally divided to bulk and interrupt EPs.
+// Pure isochronous EPs do not get any spare space as it does not make any sense.
+TU_ATTR_WEAK bool dcd_alloc_mem_for_conf(uint8_t rhport, tusb_desc_configuration_t const * desc_cfg)
```
hathach, line 1171: To be honest, I don't like the introduction of this function. defragmenting is not a real issue, it is mostly switching between open/close ISO endpoint, which could be addressed in other way like introducing `dcd_edpt_iso_open` with additional max packet size of all alternate Itf. If you could revert this change and still having the audio_test example running. I will update the ISO and dcd to better support open/close endpoint.

hathach (commented):
Thank you very much for this superb PRs with lots of works and modification. Due to its large work, I have only reviewed a small portion of this PR. I got into issue with building audio_test example, and attempt to fix it with PR to PR here https://github.com/PanRe/tinyusb/pull/3 

I got the stm32f411 enumerated as audio device and response to several of host messages. Though since the audio_test doesn't do much now, we couldn't verify its actual functionality. Though that is definitely can be done later. I will try to have some board API with I2C to output some sound and maybe we can make and `audio_speaker` example and it is beyond this PR. 

However, I don't really like the concept of `dcd_alloc_mem_for_conf()`, although it is optional, the knowledge and parsing configuration descriptors is beyond the DCD layer and should really be on the USBD instead. DCD should only worry about low level API that interact directly with endpoints. It is true that the current set of DCD API cause the fragmentation of FIFO endpoint in the case of synopsys (not with other though). However, I do have some idea to overcome this (by adding additional API, parammeter specifically to iso). This is also useful since other mcu with similar dynamic allocation scheme can also be supported as well. Therefore If possible would you please revert the dcd_synopsys to use the current allocation scheme. As long as it works I could work on the new dcd API to overcome the fragmentation without having dcd knowing too much.

PS: It probably take a few rounds of review, please be patient. Thanks again for this great PR.

src/class/audio/audio_device.h
```diff
@@ -0,0 +1,399 @@
...
+// For RX: the input stream gets decoded into its corresponding channels, where for each channel a FIFO is setup to hold its data -> see: audio_rx_done_cb().
+// For TX: the output stream is composed from CFG_TUD_AUDIO_N_CHANNELS_TX channels, where for each channel a FIFO is defined.
+// Further, it implements encoding and decoding of the individual channels (parameterized by the defines below).
+// If you don't use the FIFOs you need to handle encoding and decoding on your own in audio_rx_done_cb() and audio_tx_done_cb(). This, however, allows for optimizations.
+
+#ifndef CFG_TUD_AUDIO_TX_FIFO_SIZE
+#define CFG_TUD_AUDIO_TX_FIFO_SIZE  0                           // Buffer size per channel
+#endif
+
+#if CFG_TUD_AUDIO_TX_FIFO_SIZE && CFG_TUD_AUDIO_TX_DMA_RINGBUFFER_SIZE
```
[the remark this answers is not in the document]
PanRe: Actually this CFG_TUD_AUDIO_TX_DMA_RINGBUFFER_SIZE is not necessary any more in case of the new unmasked FIFO. I remove it completely.

src/class/audio/audio_device.h
```diff
@@ -0,0 +1,399 @@
...
+#ifndef CFG_TUD_AUDIO_FORMAT_TYPE_TX
+#define CFG_TUD_AUDIO_FORMAT_TYPE_TX                AUDIO_FORMAT_TYPE_UNDEFINED     // If this option is used, an encoding function has to be implemented in audio_device.c
+#endif
+
+#ifndef CFG_TUD_AUDIO_FORMAT_TYPE_RX
+#define CFG_TUD_AUDIO_FORMAT_TYPE_RX                AUDIO_FORMAT_TYPE_UNDEFINED     // If this option is used, a decoding function has to be implemented in audio_device.c
+#endif
+
+// Audio data format type I specifications
+#if CFG_TUD_AUDIO_FORMAT_TYPE_TX == AUDIO_FORMAT_TYPE_I
```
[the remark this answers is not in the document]
PanRe: Very well spotted! Do you think it is worth reworking the defines in audio_device.h such that `AUDIO_FORMAT_TYPE_I` is an enumeration again?

src/class/audio/audio_device.h
```diff
@@ -0,0 +1,399 @@
...
+#endif
+
+/*  This function is intended for later use once EP buffers (at least for ISO EPs) are implemented as ring buffers
+#if CFG_TUD_AUDIO_EPSIZE_IN && !CFG_TUD_AUDIO_TX_FIFO_SIZE
+uint16_t tud_audio_n_write_ep_in_buffer(uint8_t itf, const void * data, uint16_t len)
+#endif
+*/
+
+#if CFG_TUD_AUDIO_EPSIZE_IN && CFG_TUD_AUDIO_TX_FIFO_SIZE
+#if CFG_TUD_AUDIO_TX_FIFO_COUNT > 1
```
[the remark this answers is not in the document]
PanRe: Thanks for spotting - there are many configurations possible and i did not catch up checking all of them at the end!

src/class/audio/audio_device.h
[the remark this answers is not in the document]
hathach: Great 👍

src/class/audio/audio_device.h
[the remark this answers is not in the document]
hathach: Yeah, we could. But it can wait until we finalize the api. There is little different in the usage of macro and enum in this case. There is no hurry at all

src/class/audio/audio_device.h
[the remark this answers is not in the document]
hathach: > Thanks for spotting - there are many configurations possible and i did not catch up checking all of them at the end!

No problem :)

src/portable/st/synopsys/dcd_synopsys.c
PanRe: I see your reason why it would be better to have such a configuration parsing within USBD and not in DCD. This totally makes sense since other MCUs may need the same. However, i am not sure if handling ISO EPs without taking care of the remaining EPs is sufficient! Your idea of reserving space only for the ISO EPs is similar to my approach but i also take care that there is space for all used remaining EPs. So an overflow can be avoided and detected at compilation time. Every EP has its dedicated space available so you also do not need any memory management.
The very best thing would be to be able to close EPs and defragment the FIFO right away. In this case you would not need any pre-FIFO-allocation scan at all. But i am not sure if this is easily done on the fly.

If you insist i certainly revert the allocation scheme but i think it would be worth considering the alternatives.

hathach: >  Your idea of reserving space only for the ISO EPs is similar to my approach but i also take care that there is space for all used remaining EPs.

No I didn't say we will reserve space for ISO only. First of all the only endpoint that get closed and reopened is only ISO. It is unspoken truth that for all the ISO endpoints, its default interface (alternate = 0) must has no endpoint (closed ISO). Therefore ISO will likely be the last endpoints got opened. All other endpoints are opened when device receives **SET_CONFIGURED** by their driver. The ISO will need an additional **SET_INTERFACE** to actually open the atl one with corresponding packet size.

What I am trying to say is we don't have to deal with it like malloc() when open()/close() randomly, only few last endpoints are closed and re-open. When the usbd/class driver go through all the Alt Interface (e.g Audio streaming), It will know the max size of all the ATL e.g Atl 1 has size = 100, atl 2 size = 125 . The first time it open the endpoint even though with size = 100, it will also pass the max size of 125 so that dcd can reserve that space for easy close/reopen. If the FIFO size does not afford the max size of 125 for ISO, the description shouldn't include that Alt with 125 (should only support 100 e.g).

Yeah, sure it needs a bit more work now which I will be happy to do later on, however I think any higher level than endpoint should be decided by USBD/class driver.

PanRe: Ah now the picture gets clearer. Actually a smart thought! Your are right, the interfaces which need to be activated by **SET_INTERFACE** will use what is left from the other EPs. I always have this highly complicated picture in mind, where the end user may has multiple multi-interface USB applications running (similar to audio). I guess this comes from reading too many USB specifications ^^. They always suggest you may close interfaces currently not needed in order to allow a bigger interface to be opened and so on.

I can see that your approach works definitely if you have only one audio interface running. In case you have two such interfaces you need additional logic to recognize if a FIFO was already setup with a bigger space (e.g. 125). This is the memory management i meant before.

To summarize you and i try to solve the very same problem in a very similar manner. Actually you definitely too reserve space but only for ISO EPs.

> The first time it open the endpoint even though with size = 100, it will also pass the max size of 125 so that dcd can reserve that space for easy close/reopen.

This brings the need for memory management and additional APIs for ISO EPs. My approach parses everything at the beginning, sets the FIFO once and does not change anything later on. So there is no memory management needed but it needs to be done once a new configuration gets set and can not be done by a call to ep_open().

I suggest i revert my approach for now. If desired, after all it is your choice at the end, we could introduce this in a separate PR.

src/class/audio/audio_device.h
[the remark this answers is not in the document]
PanRe: Ok!

src/portable/st/synopsys/dcd_synopsys.c
hathach: > Ah now the picture gets clearer. Actually a smart thought! Your are right, the interfaces which need to be activated by **SET_INTERFACE** will use what is left from the other EPs. I always have this highly complicated picture in mind, where the end user may has multiple multi-interface USB applications running (similar to audio). I guess this comes from reading too many USB specifications ^^. They always suggest you may close interfaces currently not needed in order to allow a bigger interface to be opened and so on.

Yeah, USB docs is over-engineer for sure, but they do a really good job with 2.0. The term after 3.0/3.1 gen1 gen2 are too messy :laughing: 

> This brings the need for memory management and additional APIs for ISO EPs. My approach parses everything at the beginning, sets the FIFO once and does not change anything later on. So there is no memory management needed but it needs to be done once a new configuration gets set and can not be done by a call to ep_open().  
 
We may need one but it is not really complicated. I think for stm32, we only need to update the fifo index with a smaller size, there is probably no need to shrink + combine fifo together. Actually, your implementation still need to handle this case of  SET_INTERFACE request for changing the endpoint size as well e.g in video class the ISO endpoint size is determined by resolution in Video Streaming Data
- ALT 0 : no endpoint
- ATL 1: 800 x 600 : size = 200
- ATL 2: 1024 x 768 : size = 250
- ATL 3: 1280 x 720 : size = 300

And this switch dynamically within an same configuration descriptor depending on the user selection of the resolution. What I propose is the class driver take note of the max possible ATL = 300, then pass/set it to reserve the space, from there it could switch back and forth with ease (only change the fifo size) and it shouldn't affect the index and space of the endpoint after it (multiple audio or video + audio).

> I suggest i revert my approach for now. If desired, after all it is your choice at the end, we could introduce this in a separate PR.

I am glad to hear you think my approach make sense. Yeah, I think if needed, we could have this a its own PR. Once we have everything working (now It should since the ISO is the last), and we can work from there for optimization. 

PS: The next PR I will love to work on is getting the `audio_speaker` or `audio_microphone` so that we could visually see or rather hear the actual result.

PanRe: > Actually, your implementation still need to handle this case of SET_INTERFACE request for changing the endpoint size as well...

Thats acutally not correct, my implementation would recognize ATL 3 and reserve the required 300 in the first place. It determines the worst case i.e. the biggest required FIFO size for every EP ever needed and sets up the FIFO this way. Thus the FIFO never needs to be reconfigured even in case of open/close EPs and it does not need any management.

Regarding the examples, I can update the microphone example but a speaker example would best be given by @kasjer since he already tested the speaker case.

kasjer: Both microphone and speaker in real life require code that is capable of handling sound.
This is very hardware dependent and I don't think we can have out of the box code that will work on more then one platform.
For microphone we can have some sort of noise generator that will produce stream that host can see.
For speaker it is a little more difficult, audio data has to go somewhere.

For my part I'm using I2S microphone and I2S speaker that works with STM32F4 MCUs.
Speaker also works with NRF52 (microphone is not working due to NRF52 I2S 24 bits limitation, microphone accepts only LRCL=BCLK/64).
Speaker also works with DA1469x (16 bits stereo) (microphone not done yet due to I2S not being ready in mynewt).

I though that I will do headset instead that will have both microphone and speaker at the same time.
That way we can have an example that does not need any specific hardware and will simply route speaker 
to microphone (stereo to mono).

This partially works on Linux, Windows rejects my headset device, Mac I did not checked yet.

hathach: > Thats acutally not correct, my implementation would recognize ATL 3 and reserve the required 300 in the first place. It determines the worst case i.e. the biggest required FIFO size for every EP ever needed and sets up the FIFO this way. Thus the FIFO never needs to be reconfigured even in case of open/close EPs and it does not need any management.

Yeah, I mean I know you reserve the largest space for ISO. But we still need to resize the ep when SET_INTERFACE with different ep size is received. For stm32, maybe we can leave it with max fifo size and should be no issues, other mcus may need different approach. I kind of mixed-up some thought along the way across MCUs approach.
 
> Regarding the examples, I can update the microphone example but a speaker example would best be given by @kasjer since he already tested the speaker case.

> Both microphone and speaker in real life require code that is capable of handling sound.
> This is very hardware dependent and I don't think we can have out of the box code that will work on more then one platform.
> For microphone we can have some sort of noise generator that will produce stream that host can see.
> For speaker it is a little more difficult, audio data has to go somewhere.
> 
> For my part I'm using I2S microphone and I2S speaker that works with STM32F4 MCUs.
> Speaker also works with NRF52 (microphone is not working due to NRF52 I2S 24 bits limitation, microphone accepts only LRCL=BCLK/64).
> Speaker also works with DA1469x (16 bits stereo) (microphone not done yet due to I2S not being ready in mynewt).
> 
> I though that I will do headset instead that will have both microphone and speaker at the same time.
> That way we can have an example that does not need any specific hardware and will simply route speaker
> to microphone (stereo to mono).
> 
> This partially works on Linux, Windows rejects my headset device, Mac I did not checked yet.

Superb !!! Thank you very much, that would be great to have a visually working example makes it super easy when refactoring/enhancing let's say using DMA fifo/ EP allocation scheme :).  There is no rush though, and we don't need to have a complete portable example. Not all the boards has built-in mic/speaker anyway, and I am totally agreed with the noise/sound generator that would make the example runs on most of the boards. The speaker is optional, we can start with specific supported board, then move up later on. I could help with that :).

PanRe: > Yeah, I mean I know you reserve the largest space for ISO. But we still need to resize the ep when SET_INTERFACE with different ep size is received. For stm32, maybe we can leave it with max fifo size and should be no issues, other mcus may need different approach. I kind of mixed-up some thought along the way across MCUs approach.

That is correct, the EP needs to know its current size but the FIFO size is fixed to maximum required size so there will never occur any issues once everything was setup. I also don't know how other MCUs work here, i only use STMs. I am reverting the UAC2 driver and we handle this in a separate PR. Some additional time to allow creativity hammering out the best approach is never a bad idea! ;)

For testing purpose, currently i am testing my application by sending a saw-tooth signal (i just send increasing values until it overflows and so forth). This does not sound nice but one can check visually (e.g. via Python or other software capturing and plotting a sound snippet) if there are no missing values. You may do the same for speaker and check via a debugger the content of the receive FIFO. Those are manual testing procedures i know. The noise and especially the head-set example are a brilliant idea!

hathach:
@PanRe do you have time to revert the dcd synopsys changes ? I would like to merge it this weekend to try to see if I could do further testing with audio this weekend. If you don't have time, I could merge it as it is and do the revert my self afterward (basically overwrite it with the current one) :D. Nice thing is we will have these modification in git log whenever we want to pull it up in the future :D

PanRe:
> 
> 
> @PanRe do you have time to revert the dcd synopsys changes ? I hope to merge it this and try to see if I could do further testing with audio this weekend. If you don't have time, I could merge it as it is and do the revert my self afterward (basically overwrite it with the current one) :D. Nice thing is we will have these modification in git log whenever we want to pull it up in the future :D

Done! I had to keep some changes @kasjer introduces for ISO EPs. I also extended the example to send a saw-tooth signal. It sweeps over the complete int16 range. Hope it works as i am not able to test the examples since i only have my custom PCB here.

hathach (approved):
thank you very much and also @kasjer for this brilliant PR. This will be a very interesting & exciting feature for the stack. I cannot say thank enough to you two.

PanRe:
> thank you very much and also @kasjer for this brilliant PR. This will be a very interesting & exciting feature for the stack. I cannot say thank enough to you two.

You are very welcome. There is still much to do yet! But i am happy that the driver finally made it into the stack! :)

hathach:
> > thank you very much and also @kasjer for this brilliant PR. This will be a very interesting & exciting feature for the stack. I cannot say thank enough to you two.
> 
> 
> 
> You are very welcome. There is still much to do yet! But i am happy that the driver finally made it into the stack! :)

Yeah, lots left to be done, especially audio normally involved dsp work. But it is a great milestone 👍

ddavidebor:
Hi @PanRe @kasjer could you get in touch with me? Looking for USB Audio Class 2.0 Developers for some contract work davide.bortolami@thinksmartbox.com

Issue #522, opened 2020-09-30
2+ HID Interfaces on a single device


zlittell:
**Describe what the question is**
I see the composite example uses a keyboard and mouse on the same interface. When I try to add multiple interfaces I start to get "device not started" issues. The descriptors come in and then it stalls. Is what I am trying to do supported by TinyUSB? If so can someone who has experience with it give me some pointers on what needs to be done to get this type of device rolling.

Most other libraries have you list the report descriptors separately per interface.

Here is a GIST to play with:
https://gist.github.com/zlittell/ff8ca12fdf11228a6f95212dd6a2cb31

hathach:
90% of the work for multiple HIDs are done, though it is not very often used by most application. If you are willing to submit pr to add HID multiple interfaces support. I am happy to review. You could take a look at  multiple CDC API for reference 
https://github.com/hathach/tinyusb/blob/master/src/class/cdc/cdc_device.h#L61

zlittell:
working on this now. making good progress, so hopefully all works as planned

hathach:
> working on this now. making good progress, so hopefully all works as planned

great to hear that, the most annoying part is we have to add `itf` arguments to all of the callbacks() API, which break current applications. I should have added that in the first place.

PR #524, opened 2020-10-01
Feature multiple hid interfaces

zlittell:
**Describe the PR**
References #522 
Allows multiple hid interfaces within a single device. Includes example as well.

**Additional context**
https://github.com/hathach/tinyusb/commit/b7208d6f7e97e724853eb5d0e512a23005cc8790
This needs to be changed because it breaks all legacy hid usb_descriptors.c files. There should be a better way to do this, but I thought I would just get it done and then pose the question here. I need to be able to call tud_hid_descriptor_report_cb and pass desc_index as well as pass nothing so old stuff still works.
We could do a simple check for  CFG_TUD_HID > 1 and then just execute the right callback depending on that. Hoping theres a craftier way to accomplish this though.


src/class/hid/hid_device.c
```diff
@@ -237,7 +237,7 @@ bool hidd_control_request(uint8_t rhport, tusb_control_request_t const * request
     }
     else if (request->bRequest == TUSB_REQ_GET_DESCRIPTOR && desc_type == HID_DESC_TYPE_REPORT)
     {
-      uint8_t const * desc_report = tud_hid_descriptor_report_cb();
+      uint8_t const * desc_report = tud_hid_descriptor_report_cb((uint8_t) request->wIndex);
```
hathach, line 240: I think there is a bit of mixing-up here. The `wIndex` is actually the interface number of the HID interface and not the `desc_index`. It just happens to be the same 0, 1 order as the example. It will be different, if we have additional class before and/or between the hid interface.

What we should do here is use the wIndex to figure out the index of the targeted interface in hid pool and pass it as `itf` to the callback. It is a foor-loop comparing the wIndex agains the `itf_num` of the interface.
   
image

examples/device/hid_multipleinterface/src/usb_descriptors.c
```diff
@@ -0,0 +1,189 @@
...
+  return (uint8_t const *) &desc_device;
+}
+
+//--------------------------------------------------------------------+
+// HID Report Descriptor
+//--------------------------------------------------------------------+
+
+uint8_t const desc_hid_report1[] =
+{
+  TUD_HID_REPORT_DESC_KEYBOARD( HID_REPORT_ID(REPORT_ID_KEYBOARD) )
```
hathach, line 76: this is incorrect use of the template, since it is single report it should be simply

`uint8_t const desc_hid_report1[] =  { TUD_HID_REPORT_DESC_KEYBOARD() }`

examples/device/hid_multipleinterface/src/usb_descriptors.c
```diff
@@ -0,0 +1,189 @@
...
+};
+
+uint8_t desc_hid_report2[] =
+{
+  TUD_HID_REPORT_DESC_MOUSE   ( HID_REPORT_ID(REPORT_ID_MOUSE) )
```
hathach, line 81: same as above

examples/device/hid_multipleinterface/src/usb_descriptors.h
```diff
@@ -0,0 +1,34 @@
+/*
```
hathach, line 1: This file can be removed since it is not needed. Each interface in this example is single report and does not need report id.

src/class/hid/hid_device.h
```diff
@@ -50,33 +50,44 @@
...
 
 //--------------------------------------------------------------------+
 // Callbacks (Weak is optional)
 //--------------------------------------------------------------------+
 
 // Invoked when received GET HID REPORT DESCRIPTOR request
 // Application return pointer to descriptor, whose contents must exist long enough for transfer to complete
-uint8_t const * tud_hid_descriptor_report_cb(void);
+// TODO Talk about this change... because it is breaking. Might be better way to handle.
+uint8_t const * tud_hid_descriptor_report_cb(uint8_t desc_index);
```
hathach, line 90: `desc_index` should be changed to `itf` for correct meaning.

hathach (changes requested):
Thank you for your PR, everything look good except for the `tud_hid_descriptor_report_cb()` that needs to be updated per review. Yeah, regarding the backward compatible, Ideally all the callback has need to have additional `itf` argument for application to know which interface that event occurred all, and this sadly break the existing code. The best approach now is 
- when `CFG_TUD_HID == 1`  we will keep the current single interface API and define the callback_n() version as macro that expand to single interface one
- when ``CFG_TUD_HID > 1`  callback_n() are required and there is no single interface API

It is a bit more complicated that it should be however, it retain backward-compatible and give other more time to migrate. Eventually, we will make a big release and tell people that should upgrade to use callback_n() version (just kip itf arg if they don't use like the cdc API).

src/class/hid/hid_device.c
```diff
@@ -70,17 +70,17 @@ static inline hidd_interface_t* get_interface_by_itfnum(uint8_t itf_num)
 //--------------------------------------------------------------------+
 // APPLICATION API
 //--------------------------------------------------------------------+
-bool tud_hid_ready(void)
+bool tud_hid_n_ready(uint8_t itf)
 {
-  uint8_t const itf = 0;
+  //uint8_t const itf = 0;
```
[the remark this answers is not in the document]
zlittell: did this when playing with the feature and forgot to pull them out. Will do

examples/device/hid_multipleinterface/src/usb_descriptors.c
zlittell: Realized I copy and pasted these with the report IDs later in the day haha.

src/class/hid/hid_device.c
zlittell: We should be able to use the lower byte of wValue to determine the descriptor index. The setup packet has the correct index in it when sniffing with a beagle. I set up some traps to debug this and wValue is always 0x2200 instead of 0x2200 and then 0x2201. Trying to track down why, I feel like this is the right way to solve this problem if it isn't crazy. That way we are relying on the spec instead of a hack.

zlittell: image
I wanted to include an image here. It is super wild, I have traced the data all the way back and can't figure out where it would be manipulated. It seems to just come in on the FIFO read so how would 1 byte be wrong on the setup packet.

src/class/hid/hid_device.h
zlittell: I determined the descriptor index in hid_device. You are asking the descriptor report callback to return the correct hid descriptor based on the index requested not based on the interface.

zlittell:
I moved the issue with wValue out to its own bug for tracking and implemented the hack here so that this works for now.

src/class/hid/hid_device.h
zlittell: realized there is a difference in definition between itf_num and itf. switched this over in the latest code.

zlittell:
everything should be set up now with callbacks depending on if CFG_TUD_HID>1 or not. Let me know if I am missing anything.

src/class/hid/hid_device.c
```diff
@@ -67,20 +67,28 @@ static inline hidd_interface_t* get_interface_by_itfnum(uint8_t itf_num)
...
 }
 
+static inline uint8_t get_hid_index_by_itfnum(uint8_t itf_num)
+{
+	for (uint8_t i=0; i < CFG_TUD_HID; i++ )
+	{
+		if ( itf_num == _hidd_itf[i].itf_num ) return i;
+	}
+
+	return 0;
```
hathach, line 77: 0 is valid return, you should return either -1 or 0xff and check the return in the caller. But it is ok to leave it as it is, I will merge the usage of `get_hid_index_by_itfnum & get_interface_by_itfnum` as follow up PR as well.

.github/workflows/build.yml
```diff
@@ -35,13 +35,14 @@ jobs:
         - 'device/hid_composite'
         - 'device/hid_composite_freertos'
         - 'device/hid_generic_inout'
+        - 'device/hid_multipleinterface'
```
hathach, line 38: it is a bit late, but `hid_multiple_interface` would be a better name. But no problems, I will do rename later on.

examples/device/hid_multipleinterface/src/main.c
```diff
@@ -0,0 +1,208 @@
...
+  {
+    if ( btn )
+    {
+      int8_t const delta = 5;
+
+      // no button, right + down, no scroll pan
+      tud_hid_n_mouse_report(mouse_interface, 0, 0x00, delta, delta, 0, 0);
+
+      // delay a bit before attempt to send keyboard report
+      board_delay(10);
```
hathach, line 161: There is no need to delay since they are on separated interfaces. Previously itf is shared, so we have to wait before sending another one.

src/class/hid/hid_device.h
```diff
@@ -50,42 +50,64 @@
...
+static inline bool tud_hid_keyboard_report(uint8_t report_id, uint8_t modifier, uint8_t keycode[6]);
+static inline bool tud_hid_mouse_report(uint8_t report_id, uint8_t buttons, int8_t x, int8_t y, int8_t vertical, int8_t horizontal);
 
 //--------------------------------------------------------------------+
 // Callbacks (Weak is optional)
 //--------------------------------------------------------------------+
 
 // Invoked when received GET HID REPORT DESCRIPTOR request
 // Application return pointer to descriptor, whose contents must exist long enough for transfer to complete
+#if CFG_TUD_HID>1
```
hathach, line 89: Not very elegant approach, but it does has backward-compatible. This only serves as interim. After a couple of release, I think we could basically tell user to migrate to _n API.

hathach (approved):
Everything looks good, there are several minors review, but they are all optional. I could do a follow-up pr for that later on. Thank you for putting effort into this PR.

examples/device/hid_multipleinterface/src/usb_descriptors.c
```diff
@@ -0,0 +1,188 @@
...
+  if (itf == 0)
+	{
+		return desc_hid_report1;
+	}
+	else if (itf == 1)
+	{
+		return desc_hid_report2;
+	}
+
+		return 0;
```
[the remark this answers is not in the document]
zlittell: I might be misunderstanding what you meant here but this should be correct
image
*note these are not from the example so the descriptors aren't mouse then keyboard*

src/class/hid/hid_device.h
zlittell: yeah I figured since it was eventually going to get corrected and there weren't too many spots to do this in, it was alright being ugly and simple for now.

examples/device/hid_multipleinterface/src/usb_descriptors.c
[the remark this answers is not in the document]
hathach: sorry, i mean `mis-indent`, btw please also change it to `return NULL`

zlittell:
I was going to do another push to update some of the items mentioned but then realized I would be pushing a commit that just made the 2nd descripter const and then reduced the example buffer size to 8. Figured those two housekeeping items could be done when _n_ api and combining the helpers is done. I didn't do much, but I am glad I could contribute. I have asked a ton of questions and bugged a few people, so I feel better about myself a little now haha.

src/class/hid/hid_device.c
```diff
@@ -232,6 +232,7 @@ bool hidd_control_request(uint8_t rhport, tusb_control_request_t const * request
 
   #if CFG_TUD_HID>1
   uint8_t const hid_itf = get_hid_index_by_itfnum((uint8_t) request->wIndex);
+  TU_VERIFY(hid_itf<0xFF, 0);
```
hathach, line 235: since return value is bool, you should return false instead of 0. or you could just simply drop it and just leave 

```
TU_VERIFY(hid_itf<0xFF);
```

TU_VERFIY has a trick to have 2nd arg as optional.

hathach:
> I was going to do another push to update some of the items mentioned but then realized I would be pushing a commit that just made the 2nd descripter const and then reduced the example buffer size to 8. Figured those two housekeeping items could be done when _n_ api and combining the helpers is done. I didn't do much, but I am glad I could contribute. I have asked a ton of questions and bugged a few people, so I feel better about myself a little now haha.

You are doing great, the PR add very useful functionality for others. Especially user that focus on keyboard/mouse only, having separated interface does has an advantage to have them to use BOOT PROTOCOL which is guarantee to work on all machine (even in DOS).

zlittell:
The reason I needed the multi interface is that I have a generic HID interface that exchanges bytes and a keyboard interface. If they are the same interface I cannot capture the generic interface in my application because windows won't let you open a keyboard interface. Not sure if the same in other OS.

hathach:
> The reason I needed the multi interface is that I have a generic HID interface that exchanges bytes and a keyboard interface. If they are the same interface I cannot capture the generic interface in my application because windows won't let you open a keyboard interface. Not sure if the same in other OS.

Windows does a good job to prevent keylogger software :D . I am sure others will have a strict limit toward keyboard as well :)

hathach:
I does rename  the API to remove `_n` in the callback, and some other clean up work at #532

PR #525, opened 2020-10-02
da1469x iso support

kasjer:
**Describe the PR**
ISO endpoints for **DA146xx** were not supported so far.
To enable them and in turn allow USB audio to work following changes were made to **DA146xx** driver.
- Max endpoint size restriction was removed (64 bytes size of the internal FIFO).
It was achieved by adding DMA support and also FIFO level interrupt that can be used when DMA is in use.
(Only one endpoint in each direction can be used with DMA at a time).
- **dcd_edpt_close()** function was added that will be needed by **UAC2**
- little details were missing for ISO endpoints to work.

**Additional context**
It was tested in speaker role with #521 code. With endpoint sizes 64, 96 and 192 bytes.
It worked with DMA with 32MHz clock, and without DMA with 96MHz clock.

~TX DMA path is not present yet.~

Interrupt disabling is needed for accessing variables that decides which endpoint should use DMA.
It will still work correctly if only one endpoint for each direction will have max packed size greater then 64.

### DIFF
--- a/src/portable/dialog/da146xx/dcd_da146xx.c
+++ b/src/portable/dialog/da146xx/dcd_da146xx.c
@@ -285,15 +285,12 @@ void tusb_vbus_changed(bool present)
   }
 }
 
-static void transmit_packet(xfer_ctl_t * xfer)
+static void fill_tx_fifo(xfer_ctl_t * xfer)
 {
   int left_to_send;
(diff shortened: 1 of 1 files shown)


hathach (approved):
thank you very much for another brilliant PR :+1: :+1:

PR #527, opened 2020-10-03
Fix for TinyUSB and FreeRTOS on STM32F1

k0d:
When testing the cdc_msc_freertos example on a stm32f1 board, it doesn't work properly. This is because the IRQ for the USB is higher than the freertos one.

This fix basically sets the USB IRQ priority to a lower value which enables it to function as it should.

### DIFF
--- a/src/portable/st/stm32_fsdev/dcd_stm32_fsdev.c
+++ b/src/portable/st/stm32_fsdev/dcd_stm32_fsdev.c
@@ -280,6 +280,9 @@ void dcd_int_enable (uint8_t rhport)
   NVIC_EnableIRQ(USB_LP_CAN_RX0_IRQn);
   NVIC_EnableIRQ(USBWakeUp_IRQn);
 #elif CFG_TUSB_MCU == OPT_MCU_STM32F1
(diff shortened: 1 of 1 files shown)


src/portable/st/stm32_fsdev/dcd_stm32_fsdev.c
```diff
@@ -280,6 +280,9 @@ void dcd_int_enable (uint8_t rhport)
   NVIC_EnableIRQ(USB_LP_CAN_RX0_IRQn);
   NVIC_EnableIRQ(USBWakeUp_IRQn);
 #elif CFG_TUSB_MCU == OPT_MCU_STM32F1
+  NVIC_SetPriority(USB_HP_CAN1_TX_IRQn, 15);
```
hathach, line 283: the irq priority should be set by application, in this case, it is in bsp_init(). Setting it here is not flexible enough for all the application.

hathach (changes requested):
Thanks for PR, please move this to application i.e bsp_init() here is a similar code  for other MCU
https://github.com/hathach/tinyusb/blob/master/hw/bsp/lpcxpresso1769/lpcxpresso1769.c#L114

PR #529, opened 2020-10-06
stm32 fsdev: fix ISTR and CTR_RX/TX race conditions

gh2o:
This PR fixes two race conditions in the stm32 fsdev driver:
- USB->ISTR is a register where bits are cleared by writing 0. A non-atomic read-modify-write, as is done in reg16_clear_bits(), may inadvertently clear an interrupt flag that gets set between reading ISTR and writing ISTR, causing that interrupt to be dropped.
- Within the EPxR registers, CTR_TX and CTR_RX are also cleared by writing 0. Ensure that pcd_clear_rx_ep_ctr and pcd_clear_tx_ep_ctr write a 1 to the bit they are not clearing.

This bug occasionally occurred during high speed bidirectional data transfer on a pair of bulk endpoints with the same endpoint address.

### DIFF
--- a/src/portable/st/stm32_fsdev/dcd_stm32_fsdev.c
+++ b/src/portable/st/stm32_fsdev/dcd_stm32_fsdev.c
@@ -203,6 +203,11 @@ static inline void reg16_clear_bits(__IO uint16_t *reg, uint16_t mask) {
   *reg = (uint16_t)(*reg & ~mask);
 }
 
+// Bits in ISTR are cleared upon writing 0
+static inline void clear_istr_bits(uint16_t mask) {
+  USB->ISTR = ~mask;
+}
+
 void dcd_init (uint8_t rhport)
 {
   /* Clocks should already be enabled */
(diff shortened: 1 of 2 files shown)


src/portable/st/stm32_fsdev/dcd_stm32_fsdev.c
```diff
@@ -203,6 +203,11 @@ static inline void reg16_clear_bits(__IO uint16_t *reg, uint16_t mask) {
   *reg = (uint16_t)(*reg & ~mask);
 }
 
+// Bits in ISTR are cleared upon writing 0
+static inline void reg16_clear_istr(__IO uint16_t *reg, uint16_t mask) {
```
kasjer, line 207: Function name and comment suggests that it will work on **ISTR** register.
There is not point in passing register address as argument in my opinion.
I would also get rid of *reg16_*, it was OK when generic register was modified.
Just opinion.

gh2o: Noted. I've updated the PR.

pigrew:
Thanks for finding so many issues. I had originally copied most of the code from the stm32 HAL, with these bugs included. Since then, the HAL was updated, but this driver was never fixed.

Everything looks good to me in this PR.

In my mind for the ISTR, I didn't want to change the reserved bits since the documentation doesn't say what they do, but that does create the race condition, so what you propose is correct. ST now does the same, so it should work nicely.

pigrew (approved):

hathach:
thank you @gh2o  very much for the PR, and also @pigrew @kasjer for reviewing.

PR #531, opened 2020-10-07
mimxrt1010_evk: double stack size to allow net_lwip_webserver to run

majbthrd:
#289 reported an issue with net_lwip_webserver with the 'NXP Transdimension (18 / 43 / iMX RT)' target.

I had access to a MIMXRT1010_EVK, and the issue was nothing more than a too restrictive stack size in NXP's default stack size value in the linker script.  A minor tweak to the BSP board.mk by adding LDFLAGS to override the value remedies this.

I suspect the same change will remedy others (most notably the MIMXRT1015_EVK, which should be nearly identical to the MIMXRT1010_EVK), but I'll defer this to people who have these platforms and can confirm this.

### DIFF
--- a/hw/bsp/mimxrt1010_evk/board.mk
+++ b/hw/bsp/mimxrt1010_evk/board.mk
@@ -19,6 +19,9 @@ MCU_DIR = hw/mcu/nxp/sdk/devices/MIMXRT1011
 # All source paths should be relative to the top level.
 LD_FILE = $(MCU_DIR)/gcc/MIMXRT1011xxxxx_flexspi_nor.ld
 
+LDFLAGS += \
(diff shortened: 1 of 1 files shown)


hathach (commented):
great, thanks. Do you know where is the line that make use of stack overflow e.g local dynamic array. I try to see if we could bring it to for say global scope to avoid stack overflow like this.

majbthrd:
The heavy stack usage is pervasive throughout lwip; every TCP/IP stack that I've ever encountered has been like this.

hathach (approved):
super finding. Don't worry about the rest of the family. I will pull them out in the weekend to test with :+1:

PR #533, opened 2020-10-10
net_lwip_webserver: tweak main.c comments with smartphone suggestions

majbthrd:
The goal of net_lwip_webserver is example code that works on as many different platforms as possible.  Operation under Windows, Linux, and macOS has been tested, but smartphones seem to be under-developed and artificially picky (if not buggy).  Given the sheer cornucopia of phone out there, it is probably a losing battle, but I've added comments in main.c with a couple of suggestion on things to try.


majbthrd:
audio_test project is failing on d5035_01; this change should have had nothing whatsoever to do with that

hathach:
> audio_test project is failing on d5035_01; this change should have had nothing whatsoever to do with that

yeah, don't worry, I will fix that right away, it is due to microchip re-invent the assert() macro that conflict with the built-in one. I will probably skip that example for that board.

hathach (approved):

PR #536, opened 2020-10-13
osal_none: Fix queue definition

kasjer:
**Describe the PR**
OSAL_QUEUE_DEF was not updated while FIFO was changed
to use unmasked pointers.

Two fields that are crucial to new functionality were left
uninitialized resulting in FIFO slowing down in time.

**Additional context**


hathach (approved):
you really have an eagle eyes there :) Thank you very much.

PR #539, opened 2020-10-15
net_lwip_webserver: efficiency tweaks

majbthrd:
small code and data reduction by:
a) tweaking lwip compile-time options in examples/device/net_lwip_webserver/src/lwipopts.h
b) changing the DHCP server to not store duplicate values of what the stack already knows


hathach (approved):
look great, thank you for fine tuning it :+1:

PR #541, opened 2020-10-18
Allow use of internal FS PHY on OTG_HS interface

j4cbo:
**Describe the PR**
Allow use of internal FS PHY on OTG_HS interface

**Additional context**
Some ST parts (like STM32F74xxx / STM32F75xxx, maybe others) allow the USB_OTG_FS core to be used with either an external ULPI PHY or an internal full-speed-only (12mbps) PHY. Currently the code assumes than an ULPI PHY is used unless the chip has an internal high-speed PHY (`#if defined(USB_HS_PHYC)`), with no provision to use the internal FS PHY.

I'm not sure an #ifdef is the right way to do this, or if so, that the option name I'm using is good (`TUD_OPT_SYNOPSYS_FS_PHY`) - suggestions appreciated.

### DIFF
--- a/src/portable/st/synopsys/dcd_synopsys.c
+++ b/src/portable/st/synopsys/dcd_synopsys.c
@@ -416,9 +416,7 @@ void dcd_init (uint8_t rhport)
   if ( rhport == 1 )
   {
     // On selected MCUs HS port1 can be used with external PHY via ULPI interface
-#if defined(TUD_OPT_SYNOPSYS_FS_PHY)
-      usb_otg->GUSBCFG |= USB_OTG_GUSBCFG_PHYSEL;
-#else
+#if CFG_TUSB_RHPORT1_MODE & OPT_MODE_HIGH_SPEED
     // deactivate internal PHY
     usb_otg->GCCFG &= ~USB_OTG_GCCFG_PWRDWN;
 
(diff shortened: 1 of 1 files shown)


hathach:
could you tell me which boards and mcu you are testing with ?

j4cbo:
I'm using a custom board with STM32F750V8

src/portable/st/synopsys/dcd_synopsys.c
```diff
@@ -416,7 +416,9 @@ void dcd_init (uint8_t rhport)
   if ( rhport == 1 )
   {
     // On selected MCUs HS port1 can be used with external PHY via ULPI interface
-
+#if defined(TUD_OPT_SYNOPSYS_FS_PHY)
```
hathach, line 419: you should use the builtin highspeed test instead of introducing new macro, if `CFG_TUSB_RHPORT1_MODE` is not OPT_MODE_HIGH_SPEED, we can use the fullspeed phy  

```
#if TUD_OPT_HIGH_SPEED
```

src/portable/st/synopsys/dcd_synopsys.c
```diff
@@ -467,8 +470,13 @@ void dcd_init (uint8_t rhport)
 
   set_speed(rhport, TUD_OPT_HIGH_SPEED ? TUSB_SPEED_HIGH : TUSB_SPEED_FULL);
 
-  // Enable internal USB transceiver.
+  // Enable internal USB transceiver. Unconditional if using FS PHY on HS core.
+#if defined(TUD_OPT_SYNOPSYS_FS_PHY)
```
hathach, line 474: same as above, but you could put the `OPT_MODE_HIGH_SPEED` into the if (condition)

hathach (changes requested):
please use `OPT_MODE_HIGH_SPEED` instead, also make sure you test it on actual hardware. Since I don't have the hardware to test with.

hathach (approved):
Thank you for your PR, it looks great now.

PR #543, opened 2020-10-25
net_device: re-factor code so as to not be specific to lwIP

majbthrd:
When the net class driver was originally added to tinyusb, @hathach put in a "TODO should not include external files" comment in  ./src/class/net/net_device.h

This PR re-factors the code to remove the dependency on these include files.  It also has the benefit of making the net class driver not specific to lwIP, making it more agnostic as to which network stack is chosen by the user.

It does this by putting the lwIP-dependent portion of the code in ./examples/device/net_lwip_webserver/src/main.c

An added callback tud_network_xmit_cb() takes care of copying the packet into the net class driver's IN endpoint buffer.

P.S.: note that there are now two arguments (void *ref, uint16_t arg) to tud_network_xmit() and tud_network_xmit_cb().  In the present example, "arg" is unused, but providing "arg" enables a user wanting direct access to virtual Ethernet packets without a network stack to implement tud_network_xmit_cb() as simple as { memcpy(dst, ref, arg); return arg; } where ref is a pointer to a buffer to transmit and arg is the length.

### DIFF
--- a/examples/device/net_lwip_webserver/src/main.c
+++ b/examples/device/net_lwip_webserver/src/main.c
@@ -99,7 +99,7 @@ static err_t linkoutput_fn(struct netif *netif, struct pbuf *p)
     /* if the network driver can accept another packet, we make it happen */
     if (tud_network_can_xmit())
     {
-      tud_network_xmit(p);
+      tud_network_xmit(p, 0 /* unused for this example */);
       return ERR_OK;
     }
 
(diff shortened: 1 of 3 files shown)


hathach (approved):
Thank you very much for putting more effort to improve the example. Indeed, I was planning to make tinyusb as neutral and independent as possible with the net driver, but couldn't got the time to do so :)

PR #546, opened 2020-11-02
use same51 driver to adafruit asf4 submodule

hathach:
**Describe the PR**
same5x is added recently to asf4, it is better to re-use it than maintaining its driver ourself.

@jgressmann  may want to review this.


jgressmann:
Looks good.

hathach:
> Looks good.

thanks for the review

PR #550, opened 2020-11-05
Add CDC NCM implementation

j4cbo:
**Describe the PR**
Add implementation of CDC NCM networking interface

**Additional context**
CDC-NCM is the newest USB class for Ethernet over USB, available here: https://www.usb.org/document-library/network-control-model-devices-specification-v10-and-errata-and-adopters-agreement

It has two main advantages over the earler CDC-ECM:
- Instead of a direct mapping of one control transfer per datagram, NCM packs multiple datagrams into a transfer
buffer ("NTB"), allowing for better efficiency.
- Newer versions of Windows have built-in support for NCM, if an MS OS 2.0 descriptor pointing to the "WINNCM" driver is included; this means CDC-NCM has out-of-the-box support of all 3 of Windows, Mac, and Linux.

NCM fits more easily into a composite device than the existing RNDIS+ECM driver in src/class/net/net_device.c, which switches between the two at the configuration level; NCM is just another pair of interfaces so it's straightforward to have a composite device that includes NCM as well as other protocols.

I'm still working on hammering out bugs in this. I do have basic functionality working on all 3 of Mac (Macbook Pro), Linux (Raspberry Pi), and Windows (Parallels VM on Mac - haven't had access to a native Windows box yet) and so I'm sending it for early review, but please don't merge just yet :)

### DIFF
--- a/src/class/cdc/cdc.h
+++ b/src/class/cdc/cdc.h
@@ -69,7 +69,8 @@ typedef enum
   CDC_COMM_SUBCLASS_DEVICE_MANAGEMENT                 , ///< Device Management  [USBWMC1.1]
   CDC_COMM_SUBCLASS_MOBILE_DIRECT_LINE_MODEL          , ///< Mobile Direct Line Model  [USBWMC1.1]
   CDC_COMM_SUBCLASS_OBEX                              , ///< OBEX  [USBWMC1.1]
-  CDC_COMM_SUBCLASS_ETHERNET_EMULATION_MODEL            ///< Ethernet Emulation Model  [USBEEM1.0]
+  CDC_COMM_SUBCLASS_ETHERNET_EMULATION_MODEL          , ///< Ethernet Emulation Model  [USBEEM1.0]
+  CDC_COMM_SUBCLASS_NETWORK_CONTROL_MODEL               ///< Network Control Model  [USBNCM1.0]
 } cdc_comm_sublcass_type_t;
 
 /// Communication Interface Protocol Codes
@@ -114,7 +115,8 @@ typedef enum
   CDC_FUNC_DESC_COMMAND_SET                                      = 0x16 , ///< Command Set Functional Descriptor
   CDC_FUNC_DESC_COMMAND_SET_DETAIL                               = 0x17 , ///< Command Set Detail Functional Descriptor
   CDC_FUNC_DESC_TELEPHONE_CONTROL_MODEL                          = 0x18 , ///< Telephone Control Model Functional Descriptor
-  CDC_FUNC_DESC_OBEX_SERVICE_IDENTIFIER                          = 0x19   ///< OBEX Service Identifier Functional Descriptor
+  CDC_FUNC_DESC_OBEX_SERVICE_IDENTIFIER                          = 0x19 , ///< OBEX Service Identifier Functional Descriptor
+  CDC_FUNC_DESC_NCM                                              = 0x1A , ///< NCM Functional Descriptor
 }cdc_func_desc_type_t;
 
 //--------------------------------------------------------------------+
--- a/src/class/cdc/cdc_ncm_device.c
+++ b/src/class/cdc/cdc_ncm_device.c
@@ -1,0 +1,493 @@
+/*
+ * The MIT License (MIT)
+ *
+ * Copyright (c) 2020 Jacob Berg Potter
+ * Copyright (c) 2020 Peter Lawrence
+ * Copyright (c) 2019 Ha Thach (tinyusb.org)
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ *
+ * This file is part of the TinyUSB stack.
+ */
+
+#include "tusb_option.h"
+
+#if ( TUSB_OPT_DEVICE_ENABLED && CFG_TUD_NCM )
+
+#include "device/usbd_pvt.h"
+#include "cdc_device.h"
+#include "cdc_ncm_device.h"
+
+//--------------------------------------------------------------------+
+// MACRO CONSTANT TYPEDEF
+//--------------------------------------------------------------------+
+
+#define NTH16_SIGNATURE 0x484D434E
+#define NDP16_SIGNATURE_NCM0 0x304D434E
+#define NDP16_SIGNATURE_NCM1 0x314D434E
+
+typedef struct
+{
+  uint8_t itf_num;      // Index number of Management Interface, +1 for Data Interface
+  uint8_t itf_data_alt; // Alternate setting of Data Interface. 0 : inactive, 1 : active
+
+  uint8_t ep_notif;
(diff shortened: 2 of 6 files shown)


hathach:
Wow, thank you very much for this great PR, this NCM is definitely new to me. This looks really good at quick look, though since it is pretty big addition, It will take time for decent review, I will try to review as much as I could. Meanwhile since it is still WIP, I will mark it as draft PR, feel free to mark it as ready when you think it work reliably.

@majbthrd is probably also interested in this PR as well.

majbthrd:
Speaking personally, the things that I'd like to see are:

1) either move NCM class driver files into ./src/class/net or we move existing ./src/class/net/ into ./src/class/cdc; having some CDC network implementations be more CDC than others doesn't make sense; thoughts from @hathach ?

2) Share a single ./src/class/net/net_device.h with the same "Application API" and two sets of "INTERNAL USBD-CLASS DRIVER API": one for ECM/RNDIS and one for NCM.

3) **MOST IMPORTANTLY**: adopt the exact same "Application API" prototyped in net_device.h: tud_network_init_cb, tud_network_recv_cb, tud_network_xmit_cb, tud_network_mac_address, tud_network_recv_renew, tud_network_can_xmit, tud_network_xmit

I can't stress enough the importance of point 3.  It does not serve the TinyUSB adopter to have to develop different network stack implementations for each class driver; users should be able to run the same higher-level code that calls a neutral API and easily switch between ECM/RNDIS and NCM.

I see in the PR that handle_incoming_datagram() calls tud_ncm_receive_cb() multiple times and then immediately renews the endpoint with a ncm_receive_renew().  That's not practical from a real-world implementation perspective.  The network stack needs to process each datagram in its own time without blocking TinyUSB and only when all the datagram in the buffer have been expended should the endpoint be renewed.  Otherwise, the USB peripheral will be overwriting the data before the network stack has a chance to process it.

So, what I would suggest doing is having handle_incoming_datagram() call tud_network_recv_cb() for the first datagram, and then each tud_network_recv_renew() call provokes additional tud_network_recv_cb() calls until all datagrams have been dealt with.  Once all have been dealt with, then and only then is ncm_receive_renew() called.

IMHO, the litmus test for the PR being ready is being able to take ./examples/device/net_lwip_webserver, swap in CFG_TUD_NCM and appropriate USB descriptors, and have everything function the same (just with a different class driver).

hathach:
thanks @majbthrd for detail review, I agree the NCM and ECM/RNDIS should share the same API to make it easier for network stack to work with. @j4cbo I haven't looked at the code just yet, but whenever you think it is needed, you could also change the current usbnet ECM/RNDIS API so that it also work the same way as NCM as well.

hathach:
#557 slightly change the control transfer callback for the driver by introducing the stage (setup/data/ack) argument. You probably may want to update the driver. It does not change the driver much, but still need to update a bit.

j4cbo:
Sorry for the delay here, I've been doing some rewriting of other parts of my firmware but I'm about to pick this up again. I'm also now building on top of FreeRTOS so I'm thinking more about avoiding blocking the whole USB stack.

I think putting the new code in net/ makes sense. @hathach , what do you think about renaming `src/class/net/net_device.c` to `src/class/net/ecm_rndis.c` and `CFG_TUD_NET` to `CFG_TUD_NET_ECM_RNDIS`? Then `src/class/net/net_device.h` would be split into two files, the common application API `tud_network_*` and the NCM-specific stuff (`netd_`, moved to `ecm_rndis.h`). Sound good?

hathach:
hihi @j4cbo sorry for late response, I were off for vacation the previously and is only doing catching up until now. 

> Sorry for the delay here, I've been doing some rewriting of other parts of my firmware but I'm about to pick this up again. I'm also now building on top of FreeRTOS so I'm thinking more about avoiding blocking the whole USB stack.

Not a problem at all.

> I think putting the new code in net/ makes sense. @hathach , what do you think about renaming `src/class/net/net_device.c` to `src/class/net/ecm_rndis.c` and `CFG_TUD_NET` to `CFG_TUD_NET_ECM_RNDIS`? Then `src/class/net/net_device.h` would be split into two files, the common application API `tud_network_*` and the ECM-specific stuff (`netd_`, moved to `ecm_rndis.h`). Sound good?

To be honest, I think it is better to just add the NCM class and leave the current ECM/RNDIS as it is for now. You could introduce your own  CFG_TUD_NET_NCM or CFG_TUD_CDC_NCM (either that makes sense to you), and/or having additional cdc_ncm.c/h or net_ncm.c/h. Once we got it added, we could do the rename/refactor/re-org later on.

kripton:
Wow, the code looks very nice! @j4cbo: Any updates on this one? Or is it worth that I work my way into this to be able to fix the outstanding review comments? I'd really be interested to have an easy solution to have one tinyUSB-network that works out-of-the-box on Linux, mac and win :+1:

kripton:
@j4cbo: Is the firmware you are talking about available somewhere? My CDC-NCM device enumerates perfectly fine but since I didn't yet get my code using `tud_ncm_xmit` to compile, no packets are exchanged.
Can you give some example on how to use `tud_ncm_xmit` (also using a LwIP-based stack here)? Thanks!

kripton:
Okay, got it working. Commit still pending, need to clean up some bits. Just for quick reference:
New function:
```
void flatten(void *arg, uint8_t *buffer, uint16_t size) {
    pbuf_copy_partial(arg, buffer, size, 0);
}
```

Then modify the existing `linkoutput_fn`:
```
static err_t linkoutput_fn(struct netif *netif, struct pbuf *p)
{
    (void)netif;
      if (!tud_ready())
        return ERR_USE;

      tud_ncm_xmit(p, p->tot_len, flatten);
      return ERR_OK;
}
```

majbthrd:
@j4cbo, to help your PR be more presentable, I've [speculatively done a new branch](https://github.com/majbthrd/tinyusb/commit/5a1a6ba5c3526e822e060745be8e643e24537c3c) based off the current TinyUSB code that uses a modification of your driver.

The most important change to the driver code is to address the issue that I originally raised in Nov 2020: compatibility with the existing network API.  I had to make two minor tweaks to the existing API to accommodate your driver:

1)	tud_network_can_xmit() now takes a parameter as to the size of the would-be transmit packet
2)	tud_network_link_state_cb() was added since your driver wants to inform the user code as to the link state

The big difference between my proposed code and your PR is that instead of the user having to craft a new user application to use the NCM driver, the existing net_lwip_webserver example just works, regardless of which network driver is used.  Similarly, anyone who is writing their own TinyUSB-based network application can switch between the two drivers, using the same API, to choose which works best for their usage.

For the proposed code, picking which network driver is a matter of setting CFG_TUD_NCM in examples/device/net_lwip_webserver/src/tusb_config.h to 0 or 1.  That approach doesn't have to be set in stone; it was just a straightforward means to an end.

@kripton, since you set the process off and represent someone interested in using NCM, perhaps you could take a look at the code?

kripton:
Hi @majbthrd: Thanks for having worked on this! I've had a look at the code and used it in my project. It works just as fine as the version in this PR, with fewer changes in the library-using code. Well done, thanks!
Of course, it's still an API-breakage for everyone using `tud_network_can_xmit` I guess?

I've added the "Microsoft OS 2.0 compatible descriptor" with "WINNCM" string to my project. The device works like before on Linux. Didn't test macOS this time. On Windows, the device enumerates and shows a non-working "UsbNcm Host Device", telling me that the device failed to start with error code "0xC0000483". According to http://errorco.de/win32/ntstatus-h/status_device_hardware_error/0xc0000483/, this means "STATUS_DEVICE_HARDWARE_ERROR" and I can find this reference multiple times in the host driver (https://github.com/microsoft/NCM-Driver-for-Windows). That would be next, finding out why it works on Linux but not on Windows ...

majbthrd:
> Hi @majbthrd: Thanks for having worked on this! I've had a look at the code and used it in my project. It works just as fine as the version in this PR, with fewer changes in the library-using code. Well done, thanks!

Thanks @kripton 

My goal was keeping things as similar as possible, but if complete backwards compatibility is deemed necessary for code acceptance, this can be remedied by having a new API tud_network_can_xmit_bytes() to provide the size that the NCM driver wants and then the legacy tud_network_can_xmit() can call tud_network_can_xmit_bytes(CFG_TUD_NET_MTU) for legacy users.

kripton:
Yeah, I honestly don't mind the API change. I could have also lived with the much-changed API proposed by this PR if I can just have NCM working on Linux, macOS and Windows hosts ;). But of course less changes are to be preferred and it makes totally sense to not just ask if an xmit is possible but if an xmit with X bytes can be done. So the API change makes sense in the end.

I can confirm that the code from @majbthrd's branch is working fine on Linux (5.13.4, amd64), macOS (11.2.3, amd64) and Windows (10 Pro, 21H1, amd64). On Windows, I still have to force the driver manually but then ICMP and HTTP is working fine.

kripton:
Oh, and @j4cbo : I'd still be very interested to see your descriptor tables and/or MS OS X.Y descriptors that mad you have NCM working on Windows hosts out of the box ;)

kripton:
Sorry for the spam today but I got it working :+1: 
The problem was that I copied the ms_os_20 descriptor table from https://github.com/hathach/tinyusb/blob/master/examples/device/webusb_serial/src/usb_descriptors.c#L169 and just replaced `WINUSB` with `WINNCM`. However, then of course it still referenced `ITF_NUM_VENDOR` (which was defined in my code since I wanted to use WebUSB's "landing page" feature) but was of course wrong for NCM. When I replaced that with `ITF_NUM_CDC_NCM_CMD` it started working on Windows as well :D

Issue #549, opened 2020-11-04
MSC device cannot be unmounted on Windows


dobairoland:
**Set up**

 - **PC OS**   : Windows 10 & Arch Linux 
 - **Board**   : Custom board based on ESP32-S2
 - **Firmware**: Slightly modified version of examples/device/cdc_msc used with the latest ESP-IDF and the latests tinyusb master (075334af80303488a8c874255d725cb4ca15a7de)

**Describe the bug**

Linux can unmount the disc without any issue. The device cannot be "ejected" using the same firmware on Windows. (The same computer is used) 

**To reproduce**

Linux:
1. tud_msc_write10_cb() invoked, lba=2, offset=0 - write root directory
2. tud_msc_write10_cb() invoked, lba=0, offset=0 - write boot sector
3. tud_msc_scsi_cb() invoked. bufsize=0, SCSI_CMD_PREVENT_ALLOW_MEDIUM_REMOVAL

Windows:
1. tud_msc_write10_cb() invoked, lba=1, offset=0 - write FAT table
2. tud_msc_scsi_cb() invoked. bufsize=0, SCSI_CMD_PREVENT_ALLOW_MEDIUM_REMOVAL
3. tud_msc_start_stop_cb() invoked, power_condition=0, start=0, load_eject=1

"Safe to remove hardware" message is shown but drive remains attached and can be "ejected" again.

tannewt:
I don't think this is up to tinyusb to do. In order to not appear again you need to track the eject and change the readiness of the drive. 

The example doesn't do this because the drive is in RAM and always available. Relevant code is here: https://github.com/hathach/tinyusb/blob/master/examples/device/cdc_msc/src/msc_disk.c#L156

Here is how we do it in CircuitPython: https://github.com/adafruit/circuitpython/blob/main/supervisor/shared/usb/usb_msc_flash.c#L210

hathach:
thanks @tannewt for answering, yeah, this is application level code and beyond a usb stack to handle. @dobairoland please analyze the scsi command and its parameter when host issue the msc eject. Then act accordingly to whatever your application want to do.

dobairoland:
@tannewt Thank you for the kind help. It is working now.

@hathach I think the example should be ejectable even if the RAM drive is always ready. For example, if I have an USB flash HDD ejected then it is ejected properly and is not showing anymore in "This PC" even though I still have the disk in the USB.

hathach:
[...]

PR #551, opened 2020-11-05
Make the disk disappear on Windows after it was ejected

dobairoland:
Make the disk disappear on Windows after it was ejected. The device needs to be re-inserted or reseted to re-appear again.

This doesn't affect Linux where the device can be mounted and unmounted repeatedly.

Note that this should be applied to three more examples.

Closes https://github.com/hathach/tinyusb/issues/549

### DIFF
--- a/examples/device/cdc_msc/src/msc_disk.c
+++ b/examples/device/cdc_msc/src/msc_disk.c
@@ -28,6 +28,8 @@
 
 #if CFG_TUD_MSC
 
+static bool ejected = false;
+
 // Some MCU doesn't have enough 8KB SRAM to store the whole disk
 // We will use Flash as read-only disk with board that has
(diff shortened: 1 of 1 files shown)


examples/device/cdc_msc/src/msc_disk.c
```diff
@@ -137,7 +139,13 @@ bool tud_msc_test_unit_ready_cb(uint8_t lun)
 {
   (void) lun;
 
-  return true; // RAM disk is always ready
+  // RAM disk is ready until is not ejected
```
hathach, line 142: hmm the comment should be `until ejected` without NOT.

hathach (commented):
look good but I think the comment has a typo

hathach (approved):
thank for the PR, will merge and correct the comment as follow up

Issue #340, opened 2020-04-13
add dcd_edpt_close() API


hathach:
Related to #326, #336, #320 #317 

check list for adding new dcd API to close an endpoint.  This API is needed for supporting 
- multiple configuration and 
- Alternate Interface. 

Although the API will only be used by above usecase, it is probably simple enough to be implemented for all port. Port checklikst

- [ ] Espressif ESP32-S2
- [ ] Microchip SAMD
- [ ] Microchip SAMG 
- [ ] Nordic nRF5x 
- [ ] Nuvoton NUC120
- [ ] Nuvoton NUC121/125/126
- [ ] Nuvoton NUC505
- [ ] NXP LPC3511
- [ ] NXP LPC17/40
- [ ] NXP Transdimension (18/ iMX RT)
- [ ] Sony CXD56
- [ ] STM32 FSDev
- [ ] STM32 Synopsys
- [ ] TI MSP430
- [ ] ValentyUSB eptri (fomu)

JeffGiant:
@hathach Do you have plan to support this API for ESP32-S2? Or any update.

hathach:
no update and no plan, I will work on it when I got some time. Any PRs is welcome.

Issue #540, opened 2020-10-16
Better dcd synopsys fifo allocation for ISO  


hathach:
**Is your feature request related to a problem? Please describe.**

current synopsys EP FIFO allocation scheme does not support high ISO packet size e.g 192 in #538 . It assumes all FS endpoints packet only has up to 64 bytes. However FS ISO can have up to 1023 bytes.  
https://github.com/hathach/tinyusb/blob/master/src/portable/st/synopsys/dcd_synopsys.c#L216

We need a better scheme not only resolve this FIFO OUT, but also open/close/re-open ISO endpoints with different packet size e.g 100, 200, 300  

For a quick walkaround, we can increase the FIFO OUT to let's say 256 when AUDIO class is enabled e.g
```
// largest_epsize is the largest of all ep packet size used in the configuration descriptors
static inline uint16_t allocate_grxfsiz(uint16_t largest_epsize)
{
  return 15 + 2 x (largest_epsize/4) + 2*EP_MAX;
}

  //   Therefore GRXFSIZ = 13 + 1 + 1 + 2 x (Largest-EPsize/4) + 2 x EPOUTnum
  //   - FullSpeed (64 Bytes ): GRXFSIZ = 15 + 2 x  16 + 2 x EP_MAX = 47  + 2 x EP_MAX
  //   - Highspeed (512 bytes): GRXFSIZ = 15 + 2 x 128 + 2 x EP_MAX = 271 + 2 x EP_MAX
  //
  //   NOTE: Largest-EPsize & EPOUTnum is actual used endpoints in configuration. Since DCD has no knowledge
  //   of the overall picture yet. We will use the worst scenario: largest possible + EP_MAX
  //
  //   FIXME: for Isochronous, largest EP size can be 1023/1024 for FS/HS respectively. In addition if multiple ISO
  //   are enabled at least "2 x (Largest-EPsize/4) + 1" are recommended.  Maybe provide a macro for application to
  //   overwrite this.

#if TUD_OPT_HIGH_SPEED
  _allocated_fifo_words = allocate_grxfsiz(512);
#else
#if CFG_TUD_AUDIO
  // FIXME walkround to increase OUT FIFO limit to able to support up to ISO packet size of 256 
  _allocated_fifo_words =  allocate_grxfsiz(256);
#else
 _allocated_fifo_words =  allocate_grxfsiz(64);
#endif 
#endif
```

[...]

PR #552, opened 2020-11-06
Add dcd_edpt_close() to esp32-s2

Milili-Tech:
**Describe the PR**
Implement dcd_edpt_close() to esp32-s2. This referred to st/synopsys implementation.

**Additional context**
The examples which support esp32-s2 are tested, include a modification version of uac2_headset.

Relate to #340.

### DIFF
--- a/src/portable/espressif/esp32s2/dcd_esp32s2.c
+++ b/src/portable/espressif/esp32s2/dcd_esp32s2.c
@@ -127,7 +127,7 @@ static void bus_reset(void)
   //   Recommended value = 10 + 1 + 2 x (16+2) = 47 --> Let's make it 52
   USB0.grstctl |= 0x10 << USB_TXFNUM_S; // fifo 0x10,
   USB0.grstctl |= USB_TXFFLSH_M;        // Flush fifo
-  USB0.grxfsiz = 52;
+  USB0.grxfsiz = 15 + 2 * (256/4) + 2*EP_MAX;
 
   // Control IN uses FIFO 0 with 64 bytes ( 16 32-bit word )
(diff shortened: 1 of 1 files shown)


hathach:
thank you for your PR, can I ask whether you work for espressif. Since you seem to undo their PR https://github.com/hathach/tinyusb/pull/454 

esp32s2 has less fifo than endpoint, which is `_allocated_fifos` for, it is the number of fifos.

JeffGant:
@hathach have reverted this part back, please check.

src/portable/espressif/esp32s2/dcd_esp32s2.c
```diff
@@ -127,7 +127,7 @@ static void bus_reset(void)
   //   Recommended value = 10 + 1 + 2 x (16+2) = 47 --> Let's make it 52
   USB0.grstctl |= 0x10 << USB_TXFNUM_S; // fifo 0x10,
   USB0.grstctl |= USB_TXFFLSH_M;        // Flush fifo
-  USB0.grxfsiz = 52;
+  USB0.grxfsiz = 15 + 2 * (256/4) + 2*EP_MAX;
```
me-no-dev, line 130: can you please explain this calculation?

JeffGant: It's a workaround provided by @hathach in #540 , but actually need better solution.

hathach: this is based on recommended by stm32 guide, however, it is not optimal for non-ISO application. I plan to work on a better fifo allocation scheme later on. I think it is probably best to leave this PR pending until we get thing settled with stm32 first. 2 moving targets are harder to work with than 1.

PR #557, opened 2020-11-19
Rework class driver control transfer

hathach:
**Describe the PR**
Current class driver hooks for control transfer is only `control_request` (setup stage) and `control_complete`(data stage). There is a need to introduce `control_ack_complete()` that invoked when the zlp of ack stage is complete. However, 3 callbacks with the same signature is a bit too much,. This PR merge 3 of them into `control_xfer_cb(rhport, stage, request)`

Note: the reason to add callback for status complete is for request such as CDC tud_cdc_line_state_cb, handling it in the SETUP stage can cause a race condition with the callback **tud_cdc_line_state_cb()** invoked and its polling tud_cdc_line_state_cb() return true before the ACK stage is complete. This lead to firmware application to start sending out data while host terminal client is not considered as connected yet. The behavior is different depending on the host OS and terminal, on Linux with minicom it often cause incorrect print out e.g

 ```
Adafruit CircuitPython 6.0.0-278-g66fb09506-dirty on 2020-11-17; Saola 1 w/Wrover with ESP32S2
>>> uto-reload is o
```

PS: still work in progress, have been migrating msc, cdc, and hid so far. Create PR for feedback and will continue to migrate all drivers soon enough.

PS2: all drivers are migrated and tested, so far so good.


cr1901:
>However, 3 callbacks with the same signature is a bit too much,.

It's your code; I'm just a contributor :P. But I don't agree with this. 

From skimming your changes, putting all the callbacks into one function creates a rather large `control_xfer_cb` function. You can trim `control_xfer_cb` by  calling into the _real_ `inline`d callbacks in each branch of the `if`/`else` chain. But isn't that inlining equivalent to what you're already doing before this change :P? This seems like more work that can potentially lead to an unwieldy and large `control_xfer_cb` callback.

Of course, we should at least "try and see what happens". I could be worried for naught :).

hathach:
> > However, 3 callbacks with the same signature is a bit too much,.
> 
> It's your code; I'm just a contributor :P. But I don't agree with this.
> 
> From skimming your changes, putting all the callbacks into one function creates a rather large `control_xfer_cb` function. You can trim `control_xfer_cb` by calling into the _real_ `inline`d callbacks in each branch of the `if`/`else` chain. But isn't that inlining equivalent to what you're already doing before this change :P? This seems like more work that can potentially lead to an unwieldy and large `control_xfer_cb` callback.
> 
> Of course, we should at least "try and see what happens". I could be worried for naught :).

Thanks Will for your comment, IMHO, the stage separation setup/data/ack done by usbd can also be done by class driver if wanted to. The 1 vs 3 hooks is really good clean up especially we do support application-defined drivers now, and many doesn't not have any use for data/ack stage processing (leave as empty previously). 

Though some driver actually works better with new changes especially those with lots of multiple SET_DATA request, it could reduce the duplicated switch/case for request parsing. E.g cdc receive data in SETUP and invoke callback in the same `case SET_CONTROL_LINE`
https://github.com/hathach/tinyusb/blob/rework-class-driver-control/src/class/cdc/cdc_device.c#L359

Indeed there is potential issue with us forgetting to check the `stage == SETUP/DATA/ACK` , that is why I firstly introduce it as a pair request/complete previously. However 3 hooks is pretty much a lot, and it does make sense to me that a Control and Normal transfer each has its ONE callback.

But those are of course my thought only, and may not be ideal, that is why I always like to hear more advice from you guys.

PR #560, opened 2020-11-23
have tusb_init() return true instead of TUSB_ERROR_NONE

### LINKED ISSUE #559: tusb_init() returns incorrect value
**Affected versions**
At least 0.7.0 and latest master (589dfdb)

**Describe the bug**
It's a simple coding bug in `tusb_init()` implementation:
```
bool tusb_init(void)
{
  // skip if already initialized
  if (_initialized) return true;

#if TUSB_OPT_HOST_ENABLED
  TU_ASSERT( tuh_init() ); // init host stack
#endif

#if TUSB_OPT_DEVICE_ENABLED
  TU_ASSERT ( tud_init() ); // init device stack
#endif

  _initialized = true;

  return TUSB_ERROR_NONE;
}
```

The function returns **TUSB_ERROR_NONE** (0) while it should return **true** (1).

**Explanation**
Although not documented, from the code we see that if tusb is already initialized and `tusb_init()` is called for the second time, it returns **true**, so **true** indicates successful initialization. However, when initialization is performed successfully for the first time, it returns **TUSB_ERROR_NONE**, which has numeric value **0** that corresponds to **false** in bool.

Consider the following code:
```
if(!tusb_init()) {
  abort();
}
```

Expected behavior: `abort()` is not executed upon successful initialization.
Actual behavior: `abort()` is executed upon successful initialization.

**Proposed solution**

The obvious one is to replace `return TUSB_ERROR_NONE;` with `return true;`. However, this may break existing code that uses tusb, if someone has already implemented workaround for erroneous `tusb_init()` behavior.

P.S.
I don't fully understand the idea behind returning error code / status from this function if it uses `TU_ASSERT` to handle errors... Even if we fix it to return `true` on correct initialization, it will never be able to return `false` because of those assertions...

hathach:
**Describe the PR**
fix #559

### DIFF
--- a/src/class/bth/bth_device.c
+++ b/src/class/bth/bth_device.c
@@ -249,7 +249,7 @@ bool btd_xfer_cb(uint8_t rhport, uint8_t ep_addr, xfer_result_t result, uint32_t
     if (tud_bt_acl_data_sent_cb) tud_bt_acl_data_sent_cb((uint16_t)xferred_bytes);
   }
 
-  return TUSB_ERROR_NONE;
+  return true;
 }
 
 #endif
--- a/src/common/tusb_error.h
+++ b/src/common/tusb_error.h
@@ -54,6 +54,7 @@
     ENTRY(TUSB_ERROR_FAILED                          )\
 
 /// \brief Error Code returned
+/// TODO obsolete and to be remove
 typedef enum
 {
   ERROR_TABLE(ERROR_ENUM)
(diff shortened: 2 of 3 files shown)


hathach:
@me-no-dev may need to modify IDF since the tusb_init() has been returned the incorrect code due to legacy error code (removed currently)
https://github.com/espressif/esp-idf/blob/master/components/tinyusb/additions/src/tinyusb.c#L107

PR #571, opened 2020-12-17
Fix IAR error on memcpy and warnings.

HiFiPhile:
Signed-off-by: MasterPhi <admin@hifiphile.com>

**Describe the PR**
IAR complains on "unknown type" on memcpy, and gives warnings on undefined volatile access order

### DIFF
--- a/src/common/tusb_fifo.c
+++ b/src/common/tusb_fifo.c
@@ -30,6 +30,12 @@
 #include "osal/osal.h"
 #include "tusb_fifo.h"
 
+// Supress IAR warning
+// Warning[Pa082]: undefined behavior: the order of volatile accesses is undefined in this statement
+#if defined(__ICCARM__)
+#pragma diag_suppress = Pa082
+#endif
+
 // implement mutex lock and unlock
 #if CFG_FIFO_MUTEX
 
@@ -106,7 +112,7 @@ static void _ff_push_n(tu_fifo_t* f, void const * data, uint16_t n, uint16_t wRe
     memcpy(f->buffer + (wRel * f->item_size), data, nLin*f->item_size);
 
     // Write data wrapped around
-    memcpy(f->buffer, data + nLin*f->item_size, (n - nLin) * f->item_size);
+    memcpy(f->buffer, ((uint8_t const*) data) + nLin*f->item_size, (n - nLin) * f->item_size);
   }
 }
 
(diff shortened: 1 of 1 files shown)


src/common/tusb_fifo.c
```diff
@@ -106,7 +110,7 @@ static void _ff_push_n(tu_fifo_t* f, void const * data, uint16_t n, uint16_t wRe
     memcpy(f->buffer + (wRel * f->item_size), data, nLin*f->item_size);
 
     // Write data wrapped around
-    memcpy(f->buffer, data + nLin*f->item_size, (n - nLin) * f->item_size);
+    memcpy(f->buffer, (uint8_t*)((uintptr_t)data + nLin*f->item_size), (n - nLin) * f->item_size);
```
hathach, line 113: Maybe only `((uint8_t const*) data) + nLin*item_size)` is good enough. I am not entirely sure but `uintptr_t` seems to be optional type for some platforms. Best to keep it as portable as possible

src/common/tusb_fifo.c
```diff
@@ -131,7 +135,7 @@ static void _ff_pull_n(tu_fifo_t* f, void * p_buffer, uint16_t n, uint16_t rRel)
     memcpy(p_buffer, f->buffer + (rRel * f->item_size), nLin*f->item_size);
 
     // Read data wrapped part
-    memcpy(p_buffer + nLin*f->item_size, f->buffer, (n - nLin) * f->item_size);
+    memcpy((uint8_t*)((uintptr_t)p_buffer + nLin*f->item_size), f->buffer, (n - nLin) * f->item_size);
```
hathach, line 138: Same as above

hathach (commented):
Thanks for the pr, look great, I only want to slightly make it more portable without uintptr_t

HiFiPhile:
> make it more portable without uintptr_t

Done

src/common/tusb_fifo.c
```diff
@@ -30,6 +30,10 @@
 #include "osal/osal.h"
 #include "tusb_fifo.h"
 
+#if defined(__ICCARM__)
+#pragma diag_suppress = Pa082
```
hathach, line 34: what is the Pa082 that we need to suppress with IAR ? I would prefer to fix the warning with the code rather than suppress it with pragma

HiFiPhile: Warning is `Warning[Pa082]: undefined behavior: the order of volatile accesses is undefined in this statement`
Because both `wr_idx` and `rd_idx` are volatile, IAR don't know which one should be accessed first.

In order to fix it we need to add temporary variables:
```
uint16_t tu_fifo_count(tu_fifo_t* f)
{
    uint16_t wr_idx = f->wr_idx;
  return _tu_fifo_count(f, wr_idx, f->rd_idx);
}
```

HiFiPhile: CC @PanRe

hathach: @HiFiPhile ah, I know this warning, have been encountered it before. IAR is pretty smart  about this. There is quite a lot of wr and rd idx passing to a function at same time. So yes, it makes sense to suppress the warning as you did here. We may come back later to "fix" it whenever having time. Please also put the warning message of  `Warning[Pa082]: undefined behavior: the order of volatile accesses is undefined in this statement` so that we could know why we suppress the warning.

hathach (approved):
superb !! Thank you for your effort for fixing IAR compiling.

PR #572, opened 2020-12-18
IAR fix more warnings

HiFiPhile:
Signed-off-by: HiFiPhile <admin@hifiphile.com>

### DIFF
--- a/src/device/usbd.c
+++ b/src/device/usbd.c
@@ -517,7 +517,7 @@ void tud_task (void)
           TU_ASSERT(driver, );
 
           TU_LOG2("  %s xfer callback\r\n", driver->name);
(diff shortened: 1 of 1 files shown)


src/device/usbd.c
```diff
@@ -942,7 +943,7 @@ static bool process_get_descriptor(uint8_t rhport, tusb_control_request_t const
       // first byte of descriptor is its size
       return tud_control_xfer(rhport, p_request, (void*) desc_str, desc_str[0]);
     break;
-
+    }
```
hathach, line 946: can you move the `}` before the `break;` , my style is having break outside of `}`. I admit not many people do it that way. But it is best to be consistent.

HiFiPhile: It's also my style, just didn't pay attention :)

hathach (approved):
thank you for the pr, look great now.

PR #573, opened 2020-12-20
IAR build error fixes

HiFiPhile:
**Describe the PR**
Fix scsi_mode_sense6_t padding which cause IAR compiler error.
Add missing prototypes.

For functions in `usbd_control.c`, we can keep prototypes in both `usbd_control.c` and `usbd.c`

### DIFF
--- a/src/class/msc/msc.h
+++ b/src/class/msc/msc.h
@@ -255,7 +255,7 @@ typedef struct TU_ATTR_PACKED
 
   uint8_t : 3;
   uint8_t disable_block_descriptor : 1;
-  uint8_t : 0;
+  uint8_t : 4;
 
   uint8_t page_code : 6;
   uint8_t page_control : 2;
--- a/src/device/usbd_control.c
+++ b/src/device/usbd_control.c
@@ -140,6 +140,14 @@ bool tud_control_xfer(uint8_t rhport, tusb_control_request_t const * request, vo
 // USBD API
 //--------------------------------------------------------------------+
 
+//--------------------------------------------------------------------+
+// Prototypes
+//--------------------------------------------------------------------+
+void usbd_control_reset(void);
+void usbd_control_set_request(tusb_control_request_t const *request);
+void usbd_control_set_complete_callback( usbd_control_xfer_cb_t fp );
+bool usbd_control_xfer_cb (uint8_t rhport, uint8_t ep_addr, xfer_result_t event, uint32_t xferred_bytes);
+
 void usbd_control_reset(void)
 {
   tu_varclr(&_ctrl_xfer);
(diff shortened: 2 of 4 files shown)


src/device/usbd_control.c
```diff
@@ -140,6 +140,14 @@ bool tud_control_xfer(uint8_t rhport, tusb_control_request_t const * request, vo
 // USBD API
 //--------------------------------------------------------------------+
 
+//--------------------------------------------------------------------+
+// Prototypes
+//--------------------------------------------------------------------+
+void usbd_control_reset(void);
```
hathach, line 146: Does IAR complains without these prototypes, these are only used by `usbd.c`, typically they would have its own private header, though I try to cut-corner and put those prototypes inside the usbd.c diretly
https://github.com/hathach/tinyusb/blob/master/src/device/usbd.c#L264 .

hathach (commented):
Thank your very much for your pr to fix compling with IAR. There is only a question regarding the prototypes

src/class/msc/msc.h
```diff
@@ -255,7 +255,7 @@ typedef struct TU_ATTR_PACKED
 
   uint8_t : 3;
   uint8_t disable_block_descriptor : 1;
-  uint8_t : 0;
+  uint8_t : 4;
```
hathach, line 258: hmm, there is probably other places that have `:0` padding somewhere in the codebase, I didn't know this is not in the specs :facepalm:

src/device/usbd_control.c
HiFiPhile: If the option `Require Prototypes` is enabled, IAR emit error about this.

src/class/msc/msc.h
HiFiPhile: Yes there are many zero width bitfields :) I've also reported this bug to IAR

src/device/usbd_control.c
hathach: thanks for clarification, I guess there is no harm to have extra prototypes here.

src/class/msc/msc.h
hathach:
```
an integral constant expression with a value greater or equal to zero. When greater than zero, this is the number of bits that this bit field will occupy. The value zero is only allowed for nameless bitfields and has special meaning: it specifies that the next bit field in the class definition will begin at an allocation unit's boundary.
```

I checked a bit further, and indeed it is part of the specs (not entire sure which specs version it belongs to). Yes, it is good idea to report this to IAR.

hathach (approved):
thank you very much for your effort to fix warnings with IAR.

PR #582, opened 2021-01-03
Feature dbl m33

gsteiert:
Adding Double M33 Express board to TinyUSB.
This includes a simple driver for the neopixels.
The linker script is offset for use with TinyUF2.


hathach:
look great, could you also update the https://github.com/hathach/tinyusb/blob/master/docs/boards.md here with the board name and its link.

gsteiert:
Added board link to boards.md

hathach (approved):
It is perfecto now

PR #589, opened 2021-01-10
moved neopixel to its own files with additional features

gsteiert:
I improved the neopixel driver with additional features and moved it to its own files.


hathach (changes requested):
actually I don't prefer to have submodule for sct neopixel driver in this repo. I would just have a copy of it in the local BSP.

hathach (approved):
oops, I just realized the SCT is common to several other NXP MCUs and not specific to LPC55. Therefore it has more reason to include this as submodule as `common` source. Please excuse my lack of knowledge.

PR #590, opened 2021-01-12
Add support for Windows10 Command Prompt as a shell on GNU make.

kkitayam:
**Describe the PR**
Add support for `Windows10 Command Prompt` as a shell on GNU make.

**Additional context**
This patch has been checked for build with the following tools. I'm sorry, but I haven't checked for esp32s2.
- [GNU make of chocolatey](https://chocolatey.org/packages/make)
- [GNU Arm Embedded Toolchain](https://developer.arm.com/tools-and-software/open-source-software/developer-tools/gnu-toolchain/gnu-rm) for windows
- [MSP430-GCC-OPENSOURCE](https://www.ti.com/tool/download/MSP430-GCC-OPENSOURCE)
- [GNU RISC-V Embedded GCC](https://xpack.github.io/riscv-none-embed-gcc/install/)

I skipped the rule that `.d` files are converted to `.P` files when run at the Windows Command Prompt. Because `.P` files are not included any makefiles, I think this rule could be skip.

### DIFF
--- a/examples/rules.mk
+++ b/examples/rules.mk
@@ -81,7 +81,11 @@ uf2: $(BUILD)/$(BOARD)-firmware.uf2
 OBJ_DIRS = $(sort $(dir $(OBJ)))
 $(OBJ): | $(OBJ_DIRS)
 $(OBJ_DIRS):
+ifeq ($(UNAME),Windows)
+	@$(MKDIR) $(subst /,\,$@)
+else
 	@$(MKDIR) -p $@
+endif
 
 $(BUILD)/$(BOARD)-firmware.elf: $(OBJ)
 	@echo LINK $@
@@ -107,6 +111,7 @@ vpath %.c . $(TOP)
 $(BUILD)/obj/%.o: %.c
 	@echo CC $(notdir $@)
 	@$(CC) $(CFLAGS) -c -MD -o $@ $<
+ifneq ($(UNAME),Windows)
 	@# The following fixes the dependency file.
 	@# See http://make.paulandlesley.org/autodep.html for details.
 	@# Regex adjusted from the above to play better with Windows paths, etc.
@@ -114,6 +119,7 @@ $(BUILD)/obj/%.o: %.c
 	  $(SED) -e 's/#.*//' -e 's/^.*:  *//' -e 's/ *\\$$//' \
 	      -e '/^$$/ d' -e 's/$$/ :/' < $(@:.o=.d) >> $(@:.o=.P); \
 	  $(RM) $(@:.o=.d)
+endif
 
 # ASM sources lower case .s
 vpath %.s . $(TOP)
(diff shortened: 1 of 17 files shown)


tools/top.mk
```diff
@@ -2,16 +2,29 @@ ifneq ($(lastword a b),b)
 $(error This Makefile require make 3.81 or newer)
 endif
 
+# Detect windows or not
+# https://stackoverflow.com/questions/714100/os-detecting-makefile/52062069#52062069
+ifeq '$(findstring ;,$(PATH))' ';'
```
hathach, line 7: typically I would use the `ifeq ($(OS),Windows_NT)`, does this OS variable defined in your setup. If yes, we could just use it directly without this ifeq

hw/bsp/double_m33_express/board.mk
```diff
@@ -26,7 +26,7 @@ SRC_C += \
 	$(MCU_DIR)/drivers/fsl_reset.c \
 	$(MCU_DIR)/drivers/fsl_usart.c \
 	$(MCU_DIR)/drivers/fsl_flexcomm.c \
-	$(TOP)/lib/sct_neopixel/sct_neopixel.c 
+	lib/sct_neopixel/sct_neopixel.c
```
hathach, line 29: nice catch, I overlook this one, it should be relative to top :)

hathach (commented):
Great, thank you very much for the PR. Would you mind helping me to completely remove the deps rules as well in the PR, it is not used at all, nice catch on this :+1:

examples/rules.mk
```diff
@@ -107,13 +111,15 @@ vpath %.c . $(TOP)
...
 	@$(CC) $(CFLAGS) -c -MD -o $@ $<
+ifneq ($(UNAME),Windows)
 	@# The following fixes the dependency file.
 	@# See http://make.paulandlesley.org/autodep.html for details.
 	@# Regex adjusted from the above to play better with Windows paths, etc.
 	@$(CP) $(@:.o=.d) $(@:.o=.P); \
 	  $(SED) -e 's/#.*//' -e 's/^.*:  *//' -e 's/ *\\$$//' \
 	      -e '/^$$/ d' -e 's/$$/ :/' < $(@:.o=.d) >> $(@:.o=.P); \
 	  $(RM) $(@:.o=.d)
+endif
```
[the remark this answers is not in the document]
kkitayam: OK, I will remove it.

tools/top.mk
kkitayam: Hmm, the comment in line 5 is not good. I will update the comment.

The reason I am doing this is to detect whether POSIX-like shell (MSYS2/Cygwin) or cmd.exe. `$(OS)` is the same value, which is WIndows_NT, on these platforms. 

MSYS2 and Cygwin work fine without this PR. So, We need to replace the path delimiter from slash to backslash only when using cmd.exe. I think if we use $(OS) for this PR, it will not work with MSYS2 and CYGWIN.

kkitayam: Updated comments. Also, changed the variable name to CMDEXE by d7cfd8b.

examples/rules.mk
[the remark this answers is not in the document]
kkitayam: The rule removed by d7cfd8b.

tools/top.mk
hathach: thank you very much for clarification. That makes sense

hathach (approved):
Thank you very much for the PR. This will help lots of windows user :+1:

PR #599, opened 2021-01-22
Group boards into family

hathach:
**Describe the PR**
groups boards into family to easier to add new boards and maintain

**Additional context**
existing single board per folder still works just fine as before both with `make BOARD=` and the ci build


hathach:
force to merge, ci build failed due to rp2040 aren't fully built yet with existing examples.

PR #601, opened 2021-01-24
start the assimilation of Pi Pico into TU  

majbthrd:
hcd_edpt_xfer() calls get_dev_ep(), which may return NULL, and then unconditionally uses the return pointer.  "assert" is used elsewhere in the rp2040 driver code and its use seems warranted here too.

Including "pico/stdlib.h" seems superfluous; there appear to be no issues with its removal.

By convention, TU drivers should implement the dcd_int_handler() API.  Ensuring its presence allows users to let a fixed interrupt handler call the driver code.

Interloper macros were substituted with native TU macros like TU_MIN and TU_ARRAY_SIZE.

### DIFF
--- a/src/portable/raspberrypi/rp2040/dcd_rp2040.c
+++ b/src/portable/raspberrypi/rp2040/dcd_rp2040.c
@@ -38,8 +38,6 @@
 
 #include "device/dcd.h"
 
-#include "pico/stdlib.h"
-
 /*------------------------------------------------------------------*/
 /* Low level controller
  *------------------------------------------------------------------*/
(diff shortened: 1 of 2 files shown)


src/portable/raspberrypi/rp2040/dcd_rp2040.c
```diff
@@ -479,4 +473,10 @@ void dcd_edpt_close (uint8_t rhport, uint8_t ep_addr)
 
 }
 
+void dcd_int_handler(uint8_t rhport)
```
hathach, line 476: whoop, I just about to make this change, seems like we have the same idea. This does help with existing cross-platform app.

hathach (approved):
This looks good to me, though let's wait for @kilograham comment.

src/portable/raspberrypi/rp2040/dcd_rp2040.c
```diff
@@ -67,11 +65,7 @@ static struct hw_endpoint *hw_endpoint_get_by_addr(uint8_t ep_addr)
 }
 static void _hw_endpoint_alloc(struct hw_endpoint *ep)
 {
-    uint size = 64;
-    if (ep->wMaxPacketSize > 64)
-    {
-        size = ep->wMaxPacketSize;
-    }
+    uint size = TU_MIN(64, ep->wMaxPacketSize);
```
hathach, line 68: larger than 64 actually should throw the error, but it won't get this far. USBD will rise an TU_ASSERT()
https://github.com/hathach/tinyusb/blob/master/src/device/usbd.c#L1112

hathach:
Merge this now, since it doesn't affect the functionality, will revert if needed when comparing between forks.

PR #603, opened 2021-01-25
Pico followup for (ci) building examples

hathach:
**Describe the PR**
- Get CMakelists.txt to support both esp32s2 and rp2040 with `-DFAMILY=` along with `-DBOARD=`. Getting all of current examples compiled with rp2040.
- `*_freertos` is skipped for rp2040 since I have no idea if freeRTOS could work with rp2040 atm
- `hid_generic_inout` has issue, both hid_test.js/py couldn't send/receive data to rp2040. Not sure why, we will fix it in a separated PR.
- `net_lwip_webserver` doesn't work with rp2040, again will troubleshoot and fix it later on.
- `webusb_serial` enumerated, able to connected via chrome, but there is some issue when connecting cdc along side. Will try to fix it later.


tannewt (commented):
Will you follow up with an int_handler change?

kilograham:
wow - lots of great stuff going on here... a little bit overloaded but will try to take a look soon.

hathach:
> Will you follow up with an int_handler change?

will do, I need to get ci running first and do hand-on testing with current state of the dcd first to see what works and what not. also to make sure when changing the dcd ports, there is nothing breaking. The dcd modification is on-the-way, just need a bit of testing.

> wow - lots of great stuff going on here... a little bit overloaded but will try to take a look soon.

This PR is mostly refactor BSP build support for tinyusb example and ci to get CMake running for both rp2040 and esp32s2. There is no changes to the dcd code of rp2040 yet. I will request your review on such as PR that modify the dcd which could impact the pico-sdk. BTW, tinyusb also include pico-sdk as submodule as well for building example, look like both repo include each other :D

PR #604, opened 2021-01-25
Update Spresense SDK and fix dcd issues

kamtom480:
**Describe the PR**
Update Spresense SDK to 2.0.2.
Define Spresense EP numbers that work with the cdc_msc example.
Fix issues that caused the DTR bit not to be changed to 1.

**Additional context**
Source code for `mkspk` has been added. This tool is used to create a `.spk` file.

Changes to dcd_cxd56:
- Pass the next setup package when USB stack is ready to receive it.
- When the transfer is 0 on EP0 send complete to the USB stack
- Pass the data received after the setup package to the USB stack.


hathach (approved):
superb ! Thank you very much for the PR.

PR #610, opened 2021-01-27
Rp2040 followup2

hathach:
**Describe the PR**
Follow up to update the dcd port of rp2040 as mentioned in the feedback issue
- removed the `assigned_address` in dcd, which can be extracted from request
- change panic in dcd_remote_wakeup to pico_info() with not supported yet message.
- wrap **rp2040_usb.c** around `CFG_TUSB_MCU == OPT_MCU_RP2040` since some auto build system such as arduino will include all the C file which cause compile error on other port (due to lack of pico-sdk etc ..) 
- dcd_int_handler() wrapper is already done as part of PR #601, please check it out as well.

Note: There is also BSP modification piggy-packed as well, you could just ignore it and pay attention to `dcd_rp2040.c` and `rp2040_usb.c`. You could also test out the uploaded artifacts in ci build without the need to compile the example namely `rp2040-tinyusb-examples`.  https://github.com/hathach/tinyusb/actions/runs/515064202

### DIFF
--- a/examples/device/board_test/src/main.c
+++ b/examples/device/board_test/src/main.c
@@ -55,10 +55,6 @@ int main(void)
   {
     uint32_t interval_ms = board_button_read() ? BLINK_PRESSED : BLINK_UNPRESSED;
 
-    // uart echo
-//    uint8_t ch;
-//    if ( board_uart_read(&ch, 1) ) board_uart_write(&ch, 1);
-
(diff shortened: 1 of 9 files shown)


hathach:
Merge this now, since it doesn't affect the functionality, will revert if needed when comparing between forks.

PR #612, opened 2021-01-28
Fix midi sysex sending bug

prplz:
Fixes write buffer corruption when sending certain lengths of sysex.

All the paths in this section should set write_target_length, but this one did not.


hathach (approved):
Thank you, it looks great.

PR #627, opened 2021-02-03
Add support for SAMD11 variants of LUNA.

ktemkin:
This PR:
 
- Applies the same family/chip layout to the `samd11` as used in other families; modeled after the `samd21`.
- Adds support for SAMD11-based LUNA boards (`luna_d11`).
- Renames the SAMD21-based LUNA boards to be more specific (`luna_d21`).

Tested on all SAMD11 platforms and on the `luna_d21` platform.


hathach (approved):
It looks perfect, thank you very much

Issue #628, opened 2021-02-03
RP2040 assertion "hw_data_offset(next_buffer_ptr) <= USB_DPRAM_MAX" failed:


ndinsmore:
**Set up**
[Mandatory] Provide details of your setup help us to reproduce the issue as quick as possible  
 - **PC OS**   :  macOS 10.15 
 - **Board**   : pi pico
 - **Firmware**: examples/device/usb/UAC2_headset modified to write square wave to mic  see below
 -  Using the tinyUSB master repo not the RPI repo
 ```c++
void fill_mic_buffer(int16_t num){
    static uint8_t wave_counter = 0;
    static int16_t wave_state = 20000;
    int16_t *dst = mic_buf;
    int16_t *limit = mic_buf + num;
    
    //THis code should make a 1000 hz squarewave
    while (dst < limit) {
            wave_counter = (wave_counter == 48) ? 0 : (wave_counter+1);
            wave_state = (wave_counter == 0 ) ? (-wave_state) : wave_state;
            *dst++ = (int16_t) wave_state;
    }
}
bool tud_audio_rx_done_cb(uint8_t rhport, uint8_t *buffer, uint16_t buf_size) {
    return true;
}

bool tud_audio_tx_done_pre_load_cb(uint8_t rhport, uint8_t itf, uint8_t ep_in, uint8_t cur_alt_setting) {
    (void) rhport;     (void) itf;    (void) ep_in;    (void) cur_alt_setting;
    
    const uint16_t num_to_buf = 128;
    fill_mic_buffer(num_to_buf);
    tud_audio_write((uint8_t *) mic_buf, num_to_buf * 2 /*/ 2*/); 
    return true;
}

//--------------------------------------------------------------------+
// AUDIO Task
//--------------------------------------------------------------------+

void audio_task(void) {

    if (spk_data_size) {
        spk_data_size = 0;
    }
}
```

**Describe the bug**
Code will run then mysteriously stop,   debugging and turning on the pico_info instrumentation points to the _hw_endpoint_alloc function line 84:
https://github.com/hathach/tinyusb/blob/148a5a9181c52078d7f7738f26d09a79d060687c/src/portable/raspberrypi/rp2040/dcd_rp2040.c#L65-L100

In that function it isn't clear to me where the memory reuse happens it looks like it just keeps pushing the buffer to the next set of memory.

[...]

PR #642, opened 2021-02-10
Fixes RP2040 buffer reallocation overrun problem

ndinsmore:
**Describe the PR**
This is a simple fix for the rp2040 device that prevents drives that open and close an ep (ie. audio device) from reallocating the buffer in the DPRAM and running out of space and failing the assert.

This should close:
#628 
raspberrypi/pico-sdk/issues/85

**Additional context**

### DIFF
--- a/src/portable/raspberrypi/rp2040/dcd_rp2040.c
+++ b/src/portable/raspberrypi/rp2040/dcd_rp2040.c
@@ -166,9 +166,10 @@ static void _hw_endpoint_init(struct hw_endpoint *ep, uint8_t ep_addr, uint wMax
 
         // Now if it hasn't already been done
         //alloc a buffer and fill in endpoint control register
-        if( !(ep->configured)){
-        _hw_endpoint_alloc(ep);
-    }
+        if( !(ep->configured))
+        {
+            _hw_endpoint_alloc(ep);
+        }
     }
 
     ep->configured = true;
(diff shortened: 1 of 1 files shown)


liamfraser:
Looks like a plausible fix to me. I think the audio example may be the first one that repeatedly opens and closes an ep which is why we didn't see this?

src/portable/raspberrypi/rp2040/dcd_rp2040.c
```diff
@@ -164,9 +164,12 @@ static void _hw_endpoint_init(struct hw_endpoint *ep, uint8_t ep_addr, uint wMax
             ep->endpoint_control = &usb_dpram->ep_ctrl[num-1].out;
         }
 
-        // Now alloc a buffer and fill in endpoint control register
+        // Now if it hasn't already been done
+        //alloc a buffer and fill in endpoint control register
+        if( !(ep->configured)){
         _hw_endpoint_alloc(ep);
     }
```
lurch, line 171: Looks like the indentation needs fixing?

liamfraser: I agree. @ndinsmore can you fix the indentation and match the brace style of the rest of the file please. ie.

```
if(!(ep->configured))
{
     _hw_endpoint_alloc(ep);
}
```

liamfraser:
This looks good, but there is a larger issue here that we don't implement dcd_edpt_close. However, in this instance I feel that closing and opening the endpoint all the time is inefficient so this fix is good enough. Longer term I will raise an issue to implement dcd_edpt_close

hathach:
> This looks good, but there is a larger issue here that we don't implement dcd_edpt_close. However, in this instance I feel that closing and opening the endpoint all the time is inefficient so this fix is good enough. Longer term I will raise an issue to implement dcd_edpt_close

Thanks for reviewing, feel free to merge it when you think it is ready.

PR #644, opened 2021-02-10
Set -DCMAKE_BUILD_TYPE=Debug when building in debug for rp2040.

ipopov:
This ensures that the Pico SDK, too, is built in debug mode and with nice debugging symbols. (Otherwise, it's built by default in Release mode.)

I guess that `rules.mk` is not the most elegant place to plop this conditional... I'd love any suggestions for a better way to do it. :-)


hathach (approved):
It looks great, I don't have any better place to suggest.

PR #654, opened 2021-02-16
Add alternate bitfield padding option

xmos-jmccarthy:
Adds configuration option CFG_TUSB_ALT_BIT_PACKING_ALIGNMENT, which substitutes bitfield variable " : 0" padding syntax with an unused variable of size equal to the remaining number of bits.

This change is intended to resolve aligned access issues for some platforms.

The default behaviour is original if the option is not explicitly enabled.  Consequently, this change should not affect existing applications.

### DIFF
--- a/src/class/cdc/cdc.h
+++ b/src/class/cdc/cdc.h
@@ -277,11 +277,7 @@ typedef struct TU_ATTR_PACKED
   struct {
     uint8_t handle_call    : 1; ///< 0 - Device sends/receives call management information only over the Communications Class interface. 1 - Device can send/receive call management information over a Data Class interface.
     uint8_t send_recv_call : 1; ///< 0 - Device does not handle call management itself. 1 - Device handles call management itself.
-#if CFG_TUSB_ALT_BIT_PACKING_ALIGNMENT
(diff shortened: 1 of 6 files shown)


niepiekm (approved):

hathach:
thank you for your PR, would you mind telling me what is your platform you are working. unnamed zero size is pretty much std padding in C language.

xmos-jmccarthy:
I am working on the XCore platform.

Unnamed zero size is indeed standard; However, I found that the compiler for XCore currently will cause a hardware exception when trying to access a bitfield under certain alignment conditions.

hathach (changes requested):
Thanks for the PR, actually, the `bitfield size = 0` is due to my laziness. Since you already counted the unused size, there is no need to introduce the new macro `CFG_TUSB_ALT_BIT_PACKING_ALIGNMENT`, we should just always use it which is more explicit (therefore more portable). Would you mind updating the PR accordingly.

PS: also it is better to use the  `TU_RESERVED` instead of `unused`, it expands to unique token name. But it is optional.

hathach (approved):
look great, thank you for your PR.

PR #655, opened 2021-02-17
Fix to fifo pointer overflow mathmatics

ndinsmore:
**Describe the PR**
The basics are as follows: (non_used_index_space + max_pointer_idx)  should equal UINT16_MAX so that when you add anything to those values it rolls over.

**Additional context**
This will likely fix a bunch of difficult problems which have been itermitent.


hathach (approved):
superb !! Thank you, I have overlooked this when merging. osal_pico is probably under development in parallel with the fifo rework

PR #657, opened 2021-02-17
RP2040:Enable full size isochronous buffers

ndinsmore:
**Describe the PR**
The RP2040 spec allows full-sized isochronous buffers, this enables it.
**Additional context**
The rp2040 spec limits none ISO buffers to 64 bits.

### DIFF
--- a/src/portable/raspberrypi/rp2040/rp2040_usb.c
+++ b/src/portable/raspberrypi/rp2040/rp2040_usb.c
@@ -153,15 +153,7 @@ void _hw_endpoint_xfer_start(struct hw_endpoint *ep, uint8_t *buffer, uint16_t t
     // Fill in info now that we're kicking off the hw
     ep->total_len = total_len;
     ep->len = 0;
-    // FIXME: What if low speed
-    if(ep->transfer_type == TUSB_XFER_ISOCHRONOUS)
-    {
-        ep->transfer_size = tu_min32(total_len, ep->wMaxPacketSize);
-    }
-    else
-    {
-        ep->transfer_size = tu_min32(total_len, 64);
-    }
+    ep->transfer_size = tu_min32(total_len, ep->wMaxPacketSize);
     ep->active = true;
     ep->user_buf = buffer;
     // Recalculate if this is the last buffer
@@ -239,15 +231,7 @@ bool _hw_endpoint_xfer_continue(struct hw_endpoint *ep)
 
     // Now we have synced our state with the hardware. Is there more data to transfer?
     uint remaining_bytes = ep->total_len - ep->len;
-
-    if(ep->transfer_type == TUSB_XFER_ISOCHRONOUS)
-    {
-        ep->transfer_size = tu_min32(remaining_bytes,ep->wMaxPacketSize);
-    }
-    else
-    {
-        ep->transfer_size = tu_min32(remaining_bytes, 64);
-    }
+    ep->transfer_size = tu_min32(remaining_bytes, ep->wMaxPacketSize);
     _hw_endpoint_update_last_buf(ep);
 
     // Can happen because of programmer error so check for it


src/portable/raspberrypi/rp2040/rp2040_usb.c
```diff
@@ -154,7 +154,14 @@ void _hw_endpoint_xfer_start(struct hw_endpoint *ep, uint8_t *buffer, uint16_t t
     ep->total_len = total_len;
     ep->len = 0;
     // FIXME: What if low speed
-    ep->transfer_size = total_len > 64 ? 64 : total_len;
+    if(ep->transfer_type == TUSB_XFER_ISOCHRONOUS)
+    {
+        ep->transfer_size = total_len > ep->wMaxPacketSize ? ep->wMaxPacketSize : total_len;
```
hathach, line 159: using min function would make it easier to read

```
ep->transfer_size = tu_min32(total_len, ep->wMaxPacketSize); 
```

src/portable/raspberrypi/rp2040/rp2040_usb.c
```diff
@@ -154,7 +154,14 @@ void _hw_endpoint_xfer_start(struct hw_endpoint *ep, uint8_t *buffer, uint16_t t
...
     ep->len = 0;
     // FIXME: What if low speed
-    ep->transfer_size = total_len > 64 ? 64 : total_len;
+    if(ep->transfer_type == TUSB_XFER_ISOCHRONOUS)
+    {
+        ep->transfer_size = total_len > ep->wMaxPacketSize ? ep->wMaxPacketSize : total_len; 
+    }
+    else
+    {
+        ep->transfer_size = total_len > 64 ? 64 : total_len;
```
hathach, line 163: I think this `else{}` could be merged to the above `if()` as well if 64 stands for ep size of non-iso here. Unless there is something I don't know of. @liamfraser may shed some light on this.

src/portable/raspberrypi/rp2040/rp2040_usb.c
```diff
@@ -232,7 +239,16 @@ bool _hw_endpoint_xfer_continue(struct hw_endpoint *ep)
...
+    // ep->transfer_size = remaining_bytes > 64 ? 64 : remaining_bytes;
+
+    if(ep->transfer_type == TUSB_XFER_ISOCHRONOUS)
+    {
+        ep->transfer_size = remaining_bytes > ep->wMaxPacketSize ? ep->wMaxPacketSize : remaining_bytes;
+    }
+    else
+    {
+        ep->transfer_size = remaining_bytes > 64 ? 64 : remaining_bytes;
+    }
```
hathach, line 251: like above these line can be simply be

```
ep->transfer_size = tu_min32(total_len, ep->wMaxPacketSize); 
```

hathach (commented):
Look good to me, though I prefer to merge if() else{} with simple min function against ep max size. Let wait for more feedback from RPI team.

src/portable/raspberrypi/rp2040/rp2040_usb.c
```diff
@@ -232,7 +239,16 @@ bool _hw_endpoint_xfer_continue(struct hw_endpoint *ep)
 
     // Now we have synced our state with the hardware. Is there more data to transfer?
     uint remaining_bytes = ep->total_len - ep->len;
-    ep->transfer_size = remaining_bytes > 64 ? 64 : remaining_bytes;
+    // ep->transfer_size = remaining_bytes > 64 ? 64 : remaining_bytes;
```
hathach, line 242: would you mind removing this comment line.

src/portable/raspberrypi/rp2040/rp2040_usb.c
liamfraser: 64 is the max Non ISO EP size. We could potentially use a define for this.

hathach: The ep size can actually be smaller for example interrupt ep can have max packet size of 8 or 16. Will we use the actual mps (8) here or 64 still is correct ? If 8 is needed then we should change it to wMaxPacketSize (like iso) instead of fixed 64.

liamfraser: I think we should use wMaxPacketSize everywhere. I think that's what my "// FIXME: What if low speed" was meant for but I never got round to fixing it. I would like there to be a check that wMaxPacketSize is in a valid range for RP2040 somewhere though.

hathach: Yeah, that is what I think as well. Will update pr accordingly. usbd already perform the endpoint size check already before passing it to dcd open, so I guess it is rather OK afterwards. https://github.com/hathach/tinyusb/blob/1bba2c0fc3bce05e9fbe4ff23dda30283d08574d/src/device/usbd.c#L1111

hathach: Yeah, it changes recently to dynamically support FS/HS. It is mostly for  HS capable device to work properly with attached to let's say PICO FS host. Fullspeed device as default should have no issue, don't worry about low speed mode. It is virtually non-existent, no body is interested in it :smile: .

hathach (approved):
update the PR to min check against remaining and ep size. I  did accidentally merge from latest master and have to do the trick to change-rechange base for github to update the file difference. All good now, PR will be merged when ci complete. Thank you very much for the fix.

PR #658, opened 2021-02-18
rp2040/family.c: allow LED_PIN to be undefined

lurch:
Similar to the usage of `#ifdef UART_DEV` and `#ifdef BUTTON_BOOTSEL` in the same file.


hathach (approved):
thanks for the PR, look great.

PR #670, opened 2021-02-21
WIP : Adding Tiva C (ek-tm4c123gxl) support.

pran005:
**Describe the PR**
Porting the stack to support Tiva-C (ek-tm4c123gxl) board. Will be adding the relevant relevant drivers soon,

**Additional context**
TI doesn't provide a driver layer compliant to the CMSIS register definitions. Added the CMSIS provided system_init (clock_init) and basic drivers for board support using the CMSIS register definitions.


pran005:
Hello, 

I'm unable to see the files that I committed in hw/mcu/ti/ , am I missing something?

scott-olero:
Thank you for having a go at this. I have been struggling with it.

hathach:
@pran005 hw/mcu/ti is a submodule, you need to make an PR to its repo separated https://github.com/hathach/ti_driver , then git commit the tinyusb to update the submodule. Sorry for being more complicated than it should be. If that is too much of hassle, just zip the tiva-c and upload it here, I will push it to the ti_driver and update the tinyusb.

hathach (commented):
I am not familiar with TI MCU, but a quick glance at it, Tiva C is marketing term, I think we could just use `tm4c` as the name of the dcd and hcd. and both tm4c123x and tm4c129x share the same USB controlleer, therefore please rename the 
- `TM4C_Tiva-C/dcd_TM4C_Tiva-C.c` to `tm4c/dcd_tm4c.c'
-  `TM4C_Tiva-C/hcd_tm4c123.c` to `tm4c/hcd_tm4c.c'

I am open to any discussion on the naming,

pran005:
@hathach  I am okay with creating a PR at  https://github.com/hathach/ti_driver , although I'll need to figure out how shall I update the submodule in the tinyusb repo. 

Regarding naming, yes you're right Tiva-C is a marketing name, but I'd suggest creating to separate families of devices : tm4c123xx & tm4c129xx, the reason being there are some changes in both the silicon, including register definitions. Also some parts under tm4c129xx family have USB chapters of their reference manuals under NDA as shown:

image
 
Would it be okay if I rename it to tm4c/dcd_tm4c123xx.c  & tm4c/dcd_tm4c123xx.c for the time being? 
I don't have any boards based on tm4c129 devices, so I may not be able to test anything for that family as of now. I can plan something once the firmware for the current family is up and running.

hathach:
> @hathach I am okay with creating a PR at https://github.com/hathach/ti_driver , although I'll need to figure out how shall I update the submodule in the tinyusb repo.

Once the PR to submodule is merged, you just simply pull the submodule, then git add in the tinyusb repo, it will get updated to the latest commit like you did with normal file.

> Regarding naming, yes you're right Tiva-C is a marketing name, but I'd suggest creating to separate families of devices : tm4c123xx & tm4c129xx, the reason being there are some changes in both the silicon, including register definitions. Also some parts under tm4c129xx family have USB chapters of their reference manuals under NDA as shown in the snapshot. I have looked at this datasheet after signing an NDA with TI, I'm not sure how much can I reveal, but I can say there are certain differences/enhancements when compared to tm4c123xx devices.

From what I have seen, it is pretty much the same. Truth that each mcu has some changes to the controller IP such as stm32f4 (highspeed, full speed, on-chip phy etc ...) but it could be easy addressed by maco `#if`. I hope to have a single dcd and hcd file, the  TUSB_CFG_MCU can have OPT_MCU_TM4C123 and OPT_MCU_TM4C129 separately though. Anyway, don't worry about this much at this stage, just proceed with what you are comfortable with. We could always rename it later on.  No problems at all.

> image
> 
> Would it be okay if I rename it to tm4c/dcd_tm4c123xx.c & tm4c/dcd_tm4c123xx.c for the time being?
> I don't have any boards based on tm4c129 devices, so I may not be able to test anything for that family as of now. I can plan something once the firmware for the current family is up and running.

Yeah, we don't really need to access their controller IP full databook, the register interface is good enough, it is weird to see such as notice in the datasheet though.

pran005:
@hathach Yes the mention is weird. Also, their datasheets contain the register interface and bit definitions as well.  There are no register definitions available in the open version : [https://www.ti.com/lit/ds/symlink/tm4c129encpdt.pdf?ts=1613987608999&ref_url=https%253A%252F%252Fwww.ti.com%252Fproduct%252FTM4C129ENCPDT  ](url) . 

I think if the #if macro is fine, then we can name it tm4c* for now, and add the device specific implementation later. I'll get started on this.

hathach:
Please go ahead. Since PR is not ready for merging yet. I will convert it to draft, feel free to mark it to ready when you think it is OK

pran005:
@hathach , I have created a PR to your ti_driver repo, to add the necessary supporting files for tm4c123xx family of devices. Once that gets merged. I expect the build shall go through fine.

hathach:
> @hathach , I have created a PR to your ti_driver repo, to add the necessary supporting files for tm4c123xx family of devices. Once that gets merged. I expect the build shall go through fine.

It is merged already

> Regarding it being ready, when can I say it is ready? When I've added the relevant USB Drivers or when the bare minimal builds pass?

When some of the examples is working as proof of concept, especially the `cdc_msc` example. Once it is running, it is ready for merging, any later bugs can be addressed by follow-up PR. 

> P.S. : Sorry for the delay, it gets a little hard to contribute on weekdays due to my job. Thanks!

No problems at all, take your time. porting to a new MCU is never easy and consumes lots of time and effort.

pran005:
@hathach I need your help in one place, 2 FreeRTOS based builds are failing because of the following: 

/home/runner/work/tinyusb/tinyusb/lib/FreeRTOS/FreeRTOS/Source/include/portable.h:52:11: fatal error: portmacro.h: No such file or directory
   52 |  #include "portmacro.h"

Can you please help me understand what's the dependency I'm missing? 

Also, please tell me what flash tool shall I use, I have used OpenOCD to flash this board. How shall I be integrating it? 

Thanks

pran005:
Update : The board test seems to work fine now, the LED blinks at nearly 250ms by default, after pressing down SW1 on the board, the blinking rate slows down to 1s.

image

Additionally, the board prints out the hello statement through the Stellaris VCP (no additional circuitry required, it gets enumerated on connecting the on-board debugger to the PC). 

image

Please test the same on your end and let me know if all's well. 

P.S. : I still need some help on integrating the flashing tool into the bsp. 

Thanks

hathach:
@pran005 that looks great, `board_test` is also my first example I would run to make sure the clock is setup up correctly as well. Look like you are all set for dcd port. Don't worry about the `flash` target if it is too much of trouble, I will do it later. It is just a convenient way to flash the board, just use what you normally use (ide/gui) for now.

pran005:
@hathach That's what  I'm confused about, I use OpenOCD from the command line, the only difference is I need to open a gdb server, from a different terminal, and issue commands manually from the new terminal to actually flash the device. How can  I accommodate such a thing from board.mk ?

hathach:
You can use openocd as standalone flash without gdb http://openocd.org/doc/html/Flash-Commands.html . like the nuvoton mcu https://github.com/hathach/tinyusb/blob/master/hw/bsp/nutiny_nuc121s/board.mk#L55

pran005:
@hathach  I had a question related to one of the call backs that the driver makes to the stack, i.e.  `dcd_event_setup_received(0, setup, true);` . It says whenever we receive a "SETUP" packet then it shall be called. My question is, what do we mean by SETUP packet here, just the SETUP packet as  in figure 1, or the Data that follows the SETUP (i.e. Data packet in the SETUP stage) as in figure 2.

Figure 1: 
image

Figure 2:
image

Because you mention 8-bytes of data in the porting guide, which looks like the Data packet in the SETUP stage of a Control Transfer. Also, the IP of this MCU doesn't provide any mechanism to judge whether a SETUP token was received in device mode, we need to read what type of request (i.e. standard/class etc.)  was made by the host and take action appropriately.

hathach:
It means the 8 bytes of `Setup Data`, not the `Setup PID`, you should pullout/get it then pass it into the dcd_event_setup_received(). Hope this clear thing up.

pran005:
@hathach Thanks for the clarity. One more question I had, is there a way through which I can enable printing of some debug logs from the stack to the board's uart ?

hathach:
yeah,  you can use make `LOG=2` (or 1) which will output log to uart (if you implement the board_uart_write() ), you could redirect it to segger rtt if you could hook up an jlink, which is much faster. Detail is described here  https://github.com/hathach/tinyusb/blob/master/docs/getting_started.md

pran005:
Thanks for this!

pran005:
@hathach , Are you implementing the debug logs using nano libc and overwriting _write with board_uart_write ? I am hitting many hardfaults (bus faults) when I pass LOG=2. Doesn't happen otherwise. Am I doing something wrong? ( I see the rest of the board.mk files have -nostdlib flag).

hathach:
yes, I use nanolib with _write() for retarget and have no issues with other mcus. -nostdlib flag is more related to startup and not relevant to nanolib (libc). Do you printf in an ISR, maybe mcu doesn't like that.

pran005:
@hathach , no I'm not using an ISR to print. On further debugging I see bss global variables like _usbd_q are not zero initialized when I compile with LOG=2, which is resulting in a fault.

pran005:
Looks like something is wrong with my startup file. Can't figure out what.

hathach:
I mainly run make with `LOG=1/2` for debugging, if it becomes too difficult, I will use segger ozone as debugger. Ozone only need axf file, then you could step and breakpoint the program. I don't use much of IDE, it is very hard to keep up with all the ports with IDE, makefile works better.

pran005:
@hathach , There seems to be a serious problem in this MCU's silicon. I went on debugging trying to understand why a specific address is having spurious  values even after initialization. The situation is as follows :

1. I have two exactly identical board with same part number. 
2. The address at fault is 0x20002784 which shall contain the address of _usbd_q as confirmed by the map file below: 
image

3.  However, I see that after flashing the device, if I power cycle it. There's some weird but fixed value at this address. This value is fixed for one silicon, i.e. if I change the board I see a different corresponding to that board, which remains fixed after any number of power cycles. 

4. Below are screenshots of the values I see (the memory is in little endian) . Please notice the PC is at main(), i.e. the Vendor provided startup had already run, which should ideally have zeroed out this region (since it was used by .bss section). 

Board-1
image

Board-2 
image

->For an added confirmation the following is the value of BFAR (Bus Fault Address Register) for both the silicon

Board-1
image

Board-2
image

Again, I confirm both the boards are identical and have the same silicon part on them.  

In order to reproduce the issue, one may program the board, disconnect it from power, and then re-power it after a second. For some reason, if the power was on since the board was programmed it seems to work fine (i.e. no Hardfault). 

I'm not sure where shall I go from here. Looks like I'll need to close the PR. Please suggest if there's something that can be done.

pran005:
@hathach I've fixed the issue in the startup file. Please merge my PR in TI Driver repo to bring in the required changes.

hathach:
> @hathach I've fixed the issue in the startup file. Please merge my PR in TI Driver repo to bring in the required changes.

Is this a bug out of TI startup file. Which is really annoying !!

pran005:
> > @hathach I've fixed the issue in the startup file. Please merge my PR in TI Driver repo to bring in the required changes.
> 
> Merged, thanks. Is this a bug out of TI startup file. Which is really annoying !!

Yes! It was annoying indeed! The zerofill loop was written in C and was being optimized out. Converting it to assembly made it work!

hathach:
> Yes! It was annoying indeed! The zerofill loop was written in C and was being optimized out. Converting it to assembly made it work!

it is probably due to the LTO and/or other optimization option, if you continue to see issue like this, you could turn off the `-lto` and reduce the optimization if needed :)

pran005:
Yes, it was because of Ox optimization option. Thanks for the help :)

pran005:
@hathach , How does the communication happen stack and the driver if there more than 64 bytes of data to be transferred during control transfer? I see the stack says that the total length of configuration descriptors is 98 bytes, however, the `dcd_edpt_xfer` function just provides 64 bytes for. The driver needs to send more data on next IN packet from the host.  Is there bufer

pran005:
@hathach , How does the communication happen stack and the driver if there more than 64 bytes of data to be transferred during control transfer? I see the stack says that the total length of configuration descriptors is 98 bytes, however, the `dcd_edpt_xfer` function just provides 64 bytes for. The driver needs to send more data on next IN packet from the host.  Is there a shared buffer that I can use from the driver?

pran005:
@hathach  For some reason the checks are failing as they aren't able to find the ti_driver repo. I don't know what happened all of a sudden, can you help me with that? 

Additionally, I had a question, how does the communication happen between the stack & the driver if there are more than 64 bytes to be transferred? I see the configuration descriptor's total length is  98 bytes. However the  `dcd_edpt_xfer` passes only 64 bytes to the driver at a time. Is there a shared buffer that the driver can use to fetch the next set of bytes?

hw/bsp/ek-tm4c123gxl/board.mk
```diff
@@ -0,0 +1,44 @@
+CFLAGS += \
```
hathach, line 1: Add `DEPS_SUBMODULES = hw/mcu/ti` here to fix ci. I have opyimize ci speed by only checkout required module to save time.

https://github.com/hathach/tinyusb/blob/2adb7e719316b12c53a907153cfa0056db1abd70/hw/bsp/ea4357/board.mk#L1

hathach:
TinyUSB usbd can request transfer that is a lot larger than max packet size. It is dcd responsibility to carry out the requested length. When all the bytes is transferred, dcd invoked the dcd_event_xfer_complete() to notify the usbd

pran005:
> TinyUSB usbd can request transfer that is a lot larger than max packet size. It is dcd responsibility to carry out the requested length. When all the bytes is transferred, dcd invoked the dcd_event_xfer_complete() to notify the usbd

@hathach  but I see the configuration descriptor is 98 bytes in length, however the `len` field in `dcd_edpt_xfer` is set t 64 by the usbd. Also, no stack doesn't request any ZLP to be sent. Is there some other function / shared data structure that requests transfers?

hathach:
usbd will request next packet after 64 first bytes is transferred. It will also request xfer with length=0 as zlp as well. DCD must not auto send zlp. Sum up, dcd just need to transfer what the usbd request. The rest will be taken care by usbd. 
Note: to save sram, usbd control is transferred 64 byte each, but behavior can be configured/changed later on. Other class driver such as msc can normally request 4K or more for each transfer.

pran005:
That clears it. Thanks! I'll try calling back 'dcd_event_xfer_complete() ' to notify usbd and expect more data from it. Thanks !

pran005:
@hathach I see this MCU's USB IP is very different from the other ones, and that seems to be causing trouble. There's no way we can find if a SETUP request was sent using the IP , even a PID OUT raises the same interrupt as PID SETUP. Because of this the driver I implement, calls back to setup_received even when a PID OUT is received. 

I see even in TI's USB stack for this MCU, they are checking if the control packet was a standard request in the driver itself. Shall I add a similar check? I can check for standard requests and call PID setup for all standard requests and handle others appropriately.

hathach:
> @hathach I see this MCU's USB IP is very different from the other ones, and that seems to be causing trouble. There's no way we can find if a SETUP request was sent using the IP , even a PID OUT is treated the same as PID SETUP by the IP. Because of this the driver I implement, calls back to setup_received even when a PID OUT is received.
> 
> I see even in TI's USB stack for this MCU, they are checking if the control packet was a standard request in the driver itself. Shall I add a similar check? I can check for standard requests and call PID setup for all standard requests and handle others appropriately. I'm doing something similar for a personal learning project I'm doing (USB Development repo on my profile), that enumerates fine.

to be honest, I haven't looked into this controller details to comment on this. Though it is weird that it doesn't detect the SETUP vs OUT PID. Though if you could get it running without modification of any files besides  `dcd_tm4c123xx.c` then It would be OK. Otherwise, can you post your intended modification for discussion beforehand.

pran005:
Sure @hathach , I'll try doing so, also, just wanted to ask one question related to the stack, is it possible that in any CONTROL transfer, the stack sends a data of more than 64 bytes?  Because currently, while returning the Configuration descriptors, I see the device returns 98 bytes in a single transaction (which is more than max allowed) . Due to this the PC (host) treats it as an erroneous situation and the enumeration fails.

hathach:
that stack can send 98 bytes without any issue, it will break it into 2 packets, 
- the 1st is 64 bytes then dcd xfer complete is called
- usbd will schedule another 34 bytes, then dcd xfer complete is called
- usbd will schedule another 0 bytes  (on opposite zero endpoint) for ACK

here is the LOG=2 on other mcu.  

```
USBD Setup Received 80 06 00 02 00 00 62 00
  Get Descriptor Configuration[0]
  Queue EP 80 with 64 bytes
USBD Xfer Complete on EP 80 with 64 bytes
  Queue EP 80 with 34 bytes
USBD Xfer Complete on EP 80 with 34 bytes
  Queue EP 00 with zlp Status
```

pran005:
@hathach  Thanks for the sample logs. I'm unable to log, it slows the USB part. I'll try to use DMA or Interrupts to log, maybe that helps. Currently I am purely relying on the trace. Thanks!

pran005:
@hathach  I need some help in the following: 

1. The device seems to enumerate fine on my PC.  Can you please test it at your end and see if the enumeration is going as expected? 

2. I see a few checks are failing, I did add `DEPS_SUBMODULES += hw/mcu/ti` but ithe CI builds are still failing.

hathach:
Great news to hear that you got it working. I will try to test it out as soon as I could. As for the ci, a quick glance doesn't give me idea why it wouldn't link. Maybe there is some issues with make. I will also check it out when testing out your work. Thank you very much again.

pran005:
@hathach It's not completely working right now, I have only put in code for the Control Endpoint. I will be adding code for all other endpoints as well.  Thanks for the help :)

hathach:
thank you very much for your effort. Tiva-C dcd support is implemented by #1183 . Although this pr is superseded, the ek-tm4c123gxl bsp is cherry picked :+1: :+1:

pran005:
Hello, that's really some good news, I'm sorry I couldn't do it before, wasn't keeping well until recently. Great to hear that the implementation was added! Thank you!

hathach:
> Hello, that's really some good news, I'm sorry I couldn't do it before, wasn't keeping well until recently. Great to hear that the implementation was added! Thank you!

No problems at all, thank you very much for your PR & effort.

PR #671, opened 2021-02-22
Pass the correct speed on Spresense

kamtom480:
**Describe the PR**
Pass the correct speed after a bus reset on Spresense.


hathach (approved):
thanks for the PR, it looks good.

PR #672, opened 2021-02-23
rp2040: correctly size variables to reduce RAM usage

majbthrd:
The rp2040 driver maintains a ginormous (technical term) RAM array of struct to track endpoints, and the author liked to use a proprietary-to-rp2040 type called “uint” that translates to a uint32_t.

Since TinyUSB’s dcd_edpt_xfer() only transfers in sizes of uint16_t, there is no reason to store FOUR variables of size uint32_t in the struct (of which there are 32 total in the array) when 16-bit more than suffices.

Fixing this yields 4 x 2 x 32 = 256 bytes, which may not matter to everyone, but this project is called “TinyUSB”.

Possibly more controversial was adding a macro called TUD_OPT_RP2040_USB_MAX_ENDPOINTS that allows the ginormous array of struct to be downsized by the user in situations where not all 32 entries are needed.  Even with the above trim, each of the 32 structs is 88 bytes in size, so it consumes a lot of RAM.

I also removed a superfluous include “hardware/clocks.h” from rp2040_usb.c, since it serves no function.


majbthrd:
In lieu of adding TUD_OPT_RP2040_USB_MAX_ENDPOINTS, I unearthed an existing macro USB_MAX_ENDPOINTS in the bowels of pico-sdk that was supposed "to save RAN"... so let's actually put it to use.

src/portable/raspberrypi/rp2040/rp2040_usb.h
```diff
@@ -67,10 +67,10 @@ struct hw_endpoint
 
     // Current transfer information
     bool active;
-    uint total_len;
-    uint len;
+    uint16_t total_len;
```
hathach, line 70: total_len can actually be > 64KB in case of MSC, I intended to change the dcd_edpt_xfer() total_bytes from `uint16_t to uint32_t` however, the work is a bit overwhelming to change all the dcd for now.

src/portable/raspberrypi/rp2040/rp2040_usb.h
```diff
@@ -80,7 +80,7 @@ struct hw_endpoint
     uint8_t *user_buf;
 
     // Data needed from EP descriptor
-    uint wMaxPacketSize;
+    uint16_t wMaxPacketSize;
```
hathach, line 83: although this is correct, this structure is also used by another usb stack (one implemented the BOOTROM). TinyUSB is not the only stack that work with rp2040 mcu, therefore it still would be `uint` for compatible. I am fine with this though.

hathach (commented):
I am fine with the PR, though I think the total_len should still be uint32_t (uint16_t is kind of my mistake in the dcd API). The max packet size is only up to 1024 (therefore uint16_t), though it may be stil be `uint` for compatible with RPi's own usb stack. I will leave this to @liamfraser to decide.

liamfraser (approved):
All looks good to me. This code is all TinyUSB specific. The bootrom uses its own USB device implementation that was written before our TinyUSB port.

src/portable/raspberrypi/rp2040/dcd_rp2040.c
```diff
@@ -64,7 +64,7 @@ static struct hw_endpoint *hw_endpoint_get_by_addr(uint8_t ep_addr)
 }
 static void _hw_endpoint_alloc(struct hw_endpoint *ep)
 {
-    uint size = TU_MIN(64, ep->wMaxPacketSize);
+    uint16_t size = TU_MIN(64, ep->wMaxPacketSize);
```
hathach, line 67: TU_MIN() -> tu_min16() would be better, inline function is better than macro (although the stack uses tons of macro already :smile: :smile: )

src/portable/raspberrypi/rp2040/dcd_rp2040.c
```diff
@@ -48,8 +48,8 @@
 // Init these in dcd_init
 static uint8_t *next_buffer_ptr;
 
-// Endpoints 0-15, direction 0 for out and 1 for in.
-static struct hw_endpoint hw_endpoints[16][2] = {0};
+// USB_MAX_ENDPOINTS Endpoints, direction 0 for out and 1 for in.
+static struct hw_endpoint hw_endpoints[USB_MAX_ENDPOINTS][2] = {0};
```
hathach, line 52: 16 pairs of endpoints are lots, 8 is more common. not really need changes though :smile:

hathach (commented):
good for a merge, only an minor request to change TU_MIN() to tu_min16()

PR #676, opened 2021-02-23
rp2040: don't compile in host code when in device mode

majbthrd:
Presently, there is assorted host-only rp2040 code that gets compiled into device-only operation.  (Host and device modes are mutually exclusive.)  Making this host-only code conditional at compile time saves code and data space.

This PR alone is good for 272 fewer bytes of code and 128 bytes less data RAM consumed.


liamfraser (approved):

src/portable/raspberrypi/rp2040/rp2040_usb.h
```diff
@@ -71,11 +71,13 @@ struct hw_endpoint
     uint len;
     // Amount of data with the hardware
     uint transfer_size;
+#ifdef RP2040_USB_HOST_MODE
```
hathach, line 74: if the field order is not matter, we should merge this with line 89 into a single `#ifdef`

hathach (commented):
since it is approved by RPi team, I think it is good to merge, though I think we could merge line 74 and 89 together for easier reading.

PR #680, opened 2021-02-24
spresense: fix setup processed flag

kamtom480:
**Describe the PR**
Fix usb setup flag which sometimes caused DTR bit detection problem.

### DIFF
--- a/src/portable/sony/cxd56/dcd_cxd56.c
+++ b/src/portable/sony/cxd56/dcd_cxd56.c
@@ -329,6 +329,7 @@ bool dcd_edpt_xfer(uint8_t rhport, uint8_t ep_addr, uint8_t *buffer, uint16_t to
   {
     if (total_bytes == 0)
     {
+      usbdcd_driver.setup_processed = true;
       dcd_event_xfer_complete(0, ep_addr, 0, XFER_RESULT_SUCCESS, false);
     }
     else if (ep_addr == 0x00 && total_bytes == usbdcd_driver.outlen)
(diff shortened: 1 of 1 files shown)


kamtom480:
@hathach I merged it. I hope that's not a problem.

src/portable/sony/cxd56/dcd_cxd56.c
```diff
@@ -329,6 +329,7 @@ bool dcd_edpt_xfer(uint8_t rhport, uint8_t ep_addr, uint8_t *buffer, uint16_t to
   {
     if (total_bytes == 0)
     {
+      usbdcd_driver.setup_processed = true;
```
hathach, line 332: if you want to detect status packet with this logic, it can mis-inteprete other event. For example when receiving 64 bytes of a descriptor, to end the data transfer, ZLP is sent in the DATA state on the IN endpoint. Afterwards there is another ZLP on OUT endpoint for status.

hathach (commented):
There is a better way to catch status, it is an optional callback `dcd_edpt0_status_complete()` which is added later after your port. The usbd will call this function right after status phase is complete (full transferred). You could use this to pull setup packet without the need of `usbdcd_driver.setup_processed)`
https://github.com/hathach/tinyusb/blob/master/src/device/dcd.h#L124

hathach:
> @hathach I merged it. I hope that's not a problem.

Not a problem at all, you understand the dcd much better than me. Though looking through the chance, I think there is probably a better approach with `dcd_edpt0_status_complete()` some dcd make use of this for changing address after status complete and prepare the next setup https://github.com/hathach/tinyusb/blob/master/src/portable/microchip/samd/dcd_samd.c#L187, some doesn't need it therefore it is optional.

PS: There is another breaking API as well with supporting multiple hid interfaces https://github.com/hathach/tinyusb/pull/678, that could break your circuitpython ci. It is a easy fix, I will comment on the circuitpython PR.

kamtom480:
Thanks for your comments! This is very useful feedback. I'll try to use `dcd_edpt0_status_complete()` instead of `usbdcd_driver.setup_processed`.

PR #682, opened 2021-02-25
Update RP2040 hcd_init to have rhport argument.

liamfraser:
Fix RP2040 host compile. @hathach can we compile something for RP2040 in host mode as part of CI so this would be caught in the future?


hathach:
@liamfraser ah sorry, I forgot about rp2040 host mode, can you also add `.only.MCU_RP2040` file to the host example https://github.com/hathach/tinyusb/tree/master/examples/host/cdc_msc_hid . That will have ci to build host example with rp2040

PS: ci has some internal issues, I have cancel the run, feel free to push new commit.

hathach (approved):

hathach:
Update: host example needs to have CMakelists.txt as well to build, I will merge it first and add the host example build for rp2040 in a separated PR. Ru-run ci just now.

Issue #681, opened 2021-02-25
Better way to handle CDC_REQUEST_SEND_BREAK


amitv87:
Found few legacy apps on windows which try to send break requests and terminate the connection if request fails.

Meanwhile I have made a change in **cdc_device.c** to support **CDC_REQUEST_SEND_BREAK** by adding a new case as show in code snippet.
``` c
case CDC_REQUEST_SEND_BREAK:
  if (stage == CONTROL_STAGE_SETUP)
  {
    tud_control_status(rhport, request);
  }
  else if (stage == CONTROL_STAGE_ACK)
  {
    TU_LOG2("  Send Break\r\n");
    if ( tud_cdc_send_break_cb ) tud_cdc_send_break_cb(itf);
  }
break;
```

hathach:
Can you name those legacy apps, and how to reproduce the request.

amitv87:
I have setup an rpi pico to behave as USB to UART bridge using CDC. This works perfectly fine in almost all scenarios.

I encountered the behaviour when I tried to flash a firmware using [qflash](https://forums.quectel.com/uploads/short-url/ApQc64SUOmNIU7ibGcQNPB5prxD.pdf) on a quectel modem.
The tool works absolutely fine with ftdi, silabs, prolific etc.

Although the modem needs just the rx and tx uart lines, qflash blindly expects the break request to be acknowledged else it chooses close the connection and halts the progress. It sends break request right after the connection is opened.

Also I have no idea about any standard use case of break request.

hathach:
thanks for the context, I am ok with adding more callback, but I would also like to take the chance to understand what the request does as well. I am currently working on other works, will pull out the cdc specs later on. Meanwhile, you could submit an PR for modification (it looks good), since I prefer to review/discuss code modification via PR than issue.

amitv87:
Great, will send a pr asap.
Also thanks for writing this wonderful library.
👍

hathach:
no problem, I am glad you find this project useful. closed by #683

PR #683, opened 2021-02-25
Handle CDC_REQUEST_SEND_BREAK 

amitv87:
Related to https://github.com/hathach/tinyusb/issues/681

### DIFF
--- a/src/class/cdc/cdc_device.h
+++ b/src/class/cdc/cdc_device.h
@@ -133,6 +133,9 @@ static inline bool     tud_cdc_write_clear     (void);
 // Invoked when received new data
 TU_ATTR_WEAK void tud_cdc_rx_cb(uint8_t itf);
 
+// Invoked when received send break
+TU_ATTR_WEAK void tud_cdc_send_break_cb(uint8_t itf, uint16_t wait_ms);
+
 // Invoked when received `wanted_char`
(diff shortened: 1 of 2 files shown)


hathach:
The legacy app shouldn't send SEND_BREAK since the current default descriptor doesn't support Send_break in its ACM descriptor. I guess it doesn't read those at all. Note for application that want to make use of send_break should set this bit to 1.

image

src/class/cdc/cdc_device.h
```diff
@@ -133,6 +133,9 @@ static inline bool     tud_cdc_write_clear     (void);
 // Invoked when received new data
 TU_ATTR_WEAK void tud_cdc_rx_cb(uint8_t itf);
 
+// Invoked when received send break
+TU_ATTR_WEAK void tud_cdc_send_break_cb(uint8_t itf, uint16_t wait_ms);
```
hathach, line 137: can you move it after `tud_cdc_line_coding_cb()`, also change the `wait_ms` to `duration_ms`.

hathach (commented):
thanks, the pr looks good, only a minor cosmetic change request.

hathach:
The legacy app shouldn't send SEND_BREAK since the current default descriptor doesn't support Send_break in its ACM descriptor. I guess it doesn't read those at all. Note for application that want to make use of send_break should set this bit to 1.

image

hathach (approved):
thank you for your PR. This should fix #681

PR #684, opened 2021-02-26
add rp2040 to host example build

hathach:
**Describe the PR**
Add RP2040 to host example build. Also change `ifdef RP2040_USB_HOST_MODE` to `if TUSB_OPT_HOST_ENABLED` to be consistent with tinyusb host detection convention.


hw/bsp/rp2040/family.cmake
```diff
@@ -13,8 +13,15 @@ set(SRC_TINYUSB
...
+	${TOP}/src/host/hub.c
+	${TOP}/src/host/usbh.c
+	${TOP}/src/host/usbh_control.c
+	${TOP}/src/class/cdc/cdc_host.c
+	${TOP}/src/class/hid/hid_host.c
+	${TOP}/src/class/msc/msc_host.c
 	${TOP}/src/portable/raspberrypi/${FAMILY}/rp2040_usb.c
+	${TOP}/src/portable/raspberrypi/${FAMILY}/dcd_rp2040.c
+	${TOP}/src/portable/raspberrypi/${FAMILY}/hcd_rp2040.c
 )
```
liamfraser, line 25: We have split our cmake into a device and host mode so you don't unnecessarily compile both.

examples/host/cdc_msc_hid/CMakeLists.txt
```diff
@@ -0,0 +1,46 @@
...
+  target_include_directories(${PROJECT} PUBLIC
+    ${CMAKE_CURRENT_SOURCE_DIR}/src
+  )
+
+  # Example defines
+  target_compile_definitions(${PROJECT} PUBLIC
+    CFG_TUSB_OS=OPT_OS_PICO
+  )
+
+  target_link_libraries(${PROJECT} pico_stdlib pico_fix_rp2040_usb_device_enumeration)
```
liamfraser, line 41: The USB device enumeration fix only applies to device mode. Perhaps this should be linked in by family.cmake?

hw/bsp/rp2040/family.cmake
liamfraser: See https://github.com/raspberrypi/pico-sdk/blob/master/src/rp2_common/tinyusb/CMakeLists.txt

hathach: That is ok though, tinyusb is designed to be all included and controlled by config file. To be honest that make my  example easier 😄

examples/host/cdc_msc_hid/CMakeLists.txt
hathach: It is ok as well, since the code is not run anyway, as above. There may be otg example later on

PR #685, opened 2021-02-26
[msp430]: a way skip nanolib for MSP430

Pillar1989:
I think that's a way to remove one to-do item in your code.

### DIFF
--- a/examples/rules.mk
+++ b/examples/rules.mk
@@ -93,8 +93,7 @@ INC += $(TOP)/src
 
 CFLAGS += $(addprefix -I,$(INC))
 
-# TODO Skip nanolib for MSP430
-ifeq ($(BOARD), msp_exp430f5529lp)
+ifeq ($(findstring  msp_exp430,$(BOARD)), msp_exp430)
   LDFLAGS += $(CFLAGS) -fshort-enums -Wl,-T,$(TOP)/$(LD_FILE) -Wl,-Map=$@.map -Wl,-cref -Wl,-gc-sections
 else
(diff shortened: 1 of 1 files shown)


cr1901:
Well, that indeed works for boards from TI, but I wanted something specific to "projects using msp430 the arch", regardless of which board you're using.

And I forgot at this point why this was a "TODO" (maybe there's no easy way to extract arch using `rules.mk`?). Your patch is fine for now tho.

hathach:
not all board will have an `msp_exp430` in their name. Marketing people will prefer to have AI, Super, in the product name :smile: . @cr1901 we can do it simply by moving msp430 into its own family with shared `family.mk' e.g stm32f4 https://github.com/hathach/tinyusb/tree/master/hw/bsp/stm32f4 . Though I am too lazy to do that with a family with only 1 board :smile: . will wait until some people make PR for another msp430 .

examples/rules.mk
```diff
@@ -93,8 +93,7 @@ INC += $(TOP)/src
 
 CFLAGS += $(addprefix -I,$(INC))
 
-# TODO Skip nanolib for MSP430
-ifeq ($(BOARD), msp_exp430f5529lp)
+ifeq ($(findstring  msp_exp430,$(BOARD)), msp_exp430)
```
hathach, line 96: can you use `CROSS_COMPILE` which is more accurate in this case.

```
ifeq ($(CROSS_COMPILE), msp430-elf-)
```

hathach (commented):
can you use CROSS_COMPILE which is more accurate in this case.

Pillar1989:
It looks like you can easily modify it just a bit.

PR #687, opened 2021-02-28
SAM 54 Xplained Pro: fix inverted logic for board led

jgressmann:
This PR fixes the inverted board led logic. the LED was on by default, board_led_write(1) would turn it off.

### DIFF
--- a/hw/bsp/same54xplainedpro/same54xplainedpro.c
+++ b/hw/bsp/same54xplainedpro/same54xplainedpro.c
@@ -166,9 +166,9 @@ void board_init(void)
 #endif
 
 	// LED0 init
-	gpio_set_pin_function(BUTTON_PIN, GPIO_PIN_FUNCTION_OFF);
+	gpio_set_pin_function(LED_PIN, GPIO_PIN_FUNCTION_OFF);
 	gpio_set_pin_direction(LED_PIN, GPIO_DIRECTION_OUT);
-	gpio_set_pin_level(LED_PIN, 1);
+	board_led_write(0);
 
 #if CFG_TUSB_DEBUG >= 2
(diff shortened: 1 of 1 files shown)


hw/bsp/same54xplainedpro/same54xplainedpro.c
```diff
@@ -166,15 +166,16 @@ void board_init(void)
 #endif
 
 	// LED0 init
```
hathach, line 168: Seem like a typo, should be led pin than button

hw/bsp/same54xplainedpro/same54xplainedpro.c
```diff
@@ -166,15 +166,16 @@ void board_init(void)
...
+	gpio_set_pin_function(BUTTON_PIN, GPIO_PIN_FUNCTION_OFF);
 	gpio_set_pin_direction(LED_PIN, GPIO_DIRECTION_OUT);
-	gpio_set_pin_level(LED_PIN, 0);
+	gpio_set_pin_level(LED_PIN, 1);
```
hathach, line 171: How about `board_led_write(0)` to set led to off

hathach (changes requested):
There is led pin typo in function off

hw/bsp/same54xplainedpro/same54xplainedpro.c
jgressmann: Doh!

hw/bsp/same54xplainedpro/same54xplainedpro.c
jgressmann: Done

hathach (approved):
Look great, thank you for the pr

PR #693, opened 2021-03-03
SAMx7x (E70, S70, V70, V71) DCD Support

HiFiPhile:
**Describe the PR**
Implementation of the SAM70 DCD driver.

**Additional context**
Tested with `cdc_dual_ports`, `hid_composite`, `uac2_headset`.

TODO:
- [x]  Fix EP0 bugs
- [x]  Test with more examples
- [ ]  Clean up & format code
- [x]  Add DMA & FIFO support
- [x]  Integrate BSP support

Any help is appreciated :)


hathach:
great PR, I will help and test it out whenever I could. Please keep it updated :+1: :+1:

HiFiPhile:
With DMA and Dual Bank the throughput is quiet good. 
Running at 150MHz in ITCM, with cdc_xfer_cb commented out
```
    // TODO search for wanted char first for better performance
    for(uint32_t i=0; i<xferred_bytes; i++)
    {
      tu_fifo_write(&p_cdc->rx_ff, &p_cdc->epout_buf[i]);

      // Check for wanted char and invoke callback if needed
      if ( tud_cdc_rx_wanted_cb && ( ((signed char) p_cdc->wanted_char) != -1 ) && ( p_cdc->wanted_char == p_cdc->epout_buf[i] ) )
      {
        tud_cdc_rx_wanted_cb(itf, p_cdc->wanted_char);
      }
    }
```
I got 126us for 4096 bytes OUT transfer, that's 32.5MB/s !

Even without DMA & Dual Bank I got about 12.5MB/s.

Unfortunately I can't run `cdc_dual_ports` in HS mode with dual bank as the packet ram (4KB) is too small.

hathach:
@HiFiPhile though are pretty good number, and we haven't got into the speed optimization of development stage just yet :smile:

perigoso:
whats missing on this PR? I have a device that will end up using this driver, and it would be convenient for me to have it merged, can i help in any way?

HiFiPhile:
Hi, at the moment this PR is tested on bulk transfer but other types of transfer remain untested (should work)

I've done some more work about fifo support but also remain untested.

HiFiPhile:
>  I have a device that will end up using this driver, and it would be convenient for me to have it merged, can i help in any way?

I'll push all works, could you test BSP integration and build with gcc ? Since I'm developing with IAR and my own BSP layer.

perigoso:
I can build with GCC, but the only board i have with this device is custom, so Ill have to use my own BSP as well

perigoso:
Got it to compile fine on gcc, just needed to swap some macros that are deprecated in the sam cmsis, i guess it doesnt error in IAR, as for actually testing in hardware I can't do it just yet, I'll Pr the changes in your branch

HiFiPhile:
Today I tested MSC also works, maybe this weekend I'll test more examples.

perigoso:
Sorry for all the force pushes, kept overlooking somethings that made the build check fail :sweat_smile:

src/portable/microchip/sam7x/dcd_sam7x.c
```diff
@@ -0,0 +1,710 @@
...
+  (void) rhport;
+  USBHS->USBHS_DEVCTRL |= USBHS_DEVCTRL_RMWKUP;
+}
+
+// Connect by enabling internal pull-up resistor on D+/D-
+void dcd_connect(uint8_t rhport)
+{
+  (void) rhport;
+  uint32_t irq_state = __get_PRIMASK();
+  __disable_irq();
```
hathach, line 157: do we really need to disable the whole irq, can we just disable the USB IRQ for this.

src/tusb_option.h
```diff
@@ -61,6 +61,8 @@
 #define OPT_MCU_SAME5X            203 ///< MicroChip SAM E5x
 #define OPT_MCU_SAMD11            204 ///< MicroChip SAMD11
 #define OPT_MCU_SAML22            205 ///< MicroChip SAML22
+#define OPT_MCU_SAM7X             206 ///< MicroChip SAME70, S70, V70, V71 family
```
hathach, line 64: there is actually an SAM7X which is different mcu, https://www.microchip.com/developmenttools/ProductDetails/PartNO/AT91SAM7X-EK

Maybe call it `SAMX7X`, I am not good for naming, any other suggestion is welcome.

HiFiPhile: Yes `SAMX7X` works... I can't figurer out something better neither 😅

src/portable/microchip/sam7x/dcd_sam7x.c
HiFiPhile: Maybe, I just copied ASF's code.

HiFiPhile: But ` __disable_irq()` is still needed in https://github.com/hathach/tinyusb/blob/f039607afcf7473ea3869f2e2e7cd763d13f8de0/src/portable/microchip/sam7x/dcd_sam7x.c#L584

hathach: that is quite a lot, we don't disconnect/connect often, but that is a bit weird to go into critical section for this. But let just leave it as it is. We can revisit it later.

src/tusb_option.h
perigoso: ups my bad, had no knowledge of sam7x being a thing, and didnt think to search.

hathach: not a problem at all. I like your naming more, but unfortunately, avoiding confusion has more priority :)

hathach:
@HiFiPhile let me know when you think the PR is ready for testing, or just mark it as Ready. I will pull out my same71 explained to test with.

HiFiPhile:
I got some problem testing HID example, seems like HID descriptor was not sent correctly ?
```
USBD init
HID init
USBD Bus Reset : Full Speed

  Get Descriptor Device
  Queue EP 80 with 18 bytes ... OK
USBD Resume
  Queue EP 00 with 0 bytes ... OK
USBD Resume
USBD Bus Reset : Full Speed
USBD Resume

USBD Setup Received 00 05 01 00 00 00 00 00
  Set Address
USBD Resume
USBD Resume

USBD Setup Received 80 06 00 01 00 00 12 00
  Get Descriptor Device
  Queue EP 80 with 18 bytes ... OK
USBD Resume
  Queue EP 00 with 0 bytes ... OK
USBD Resume
USBD Resume

USBD Setup Received 80 06 00 02 00 00 FF 00
  Get Descriptor Configuration[0]
  Queue EP 80 with 34 bytes ... OK
USBD Resume
USBD Xfer Complete on EP 80 with 34 bytes
  Queue EP 00 with 0 bytes ... OK
USBD Resume
USBD Resume

USBD Setup Received 80 06 03 03 09 04 FF 00
  Get Descriptor String[3]
  Queue EP 80 with 14 bytes ... OK
USBD Resume
USBD Xfer Complete on EP 80 with 14 bytes
  Queue EP 00 with 0 bytes ... OK
USBD Resume
USBD Resume

USBD Setup Received 80 06 00 03 00 00 FF 00
  Get Descriptor String[0]
  Queue EP 80 with 4 bytes ... OK
USBD Resume
USBD Xfer Complete on EP 80 with 4 bytes
  Queue EP 00 with 0 bytes ... OK
USBD Resume
USBD Resume

USBD Setup Received 80 06 02 03 09 04 FF 00
  Get Descriptor String[2]
  Queue EP 80 with 30 bytes ... OK
USBD Resume
USBD Xfer Complete on EP 80 with 30 bytes
  Queue EP 00 with 0 bytes ... OK
USBD Resume
USBD Resume

USBD Setup Received 80 06 00 06 00 00 0A 00
  Get Descriptor Device Qualifier
  Stall EP0
USBD Resume

USBD Setup Received 80 06 00 01 00 00 12 00
  Get Descriptor Device
  Queue EP 80 with 18 bytes ... OK
USBD Resume
  Queue EP 00 with 0 bytes ... OK
USBD Resume
USBD Resume

USBD Setup Received 80 06 00 02 00 00 09 00
  Get Descriptor Configuration[0]
  Queue EP 80 with 9 bytes ... OK
USBD Resume
USBD Xfer Complete on EP 80 with 9 bytes
  Queue EP 00 with 0 bytes ... OK
USBD Resume
USBD Resume

USBD Setup Received 80 06 00 02 00 00 22 00
  Get Descriptor Configuration[0]
  Queue EP 80 with 34 bytes ... OK
USBD Resume
USBD Xfer Complete on EP 80 with 34 bytes
  Queue EP 00 with 0 bytes ... OK
USBD Resume
USBD Resume

USBD Setup Received 00 09 01 00 00 00 00 00
  Set Configuration
  Open EP 81 with Size = 16
  HID opened
  Queue EP 80 with 0 bytes ... OK
USBD Resume
USBD Resume

USBD Setup Received 21 0A 00 00 00 00 00 00
  HID control request
  Queue EP 80 with 0 bytes ... OK
USBD Resume
USBD Resume

USBD Setup Received 81 06 00 22 00 00 2F 01
  Get Descriptor  HID control request
  Queue EP 80 with 64 bytes ... OK
USBD Resume
USBD Xfer Complete on EP 80 with 64 bytes
  Queue EP 80 with 64 bytes ... OK
USBD Resume
USBD Xfer Complete on EP 80 with 64 bytes
  Queue EP 80 with 64 bytes ... OK
USBD Resume
USBD Xfer Complete on EP 80 with 64 bytes
  Queue EP 80 with 47 bytes ... OK
USBD Resume
USBD Xfer Complete on EP 80 with 47 bytes
  HID control complete
  Queue EP 00 with 0 bytes ... OK
USBD Resume
USBD Resume

USBD Setup Received 81 06 00 22 00 00 2F 01
  Get Descriptor  HID control request
  Queue EP 80 with 64 bytes ... OK
USBD Resume
USBD Xfer Complete on EP 80 with 64 bytes
  Queue EP 80 with 64 bytes ... OK
USBD Resume
USBD Xfer Complete on EP 80 with 64 bytes
  Queue EP 80 with 64 bytes ... OK
USBD Resume
USBD Xfer Complete on EP 80 with 64 bytes
  Queue EP 80 with 47 bytes ... OK
USBD Resume
USBD Xfer Complete on EP 80 with 47 bytes
  HID control complete
  Queue EP 00 with 0 bytes ... OK
USBD Resume
USBD Resume

USBD Setup Received 81 06 00 22 00 00 2F 01
  Get Descriptor  HID control request
  Queue EP 80 with 64 bytes ... OK
USBD Resume
USBD Xfer Complete on EP 80 with 64 bytes
  Queue EP 80 with 64 bytes ... OK
USBD Resume
USBD Xfer Complete on EP 80 with 64 bytes
  Queue EP 80 with 64 bytes ... OK
USBD Resume
USBD Xfer Complete on EP 80 with 64 bytes
  Queue EP 80 with 47 bytes ... OK
USBD Resume
USBD Xfer Complete on EP 80 with 47 bytes
  HID control complete
  Queue EP 00 with 0 bytes ... OK
USBD Resume
USBD Suspend
```

hathach:
There is also tons of Resume signal without the Suspend before that. I will help to troubleshoot whenever I could

HiFiPhile:
What's strange is the `desc_hid_report`, first 3* 64 bytes packets are good but last 47 bytes is wrong with Bushound:
```
Device  Length    Phase  Data                                                Description       Delta  Cmd.Phase.Ofs(rep)
------  --------  -----  --------------------------------------------------  ----------------  -----  ------------------
  48.0            CTL    81 06 00 22  00 00 2f 01                            GET DESCRIPTOR      9us        10.1.0        
  48.0       238  IN     05 01 09 06  a1 01 85 01  05 07 19 e0  29 e7 15 00  ............)...   12ms        10.2.0        
                         25 01 95 08  75 01 81 02  95 01 75 08  81 01 05 07  %...u.....u.....               10.2.16       
                         19 00 2a ff  00 15 00 26  ff 00 95 06  75 08 81 00  ..*....&....u...               10.2.32       
                         05 08 19 01  29 05 95 05  75 01 91 02  95 01 75 03  ....)...u.....u.               10.2.48       
                         91 01 c0 05  01 09 02 a1  01 85 02 09  01 a1 00 05  ................               10.2.64       
                         09 19 01 29  05 15 00 25  01 95 05 75  01 81 02 95  ...)...%...u....               10.2.80       
                         01 75 03 81  01 05 01 09  30 09 31 15  81 25 7f 95  .u......0.1..%..               10.2.96       
                         02 75 08 81  06 09 38 15  81 25 7f 95  01 75 08 81  .u....8..%...u..               10.2.112      
                         06 05 0c 0a  38 02 15 81  25 7f 95 01  75 08 81 06  ....8...%...u...               10.2.128      
                         c0 c0 05 0c  09 01 a1 01  85 03 15 00  26 ff 03 19  ............&...               10.2.144      
                         00 2a ff 03  95 01 75 10  81 00 c0 05  01 09 05 a1  .*....u.........               10.2.160      
                         
                         01 85 04 05  01 09 30 09  31 09 32 09  35 09 33 09  ......0.1.2.5.3.               10.2.176      
                         15 00 c0 25  7f 95 06 75  08 81 02 05  01 09 39 15  ...%...u......9.               10.2.192      
                         01 25 08 35  00 46 3b 01  95 01 75 08  81 02 05 09  .%.5.F;...u.....               10.2.208      
                         19 01 29 20  15 00 25 01  95 20 75 01  81 02        ..) ..%.. u...                 10.2.224  
```
Which is supposed to be 
```
USBD Setup Received 81 06 00 22 00 00 2F 01
  Get Descriptor  HID control request
  Queue EP 80 with 64 bytes ... Xfer: 05 01 09 06 A1 01 85 01 05 07 19 E0 29 E7 15 00 25 01 95 08 75 01 81 02 95 01 75 08 81 01 05 07 19 00 2A FF 00 15 00 26 FF 00 95 06 75 08 81 00 05 08 19 01 29 05 95 05 75 01 91 02 95 01 75 03
OK
USBD Xfer Complete on EP 80 with 64 bytes
  Queue EP 80 with 64 bytes ... Xfer: 91 01 C0 05 01 09 02 A1 01 85 02 09 01 A1 00 05 09 19 01 29 05 15 00 25 01 95 05 75 01 81 02 95 01 75 03 81 01 05 01 09 30 09 31 15 81 25 7F 95 02 75 08 81 06 09 38 15 81 25 7F 95 01 75 08 81
OK
USBD Xfer Complete on EP 80 with 64 bytes
  Queue EP 80 with 64 bytes ... Xfer: 06 05 0C 0A 38 02 15 81 25 7F 95 01 75 08 81 06 C0 C0 05 0C 09 01 A1 01 85 03 15 00 26 FF 03 19 00 2A FF 03 95 01 75 10 81 00 C0 05 01 09 05 A1 01 85 04 05 01 09 30 09 31 09 32 09 35 09 33 09
OK
USBD Xfer Complete on EP 80 with 64 bytes
  Queue EP 80 with 47 bytes ... Xfer: 34 15 81 25 7F 95 06 75 08 81 02 05 01 09 39 15 01 25 08 35 00 46 3B 01 95 01 75 08 81 02 05 09 19 01 29 20 15 00 25 01 95 20 75 01 81 02 C0
```

HiFiPhile:
I'm wondering if it's some side effect when Segger RTT access DTCM trying to search `_SEGGER_RTT` buffer.
No longer to reproduce the issue after fixed RTT address manually:
```
__jlinkExecCommand(SetRTTAddr 0x20400000);
```

HiFiPhile:
I've also (finally) tested ISO transfer. I think it's ready for test.

examples/device/hid_composite/src/tusb_config.h
```diff
@@ -69,7 +69,7 @@
 #endif
 
 // CFG_TUSB_DEBUG is defined by compiler in DEBUG build
-// #define CFG_TUSB_DEBUG           0
+#define CFG_TUSB_DEBUG           0
```
hathach, line 72: possibly typo, this should be reverted

examples/device/cdc_msc/src/msc_disk.c
```diff
@@ -23,7 +23,7 @@
  *
  */
 
-#include "bsp/board.h"
+//#include "bsp/board.h"
```
hathach, line 26: possibly typo, this should be reverted

src/class/cdc/cdc_device.c
```diff
@@ -481,4 +481,18 @@ bool cdcd_xfer_cb(uint8_t rhport, uint8_t ep_addr, xfer_result_t result, uint32_
...
+  return &_cdcd_itf[itf].rx_ff;
+}
+
+// Get the transmit FIFO (for DMA transfer)
+tu_fifo_t* tud_cdc_n_get_tx_ff (uint8_t itf)
+{
+  TU_ASSERT(itf < CFG_TUD_CDC);
+  return &_cdcd_itf[itf].tx_ff;
+}
+
```
hathach, line 497: this should be on its pr, please revert

src/class/cdc/cdc_device.h
```diff
@@ -104,6 +104,12 @@ uint32_t tud_cdc_n_write_available (uint8_t itf);
 // Clear the transmit FIFO
 bool tud_cdc_n_write_clear (uint8_t itf);
 
+// Get the Receive FIFO (for DMA transfer)
+tu_fifo_t* tud_cdc_n_get_rx_ff (uint8_t itf);
+
+// Get the transmit FIFO (for DMA transfer)
+tu_fifo_t* tud_cdc_n_get_tx_ff (uint8_t itf);
```
hathach, line 111: should be reverted since it has its own pr.

src/class/cdc/cdc_device.h
```diff
@@ -241,6 +247,18 @@ static inline bool tud_cdc_write_clear(void)
...
+{
+  return tud_cdc_n_get_rx_ff(0);
+}
+
+// Get the transmit FIFO
+static inline tu_fifo_t* tud_cdc_get_tx_ff (void)
+{
+  return tud_cdc_n_get_tx_ff(0);
+}
+
```
hathach, line 261: should be reverted since it has its own pr.

hathach (changes requested):
Thank you for your update. Everything looks great, I have an same70 explained, that could be used to test with. Though please revert the changes for exposing cdc fifo rx/tx, I think you probably got mixed up with your local branches.

src/class/cdc/cdc_device.h
HiFiPhile: Oops.. now it's better.

hathach:
I have tested this out on my same70 xplained board. It works well with `board_test work` for led & button. However  on  `cdc_msc` & `hid_composite` examples, but I couldn't get it enumerated at all. 

here is my compile output (gcc9.3.1) , size seems to be a bit small

```
LINK _build/same70_xplained/cdc_msc.elf
CREATE _build/same70_xplained/cdc_msc.bin
CREATE _build/same70_xplained/cdc_msc.hex

  11080    8244    5060   24384    5f40 _build/same70_xplained/cdc_msc.elf
```

analyzer show that same70 doesn't even ACK the setup package, there is probably something with hardware/phy issue. Are you also doing test with same70_xplained or other board !
image

HiFiPhile:
> Are you also doing test with same70_xplained or other board 

I test on `same70-xplained` and my own hardware.

It seems in `board_init` it lacks the switch to external xtal, and `UDP_Handler` should be `USBHS_Handler`

I'll try to update the BSP with my clock config:
```
/* Marco for register access */
#ifndef CLEAR_REG
#define CLEAR_REG(REG)        ((REG) = (0x0))
#endif

#ifndef WRITE_REG
#define WRITE_REG(REG, VAL)   ((REG) = (VAL))
#endif

#ifndef READ_REG
#define READ_REG(REG)         ((REG))
#endif

#ifndef MODIFY_REG
#define MODIFY_REG(REG, CLEARMASK, SETMASK)  WRITE_REG((REG), (((READ_REG(REG)) & (~(CLEARMASK))) | (SETMASK)))
#endif

typedef enum
{
    MCK_Freq_10M  = 0 << 24 | 3 << 16 | 40 << 8 | 4,
    MCK_Freq_20M  = 0 << 24 | 3 << 16 | 40 << 8 | 3,
    MCK_Freq_50M  = 2 << 24 | 3 << 16 | 50 << 8 | 2,
    MCK_Freq_100M = 4 << 24 | 3 << 16 | 50 << 8 | 1,
    MCK_Freq_150M = 6 << 24 | 1 << 16 | 25 << 8 | 1
} MCK_Freq_t;


void Clock_Config(MCK_Freq_t Freq)
{
    uint8_t fws = Freq >> 24;
    uint8_t div = Freq >> 16;
    uint8_t mul = Freq >> 8;
    uint8_t pres = Freq;
    // Flash Wait State, 4 for 100MHz, 6 for 150MHz
    MODIFY_REG(EFC->EEFC_FMR, 0, EEFC_FMR_FWS(fws));
    MODIFY_REG(PMC->CKGR_MOR, CKGR_MOR_MOSCXTBY, CKGR_MOR_KEY_PASSWD | CKGR_MOR_MOSCXTST(0x80));
    MODIFY_REG(PMC->CKGR_MOR, 0, CKGR_MOR_KEY_PASSWD | CKGR_MOR_MOSCXTEN);
    while (!(PMC->PMC_SR & PMC_SR_MOSCXTS)) {
        /* Wait until XTAL stabilize */
    }
    MODIFY_REG(PMC->CKGR_MOR, 0, CKGR_MOR_KEY_PASSWD | CKGR_MOR_MOSCSEL);
    // Enable PLL, Div [div], Mul [mul]
    PMC->CKGR_PLLAR = CKGR_PLLAR_ONE | CKGR_PLLAR_MULA(mul - 1) | CKGR_PLLAR_DIVA(div) | CKGR_PLLAR_PLLACOUNT(0x3fU);
    while (!(PMC->PMC_SR & PMC_SR_LOCKA)) {
        /* Wait until PLLACK stabilize */
    }
    
    // Select PLLACK as MCLK, Div [2^pres]
    MODIFY_REG(PMC->PMC_MCKR, PMC_MCKR_CSS_Msk | PMC_MCKR_PRES_Msk, PMC_MCKR_CSS_PLLA_CLK | PMC_MCKR_PRES(pres));

    // Config SySTick to generate 1ms tick
    SystemCoreClock = 12000000 / div * mul / (1 << pres);
    SysTick->CTRL |= SysTick_CTRL_CLKSOURCE_Msk;
    SysTick->LOAD  = (uint32_t)(SystemCoreClock / 1000 - 1);
    SysTick->VAL   = 0UL;
    SysTick->CTRL  = SysTick_CTRL_CLKSOURCE_Msk |
                   SysTick_CTRL_TICKINT_Msk   |
                   SysTick_CTRL_ENABLE_Msk;
    NVIC_SetPriority (SysTick_IRQn,2);
}
```

hathach:
@HiFiPhile thanks for the update, it is better, but still seems to be stuck after enumeration (descriptor is fine), msc disk does not show up, cdc connection doesn't quite work. I also found the usb bus does not work when I tried to hook it with hardware analyzer, which is weird !!

HiFiPhile:
I think something is wrong with the redundant init of clock, I removed the clock init in dcd and there is no more enumeration failure.

Once again EP0 bugged, OUT packet miss-aligned by one transfer.
```
USBD Setup Received 21 20 00 00 02 00 07 00
  CDC control request
  Set Line Coding
  Queue EP 00 with 7 bytes ...
NULL
  CDC control complete
  Queue EP 80 with 0 bytes ...

USBD Setup Received A1 21 00 00 02 00 07 00
  CDC control request
  Get Line Coding
  Queue EP 80 with 7 bytes ...
USBD Xfer Complete on EP 80 with 7 bytes
  CDC control complete
  Queue EP 00 with 0 bytes ...
USBD Xfer Complete on EP 00 with 7 bytes
usbd_control_xfer_cb 171: ASSERT FAILED
```

I don't know how I managed to run multiple examples with this issue, seems it's not quite ready...

perigoso:
Clock initiation is supposed to be done outside tinyusb anyway is it not?

hathach:
> Clock initiation is supposed to be done outside tinyusb anyway is it not?

That also depends on the port, if the clock depends on bsp component such as xtal, and/or clock source etc. Generally, having it out of dcd make dcd  more portable e.g easier to support other mcu such as sam3u etc..

@HiFiPhile I am currently busy on other work and couldn't help to fix this. We could merge this as it is and create an issue for it so that we could come back later on when having time. Of course, it is up to you if you want to work more on this.

HiFiPhile:
You know what, I fight for NAKed transfer for 2 days with a bad cable ! 

This cable works on my lpc4357 so I assume it's ok but in fact the connector of same70-xplained is a little loose lol...

Yes we can merge at current state as the data transfer works, I put a workaround for EP0 issue caused by spurious OUT irq, we can come back later to investigate.

hathach:
> You know what, I fight for NAKed transfer for 2 days with a bad cable !
> 
> This cable works on my lpc4357 so I assume it's ok but in fact the connector of same70-xplained is a little loose lol...

It is probably more than the cable, I found that the bus is very sensitive. I could see it get passed enumeration if attached to the PC, but same70 refuse to communicate when using hardware analyzer. Which indicate it has some bus/phy issue.
 
> Update: With [eec927e](https://github.com/hathach/tinyusb/commit/eec927ea959d6d972c62d60fa634df227e196b36) I think EP0 issue is finally fixed. There is no more assert hit and CDC Set/Get line coding works correctly.

I have pulled and tried the latest, unfortunately, it doesn't improve much, same result as previously with my ubuntu machine. CDC connection and echo does not work, msc drive doesn't show up and couldn't captured usb bus with analyzer

> Yes we can merge at current state as the data transfer works, I put a workaround for EP0 issue caused by spurious OUT irq, we can come back later to investigate.

Let me know if you are done working on same70, we will merge it for now as reference.

HiFiPhile:
> I have pulled and tried the latest, unfortunately, it doesn't improve much, same result as previously with my ubuntu machine.

Could you force it in full speed mode to see if it helps ?

hathach:
I force fullspeed by commenting out `OPT_MODE_FULL_SPEED` in descriptor but it doesn't help, it doesn't  get passed enumeration at all.

HiFiPhile:
Try change only `BOARD_DEVICE_RHPORT_SPEED` in tusb_conf.h, I use TUD_OPT_HIGH_SPEED in dcd init.

I've no problem running high speed with my AMD Arch pc, could you try my compiled `cdc_dual_ports` 
[same70.zip](https://github.com/hathach/tinyusb/files/6853645/same70.zip)

hathach:
@HiFiPhile with latest push the cdc_msc now work great yeah !!! For analyzer, you are right about the cable, it work with very short cable (0.3m length), while other 0.5-1m doesn't, even that it works with other HS board such as lpc18xx. I guess it is more peaky than others.

Finally it works, though I go further to test with net lwip example, but it failed that one. Look like setup is still missing, it is probably overrun or race condition. 

image

hathach (approved):
Superb work !! Thank you very much for your PR and effort to troubleshoot the issue with the PR.  I have tested cdc_msc, hid_composite example both work greats. There is still issue with net_lwip_webserver and webusb examle, it is probably due to race condition and/or the missing setup etc ... Which we could definitely fix later with followup PRs. The race condition is more severe with HS since bus is much faster and so does the mcu.

HiFiPhile:
>  do you have any pending code or thing that want to push.

I'll take a look of net_lwip_webserver if I can make a quick fix, otherwise we can come back later.

HiFiPhile:
It's simply because SAME70 need different EP for IN and OUT. 

Now WebUSB echo works but I can't see `\r\nTinyUSB WebUSB device example\r\n` which is received in BusHound.

hathach:
> It's simply because SAME70 need different EP for IN and OUT.
> 
> Now WebUSB echo works but I can't see `\r\nTinyUSB WebUSB device example\r\n` which is received in BusHound.

Ah thanks, I didn't notice that. Great, merge now, #975 will make it easier for cross-platform application to detect these requirement

fjgpereira:
I managed to get tinyusb/CDC workin on a SAMS70, but had to adapt the code to
the new microchip include files from Harmony3 - XC32.
Detected a few cache/DMA issues that were solved with 2 clean/invalidate cache instructons.
Thank you very much.

HiFiPhile:
Hi @fjgpereira ,

>  had to adapt the code to the new microchip include files from Harmony3 - XC32.

So with Harmony3 the register defines are different ?

> Detected a few cache/DMA issues that were solved with 2 clean/invalidate cache instructions.

In the DCD driver I haven't included cache clean/invalidate since it's configuration dependant. For example in my case I use DTCM/ITCM and set normal sram to non-cached.

I can add a define to enable them when needed.

hathach:
maybe you could add an PR for cache, it is typically required on M7 mcu. I did something similar for rt10xx 
https://github.com/hathach/tinyusb/blob/master/src/portable/nxp/transdimension/dcd_transdimension.c#L48

fjgpereira:
Hi,

The include files i found on the Harmony3 project seem to have suffered a few changes, but for the most part the symbols are the same.
For instance most macros now require a value/parameter, etc, the USBHS_DEVDMA array now is spelled in upercase, etc.

In case it has any use, I placed a patch/diff file with the changes at [http://inocam.com/~fjp/dcd_samx7x_H3.diff]

Regarding cache, everything looked to be working fine with ICache enabled and DCache disabled, but to enable DCache
required the invalidate/clean instructions. Maybe it can be placed inside #ifdefs.

I did not investigate the reason why, but It was also necessary to add a __DSB() barrier 
to dcd_event_setup_received() on usbd.c after a call to memcpy(), or else Setup events (GetStatus, etc.) were
not processed correctly.

HiFiPhile:
> maybe you could add an PR for cache, it is typically required on M7 mcu. I did something similar for rt10xx
> https://github.com/hathach/tinyusb/blob/master/src/portable/nxp/transdimension/dcd_transdimension.c#L48

I think it's better to add a on/off switch, also for dcd_transdimension.

Clean/invalidate cache will cause hardfault in case of it is not enabled.

How I edited your comment :D

hathach:
> I think it's better to add a on/off switch, also for dcd_transdimension.
> Clean/invalidate cache will cause hardfault in case of it is not enabled.
> How I edited your comment :D

I didn't see any issue with rt10xx so far, I don't remember caching within stock example. If it is an issue on other MCU, it is better to detect if cached is enabled or not. Adding an macro is easy, but that config file is growing and already complicated to user (than I would like it to).

fjgpereira:
Don't know about Microchip plans, but I looked into the include files supplied with XC32 compiler and found that it also uses the new naming conventions.

I didn't check if there are any standard macros to detect the Microchip tools/versions, but as the registers pointer names are different, maybe you can use something like this:

#ifdef USBHS_REGS
#define   USBHS     USBHS_REGS
#define   PMC         PMC_REGS
#  define __MP__    (1)
#else 
#  defne __MP__
#endif

and then add __MP__ after each macro.

HiFiPhile:
@hathach I think it's also possible to include register defines in tinyUSB, like dcd_transdimension.

hathach:
It is possible, but (transdimention) tdi is a bit extreme since all naming and register address and mapping is totally different between lpc and rt10xx. It is probably needed if we also bundle sam3u etc into the same dcd. However, it will depend on actual register naming/mapping from microchip. The naming here is not that bad, it seems only take an extra argument for some macros. Anyway, I would prefer to review actual code before making any decision at all.

HiFiPhile:
> I did not investigate the reason why, but It was also necessary to add a __DSB() barrier
to dcd_event_setup_received() on usbd.c after a call to memcpy(), or else Setup events (GetStatus, etc.) were
not processed correctly.

How did you configure the MPU regions ?

HiFiPhile:
I'm making changes in #985 .

fjgpereira:
Actually most of the CPU initialization code and peripheral libraries is generated automatically by Harmony3 and I usually only look inside that code when I find some bug or hardware feature not implemented by the default code generated automatically, so I probably am not able to help.

However, MPU is disabled in the Harmony3 graphical tool, but I didn't check if it applies any default configuration.
Looking at the generated code, I also don't find any MPU initialization
(that is: I cannot find the MPU init code that can be found on H3 examples that make use of the MPU)

Regarding that problem on dcd_evfent_setup_received(), I initially noticed that the TU_LOG2_VAR() debug messages were displaying wrong bmRequestType/direction bits on GetStatus requests. Then I noticed the problem did not happen when some log message was inserted after the memcpy(), that lead to conclude that any code that caused "delay" in that spot would solve the problem, so I guess it can be related to cache coherency issues.

HiFiPhile:
>  I also don't find any MPU initialization

That's strange... so basically Harmony3 enable cache without prior MPU region configuration ?

But by default the memory type of USB FIFO is 'Device Memory' so it's not cached...

fjgpereira:
> That's strange... so basically Harmony3 enable cache without prior MPU region configuration ?

I think that is the default Harmony3 settings for new projects.

When the MPU is enabled, it does offer a recommended configuration with 6 regions, ITCM, DTCM, SRAM, EBI_SMC, EBI_SRAM and QSPI, ranging from 0x0 to 0x8ffffffff, but I don't know what applies to 0xA0100000 without specifically defining another region.

perigoso:
And that is why Harmony is not used

fjgpereira:
> And that is why Harmony is not used

So far I've found a few bugs in the auto-generated code, specially with unusual device configurations, but all in all, I think it has saved me a lot of time. However, I almost only use CPU and peripheral initialization code, avoiding other libraries and RTOS.
The tools for pin assignment were specially useful to draw the board schematics, as it helped choose peripherals and assign pins to each function. I think it helped get the first prototype PCBs working without mistakes.

HiFiPhile:
> And that is why Harmony is not used

> I think it has saved me a lot of time

I think it's ture for every MCU companion software I've used (STM32Cube, MCUXpresso, Simplicity studio, etc.) while pin tool is practical for PCB planning, their generated code is very pathetic. I always ended with direct register access.

perigoso:
Sure you can save a lot of time, and if you're lucky you can have a good experience with it. But you run the risk of having a veeery bad time, it's just not reliable.

fjgpereira:
> Sure you can save a lot of time, and if you're lucky you can have a good experience with it. But you run the risk of having a veeery bad time, it's just not reliable.

I suspected that and decided to use just a minimal set of features.
That's why I decided to try tinyusb instead of the Harmony's USB stack
that has lots of dependencies with other libraries and rtos.
For the same reason, avoided the E70 GMAC and corresponding TCP/IP stack and
instead chose an WizNet W6100 for TCP/IP.

PR #700, opened 2021-03-04
fix enum walkaround forever check for SE0 when pull up is disabled

hathach:
**Describe the PR**
This issue is discovered while troubleshooting https://github.com/adafruit/circuitpython/issues/4066#issuecomment-789524956

Part of enumeration walkaround is waiting for bus reset to complete i.e line state != SE0. 
https://github.com/raspberrypi/pico-sdk/blob/master/src/rp2_common/pico_fix/rp2040_usb_device_enumeration/rp2040_usb_device_enumeration.c#L61

However, when pull-up is disabled e.g application call `tud_disconnect()` to soft disconnect from Host (while still physically plugged in). SE0 will be permanent present causing the walkround forever checking for the state unitl `tud_connect()` (pullup is enabled) or cpu reset. Worse if `PICO_TIME_DEFAULT_ALARM_POOL_DISABLED = 1`, the walkaround will blockingly wait --> sort of hanged there.

To reproduce, add this to any of the example, press the button, rp2040 will disconnect but forever hanged if PICO_TIME_DEFAULT_ALARM_POOL_DISABLED=1
```
if ( board_button_read() )
{
  printf("Disconnect\n");
  tud_disconnect();
  board_delay(2000);
  printf("Re-connect\n");
  tud_connect();
}
```

**Solution**

Only run enumeration walkaround if pull-up is enabled. 

**Additional Context**

I add a bit of TODO for VBus detect and Suspend & Resume later on. Note to myself: the pico schematic use GPIO24 which is is actually VBUS over current detection (FUN9), will need a jump wire to test later on.

### DIFF
--- a/hw/bsp/rp2040/family.c
+++ b/hw/bsp/rp2040/family.c
@@ -80,8 +80,8 @@ bool __no_inline_not_in_flash_func(get_bootsel_button)() {
 #if defined(LOGGER_RTT)
 
 // Logging with RTT
-// If RTT Control Block is not found by 'Auto Detection`
-// try to use 'Search Range` with '0x20000000 0x10000'
+// - If RTT Control Block is not found by 'Auto Detection` try to use 'Search Range` with '0x20000000 0x10000'
+// - SWD speed is rather slow around 1000Khz
 
 #include "pico/stdio/driver.h"
 #include "SEGGER_RTT.h"
--- a/hw/bsp/rp2040/family.mk
+++ b/hw/bsp/rp2040/family.mk
@@ -1,3 +1,4 @@
 FAMILY_SUBMODULES = hw/mcu/raspberrypi/pico-sdk
 
 JLINK_DEVICE = rp2040_m0_0
+PYOCD_TARGET = rp2040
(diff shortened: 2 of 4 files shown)


liamfraser (approved):

liamfraser:
Looks good to me. In terms of VBUS, you can use a level high / level low interrupt on any GPIO to provide a vbus changed interrupt. That is to say, you don't need to use vbus detect from the actual usb controller input.

hathach:
> Looks good to me. In terms of VBUS, you can use a level high / level low interrupt on any GPIO to provide a vbus changed interrupt. That is to say, you don't need to use vbus detect from the actual usb controller input.

Does it trigger the VBUS_DETECT interrupt in sie_status. I would hop to handle it internally by the stack without user to call any API. For example, on nRF52, the PHY is still disabled after tud_init() and only enabled after VBUS is detected. Then it got disabled again after Disconnection is detected to save even more power. I will do more testing with that scheme later on with rp2040.

liamfraser:
If you look at the gpio function table, only some gpios are connected to USB VBUS DET (which can drive the VBUS_DETECT interrupt status). I am saying that you could use any GPIO interrupt to add and remove vbus detect in software (by forcing and removing vbus detect like we do now when the interrupt fires). This effectively lets you use any GPIO as vbus detect.

hathach:
That is what I guess as well. It will needs an extra API for tinyusb stack to interact with application regarding vbus so it could distinguish between disconnected and suspended as mentioned in the datasheet. Thanks for your advice, I will look at this later on.

PR #708, opened 2021-03-05
Add back cmsis5 as submodule

hathach:
**Describe the PR**
Add back lib/CMSIS_5 as submodules https://github.com/hathach/tinyusb/discussions/705 . Also slightly optimize ci submodules checkout


dhalbert:
Does this mean that we now need to update submodules recursively in order to fetch this. In CircuitPython we have been disrecommending `git submodule update --recursive` because it pulled in extra tinyusb submodules that weren't needed.

hathach:
@dhalbert circuitpython or any super project only need to check out tinyusb and shouldn’t check out any submodules of tinyusb to compile. Super project should only use src folder of tinyusb, any other submodules/folder are subject to change

PR #713, opened 2021-03-10
Call One time tu_fifo_write_n on cdcd_xfer_cb

HiFiPhile:
**Describe the PR**
I thought it was included in old PRs but seems not...

In my CDC test with SAME70, throughput increased from 800KB/s to 27MB/s.

### DIFF
--- a/src/class/cdc/cdc_device.c
+++ b/src/class/cdc/cdc_device.c
@@ -95,7 +95,8 @@ static void _prep_out_transaction (cdcd_interface_t* p_cdc)
   // fifo can be changed before endpoint is claimed
   available = tu_fifo_remaining(&p_cdc->rx_ff);
 
-  if ( available >= sizeof(p_cdc->epout_buf) )  {
+  if ( available >= sizeof(p_cdc->epout_buf) )
+  {
     usbd_edpt_xfer(rhport, p_cdc->ep_out, p_cdc->epout_buf, sizeof(p_cdc->epout_buf));
   }else
   {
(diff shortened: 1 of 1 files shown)


hathach (changes requested):
thanks for the PR, yeah, this is one of the improvement I planed to do but totally forgot. However, the implementation is not actually quite right. The driver must invoke the wanted_cb() **right at the time it push the wanted_char into the fifo**. which allow the application callback that call read() and see the wanted_chr as the **LAST** character in the fifo.

For example, to  catch the `\n` then  call read() to get the whole line. ~~Else application must skip dozen of characters~~. which can potentially has another '\n' ( received multi-line within a packet) and cause some confusion when the callback is invoked with more data than it could expect. 

Solution: we should still keep the loop however instead of writing one byte one, we keep an count and write_n(count) when we see the wanted (or write all if found none).

src/class/cdc/cdc_device.c
```diff
@@ -432,25 +432,25 @@ bool cdcd_xfer_cb(uint8_t rhport, uint8_t ep_addr, xfer_result_t result, uint32_
...
-      tu_fifo_write(&p_cdc->rx_ff, &p_cdc->epout_buf[i]);
-
+    if (tud_cdc_rx_wanted_cb && ( ((signed char) p_cdc->wanted_char) != -1)) {
       // Check for wanted char and invoke callback if needed
-      if ( tud_cdc_rx_wanted_cb && ( ((signed char) p_cdc->wanted_char) != -1 ) && ( p_cdc->wanted_char == p_cdc->epout_buf[i] ) )
-      {
-        tud_cdc_rx_wanted_cb(itf, p_cdc->wanted_char);
+      for (uint32_t i=0; i<xferred_bytes; i++) {
+        if ( p_cdc->wanted_char == p_cdc->epout_buf[i] ) {
+          tud_cdc_rx_wanted_cb(itf, p_cdc->wanted_char);
```
hathach, line 442: hmm, maybe it is not worth to have `wanted` as the absolute last character. Maybe we can implement it this way. Though we may need to check if data is still available before invoking callback (much like the cdc_rx_cb() below) since previous callback can potentially read() all the data.

hathach (commented):
After giving it more thought, I think it is not worth to have wanted as the last char in the fifo at the time of the callback. It doesn't occur often and also doesn't have much impact on existing usage (application may need a minor of adjustment). Let's do it this way, though we still need to check if there is still data in the fifo before invoking the wanted callback, else it could confuse the app.

src/class/cdc/cdc_device.c
HiFiPhile: After some thought I feel more hesitate about how to implement it :>

While I use UART extensively in our product, I always use a packet so there is no `last char` problem.

We also need to make sure data is written (fifo not full) before call the cb. If the application relies on `tud_cdc_rx_wanted_cb`, there could be a deadlock where the fifo is full but no wanted char received.

If it's a personal project I'd delete this cb and let the application to do all the decoding :) I think we can ask more people about what they think the best way to do it.

hathach: Oh, I didn't think about the edge case where fifo is full and wanted char has yet to come. Maybe we should have an `rx_full_cb()`. However, it should not be affected by existing code or your PR. After all, if application want to process a large data with wanted ch (e.g command with newline), they should tune up the fifo to large enough. Anyway, I think we should just continue with this PR, would you mind to add the check if fifo not empty before invoked the wanted callback. That would be enough for the merge. If needed the rx_full_cb() can be added as separated PR.

HiFiPhile: I've added fifo empty check in last commit.

hathach (approved):
Perfert !! though I just realized the fifo_empty() did a quicker check than computing the actual count. and changes to both callback check to !fifo_empty().

PR #723, opened 2021-03-16
fix for nrf52840 epout dma race condition

khpeterson:
Fix for https://github.com/adafruit/Adafruit_nRF52_Bootloader/issues/195.

Addresses race condition between writing SIZE.EPOUT=0 and xfer->data_received flag.  Since NR52840 always auto ACKs after DMA completion the code should always wait for data_received to be sure the transfer is complete.  This doesn't work when starting a sequence of transfers so timeout and write SIZE.EPOUT in this case.  If one could differentiate the start/end of a packet sequence at this level of the code then the timeout could be avoided (though always a good idea when spin waiting on anything).

### DIFF
--- a/src/portable/nordic/nrf5x/dcd_nrf5x.c
+++ b/src/portable/nordic/nrf5x/dcd_nrf5x.c
@@ -273,6 +273,18 @@ bool dcd_edpt_xfer (uint8_t rhport, uint8_t ep_addr, uint8_t * buffer, uint16_t 
 
   xfer_td_t* xfer = get_td(epnum, dir);
 
+  if (dir == TUSB_DIR_OUT) {
+    // hw auto ack'd xfer so wait until it's done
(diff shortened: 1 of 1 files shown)


hathach:
Thank you for your investigation and PR. I am pulling out the nrf52840 datasheet for further reading and testing. Will try to find an way to best avoid blocking wait if possible.

khpeterson:
That’s my conclusion too, only write SIZE.EPOUT[epnum] once to get things started.  I tried that while debugging (by omitting the spin wait in my PR) but ended up in a hung state.  From what I could tell it was necessary to kick of the DMA by writing SIZE.EPOUT[epnum] at the beginning of each packet sequence (e.g. for each firmware packet of roughly 512 bytes sent by the host).  Anyway, I didn’t try too hard so I was probably missing something else that caused the hang.

hathach:
closed in prefer of #733 with non-blocking approach. Thank you very much of raising the issue

PR #724, opened 2021-03-16
USB TMC Updates

xmos-jmccarthy:
**Describe the PR**
Adds alignment to buffers in usbtmc_interface_state_t struct.
Updates the visaQuery.py script for PyVisa module import deprecation and to support Linux.

**Additional context**
It was found that the CFG_TUSB_MEM_ALIGN attributes were missing from the buffers in the usbtmc_interface_state_t struct.  This caused exceptions due to alignment issues on the XCore platform.  This addition is untested on other platforms.
Additionally the existing visaQuery.py script was hardcoded for use with Windows.  Linux support was added.

### DIFF
--- a/examples/device/usbtmc/visaQuery.py
+++ b/examples/device/usbtmc/visaQuery.py
@@ -1,6 +1,6 @@
 #!/usr/bin/env python3
 
-import visa
+import pyvisa as visa
 import time
 import sys
 
@@ -36,8 +36,8 @@ def test_trig():
 	time.sleep(0.3) # SRQ may have some delay
 	assert (inst.read_stb() & 0x40), "SRQ not set after 0.3 seconds"
 	assert (inst.read_stb() == 0)
-	
-	
+
+
 def test_mav():
 	inst.write("delay 50")
 	inst.read_stb() # clear STB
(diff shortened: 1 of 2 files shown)


hathach (commented):
the modificaiton of `usbtmc_device.c` looks good to me. Though can you also add the CFG_TUSB_MEM_SECTION to the usbtmc_state as well at line 128, that will help some mcu that has picky DMA.

```
CFG_TUSB_MEM_SECTION static usbtmc_interface_state_t usbtmc_state =
{
    .itf_id = 0xFF,
};
```

For the `visaQuery.py` changes, I have no idea, @pigrew do you have time to help reviewing this.

pigrew:
My putting a particular RM DLL path was a bad idea, in retrospect. It was a hacky way for me to get it working on my laptop, and I shouldn't have done it. Also, it may not work for all installs of Windows.

What are your thoughts about if we remove all of the os-detection, and just have a call to:

```rm = visa.ResourceManager()```

With that, one can create a `.pyvisarc` with the proper library path based on which VISA library the user wants (if pyVISA's auto-selection doesn't work)?

(And, in general, it's good to see other people trying out the class driver. If you think there should be API changes, let me know.)

xmos-jmccarthy:
PR is updated to only add the CFG_TUSB_MEM_ALIGN and CFG_TUSB_MEM_SECTION attributes to the usbtmc implementation.

I will log an issue for how the test script is used for it to be handled external to this PR.

pigrew:
@xmos-jmccarthy,

Thanks. I'll create a PR to edit the python script.

@hathach,

I'm happy to merge this PR.

hathach (approved):
thanks @xmos-jmccarthy for your PR and @pigrew for reviewing it.

PR #733, opened 2021-03-18
fix race condition that could cause drop packet of Bulk OUT transfer

hathach:
**Describe the PR**
related to #723, fix issue raised by adafruit/Adafruit_nRF52_Bootloader#195.
- NRF_USBD->SIZE.EPOUT[epnum] only need to write once to enable Bulk/Interrupt transfer. We only need to do it in dcd_edpt_open() and dcd_edpt_clear_stall()
- Separate Control and Bulk/Int OUT hanlding
- Some minor comment and clean up

@khpeterson I have tested and tried it with your dfu file. Seem to work well without blocking wait. Would you mind trying this out.

### DIFF
--- a/src/portable/nordic/nrf5x/dcd_nrf5x.c
+++ b/src/portable/nordic/nrf5x/dcd_nrf5x.c
@@ -53,13 +53,11 @@ enum
 
 enum
 {
-  // Endpoint number is fixed (8) for ISOOUT and ISOIN.
-  EP_ISO_NUM = 8,
-  // CBI endpoints count
-  EP_COUNT = 8
+  EP_ISO_NUM   = 8, // Endpoint number is fixed (8) for ISOOUT and ISOIN
+  EP_CBI_COUNT = 8  // Control Bulk Interrupt endpoints count
 };
 
-// Transfer descriptor
+// Transfer Descriptor
 typedef struct
 {
   uint8_t* buffer;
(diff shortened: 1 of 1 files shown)


khpeterson:
Running repeats now...

hathach:
> Running repeats successfully for an hour now - looks good to me. Good job on the quick fix.

I am looking to your final number. Btw, the race condition probably get worse with the zero length packet. CDC host driver normally will send ZLP when all data is sent and is multiple of 64 (endpoint size). Hope we finally fix the issue.

khpeterson:
OK, I will run my repeat script overnight tonight and send you the result tomorrow morning.

On Mar 18, 2021, at 10:26 AM, Ha Thach ***@***.******@***.***>> wrote:



Running repeats successfully for an hour now - looks good to me. Good job on the quick fix.

I am looking to your final number. Btw, the race condition probably get worse with the zero length packet. CDC host driver normally will send ZLP when all data is sent and is multiple of 64 (endpoint size). Hope we finally fix the issue.

hathach:
1 hours is good enough for me, don't leave it running too long, it could degrade your nrf flash

khpeterson:
OK, just finished a little more than 1 hour of repeats - 110 trials, no failures.  Looks good to me

On Mar 18, 2021, at 10:31 AM, Ha Thach ***@***.******@***.***>> wrote:



1 hours is good enough for me, don't leave it running too long, it could degrade your nrf flash

hathach:
@khpeterson thank you very much for testing and raising this very hard to catch issue :+1:

PR #751, opened 2021-03-24
Allow TU_BREAKPOINT() to break for a Cortex-M0+ also

ned-pcs:
**Describe the PR**
The Cortex-M0+ (Arm architecture 6M) also supports the
DHCSR.C_DEBUGEN bit (bit 0) to tell whether a debugger is attached.
This PR adds that architecture to TU_BREAKPOINT().

**Additional context**

### DIFF
--- a/src/common/tusb_verify.h
+++ b/src/common/tusb_verify.h
@@ -83,8 +83,8 @@
   #define _MESS_FAILED() do {} while (0)
 #endif
 
-// Halt CPU (breakpoint) when hitting error, only apply for Cortex M3, M4, M7
-#if defined(__ARM_ARCH_7M__) || defined (__ARM_ARCH_7EM__)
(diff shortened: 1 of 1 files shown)


hathach:
have you tried it on actual hardware. I could be wrong, but I remember both m0 and m0+ doesn't support bit. 

https://github.com/ARM-software/CMSIS_5/blob/b7b26f50d00072812aec8453f643e24bafedccb5/CMSIS/Core/Include/core_cm0plus.h

CoreDebug on M3
https://github.com/ARM-software/CMSIS_5/blob/b7b26f50d00072812aec8453f643e24bafedccb5/CMSIS/Core/Include/core_cm3.h#L1384

ned-pcs:
> Which mcu you have tested this on. I could be wrong, but I remember both m0 and m0+ doesn't support bit.

Actually, I think you're right. I was misled by the Arm architecture docs.

PR #519, opened 2020-09-24
added DFU skeleton

amacard:
**Describe the PR**
Added a DFU skeleton to help the implementation of DFU feature.

**Additional context**
Now you just need to develop the DFU implementation on your software by creating the defined WEAK function on `dfu_rt_device.h`

Issue #29, opened 2018-12-26
DFU class


hathach:
should be used with dfu-util

gimbas:
What is the ETA for this? is anyone planning on picking this up?

I wanted to make a dfu bootloader, should i just implement it myself in said bootloader?

chegewara:
https://github.com/hathach/tinyusb/tree/master/examples/device/dfu_rt

gimbas:
I get that it is partly implemented already, but the most important parts are missing, or am i equivocated?

hathach:
Please don't ask about ETA, there is none on any features/issues. People (or at least me) implement one whenever they think it is appropriate/needed and having enough time to do so.

gimbas:
I didn't want anyone to tell me when it would be done, I souldn't have said ETA, I just wanted to know if anyone was working on it, or planning on picking it up.

Thank you anyway.

hathach:
implemented since #950

PR #754, opened 2021-03-26
Add DFU Class per Version 1.1 Spec

xmos-jmccarthy:
Please review.  I have some final testing to complete, and then some additional cleanup I would like to do.  I attempted to best follow the existing structure and styling.  Please let me know if anything is misplaced, desired debug missing, or any other general hygiene issue.

**Describe the PR**
This PR completes the DFU Class, resolving feature Issue #29.  It extends the prior DFU Class stub, which implemented only a DFU RT detach flow.

**Additional context**
The class has been validated with dfu-util to complete both download and uploads.  There are numerous callbacks that the application must implement, due to the application specific nature of DFU, so I am unsure on what to change the default demo to.  After my testing and cleanup per your requests are complete I will determine how to change the existing demo.

Merging of this PR would invalidate PR #519.

**Future Enhancements**
One area of improvement is in how the upload and download data step is handled.  Currently, the DFU driver passes the pointer to it's internal state buffer of CFG_TUD_DFU_TRANSFER_BUFFER_SIZE bytes.  This is functional, but if I have time I would like to add an alternative zero copy mode so the application can provide this pointer.  That way there is 1 less memcpy, as usbd_control_xfer_cb would be doing it's memcpy directly into the application buffer.  The current setup is safer, as it guarantees the buffer is large enough (assuming the usb descriptor has a transfer size <= CFG_TUD_DFU_TRANSFER_BUFFER_SIZE).


hathach:
great work as usual, please give me a bit of time, I will try to review it asap ( in the middle of something else :sweat_smile:  )

hathach (commented):
Thank you very much for your PR. Sorry for late review, I haven't fully go through it yet. Though there is one thing I spotted. Since DFU mode & runtime is mutual exclusive. I would like to have it as separated  driver. So that application that enable  the rutnime does not need to also include the DFU mode which require more RAM & ROM. Could you separate the dfu functionalities into `dfu _device.h/c`, we will leave the dfu_rt code as it is (common enum/typedef move to dfu.h is spot-on)

xmos-jmccarthy:
> Could you separate the dfu functionalities into `dfu _device.h/c`, we will leave the dfu_rt code as it is (common enum/typedef move to dfu.h is spot-on)

I will separate out class functionality of DFU mode from RT.

hathach:
> > Could you separate the dfu functionalities into `dfu _device.h/c`, we will leave the dfu_rt code as it is (common enum/typedef move to dfu.h is spot-on)
> 
> I will separate out class functionality of DFU mode from RT.

Thanks, I will do more review this week, need to pull out the DFU specs, kind of forgot it entirely ~.~ . This will be very helpful to bootloader project like https://github.com/adafruit/tinyuf2 . In fact, I will probably use tinyuf2 to test out this PR, since the flashing API is already working there, so hooking it up with DFU is relatively easy.

src/class/dfu/dfu_rt_device.c
```diff
@@ -26,42 +26,53 @@
...
+  _dfu_state_ctx.status = DFU_STATUS_OK;
+  _dfu_state_ctx.attrs = tud_dfu_runtime_init_attrs_cb();
+  dfu_debug_print_context();
 }
 
 void dfu_rtd_reset(uint8_t rhport)
 {
-  (void) rhport;
+  if (((_dfu_state_ctx.attrs & DFU_FUNC_ATTR_WILL_DETACH_BITMASK) == 0)
+     && (_dfu_state_ctx.state == DFU_REQUEST_DETACH))
```
hathach, line 67: should be `state == APP_DETACH`

src/class/dfu/dfu_rt_device.c
```diff
@@ -106,34 +117,67 @@ bool dfu_rtd_control_xfer_cb(uint8_t rhport, uint8_t stage, tusb_control_request
...
-  (void) ep_addr;
-  (void) result;
-  (void) xferred_bytes;
-  return true;
+  _dfu_state_ctx.status = status;
+}
+
+void tud_dfu_runtime_detach_timer_elapsed()
+{
+  if (_dfu_state_ctx.state == DFU_REQUEST_DETACH)
```
hathach, line 177: should be `state == APP_DETACH`

src/class/dfu/dfu_rt_device.c
```diff
@@ -106,34 +117,67 @@ bool dfu_rtd_control_xfer_cb(uint8_t rhport, uint8_t stage, tusb_control_request
...
+  switch (request->bRequest)
   {
     case DFU_REQUEST_DETACH:
-      tud_control_status(rhport, request);
-      tud_dfu_runtime_reboot_to_dfu_cb();
+    {
+      if (_dfu_state_ctx.state == APP_IDLE)
+      {
+        _dfu_state_ctx.state = APP_DETACH;
+        if ((_dfu_state_ctx.attrs & DFU_FUNC_ATTR_WILL_DETACH_BITMASK) == 1)
```
hathach, line 128: `== 1` must be dropped, it is bitwise, and will return the WILL_DETACH which is (1<<3) = 8

src/class/dfu/dfu_rt_device.h
```diff
@@ -29,36 +29,52 @@
...
+// Invoke when the dfu runtime detach timer has elapsed
+void tud_dfu_runtime_detach_timer_elapsed();
 
-//--------------------------------------------------------------------+
-// Application Callback API (weak is optional)
-//--------------------------------------------------------------------+
+// Invoked when a nonstandard request is received
+// Use may be vendor specific.
+// Return false to stall
+TU_ATTR_WEAK bool tud_dfu_runtime_req_nonstandard_cb(uint8_t rhport, uint8_t stage, tusb_control_request_t const * request);
```
hathach, line 64: please remove, vendor need to write their own driver. The specs doesn't mention any of the vendor request.

src/class/dfu/dfu_rt_device.c
```diff
@@ -106,34 +117,67 @@ bool dfu_rtd_control_xfer_cb(uint8_t rhport, uint8_t stage, tusb_control_request
...
+    case DFU_REQUEST_GETSTATE:
+    {
+      tud_control_xfer(rhport, request, &_dfu_state_ctx.state, 1);
+    }
+    break;
+
+    default:
+    {
+      TU_LOG2("  DFU Nonstandard Runtime Request: %u\r\n", request->bRequest);
+      return ( tud_dfu_runtime_req_nonstandard_cb ) ? tud_dfu_runtime_req_nonstandard_cb(rhport, stage, request) : false;
```
hathach, line 162: please removed, we don't support vendor request here. Vendor need to implement their own if needed. Here is how raspberrypi add their own. https://github.com/raspberrypi/pico-sdk/pull/197

src/class/dfu/dfu_rt_device.h
```diff
@@ -29,36 +29,52 @@
...
+void tud_dfu_runtime_set_status(dfu_mode_device_status_t status);
 
-// DFU Protocol
-typedef enum
-{
-  DFU_PROTOCOL_RT  = 1,
-  DFU_PROTOCOL_DFU = 2,
-} dfu_protocol_type_t;
+// Invoked when a DFU_DETACH request is received and bitWillDetach is set
+void tud_dfu_runtime_reboot_to_dfu_cb();
```
hathach, line 47: no parameter must has `void` to prevent warnings from gcc.

```C
void tud_dfu_runtime_reboot_to_dfu_cb(void)
```

src/class/dfu/dfu_rt_device.h
```diff
@@ -29,36 +29,52 @@
...
 
-
 //--------------------------------------------------------------------+
-// Common Definitions
+// Application Callback API (weak is optional)
 //--------------------------------------------------------------------+
+// Allow the application to update the status as required.
+// Is set to DFU_STATUS_OK during internal initialization and USB reset
+// Value is not checked to allow for custom statuses to be used
+void tud_dfu_runtime_set_status(dfu_mode_device_status_t status);
```
hathach, line 44: what is your scenario to have the need to set runtime status. I am opted to remove if it is just-in-case scenario.

src/class/dfu/dfu_rt_device.h
```diff
@@ -29,36 +29,52 @@
...
+//   DFU_FUNC_ATTR_MANIFESTATION_TOLERANT_BITMASK
+//   DFU_FUNC_ATTR_WILL_DETACH_BITMASK
+// Note: This should match the USB descriptor
+uint8_t tud_dfu_runtime_init_attrs_cb();
 
-// Invoked when received new data
-TU_ATTR_WEAK void tud_dfu_runtime_reboot_to_dfu_cb(void);
+// Invoked during a DFU_GETSTATUS request to get for the string index
+// to the status description string table.
+TU_ATTR_WEAK uint8_t tud_dfu_runtime_get_status_desc_table_index_cb();
```
hathach, line 77: again, I think runtime only support 2 request: 
1.GET_STATUS which return APP_IDLE and
2.DETACH which reboot immediately.

src/device/usbd.c
```diff
@@ -182,7 +182,31 @@ static usbd_class_driver_t const _usbd_driver[] =
...
+    .init             = dfu_mode_init,
+    .reset            = dfu_mode_reset,
+    .open             = dfu_mode_open,
+    .control_xfer_cb  = dfu_mode_control_xfer_cb,
+    .xfer_cb          = NULL,
+    .sof              = NULL
+  },
+  #endif
+
+  #if CFG_TUD_DFU_RUNTIME_AND_MODE
```
hathach, line 202: hmm, what is the scenario you have in mind for this `RUNTIME_AND_MODE`. Can you elaborate it more, I thought it is either dfu runtime or dfu mode (bootloader)

hathach (commented):
Thank you very much for your PR. I haven't go through all the changes just yet. Though I notice several changes that we could address first with DFU Runtime, still haven't looked at actual DFU mode yet. 
1.  There is couple typo when comparing runtime state with `DFU_REQUEST_DETACH`, should be  `APP_DETACH`
2. Bitwise result must not be equal-check with 1, it is good as just bitwise alone
3. the updated dfu runtime is more complicated than it should be. I think the existing/previous code is good enough.

PS: Adding a new driver is lots of works, we may need to go through a few rounds of reviews before merging. Please be patient.

src/class/dfu/dfu_rt_device.c
```diff
@@ -106,34 +117,67 @@ bool dfu_rtd_control_xfer_cb(uint8_t rhport, uint8_t stage, tusb_control_request
...
-      tud_dfu_runtime_reboot_to_dfu_cb();
+    {
+      if (_dfu_state_ctx.state == APP_IDLE)
+      {
+        _dfu_state_ctx.state = APP_DETACH;
+        if ((_dfu_state_ctx.attrs & DFU_FUNC_ATTR_WILL_DETACH_BITMASK) == 1)
+        {
+          tud_dfu_runtime_reboot_to_dfu_cb();
+        } else {
+          tud_dfu_runtime_detach_start_timer_cb(request->wValue);
```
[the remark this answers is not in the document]
xmos-jmccarthy: Sounds good.  I will remove the RT mode will_detach capability, which should allow this to go back to the original state.

src/class/dfu/dfu_rt_device.h
xmos-jmccarthy: This was purely for a just-in-case.  I will remove.

src/device/usbd.c
xmos-jmccarthy: You are correct in that dfu runtime and dfu mode are exclusive.  After I split the RT and Mode, I ran into an issue where because the tusb_config.h had:
```
#define CFG_TUD_DFU_RUNTIME 1
#define CFG_TUD_DFU_MODE    1
```
Which resulted in both DFU classes being initialized when `tud_task()` was called.

One possible solution to this would be to have separate builds, one for the app with RT and another for the bootloader with Mode.  In the event than an application does not want to have 2 separate builds, I added this this `RUNTIME_AND_MODE` "class" which allows for the code for both to exist in a single build, but only 1 class is initialized at runtime.  When this DFU class is initialized, it will ask the app if it should run as RT or Mode.  After this check, the class will behave only as RT or Mode.

The name can be changed to prevent confusion.  CFG_TUD_DFU_DYNAMIC or something.

src/class/dfu/dfu_rt_device.c
```diff
@@ -26,42 +26,53 @@
...
+static void dfu_rt_getstatus_reply(uint8_t rhport, tusb_control_request_t const * request);
 
 //--------------------------------------------------------------------+
 // USBD Driver API
 //--------------------------------------------------------------------+
 void dfu_rtd_init(void)
 {
+  _dfu_state_ctx.state = APP_IDLE;
+  _dfu_state_ctx.status = DFU_STATUS_OK;
+  _dfu_state_ctx.attrs = tud_dfu_runtime_init_attrs_cb();
```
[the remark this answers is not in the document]
xmos-jmccarthy: Perfect.  This particular line is gone now, for the runtime simplification; however, I have made this change in dfu mode.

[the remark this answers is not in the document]
xmos-jmccarthy: Resolved in 2e2dc7b

src/class/dfu/dfu_rt_device.c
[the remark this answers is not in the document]
xmos-jmccarthy: Resolved in 2e2dc7b

src/class/dfu/dfu_rt_device.c
```diff
@@ -106,34 +117,67 @@ bool dfu_rtd_control_xfer_cb(uint8_t rhport, uint8_t stage, tusb_control_request
...
-      // status = OK, poll timeout = 0, state = app idle, istring = 0
-      uint8_t status_response[6] = { 0, 0, 0, 0, 0, 0 };
-      tud_control_xfer(rhport, request, status_response, sizeof(status_response));
+      dfu_status_req_payload_t resp;
+      resp.bStatus = _dfu_state_ctx.status;
+      memset((uint8_t *)&resp.bwPollTimeout, 0x00, 3);  // Value is ignored
+      resp.bState = _dfu_state_ctx.state;
+      resp.iString = ( tud_dfu_runtime_get_status_desc_table_index_cb ) ? tud_dfu_runtime_get_status_desc_table_index_cb() : 0;
+
+      tud_control_xfer(rhport, request, &resp, sizeof(dfu_status_req_payload_t));
```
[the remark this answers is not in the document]
xmos-jmccarthy: Resolved in 2e2dc7b

src/class/dfu/dfu_rt_device.c
xmos-jmccarthy: Resolved in 2e2dc7b
No longer applicable

src/class/dfu/dfu_rt_device.c
xmos-jmccarthy: Resolved in 2e2dc7b
No longer applicable.

src/class/dfu/dfu_rt_device.c
xmos-jmccarthy: Resolved in 2e2dc7b
No longer applicable.

src/class/dfu/dfu_rt_device.h
xmos-jmccarthy: Resolved in 2e2dc7b

src/class/dfu/dfu_rt_device.c
xmos-jmccarthy: Resolved in 2e2dc7b

src/class/dfu/dfu_rt_device.h
xmos-jmccarthy: Resolved in 2e2dc7b

src/class/dfu/dfu_rt_device.h
xmos-jmccarthy: Resolved in ae851bb

src/device/usbd.c
hathach: I don't think we would need to ask the app, I wonder why would bptj CFG_TUD_DFU_RUNTIME and CFG_TUD_DFU_MODE are enabled at the same time. And what would be the issue with both dfu mode and dfu rt are initialized. It really doesn't matter, the usb configuration descriptor will dictate which interface e.g itf = 0 for the DFU mode and which is used for DFU runtime should they are both enabled.

Please remove the the `RUNTIME_AND_MODE` we will treat each one as separated driver like msc vs cdc.

hathach (commented):
Please remove the the `RUNTIME_AND_MODE` we will treat DFU_RUNTIME and DFU as separated driver like msc vs cdc. There is no conflict when both enabled.

src/device/usbd.c
xmos-jmccarthy: Resolved in 7b45b38

xmos-jmccarthy:
I am expecting that you would like the nonstandard requests removed from the DFU mode, but I will wait until your review before I make this change.

To adhere to the spec, the finite state machine is verbose, with a fair bit of duplication.  I opted for readability rather than compactness, which is why some things, such as get_state and get_status requests are in most states, rather than being checked at the request level.

hathach:
Thank you very much for you update, I will review it asap :)

src/class/dfu/dfu_mode_device.h
```diff
@@ -0,0 +1,131 @@
...
+
+
+//--------------------------------------------------------------------+
+// Application Callback API (weak is optional)
+//--------------------------------------------------------------------+
+// Invoked when a reset is received to check if firmware is valid
+bool tud_dfu_mode_firmware_valid_check_cb(void);
+
+// Invoked when the device must reboot to dfu runtime mode
+void tud_dfu_mode_reboot_to_rt_cb(void);
```
hathach, line 46: I think we should remove this callback, nowadays all MCU can perform soft reset to application mode

src/class/dfu/dfu_mode_device.h
```diff
@@ -0,0 +1,131 @@
...
+#ifdef __cplusplus
+  extern "C" {
+#endif
+
+
+//--------------------------------------------------------------------+
+// Application Callback API (weak is optional)
+//--------------------------------------------------------------------+
+// Invoked when a reset is received to check if firmware is valid
+bool tud_dfu_mode_firmware_valid_check_cb(void);
```
hathach, line 43: should be removed as well.

src/class/dfu/dfu_mode_device.h
```diff
@@ -0,0 +1,131 @@
...
+void tud_dfu_mode_start_poll_timeout_cb(uint8_t *ms_timeout);
+
+// Must be called when the poll_timeout has elapsed
+void tud_dfu_mode_poll_timeout_done(void);
+
+// Invoked when a DFU_DNLOAD request is received
+// This callback takes the wBlockNum chunk of length length and provides it
+// to the application at the data pointer.  This data is only valid for this
+// call, so the app must use it not or copy it.
+void tud_dfu_mode_req_dnload_data_cb(uint16_t wBlockNum, uint8_t* data, uint16_t length);
```
hathach, line 93: IMHO, this callback should return the poll_timeout, instead of using `tud_dfu_mode_get_poll_timeout_cb()`. It will make thing easier for application, since app have the data length to write, it should be able to determine how long it need to program that chunk. Also we have less API to worry about.

src/class/dfu/dfu_mode_device.h
```diff
@@ -0,0 +1,131 @@
...
+TU_ATTR_WEAK void tud_dfu_mode_usb_reset_cb(uint8_t rhport, dfu_mode_state_t *state);
+
+// Invoked during a DFU_GETSTATUS request to set the timeout in ms to use
+// before the subsequent DFU_GETSTATUS requests.
+// The purpose of this value is to allow the device to tell the host
+// how long to wait between the DFU_DNLOAD and DFU_GETSTATUS checks
+// to allow the device to have time to erase, write memory, etc.
+// ms_timeout is a pointer to array of length 3.
+// Refer to the USB DFU class specification for more details
+TU_ATTR_WEAK void tud_dfu_mode_get_poll_timeout_cb(uint8_t *ms_timeout);
```
hathach, line 76: should be removed, see comment in `tud_dfu_mode_req_dnload_data_cb()`

src/class/dfu/dfu_mode_device.h
```diff
@@ -0,0 +1,131 @@
...
+
+// Invoked when a DFU_DNLOAD request is received
+// This should start a timer for ms_timeout ms
+// When the timer has elapsed, tud_dfu_runtime_poll_timeout_done must be called
+// NOTE: This callback should return immediately, and not implement the delay
+// internally, as this will hold up the class stack from receiving any packets
+// during the DFU_DNLOAD_SYNC and DFU_DNBUSY states
+void tud_dfu_mode_start_poll_timeout_cb(uint8_t *ms_timeout);
```
hathach, line 84: should be removed, it requires application to have timer to operate DFU. see comment at `tud_dfu_mode_poll_timeout_done()`

src/class/dfu/dfu_mode_device.h
```diff
@@ -0,0 +1,131 @@
...
+
+// Must be called when the poll_timeout has elapsed
+void tud_dfu_mode_poll_timeout_done(void);
```
hathach, line 87: instead of using timer based, this API should be renamed to `tud_dfu_download_complete()` which is called when the flashing is complete by application (started by tud_dfu_mode_req_dnload_data_cb() ). From what I understand the poll timeout is just a suggestion for host to time its next DOWNLOAD request. There is no need to keep the timer on. If host decide to request before we actually complete, we can simply STALL the request to reject it.

hathach (commented):
Thank you for your PR again. There are only several comments, but all is minor (except the timeout change), not really require much changing. Besides that I have 2 major requests
1. Could you simplify the `dfu_mode` to just `dfu` (both in API and file name). I think it will be simpler and easier to read. If you think it is too much of a hassle, then I could help with the rename
2. As part of adding new class, it typically requires to have an example to (a) to test the driver and (b) demonstrate to user on how to use the API. Example is also my excuse for not writing fancy doc, since I simply just point user to example :smile: . In this case, we don't have to do any actual flashing. We only need to dump/print out the download data, and response to upload as well as other requests from the dfu-util.

PS: I am sorry if the review is a bit overwhelming, I try my best to do so. For some of the changes, If anything is unclear, or you have any other idea feel free to suggest. I am open to all  dicussion.

xmos-jmccarthy:
>     1. Could you simplify the `dfu_mode` to just `dfu` (both in API and file name). I think it will be simpler and easier to read. If you think it is too much of a hassle, then I could help with the rename
I can do this.  It's funny that you asked for that.  When I separated the RT and DFU mode portions I had everything named DFU, and then added MODE so it was consistent with RT.  I will change things from dfu_mode to dfu as I go through the rest of your comments.

>     2. As part of adding new class, it typically requires to have an example to (a) to test the driver and (b) demonstrate to user on how to use the API. Example is also my excuse for not writing fancy doc, since I simply just point user to example smile . In this case, we don't have to do any actual flashing. We only need to dump/print out the download data, and response to upload as well as other requests from the dfu-util.
Yes, adding an example was on the TODO list.  My testing with dfu-util was with dummy handlers that just printed the contents being downloaded.  The upload test was to a file that I could just open which would have the arbitrary string "Hello world"  I will add this demo once I have addressed the other comments.

hathach:
> I can do this. It's funny that you asked for that. When I separated the RT and DFU mode portions I had everything named DFU, and then added MODE so it was consistent with RT. I will change things from dfu_mode to dfu as I go through the rest of your comments.

Actually, I meant to split the runtime from the actual DFU mode, have them as separated drivers. Though words can be misleading, English is not my 1st language after all. That is ok if you don't want to make the rename, I could do it myself. Not a deal breaker at all.  

> Yes, adding an example was on the TODO list. My testing with dfu-util was with dummy handlers that just printed the contents being downloaded. The upload test was to a file that I could just open which would have the arbitrary string "Hello world" I will add this demo once I have addressed the other comments.

Thanks, that would be helpful

src/class/dfu/dfu_mode_device.c
```diff
@@ -0,0 +1,594 @@
...
+        case DFU_IDLE:
+        case DFU_DNLOAD_SYNC:
+        case DFU_DNBUSY:
+        case DFU_DNLOAD_IDLE:
+        case DFU_MANIFEST_SYNC:
+        case DFU_MANIFEST:
+        case DFU_MANIFEST_WAIT_RESET:
+        case DFU_UPLOAD_IDLE:
+        {
+          _dfu_state_ctx.state = (tud_dfu_mode_firmware_valid_check_cb()) ?  APP_IDLE : DFU_ERROR;
```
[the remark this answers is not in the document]
xmos-jmccarthy: The valid firmware check is a bit of a catch all here.

Reset in the dfuMANIFEST-WAIT-RESET state should result in the firmware validity being checked, and the device rebooting if valid, or change to DFU_ERROR.  Reset in any other state should result in rebooting to RT, which the app can then decide if that is a hard/soft reboot or reenumeration.

src/class/dfu/dfu_mode_device.c
```diff
@@ -0,0 +1,594 @@
...
+static void dfu_mode_req_dnload_reply(uint8_t rhport, tusb_control_request_t const * request);
+static bool dfu_mode_state_machine(uint8_t rhport, tusb_control_request_t const * request);
+
+
+//--------------------------------------------------------------------+
+// USBD Driver API
+//--------------------------------------------------------------------+
+void dfu_moded_init(void)
+{
+  _dfu_state_ctx.state = APP_DETACH;    // After init, reset will occur.  We want to be in APP_DETACH to move to DFU_IDLE
```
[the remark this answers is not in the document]
xmos-jmccarthy: When the start starts the driver is init and the host sends a reset.  Starting in APP_DETACH is so the initial RESET doesn't tell the app to reboot/reenumerate back in RT.  Depending on what changes in reset, I will update this appropriately.

src/class/dfu/dfu_mode_device.h
xmos-jmccarthy: This callback is named reboot, but could be renamed to prevent confusion.
When this is called the application should reboot hard/soft or reenumerate back to runtime mode.  This is a callback since it can occur when dfu-util decides to reset the device or when the firmware is accepted locally by the device.

src/class/dfu/dfu.h
```diff
@@ -0,0 +1,198 @@
...
+  uint8_t iString;
+} dfu_status_req_payload_t;
+
+TU_VERIFY_STATIC( sizeof(dfu_status_req_payload_t) == 6, "size is not correct");
+
+
+//--------------------------------------------------------------------+
+// Debug
+//--------------------------------------------------------------------+
+#if CFG_TUSB_DEBUG >= 2
```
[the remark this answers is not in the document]
xmos-jmccarthy: Resolved in 88dea7a

src/class/dfu/dfu_mode_device.h
```diff
@@ -0,0 +1,131 @@
...
+
+// Invoked when a DFU_UPLOAD request is received
+// This callback must populate data with up to length bytes
+// Return the number of bytes to write
+uint16_t tud_dfu_mode_req_upload_data_cb(uint16_t block_num, uint8_t* data, uint16_t length);
+
+// Invoked when a nonstandard request is received
+// Use may be vendor specific.
+// Return false to stall
+TU_ATTR_WEAK bool tud_dfu_mode_req_nonstandard_cb(uint8_t rhport, uint8_t stage, tusb_control_request_t const * request);
```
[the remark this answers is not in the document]
xmos-jmccarthy: Resolved in 0936a76

src/class/dfu/dfu_mode_device.h
```diff
@@ -0,0 +1,131 @@
...
+// to the status description string table.
+TU_ATTR_WEAK uint8_t tud_dfu_mode_get_status_desc_table_index_cb(void);
+
+// Invoked during a USB reset
+// Lets the app perform custom behavior on a USB reset.
+// If not defined, the default behavior remains the DFU specification of
+// Checking the firmware valid and changing to the error state or rebooting to runtime
+// Note: If used, the application must perform the reset logic for all states.
+// Changing the state to APP_IDLE will result in tud_dfu_mode_reboot_to_rt_cb being called
+TU_ATTR_WEAK void tud_dfu_mode_usb_reset_cb(uint8_t rhport, dfu_mode_state_t *state);
```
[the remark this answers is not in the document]
xmos-jmccarthy: Resolved in 18e9d25

src/class/dfu/dfu_mode_device.h
```diff
@@ -0,0 +1,131 @@
...
+void tud_dfu_mode_reboot_to_rt_cb(void);
+
+// Invoked during initialization of the dfu driver to set attributes
+// Return byte set with bitmasks:
+//   DFU_FUNC_ATTR_CAN_DOWNLOAD_BITMASK
+//   DFU_FUNC_ATTR_CAN_UPLOAD_BITMASK
+//   DFU_FUNC_ATTR_MANIFESTATION_TOLERANT_BITMASK
+//   DFU_FUNC_ATTR_WILL_DETACH_BITMASK
+// Note: This should match the USB descriptor
+uint8_t tud_dfu_mode_init_attrs_cb(void);
```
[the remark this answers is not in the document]
xmos-jmccarthy: Yes this is now populated implicitly off of the descriptor
Resolved in 289af58

src/class/dfu/dfu_mode_device.c
```diff
@@ -0,0 +1,594 @@
...
+  // Ensure this is DFU Mode
+  TU_VERIFY((itf_desc->bInterfaceSubClass == TUD_DFU_APP_SUBCLASS) &&
+            (itf_desc->bInterfaceProtocol == DFU_PROTOCOL_DFU), 0);
+
+  uint8_t const * p_desc = tu_desc_next( itf_desc );
+  uint16_t drv_len = sizeof(tusb_desc_interface_t);
+
+  if ( TUSB_DESC_FUNCTIONAL == tu_desc_type(p_desc) )
+  {
+    tusb_desc_dfu_functional_t *dfu_desc = (tusb_desc_dfu_functional_t *)p_desc;
```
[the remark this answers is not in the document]
xmos-jmccarthy: Resolved in e54d9d1

src/class/dfu/dfu_mode_device.c
```diff
@@ -0,0 +1,594 @@
...
+    case DFU_REQUEST_ABORT:
+    {
+      return dfu_mode_state_machine(rhport, request);
+    }
+    break;
+
+    default:
+    {
+      TU_LOG2("  DFU Nonstandard Request: %u\r\n", request->bRequest);
+      return ( tud_dfu_mode_req_nonstandard_cb ) ? tud_dfu_mode_req_nonstandard_cb(rhport, stage, request) : false;
```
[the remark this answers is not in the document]
xmos-jmccarthy: Resolved in 0936a76

src/device/usbd.h
```diff
@@ -542,6 +542,17 @@ TU_ATTR_WEAK bool tud_vendor_control_xfer_cb(uint8_t rhport, uint8_t stage, tusb
   /* Function */ \
   9, DFU_DESC_FUNCTIONAL, _attr, U16_TO_U8S_LE(_timeout), U16_TO_U8S_LE(_xfer_size), U16_TO_U8S_LE(0x0101)
 
+// Length of template descriptr: 18 bytes
+#define TUD_DFU_MODE_DESC_LEN (9 + 9)
+
+// DFU runtime descriptor
+// Interface number, string index, alternate setting, attributes, detach timeout, transfer size
+#define TUD_DFU_MODE_DESCRIPTOR(_itfnum, _stridx, _alt_setting, _attr, _timeout, _xfer_size) \
```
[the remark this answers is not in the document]
xmos-jmccarthy: Resolved in 45e401e

src/class/dfu/dfu_mode_device.h
```diff
@@ -0,0 +1,131 @@
...
+
+// Invoked during a DFU_GETSTATUS request to get for the string index
+// to the status description string table.
+TU_ATTR_WEAK uint8_t tud_dfu_mode_get_status_desc_table_index_cb(void);
```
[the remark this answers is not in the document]
xmos-jmccarthy: The application would know the string table that it would have to set with the USB descriptors.  Any use of this is nonstandard though, so removed.
Resolved in cc440ad

src/class/dfu/dfu_mode_device.c
```diff
@@ -0,0 +1,594 @@
...
+
+  resp.bStatus = _dfu_state_ctx.status;
+  if ( tud_dfu_mode_get_poll_timeout_cb )
+  {
+    tud_dfu_mode_get_poll_timeout_cb((uint8_t *)&resp.bwPollTimeout);
+  } else {
+    memset((uint8_t *)&resp.bwPollTimeout, 0x00, 3);
+  }
+  resp.bState = _dfu_state_ctx.state;
+  resp.iString = ( tud_dfu_mode_get_status_desc_table_index_cb ) ? tud_dfu_mode_get_status_desc_table_index_cb() : 0;
```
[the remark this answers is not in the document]
xmos-jmccarthy: The application would know the string table that it would have to set with the USB descriptors.  Any use of this is nonstandard though, so removed.
Resolved in cc440ad

src/class/dfu/dfu_mode_device.c
```diff
@@ -0,0 +1,594 @@
...
+
+  return drv_len;
+}
+
+// Invoked when a control transfer occurred on an interface of this class
+// Driver response accordingly to the request and the transfer stage (setup/data/ack)
+// return false to stall control endpoint (e.g unsupported request)
+bool dfu_moded_control_xfer_cb(uint8_t rhport, uint8_t stage, tusb_control_request_t const * request)
+{
+  if ( (stage == CONTROL_STAGE_DATA) && (request->bRequest == DFU_DNLOAD_SYNC) )
```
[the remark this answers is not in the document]
xmos-jmccarthy: There is a bug here.  I have refactored it to be the correct behavour.  request->bRequest should not have been there.
For additional review in 8c80dde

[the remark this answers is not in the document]
xmos-jmccarthy: Correction, b8e5885

src/class/dfu/dfu_mode_device.c
```diff
@@ -0,0 +1,594 @@
...
+
+  if ( tud_dfu_mode_get_poll_timeout_cb )
+  {
+    tud_dfu_mode_get_poll_timeout_cb((uint8_t *)&bwPollTimeout);
+  }
+
+  tud_dfu_mode_start_poll_timeout_cb((uint8_t *)&bwPollTimeout);
+
+  // TODO: I want the real xferred len, not what is expected.  May need to change usbd.c to do this.
+  tud_dfu_mode_req_dnload_data_cb(_dfu_state_ctx.last_block_num, (uint8_t *)&_dfu_state_ctx.transfer_buf, _dfu_state_ctx.last_transfer_len);
```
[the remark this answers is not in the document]
xmos-jmccarthy: Please review b8e5885

src/class/dfu/dfu_mode_device.h
xmos-jmccarthy: Removed.
Please review b8e5885

src/class/dfu/dfu_mode_device.h
xmos-jmccarthy: Removed.
Please review b8e5885

src/class/dfu/dfu_mode_device.h
xmos-jmccarthy: Removed.
Please review b8e5885

src/class/dfu/dfu_mode_device.h
xmos-jmccarthy: Refactored.
Please review b8e5885

hathach:
Thank you @xmos-jmccarthy for addressing previous feedback, Let me know if you want me to review the updated code, or wait until you push the dfu example.

xmos-jmccarthy:
Please review the changes in the latest commit.  There are a few unresolved comments that I would like more feedback/consideration on before I make them.

Once the class driver is complete I  will add in an in repo example.

src/class/dfu/dfu_mode_device.c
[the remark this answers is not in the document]
hathach: from my point of view, USB DFU will be only used as transport layer as bootloader. There are  probably more transport such as cdc/uart/msc/hid as well. What DFU should do is invoked callback when receiving the final packet `tud_dfu_device_data_done_check_cb`. That would be then up to application/bootloader to decide whether it want to perform checksum and reset to app or failed the DFU.  For BUS_RESET, I think we just treat it as data reset for now.

src/class/dfu/dfu_mode_device.h
hathach: I think we can safely remove the comment now `This data is only valid for this call, so the app must use it not or copy it.` We have an async API now, dfu invoke callback when receive data. And app call complete API to transition out of BUSY state.

src/class/dfu/dfu_device.c
```diff
@@ -0,0 +1,648 @@
...
+      {
+        _dfu_state_ctx.state = APP_IDLE;
+      }
+      break;
+    }
+  }
+
+  if(_dfu_state_ctx.state == APP_IDLE)
+  {
+    tud_dfu_reboot_to_rt_cb();
```
hathach, line 183: We should remove this API as well, just let the bootloader to decide this by itself. Overall, I think BUS_RESET should just reset the dfu state/variables.

src/class/dfu/dfu_device.c
```diff
@@ -0,0 +1,648 @@
...
+    break;
+  }
+
+  return true;
+}
+
+static uint16_t dfu_req_upload(uint8_t rhport, tusb_control_request_t const * request, uint16_t block_num, uint16_t wLength)
+{
+  TU_VERIFY( wLength <= CFG_TUD_DFU_TRANSFER_BUFFER_SIZE);
+  uint16_t retval = tud_dfu_req_upload_data_cb(block_num, (uint8_t *)&_dfu_state_ctx.transfer_buf, wLength);
```
hathach, line 279: it is a typo, should be `(uint8_t *) _dfu_state_ctx.transfer_buf` without reference operator. This applies for all the usage of transfer_buf

src/class/dfu/dfu_device.c
```diff
@@ -0,0 +1,648 @@
...
+    case DFU_REQUEST_DETACH:
+    case DFU_REQUEST_UPLOAD:
+    case DFU_REQUEST_GETSTATUS:
+    case DFU_REQUEST_CLRSTATUS:
+    case DFU_REQUEST_GETSTATE:
+    case DFU_REQUEST_ABORT:
+    {
+      if(stage == CONTROL_STAGE_SETUP)
+      {
+        return dfu_state_machine(rhport, request);
```
hathach, line 260: putting all the requests combo with all the states into this make the `dfu_state_machine()` much more complicated than it should be. We should simplify the dfu_state_machine() by breaking down and handle request that applies to all state such as GETSTATUS and GETSTATE

src/class/dfu/dfu_device.c
```diff
@@ -0,0 +1,648 @@
...
+  {
+    _dfu_state_ctx.state = DFU_DNLOAD_SYNC;
+  } else if (_dfu_state_ctx.state == DFU_MANIFEST)
+  {
+    _dfu_state_ctx.state = ((_dfu_state_ctx.attrs & DFU_FUNC_ATTR_MANIFESTATION_TOLERANT_BITMASK) != 0)
+                           ? DFU_MANIFEST_WAIT_RESET : DFU_MANIFEST_SYNC;
+  }
+}
+
+static bool dfu_state_machine(uint8_t rhport, tusb_control_request_t const * request)
```
hathach, line 330: IMHO, dfu_state_machine() is written more complicated than it should be. To be honest, I don't know what would suggest here. Maybe when adding example we could understand this better.

xmos-jmccarthy: Yes, so this concern one of my prior comments.  I am very open to changing it, but here is some background for how it came to be.

I opted for the current setup where there is a very verbose state machine, but it is explicit in the behaviour to expect on any given request.  An earlier iteration had all behaviour inside of dfu_moded_control_xfer_cb.  This was incredibly difficult to follow, since each request ended up containing a switch for every single possible state, where most resulted in a stall.

I then migrated to a setup where some common commands such as DFU_REQUEST_GETSTATUS and DFU_REQUEST_GETSTATE where handled in dfu_moded_control_xfer_cb in the bRequest switch, and the rest where in the dfu_state_machine function.  This was better, but there still was some ugliness due to DFU_MANIFEST, DFU_MANIFEST_WAIT_RESET, and DFU_DNBUSY supposed to stall when these requests occur or DFU_MANIFEST_SYNC having different behaviour on a DFU_REQUEST_GETSTATUS if DFU_FUNC_ATTR_MANIFESTATION_TOLERANT_BITMASK is set.

src/class/dfu/dfu_device.c
xmos-jmccarthy: See comment on dfu_state_machine

src/class/dfu/dfu_device.c
xmos-jmccarthy: Resolved in f830800

src/class/dfu/dfu_device.c
xmos-jmccarthy: Resolved in f830800

src/class/dfu/dfu_mode_device.h
xmos-jmccarthy: Resolved in f830800

src/class/dfu/dfu_mode_device.c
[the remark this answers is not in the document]
xmos-jmccarthy: Resolved in f830800

src/class/dfu/dfu_mode_device.c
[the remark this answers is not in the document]
xmos-jmccarthy: Resolved in f830800

src/class/dfu/dfu_device.c
```diff
@@ -0,0 +1,616 @@
...
+      switch (request->bRequest)
+      {
+        case DFU_REQUEST_GETSTATUS:
+        {
+          if ((_dfu_state_ctx.attrs & DFU_FUNC_ATTR_MANIFESTATION_TOLERANT_BITMASK) != 0)
+          {
+            _dfu_state_ctx.state = DFU_MANIFEST;
+            dfu_req_getstatus_reply(rhport, request);
+          } else {
+            if ( tud_dfu_firmware_valid_check_cb() )
```
hathach, line 478: I think we should just remove this `tud_dfu_firmware_valid_check_cb()`

Side note:  we don't really need to check for the MANIFESTATION_TOLERANT here. The reason is that if we could have CPU checking this code, we already able to communicate via USB and therefore imply the MANIFESTATION_TOLERANT = 1 already.

src/class/dfu/dfu_device.c
hathach: I see, it is your point of view for implementation. Though it is still a bit complicated for my taste. However, you don't have to change it now. We will see how it turn out when adding the example. And I will try to move code around if needed as follow up later on. I do have some renaming in my mind. It is too minor mentioned in reviews.

hathach (commented):
Thanks for quickly address the feedback. I think we should remove `tud_dfu_firmware_valid_check_cb()` API, then we could move on with adding example to test our the class driver.

src/class/dfu/dfu_device.c
hathach: Yeah, that is the approach. Bootloader application is the one in charge of performing checksum, jump app. From my experience working with a couple of bootloader project [tinyuf2](https://github.com/adafruit/tinyuf2) and [Adafruit_nRF52_Bootloader](https://github.com/adafruit/Adafruit_nRF52_Bootloader) . DFU class will only be one of supported transport together with msc (drag and drop firmware), HID, WebUSB, CDC, or even BLE/WIFI OTA. Typically each transport will let the bootloader know when all the data is transfer, bootloader witll finish all the writing, then do some checksum, reset peripheral, and reboot too application its-self. It won't even need any further help from host llike GETSTATUS (from most MCU I have worked with).

hathach (commented):
Thanks for working on the PR, please take your time to add the example (there is no rush). I am sure that will help to finalize the API for application stack interaction. I am looking forward to follow / help with your further commits.

xmos-jmccarthy:
I added an example app for review.  I added a separate example, rather than updating the RT example.  The reason for this was to avoid requiring that the app reboot/support dynamic configuration (my device does not).  dfu-util will give a warning that the device already is in dfu mode, but will still perform the upload or download requested.

I unfortunately do not have my platform ported to run within the TinyUSB repo so I was unable to test, as we discussed earlier.  I have tested upload and download of real firmware on my platform to verify the transport correctness.

As discussed earlier, DFU_MANIFEST isn't needed, so I am still in favor with removal to simplify the state machine.  I can do this is you'd like, or you can during your move around/standardization.

hathach:
Thanks for the update,  Adding it as separated example is spot on. DFU runtime is application, while DFU is actually bootloader. I am more than welcome anything that could help with the simplification (the stack is too complicated already). I will do an actual test and review as soon as I could

hathach:
sorry I have been super busy with lots of work for Arduino TinyUSB Lib for rp2040. I will try to check this out asap (soon)..

hathach (approved):
superb! thank you @xmos-jmccarthy  very much for the PR. I am really appreciated your effort and patient throughout the whole PR reviews. It has been a bit busy due to the release of rp2040 in recent time. And I couldn't keep up with all the fix and pr.  The example work great with both download and upload. Bravo !!! :tada: :tada:  . And it is good for the merge now. Later on when integrating into another bootloader, I will do more actual testing on hardware. Should there is any improvement, I will tag you for review/feedback.

PS: I have updated the PR  to fix build issue on CI.

PR #758, opened 2021-03-28
Add macro for unaligned access.

HiFiPhile:
**Describe the PR**
A proposition for unaligned access macro.

Then we can refactor unaligned accesses, for example:
```
uint16_t total_len;
// Use offsetof to avoid pointer to the odd/misaligned address
memcpy(&total_len, (uint8_t*) desc_bos + offsetof(tusb_desc_bos_t, wTotalLength), 2);
```
Into
```
uint16_t total_len;
TU_UNALIGNED_GET(total_len, &desc_bos->wTotalLength, uint16_t);
```

**Additional context**
It *should* works on TI compiler, need test.

### DIFF
--- a/src/common/tusb_compiler.h
+++ b/src/common/tusb_compiler.h
@@ -81,6 +81,27 @@
   #define TU_BSWAP16(u16) (__builtin_bswap16(u16))
   #define TU_BSWAP32(u32) (__builtin_bswap32(u32))
 
+  #define TU_UNALIGNED_GET(var, ptr, type)  \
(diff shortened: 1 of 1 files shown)


hathach:
thank you for your PR, it looks good. However, I think this shouldn't be part of the compiler. I am thinking to have an typdef for tu_unaligned_uint32_t then using static inline function for the assignment. I will make an PR to demonstrate the idea.

hathach:
closed in favor of #772

PR #759, opened 2021-03-29
Add releases to Mynewt repository.yml

sjanc:
This makes it possible to setup Mynewt project which checkouts
specified version of tinyusb.


hathach (approved):
thank you for your PR. Look good

PR #760, opened 2021-03-29
add suffix _s to object of assembly file

### LINKED ISSUE #755: Buildsystem does not allow a c and assembly source with the same name
Buildsystem does not allow a c and assembly source with the same name, where only one is supposed to be compiled. 

I found this issue while trying to implement an example for the efm32 family in #749. 

The Cmsis provided by silabs as both a c implementation and an assembly implementation for the startup, only one should be compiled, but the buildsystem right now tries to compile both, regardless of which is included. 

I have followed up the issue and think it is in `rules.MK` where both c and s files are compiled into a single object list, then when the objects are called for compilation, it resolves to both a c file and a s file. 

I will be working on a fix soon.

hathach:
**Describe the PR**
Fix #755 where there is .c and .s with the same name in the same folder.

### DIFF
--- a/examples/rules.mk
+++ b/examples/rules.mk
@@ -52,8 +52,8 @@ SRC_S := $(SRC_S:.S=.s)
 
 # Due to GCC LTO bug https://bugs.launchpad.net/gcc-arm-embedded/+bug/1747966
 # assembly file should be placed first in linking order
-# '_s' suffix is added to object of assembly file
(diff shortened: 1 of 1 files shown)


perigoso:
looks good, ci is ok, and I tested locally with the problematic board and all worked as expected.

hathach:
thanks for confirmation, merge now

PR #593, opened 2021-01-18
Edpt iso xfer

PanRe:
**Describe the PR**
Implementation of a FIFO instead of a linear buffer to be used as ISO EP buffer as discussed in #504.

**Additional context**
This is an initial draft and not yet finalized. It works and uses the audio driver for testing. I added `usbd_edpt_ISO_xfer()` to schedule a transmission by use of a FIFO and implemented it for STM32 by adding `dcd_edpt_ISO_xfer()` in dcd_synopsis.c.
I further needed to enhance the functionality of tu_usb_fifo as:
- add support for copying from one FIFO directly into another one
- add "copy modes" which are required to tell the FIFO to copy to/from a constant src/dst pointer. This is required to copy data corretly to/from the STM32 USB hardware buffer.

It does not break any standard behaviour and if approved it would allow for an improvement of ISO EP relating drivers. However, `dcd_edpt_ISO_xfer()` would need to be implemented for all other MCUs as well, which i am not profound of. It would be great if somebody else could do it.

PR #768, opened 2021-04-03
Fixed very slow implementation of _ff_mod

phdussud:
**Describe the PR**
Fixed very slow implementation. Symptoms are that the USB transfers take more and more time, then become fast again as the index (uint16_t)  wraps around

I appreciate the fact that some MPUs don't have a hardware divide, so I provided a fast path if the fifo depth is a power of 2.

### DIFF
--- a/src/common/tusb_fifo.c
+++ b/src/common/tusb_fifo.c
@@ -87,8 +87,15 @@ bool tu_fifo_config(tu_fifo_t *f, void* buffer, uint16_t depth, uint16_t item_si
 
 static inline uint16_t _ff_mod(uint16_t idx, uint16_t depth)
 {
-  while ( idx >= depth) idx -= depth;
-  return idx;
+  //detect power of 2 case
+  if (depth & (depth-1)) {
+    return idx % depth;
+  }
+  else 
+  {
+    //power of 2 case, fast path
+    return idx & (depth-1);
+  }
 }
 
 // send one item to FIFO WITHOUT updating write pointer


hathach (commented):
thank for the pr, could you provide the before and after timing measurement based on your test setup. It is not clear to me how much it is boost the performance here.

phdussud:
I have implemented a JTAG probe that receives over 6500 USB packets to program a FPGA. During my tests the wall clock programming goes from 2.5s to 4.5s as I repeat the programming and then goes back to 2.5s at the 10th programming and the cycle repeats. My fifo depth is 64, so mod will loop a 1000 times in the worst case. After the fix, the wall clock time is stable at 2.4s
This is on a Raspberry pico board.
Thanks,
Patrick
________________________________

phdussud:
By the way, it I pulled my hair for a while before discovering the root cause of the slowdown. I don’t know how to profile code on a Cortex-M0+ so I had to convince myself that it was communication problem first (using a LA I saw that the JTAG output was normal, except the programming bursts were more and more spaced out). I then looked at the state of my USB interface for the first time and for the 9th time. I noticed that the read and write pointers were super large on the 9th time. That surprised me so I looked at the implementation and discovered the mod problem.

Patrick

src/common/tusb_fifo.c
```diff
@@ -87,8 +87,15 @@ bool tu_fifo_config(tu_fifo_t *f, void* buffer, uint16_t depth, uint16_t item_si
 
 static inline uint16_t _ff_mod(uint16_t idx, uint16_t depth)
 {
-  while ( idx >= depth) idx -= depth;
```
hathach, line 90: Thank you for detail explanation, hmm, somehow I remembered the while loop only need to do it maximum twice. Maybe many things has changed (my memory is fading). @PanRe since we use fifo as main buffering for audio/iso transfer, we don't have the option to go power of 2 only right ?

src/common/tusb_fifo.c
```diff
@@ -87,8 +87,15 @@ bool tu_fifo_config(tu_fifo_t *f, void* buffer, uint16_t depth, uint16_t item_si
...
-  return idx;
+  //detect power of 2 case
+  if (depth & (depth-1)) {
+    return idx % depth;
```
hathach, line 92: please keep the previous while() loop, the modulo take lots more time than while minus.

hathach (commented):
Could you also print out the idx value at the 1st and 9th, I am curious to see what is its  value at that time.

src/common/tusb_fifo.c
PanRe: That is very interesting as the while loop is to be executed twice at maximum! The fifo was designed not to need more that that and the while loop was only used to keep it general (in the first PR i did not use the while loop, rather two subtractions but @hathach prefered the while loop).

Could you check the values of idx when the slow-down happens? And depth also?

PanRe: Restricting to power of two was never an option for you as i remember ^^. It would simplify a lot of things, however, restricting to a power of two length is a hard limit and not very practical for audio applications... i would not do it unless there is a serious bug we can not solve otherwise!

hathach: yeah, I remembered that. It is actually better to have arbitrary depth, that will allow us to fit the stack on memory constraint memory. Thanks for reminding, I need more SRAM for my brain !!! :sweat_smile: :sweat_smile:

PanRe: When you are doing so, please give all the variables of the fifo e.g. non_used_index_space etc.!

hathach:
Make sure you use this repo master for testing instead of the fork in raspberrypi repo. If not then switch branch and try again

src/common/tusb_fifo.c
hathach: Hmm,he may use the old code fron raspberrypi fork. There is a bug fix regarding fifo used in pico

PanRe: I think i know... maybe. There is a bug in fifo.c which i fixed within #593. In [line 564](https://github.com/hathach/tinyusb/blob/ab4d30fd6bca02c73eb9b4ff82db0b2b0f403344/src/common/tusb_fifo.c#L564) the command ` else if (count > f->depth)` is wrong! It should be `else if (n >= f->depth)`. This would explain why he needs exactly double the time at some iterations! This is a bug which I verified when i worked on #593! You may try this! :)

src/common/tusb_fifo.c
phdussud: The idx value is 62 the first time and 65472 on the 9th time. As I said, with the version I debugged the idx only grows until it wraps around. 
the depth is 64.

src/common/tusb_fifo.c
phdussud: It seems that your intent was to bound the idx (read and write) to 2*(_depth)-1. It isn't the case, at least on the pico branch. the read and write values are only ever incremented and they eventually wrap around their max uint16_t values. 
So the real correct fix would be to clamp their values to their mod depth.

PanRe: Ah thats it... ok i don't know the pico branch but the FIFO was designed as you mentioned it above!

phdussud: @PanRe : I don't think your PR fixed this. The read pointer and the write pointers are not limited. Your PR will make the write pointer be equal to the read pointer but it does not limit the read pointer. The root cause if that the function _tu_fifo_correct_read_pointer is not called if cnt is not higher than the depth. For my case it is never true. Somewhere,  there needs to be a check that the read and write pointers are not greater than their max values, regardless how many bytes are written or read.

PanRe: Do you use the master branch? Because there in `advance_pointer()` or [here](https://github.com/hathach/tinyusb/blob/1d20c84798305043c2783725a496070c39d2d966/src/common/tusb_fifo.c#L145) the pointers are limited to the mentioned maximum!

The fix i mentioned before fixes a different bug ^^

phdussud: This code is identical on the pico branch 
It seems that in the pico branch at least, max_pointer_idx and non_used_index_space get ovewritten to 0 after being initialized to the right values. I need to track this down further. 
I will let you know what I find.

hathach: @phdussud please try again with the master branch, as mentioned above, there is relevant fixed for this exact queue issue for pico.

phdussud: Ok. The problem isn't with the vendor fifos but with _usbd_qdef which isn't correctly initialized. The fifo that is embedded in the _usbd_qdef has its max_pointer_idx and non_used_index_space left to 0. 
// DCD Event

// Event queue
// OPT_MODE_DEVICE is used by OS NONE for mutex (disable usb isr)
OSAL_QUEUE_DEF(OPT_MODE_DEVICE, _usbd_qdef, CFG_TUD_TASK_QUEUE_SZ, dcd_event_t);
static osal_queue_t _usbd_q;

hathach: yeah, that one exactly.

phdussud: Ah, cool.

hathach (commented):
Thank you for you PR. This is already fixed by #655 . Thank you for @PanRe for joining the review.

phdussud:
I can confirm that everything is good when I switch to the master branch. Thank you very much for your help!!

hathach:
> I can confirm that everything is good when I switch to the master branch. Thank you very much for your help!!

thanks, please test again this repo for your next issue/pr. There has been quite a bit of fixes for rp2040 since it is released in sdk.

PR #772, opened 2021-04-06
add tu_unaligned_read32/write32

hathach:
**Describe the PR**
Alternative to #758 but prefer to using static inline instead of macros. Also take into account that ARMv7, v8 (M3-M7 , M23-M33) can do unaligned access natively. 

@HiFiPhile  @PanRe this is what I had in mind in previous discussion regarding unaligned access. Let me know what you think.

### DIFF
--- a/src/common/tusb_common.h
+++ b/src/common/tusb_common.h
@@ -111,7 +111,7 @@ static inline uint16_t tu_unaligned_read16(const void* mem)
 
 static inline void tu_unaligned_write16(void* mem, uint16_t value)
 {
-  tu_unaligned_uint16_t const* ua16 = (tu_unaligned_uint16_t const*) mem;
+  tu_unaligned_uint16_t* ua16 = (tu_unaligned_uint16_t*) mem;
   ua16->val = value;
 }
 
--- a/src/common/tusb_common.h
+++ b/src/common/tusb_common.h
@@ -81,50 +81,50 @@
 #define tu_varclr(_var)          tu_memclr(_var, sizeof(*(_var)))
 
 //------------- Bytes -------------//
-static inline uint32_t tu_u32(uint8_t b1, uint8_t b2, uint8_t b3, uint8_t b4)
+TU_ATTR_ALWAYS_INLINE static inline uint32_t tu_u32(uint8_t b1, uint8_t b2, uint8_t b3, uint8_t b4)
 {
-  return ( ((uint32_t) b1) << 24) + ( ((uint32_t) b2) << 16) + ( ((uint32_t) b3) << 8) + b4;
+  return ( ((uint32_t) b1) << 24) | ( ((uint32_t) b2) << 16) | ( ((uint32_t) b3) << 8) | b4;
 }
 
-static inline uint16_t tu_u16(uint8_t high, uint8_t low)
+TU_ATTR_ALWAYS_INLINE static inline uint16_t tu_u16(uint8_t high, uint8_t low)
 {
-  return (uint16_t)((((uint16_t) high) << 8) + low);
+  return (uint16_t) ((((uint16_t) high) << 8) | low);
 }
 
-static inline uint8_t tu_u16_high(uint16_t u16) { return (uint8_t) (((uint16_t) (u16 >> 8)) & 0x00ff); }
-static inline uint8_t tu_u16_low (uint16_t u16) { return (uint8_t) (u16 & 0x00ff); }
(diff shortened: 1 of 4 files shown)


PanRe:
I guess @HiFiPhile knows more of this topic but in any case, could you provide an unaligned access to int16 too? This is required for fast half word copies in the audio encoding/deconding! Thanks :)

HiFiPhile:
With a quick looking it should work :)

With IAR there is a little inconvenient is with inline function I can't disable ` Warning[Pa039]: use of address of unaligned structure member`.
So we need to continue to use `(uint8_t*) desc_bos + offsetof(tusb_desc_bos_t, wTotalLength)` instead of `&desc_bos->wTotalLength`.

hathach:
> With a quick looking it should work :)
> 
> With IAR there is a little inconvenient is with inline function I can't disable ` Warning[Pa039]: use of address of unaligned structure member`.
> So we need to continue to use `(uint8_t*) desc_bos + offsetof(tusb_desc_bos_t, wTotalLength)` instead of `&desc_bos->wTotalLength`.

can the pragma suppress be added into the inline ? we can have temporarily add that for now and find an better/general way later on.

hathach:
> I guess @HiFiPhile knows more of this topic but in any case, could you provide an unaligned access to int16 too? This is required for fast half word copies in the audio encoding/deconding! Thanks :)

Yeah, I was initially want to add 16-bit version, but put it off since I am not sure if we would use it at all. Though since there are all inline, there is no generated code if not used. So here it is

HiFiPhile:
With `__attribute__((always_inline))` the warning can be disable using pragma:
```
__attribute__((always_inline)) static inline uint32_t tu_unaligned_read32(const void* mem)
{
  _Pragma ("diag_suppress=Pa039");
  tu_unaligned_uint32_t const* ua32 = (tu_unaligned_uint32_t const*) mem;
  return ua32->val;
  _Pragma ("diag_default=Pa039");
}
```
Without `always_inline` sometimes IAR don't like to inline the function especially in low optimization, the pragma doesn't work since the warning is issued on function caller.

hathach:
@HiFiPhile I added the TU_ATTR_ALWAYS_INLINE, and applied it to basic inline function. Though I don't like to suppress and unsuppress it here. It makes code look less readable, also not all unaligned access need to suppress this e.g uint8_t/*void*. I am fine with the offsetof() at the caller. 

```C
#ifdef __ICCARM__
  #define IAR_PA309_SUPPRESS    _Pragma("diag_suppress=Pa039")
  #define IAR_PA309_UNSUPPRESS  _Pragma ("diag_default=Pa039")
#else
  #define IAR_PA309_SUPPRESS
  #define IAR_PA309_UNSUPPRESS
#endif

#if TUP_ARCH_STRICT_ALIGN

typedef struct { uint16_t val; } TU_ATTR_PACKED tu_unaligned_uint16_t;
typedef struct { uint32_t val; } TU_ATTR_PACKED tu_unaligned_uint32_t;

TU_ATTR_ALWAYS_INLINE static inline uint32_t tu_unaligned_read32(const void* mem)
{
  IAR_PA309_SUPPRESS
  tu_unaligned_uint32_t const* ua32 = (tu_unaligned_uint32_t const*) mem;
  return ua32->val;
  IAR_PA309_UNSUPPRESS
}

TU_ATTR_ALWAYS_INLINE static inline void tu_unaligned_write32(void* mem, uint32_t value)
{
  IAR_PA309_SUPPRESS

  tu_unaligned_uint32_t* ua32 = (tu_unaligned_uint32_t*) mem;
  ua32->val = value;

  IAR_PA309_UNSUPPRESS
}

TU_ATTR_ALWAYS_INLINE static inline uint16_t tu_unaligned_read16(const void* mem)
{
  IAR_PA309_SUPPRESS

  tu_unaligned_uint16_t const* ua16 = (tu_unaligned_uint16_t const*) mem;
  return ua16->val;

  IAR_PA309_UNSUPPRESS
}

TU_ATTR_ALWAYS_INLINE static inline void tu_unaligned_write16(void* mem, uint16_t value)
{
  IAR_PA309_SUPPRESS

  tu_unaligned_uint16_t* ua16 = (tu_unaligned_uint16_t*) mem;
  ua16->val = value;

  IAR_PA309_UNSUPPRESS
}

#else

TU_ATTR_ALWAYS_INLINE static inline uint32_t tu_unaligned_read32  (const void* mem)
{
  IAR_PA309_SUPPRESS
  return *((uint32_t*) mem);
  IAR_PA309_UNSUPPRESS
}
```

PR #776, opened 2021-04-09
Add support for SAM L22 family

joeycastillo:
**Describe the PR**
This pull request adds support for the SAM L22 family of microcontrollers.

**Additional context**
The SAM L22 is similar enough to the SAM D11/D21 that it's able to use the same implementation in dcd_samd.c, so I just added it to the platform checks in that file. Tested with just about all of the examples on the two boards I'm developing with, a [SAM L22 Feather](https://github.com/joeycastillo/Feather-Projects/tree/main/SAML22%20Feather) and the [Sensor Watch](https://github.com/joeycastillo/Sensor-Cap/tree/main/PCB/Main%20Boards) (OSO-SWAT-A1-02). 

I know that the porting guidelines say to use a readily available development board, but I don't have a SAM L22 Xplained board on hand. Both of these boards are open source designs, though.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -48,6 +48,7 @@ jobs:
         - 'samd11'
         - 'samd21'
         - 'samd51'
+        - 'saml22'
         - 'stm32f4'
         - 'stm32f7'
         - 'stm32h7'
--- a/docs/boards.md
+++ b/docs/boards.md
@@ -47,6 +47,11 @@ This code base already had supported for a handful of following boards (sorted a
 
 - [Microchip SAMG55 Xplained Pro](https://www.microchip.com/DevelopmentTools/ProductDetails/PartNO/ATSAMG55-XPRO)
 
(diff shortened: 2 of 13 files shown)


hathach (commented):
Super! This looks great, I have no problems with board that are in the making. It would be great if you could 
1. update [boards.md](https://github.com/hathach/tinyusb/blob/master/docs/boards.md) with your boards link just to help people looking up schematcis/info.
2. Update ci build to also include saml22 [build.yml](https://github.com/hathach/tinyusb/blob/1d20c84798305043c2783725a496070c39d2d966/.github/workflows/build.yml#L50) . Depending on the flash/ram some heavy example would fail, but we could easily add `.skip.MCU_SAML22` file to those to skip building it

joeycastillo:
Thanks for the feedback! I made those changes, and all of the examples built with no issues.

hathach (approved):
Perfect, thank you very much for your PRs.

PR #783, opened 2021-04-12
WIP: Add new Espressif target esp32s3 for tinyUSB

alisitsyn:
## What happened in this PR
This PR adds support of a new Espressif chip ESP32-S3 into tinyusb project.
The chip peripheral is very similar compare to ESP32-S2 target and support is added as a family driver to allow following extension of the family and group the controllers with similar USB peripheral.

1. Update the device controller driver for S family esp32sx:
`tinyusb/src/portable/espressif/esp32sx/dcd_esp32sx.c`
2. Add new esp32s3 based board in 
`tinyusb/hw/bsp/esp32s3_addax_1\*`
3. Add new chip target and family in board.mk files for all Espressif boards:
```
CHIP_TARGET = esp32s3
CHIP_FAMILY = esp32sx
VENDOR = espressif
```
4. Update of the cmake wrapper in `\tinyusb\examples\rules.mk` file to specify target chip and add new make targets to flash board using dfu-flash utility.
5. Fix the examples: `hid_composite_freertos`, `cdc_msc_freertos`, `board_test` to allow selection of different target boards.
6. The compilation is performed using the same `make BOARD=esp32s3_addax_1` command.
The compiled example can be flashed to the target board using the commands below:
`make BOARD=esp32s3_addax_1 dfu` - generates the dfu image for software
`make BOARD=esp32s3_addax_1 dfu-flash` -flashes the dfu image into target board

[dfu-flash documentation for ESP32-S2](https://docs.espressif.com/projects/esp-idf/en/latest/esp32s2/api-guides/usb-console.html)

The `soc/hal` layers in ESP-IDF are already updated to support the new chip.


me-no-dev:
Hi @alisitsyn :) There seem to be changes included that need to be reverted, like submodules being removed. Also to make the changes more clear, is it possible to not include the new example?

alisitsyn:
Hi @me-no-dev,
> There seem to be changes included that need to be reverted, like submodules being removed. 

The branch update is in progress (rebase to the latest master) and will be ready after testing soon.

> Also to make the changes more clear, is it possible to not include the new example?

My intention is not addition of a new example but update the existing espressif examples to be able to compile them for new added board based on new esp32s3 chip - `esp32s3_addax_1`.

examples/rules.mk
```diff
@@ -2,20 +2,26 @@
...
 
-.PHONY: all clean flash
-.DEFAULT_GOAL := all
+# Espressif IDF use CMake build system, this add wrapper target to call idf.py
+.PHONY: all clean build flash monitor dfu-flash dfu
+.DEFAULT_GOAL := build
 
-all:
-	idf.py -B$(BUILD) -DBOARD=$(BOARD) build
+all flash monitor dfu-flash dfu :
```
hathach, line 13: nice, I didn't know we could do this

examples/rules.mk
```diff
@@ -2,20 +2,26 @@
 # Common make rules for all examples
 # ---------------------------------------
 
-ifeq ($(CROSS_COMPILE),xtensa-esp32s2-elf-)
-# Espressif IDF use CMake build system, this add wrapper target to call idf.py
+ifeq ($(findstring esp32s, $(BOARD)), esp32s)
+$(info Vendor: $(VENDOR), family: $(CHIP_FAMILY), target: $(CHIP_TARGET))
```
hathach, line 6: not all board has esp32s2 or esp32s3 in their name, we should just use the CHIP_FAMILY. It is good enough.

hw/bsp/esp32s2_kaluga_1/board.mk
```diff
@@ -1,2 +1,9 @@
 # Cross Compiler for ESP32
 CROSS_COMPILE = xtensa-esp32s2-elf-
+
+CFLAGS += \
+    "-DCFG_TUSB_MCU=OPT_MCU_ESP32S2"
```
hathach, line 5: seem like this line isn't needed, I didn't see it on other board.mk

hathach (changes requested):
Thank you very much for your PR. It is great to have the new ESP32-S3 chip added. The changes look good, though I have a few feedback
1. Please revert the submodules removal. They are important and needed to build for other platforms :) 
2. Can you move the esp32s3 into it own family `hw/bsp/esp32s3` similar to esps2, it will help to group similar driver such as board_init() and ws2812 driver.

CI also needs to be update to build with esp32s3, though I could help with that.

Since PR status is WIP, I converted it to Draft. Feel free to push new update, I will try to follow up with feedback. Whenever you think it is ready for a merge, just Mark is as ready to covert it back.

examples/rules.mk
alisitsyn: The commit I was based on contains the names of the espressif boards started with target prefix. This looked useful for me. This was updated accordingly as it is in latest master.

alisitsyn:
@hathach,

Thank you for your feedback and advices.

> Please revert the submodules removal. They are important and needed to build for other platforms.

It seems that submodules were removed incidentally by script.

> Can you move the esp32s3 into it own family hw/bsp/esp32s3 similar to esps2, it will help to group similar driver such as board_init() and ws2812 driver.

I am going to move it into the separated family folder `hw/bsp/esp32sx` how it was done in the latest master. All examples will inherit the same standard board functionality and ws2812 driver (exists on esp32s3_addax_1 board also).

> CI also needs to be updated to build with `esp32s3`, though I could help with that.

I updated the CI scripts to work with `esp32sx` family instead of `esp32s2` based on latest master. The branch will be updated soon. I will need your advises soon.

examples/rules.mk
hathach: name like Adafruit funhouse or tinyS2 etc. won't help esp32s2 or s3. It will cause issue with other boards in the future. The $(VENDOR) variable will probably be removed as well. I don't see the use of it in the make. Board name is already uniquely defined.  If you think `esp32s2_saola_1` does not reflect the espressif as vendor, we could rename it to `espressif_saola_1`

hathach:
> It seems that submodules were removed incidentally by script.

no problem at all, just make sure you revert it 

> I am going to move it into the separated family folder `hw/bsp/esp32sx` how it was done in the latest master. All examples will inherit the same standard board functionality and ws2812 driver (exists on esp32s3_addax_1 board also).

I actually think it is better to keep family separated as `esp32s2` and `esp32s3` (rather than esp32sx), although IDF does an great job of driver abstraction, It is still useful and more obvious to end-user. Yeah, I know there is a bit of code duplication but I am fine with it. Other mcu such as stm32 f2 f3 f4 is actually quite similar, but separated them make it easier to maintain and manage. Hope this makes sense

> I updated the CI scripts to work with `esp32sx` family instead of `esp32s2` based on latest master. The branch will be updated soon. I will need your advises soon.

Great, I am looking forward your next push.

examples/rules.mk
alisitsyn: Ah, I see now. The board name is enough to determine all required information as `$(CHIP_TARGET), $(FAMILY)` and so on...

alisitsyn:
> I actually think it is better to keep family separated as esp32s2 and esp32s3 (rather than esp32sx), although IDF does an great job of driver abstraction, It is still useful and more obvious to end-user. Yeah, I know there is a bit of code duplication but I am fine with it. Other mcu such as stm32 f2 f3 f4 is actually quite similar, but separated them make it easier to maintain and manage. Hope this makes sense

alisitsyn:
@me-no-dev , Unfortunately I can not add you to reviewers. Could you take a look to update and give me your comments? Thank you.

hathach:
> Actually the ESP32S (like F4 for STM) is the family and ESP32-S2, ESP32-S3 are actual chips inside this family. I personally think that this is logically correct. It is not an issue to follow your approach from make standpoint. As I understand the structure will be like below. Is this correct?
> 
> ```
> tinyusb_esp\hw\bsp\
>                     esp32s2\ < --- esp32s2 family boards
>                         boards\
>                             adafruit_feather_esp32s2
>                             adafruit_magtag_29gray
>                             adafruit_metro_esp32s2
>                             espressif_kaluga_1
>                             espressif_saola_1
>                     esp32s3 < --- esp32s2 family boards
>                         boards\
>                             espressif_addax_1can temporarily

Yeah, this is correct. Normally MCUs within families is only slightly different in ram & rom, pinout (and maybe an extra peripheral). I thought esp32s3 is totally different with extra core. Though you guys at espressif apparently understand it better, if you think their cores and peripheral is similar enough to be in the same group. So yes, please go ahead with this. we can call it `esp32sx`.

> tinyusb_esp\src\portable\espressif\ < ---- dcd driver
>                                     esp32s2\
>                                     esp32s3\

In both cases, the dcd file should only be one file `dcd_esp32sx.c`, dcd is only different if mcus use totally different USB IP. In this case, both s2 and s3 (and other stm32) share the same synopsys IP. 

> The CI test for `espressif_addax_1` board is failed because the actual update of `soc/hal` drivers in ESP-IDF is not merged.

This wouldn't an issue, can you tell me at which branch it works, I will see if I could quickly modify ci to get it build (also have some local testing as well). If it is not easy, we could temporarily disable its build.

hathach:
> @me-no-dev , Unfortunately I can not add you to reviewers. Could you take a look to update and give me your comments? Thank you.

Hmm, that is weird, I am sure he is one of repo contributors, in any case, I have added him as reviewer.

hathach (commented):
This looks good already, please keep it that way. No need to separate s2 and s3. Ignore my previous comment, since I thought s2 and s3 are totally different due to addition of new core and ble :sweat_smile: . This look great already. I will try to have some hand-on testing and will provide more feedback soon.

alisitsyn:
@hathatch,

> Normally MCUs within families is only slightly different in ram & rom, pinout (and maybe an extra peripheral). I thought esp32s3 is totally different with extra core. Though you guys at espressif apparently understand it better, if you think their cores and peripheral is similar enough to be in the same group. So yes, please go ahead with esp32sx

You are right about the ESP32-S3. It has enough differences (cores, different peripheral and memory changes) and I can see the reason to divide them into two different family folders.

> In both cases, the dcd file should only be one file dcd_esp32sx.c, dcd is only different if mcus use totally different USB IP. In this case, both s2 and s3 (and other stm32) share the same synopsys IP.

The USB IP are almost the same, so agree it is better to keep the one device controller driver for them. I also saw the https://github.com/hathach/tinyusb/issues/382 but it is better to keep existing driver for now.

> This wouldn't an issue, can you tell me at which branch it works, I will see if I could quickly modify ci to get it build (also have some local testing as well). If it is not easy, we could temporarily disable its build.

The branch with the required update is in the internal system. I can provide the patch to fix the actual hal/soc in esp-idf master.
I will contact our team to provide the better way to solve this and will contact you then.

Thank you for your comments.

hathach:
> You are right about the ESP32-S3. It has enough differences (cores, different peripheral and memory changes) and I can see the reason to divide them into two different family folders. I think it is better to get team opinion about this, cc @me-no-dev.

Actually, it is not a big deal, this bsp family folder thing is just an way to organize boards for testing stock examples in tinyusb repo. It should not make much of differences. Since most end-user people will just IDF platform, arduino or micropython/circuitpython. It is not really that important and can be easily changes later on if needed. One of the tip to see if we should split or not is when you see lots of #ifdef in source/cmake etc .. I will leave it up to you to decide, I am fine with either way.

> The USB IP are almost the same, so agree it is better to keep the one device controller driver for them. I also saw the #382 but it is better to keep existing driver for now.

Yup, later on I would like to abstract the synopsys driver enough so that stm and esp32s can share functions (much like ehci, ohci). Currently stm32f receive lots of improvement and enhancement regarding iso, fifo scheme/optimization. But it doesn't apply to esp32s just yet due to lack of time for testing. But that is for later PRs.

> The branch with the required update is in the internal system. I can provide the patch to fix the actual hal/soc in esp-idf master.
> I will contact our team to provide the better way to solve this and will contact you then.
> 
> Thank you for your comments. I will contact you ASAP.

Actually, I am fine with testing it, we can just comment out the build for esp32s3 in ci matrix board build for now, so that ci won't block other PRs. We could always enable later on by a flip of the comment. Our focus should only on the file changes. Give me a bit of time, I am still catching up with review :)

alisitsyn:
> Actually, it is not a big deal, this bsp family folder thing is just an way to organize boards for testing stock examples in tinyusb repo. It should not make much of differences. Since most end-user people will just IDF platform, arduino or micropython/circuitpython. It is not really that important and can be easily changes later on if needed.

I agree with your opinion and pushed new commit to update PR according to our discussion. Please take a look when possible.

> Yup, later on I would like to abstract the synopsys driver enough so that stm and esp32s can share functions (much like ehci, ohci). Currently stm32f receive lots of improvement and enhancement regarding iso, fifo scheme/optimization. But it doesn't apply to esp32s just yet due to lack of time for testing. But that is for later PRs.

This is a good idea and I can take a look to update later in order to apply changes to espressif driver.

> We could always enable later on by a flip of the comment. Our focus should only on the file changes. Give me a bit of time, I am still catching up with review

I rebased on master and pushed new commit to this PR to group boards using the target name as family. The esp32s3 based board was commented out in board matrix. Still have issues with `raspberry_pi_pico` but the failures are not related to update.
Please review the changes when you have time. 

Thanks you!

examples/make.mk
```diff
@@ -27,6 +27,9 @@ endif
 # Board within family
 ifeq ($(BOARD_PATH),)
   BOARD_PATH := $(subst $(TOP)/,,$(wildcard $(TOP)/hw/bsp/*/boards/$(BOARD)))
+ifneq ($(wildcard $(TOP)/hw/bsp/*/boards/$(BOARD)/board.mk),)
+    include $(TOP)/hw/bsp/*/boards/$(BOARD)/board.mk
+endif
```
hathach: the `board.mk` should not be included here. The reason is that depending on the family, it may need some variables define before the include of board.mk (e.g board linker overwrite family linker etc.. ) and some after. So it is best to remove and let the family.mk to decide.

examples/device/board_test/CMakeLists.txt
```diff
@@ -7,7 +7,7 @@ set(TOP "../../..")
 get_filename_component(TOP "${TOP}" REALPATH)
 
 # Check for -DFAMILY=
-if(FAMILY STREQUAL "esp32s2")
+if(FAMILY MATCHES "^(esp32s[2-3])*")
```
hathach: regex is not quite right, `*` allow some odds such as `esp32s2esp32s3` match as well. We could simply have it as `"^esp32s[2-3]"`

hathach (approved):
Great work, everything look good now. I have a couple of minor comments. Since I have do a merge from (master) to fix rp2040 build, I take the chance to update the PR according to the comment as well.

examples/make.mk
alisitsyn: > 
> 
> the `board.mk` should not be included here. The reason is that depending on the family, it may need some variables define before the include of board.mk (e.g board linker overwrite family linker etc.. ) and some after. So it is best to remove and let the family.mk to decide.
This approach with included `board.mk` here was used in previous approach (with esp32sx family) to override the chip target settings. In this update it is not required any more. I removed the `board.mk` from espressif chip board folders but forgot to remove these lines from `make.mk`.

examples/device/board_test/CMakeLists.txt
alisitsyn: > 
> 
> regex is not quite right, `*` allow some odds such as `esp32s2esp32s3` match as well. We could simply have it as `"^esp32s[2-3]"`

agree, in some very rare cases it can cause some issues.

alisitsyn:
> I take the chance to update the PR according to the comment as well. I think this PR is ready to merge, unless you still plan to push more update ( let me know if you do ).

I do not have anything to update. Please merge this PR ASAP in spite of rp2040 failures. My esp-idf MR will be merged then. Thank you.

hathach:
Merged now, thank you very much for your PR :+1: :+1:

alisitsyn:
> 
> 
> Merged now, thank you very much for your PR 👍 👍

Thank you very much for your support. 👍

hathach:
> > Merged now, thank you very much for your PR +1 +1
> 
> Thank you very much for your support.

I am the one to say thank :)

alisitsyn:
Hello @hathach,

There was the delay with merging of [esp-idf tinyusb support on esp32s3](https://github.com/espressif/esp-idf/commit/ea6710ce988f643376850091455dd9c402b96b26) , but it was finally merged. Could you help to reenable CI test support for `esp32s3_addax_1` board? Please let me know if you have any questions. Thank you.

hathach:
@alisitsyn no problem, done via https://github.com/hathach/tinyusb/pull/840

alisitsyn:
@hathach,

Thank you very much! :+1:

hathach:
No problem, you're welcome

PR #786, opened 2021-04-15
Changed LED_PIN to 17 on Seeeduino Xiao

kapacuk:
The LEDs on Seeduino Xiao are connected to PA17, PA18, and PA19:

https://files.seeedstudio.com/wiki/Seeeduino-XIAO/res/Seeeduino-XIAO-v1.0-SCH-191112.pdf

There is no LED on pin 26.

### DIFF
--- a/hw/bsp/samd21/boards/seeeduino_xiao/board.h
+++ b/hw/bsp/samd21/boards/seeeduino_xiao/board.h
@@ -32,8 +32,8 @@
 #endif
 
 // LED
-#define LED_PIN               26
-#define LED_STATE_ON          1
+#define LED_PIN               17
+#define LED_STATE_ON          0
 
 // Button
 #define BUTTON_PIN            9 // PA4 pin D1 on seed input


hw/bsp/samd21/boards/seeeduino_xiao/board.h
```diff
@@ -32,7 +32,7 @@
 #endif
 
 // LED
-#define LED_PIN               26
+#define LED_PIN               17
```
hathach, line 35: From schematic, it seems like the led is active low. Could you also change the `LED_STATE_ON` to 0 as well.

kapacuk: You're right, I'll do that.

kapacuk: I've changed LED_STATE_ON to 0 and amended my last commit. I hope I've done it right, but if not I can raise another PR.

hathach: That is ok, I don't really mind the git log history. The commit message and actual code changes are far more important.

hathach (approved):
Thank you very much for your PR. Maybe the board revision changes or something. I am glad you spotted it.

PR #787, opened 2021-04-15
Added support for Black Magic Probe

kapacuk:
Added flash-bmp and debug-bmp make targets to allow flashing and debugging with the [Black Magic Probe](https://github.com/blacksphere/blackmagic)

### DIFF
--- a/examples/rules.mk
+++ b/examples/rules.mk
@@ -170,6 +170,12 @@ flash-pyocd: $(BUILD)/$(PROJECT).hex
 	pyocd flash -t $(PYOCD_TARGET) $<
 	pyocd reset -t $(PYOCD_TARGET)
 
+# flash with Black Magic Probe
+flash-bmp: $(BUILD)/$(PROJECT).elf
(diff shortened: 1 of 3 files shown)


examples/rules.mk
```diff
@@ -170,6 +170,12 @@ flash-pyocd: $(BUILD)/$(PROJECT).hex
 	pyocd flash -t $(PYOCD_TARGET) $<
 	pyocd reset -t $(PYOCD_TARGET)
 
+# flash with Black Magic Probe
+flash-bmp: $(BUILD)/$(PROJECT).elf
+	$(GDB) --batch -ex 'target extended-remote /dev/ttyBmpGdb' -ex 'monitor swdp_scan' -ex 'attach 1' -ex load  $<
```
hathach, line 175: does the `/dev/ttyBmpGdb` name changes depending on the os (windows/macos/linux) like CDC Serial. If it does we may need to have user to specify to `SERIAL=` or `BMP=` similar to other dfu via serial. Of course, we could have with default value that works on most platform with `BMP ?=`

kapacuk: Yes, it'll be different on different systems. I'll add the BMP variable as you suggested.

hathach (approved):
thank you for your PR, it look good to me. I did a bit of minor update as well

PR #788, opened 2021-04-16
Change LED_STATE_ON to 0 for stm32f401 and stm32f411 blackpill boards

kapacuk:
According to the [blackpill schematics](https://github.com/WeActTC/MiniSTM32F4x1/blob/master/HDK/MiniF4x1Cx_V30%20SchDoc.pdf), the LED pin is active low on both F401 and F411 boards. It seems to be the same across all revisions from V1.2 to V3.0.


hathach (approved):
ah right, you are spot-on. Thank you for your PR.

PR #789, opened 2021-04-17
When used with FreeRTOS, lower the interrupt priority for OTG_FS_IRQn

kapacuk:
None of the FreeRTOS examples currently work with STM32F4, they all trigger this assert in `lib/FreeRTOS-Kernel/portable/GCC/ARM_CM4F/port.c`:

`764                     configASSERT( ucCurrentPriority >= ucMaxSysCallPriority ); `

because the interrupt priority for OTG_FS is too high (that is, the value of `ucCurrentPriority` is too low).

The (almost) correct call to `NVIC_SetPriority` was commented out in `hw/bsp/stm32f4/family.c`. 

I have a suspicion that other Cortex-M chips might be affected too, but I can only test Seeeduino Xiao and STM32F4 blackpill.

### DIFF
--- a/hw/bsp/stm32f4/family.c
+++ b/hw/bsp/stm32f4/family.c
@@ -51,7 +51,7 @@ void board_init(void)
   SysTick_Config(SystemCoreClock / 1000);
 #elif CFG_TUSB_OS == OPT_OS_FREERTOS
   // If freeRTOS is used, IRQ priority is limit by max syscall ( smaller is higher )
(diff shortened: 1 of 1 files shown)


hathach (approved):
Thank for the fix, yeah, this is very common issue when using cortex m with freertos. I have this nvic setprio for most of family, but apparently not all of them. This is downside of having too many boards/ports, that sometime I couldn't do all the hand-on testing on actual hardware

Note: The issue is only occurred if default priority (on reset) of USB is too high, somehow my stmf411disco board work just fine without this patch.

PR #797, opened 2021-04-19
Create a new make rule for checking out submodules.

kapacuk:
Currently, if you do a fresh checkout of the tinyusb project and try to build a FreeRTOS example you get an error like this on the first run:

```
......
LINK _build/seeeduino_xiao/seeeduino_xiao-cdc_msc_freertos.elf
arm-none-eabi-gcc: error: _build/seeeduino_xiao/obj/lib/FreeRTOS-Kernel/list.o: No such file or directory
arm-none-eabi-gcc: error: _build/seeeduino_xiao/obj/lib/FreeRTOS-Kernel/queue.o: No such file or directory
arm-none-eabi-gcc: error: _build/seeeduino_xiao/obj/lib/FreeRTOS-Kernel/tasks.o: No such file or directory
arm-none-eabi-gcc: error: _build/seeeduino_xiao/obj/lib/FreeRTOS-Kernel/timers.o: No such file or directory
```

You have to run make again, which is a nuisance.

When you run make the first time the files `list.c`, `queue.c`, `tasks.c` and `timers.c` don't exist under `lib/FreeRTOS-Kernel`, so there are no rules to build the corresponding `*.o` files.

This patch creates a new rule for checking out submodules, instead of doing it in the `$(shell)` function. The rule initialises every submodule directory and create an empty submodule.mk file in it. These empty files are included into the main makefile, forcing make to re-read everything and rebuild all rules after this stage.  As a result, make works successfully on the first run.

### DIFF
--- a/examples/make.mk
+++ b/examples/make.mk
@@ -46,9 +46,11 @@ else
 endif
 
 # Fetch submodules depended by family
-fetch_submodule_if_empty = $(if $(wildcard $(TOP)/$1/*),,$(info $(shell git -C $(TOP) submodule update --init $1)))
 ifdef DEPS_SUBMODULES
-  $(foreach s,$(DEPS_SUBMODULES),$(call fetch_submodule_if_empty,$(s)))
+$(foreach s,$(DEPS_SUBMODULES), $(TOP)/$(s)/submodule.mk):
+	for module in $(DEPS_SUBMODULES); do git -C $(TOP) submodule update --init $$module && touch $(TOP)/$$module/submodule.mk; done
+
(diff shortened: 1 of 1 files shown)


kapacuk:
hm.. let me fix that build failure first.

kapacuk:
This does not work very well with `make -j`. Back to the drawing board..

hathach:
yeah, the fetch submodules on-the-fly do not work well with -j option. And gcc may try to split task and I don't know, it doesn't seem to work well with dramatics file changes, similar to `make -j clean all` won't work. I was thinking of having submodule target, though I think just press up for previous target is much quicker. This is one-time issue, not a deal breaker though. Don't spend too much time on it :)

PR #798, opened 2021-04-20
Uac2 fix cs as int alt set 0

PanRe:
**Describe the PR**
Fix bug where the pointer `uint8_t * alt_setting` was reset to zero when the audio interface was opened by host. Refactor some static functions for faster checks and renamed old depracted and misleading variable names.


src/class/audio/audio_device.c
```diff
@@ -676,30 +682,30 @@ static bool audiod_decode_type_I_pcm(uint8_t rhport, audiod_interface_t* audio,
  *  Write data to buffer. If it is full, new data can be inserted once a transmit was scheduled. See audiod_tx_done_cb().
  *  If TX FIFOs are used, this function is not available in order to not let the user mess up the encoding process.
  *
- * \param[in]       itf: Index of audio function interface
+ * \param[in]       audio_fct_idx: Index of audio function interface
```
hathach, line 685: audio_fct_idx is too long, audio prefix make it harder to read, how about `func_id`or `fid`

src/class/audio/audio_device.c
```diff
@@ -1424,16 +1433,16 @@ static bool audiod_set_interface(uint8_t rhport, tusb_control_request_t const *
   // 3. Open new EP
 
   uint8_t const itf = tu_u16_low(p_request->wIndex);
-  uint8_t const alt = tu_u16_low(p_request->wValue);
+  volatile uint8_t const alt = tu_u16_low(p_request->wValue);
```
hathach, line 1436: `volatile` should be removed, it is redundant

hathach (commented):
Thank you for the PR, the review is nothing big, just a minor comment on variable naming and a redundant volatile.

src/class/audio/audio_device.c
PanRe: Sorry for that, left from debugging!

src/class/audio/audio_device.c
PanRe: I go with `func_id`!

hathach: I would go with that as well

src/class/audio/audio_device.c
hathach: no problem at all :)

hathach (approved):
it looks perfect now.

Issue #732, opened 2021-03-17
Generic USB TMC demo support


xmos-jmccarthy:
The current tmc demo has a script called visaQuery.py that explicitly uses a Windows 64bit dll.  Consequently, the current demo cannot be run on other platforms without local changes.

> My putting a particular RM DLL path was a bad idea, in retrospect. It was a hacky way for me to get it working on my laptop, and I shouldn't have done it. Also, it may not work for all installs of Windows.
>
> What are your thoughts about if we remove all of the os-detection, and just have a call to:
>
> 
> ```rm = visa.ResourceManager()```
> 
> With that, one can create a `.pyvisarc` with the proper library path based on which VISA library the user wants (if pyVISA's auto-selection doesn't work)?
> 
> (And, in general, it's good to see other people trying out the class driver. If you think there should be API changes, let me know.)

_Originally posted by @pigrew in https://github.com/hathach/tinyusb/issues/724#issuecomment-801223337_

PR #803, opened 2021-04-23
USBTMC: remove reference to visa DLL (which only was for windows, and…

pigrew:
USBTMC: Remove reference reference to windows path of VISA DLL (let pyvisa auto-select it).

If manual selection is needed, user may create .pyvisarc file with as documented by pyvisa.

Closes: #732


hathach (approved):
look good, thank you for your PR.

PR #804, opened 2021-04-23
Implement functions to allow for DMA usage in audio driver.

PanRe:
**Describe the PR**
Follow the discussion in #593, this attempts to implement functions to allow for DMA usage in audio functions.

- Add tud_audio_n_get_ep_out_ff(), tud_audio_n_get_ep_in_ff(),
tud_audio_n_get_rx_support_ff(), and tud_audio_n_get_tx_support_ff()
- Change get_linear_read/write_info() to return linear and wrapped part
at once
- Adjusted affected code in audio_device.c and tested with
audio_4_channel.

### DIFF
--- a/src/common/tusb_fifo.c
+++ b/src/common/tusb_fifo.c
@@ -883,7 +883,7 @@ void tu_fifo_advance_read_pointer(tu_fifo_t *f, uint16_t n)
 
 /******************************************************************************/
 /*!
-   @brief Get linear read info
+   @brief Get read info
 
    Returns the length and pointer from which bytes can be read in a linear manner.
    This is of major interest for DMA transmissions. If returned length is zero the
@@ -895,24 +895,18 @@ void tu_fifo_advance_read_pointer(tu_fifo_t *f, uint16_t n)
    wrapped part!
    @param[in]       f
                     Pointer to FIFO
-   @param[in]       offset
-                    Number of ITEMS to ignore before start writing
-   @param[out]      **ptr
-                    Pointer to start writing to
-   @param[in]       n
-                    Number of ITEMS to read from buffer
-   @return          len
-                    Length of linear part IN ITEMS, if zero corresponding pointer ptr is invalid
+   @param[out]      *info
+                    Pointer to struct which holds the desired infos
  */
 /******************************************************************************/
-tu_fifo_linear_wr_info tu_fifo_get_linear_read_info(tu_fifo_t *f, uint16_t offset, uint16_t n)
+void tu_fifo_get_read_info(tu_fifo_t *f, tu_fifo_buffer_info_t *info, uint16_t n)
 {
   // Operate on temporary values in case they change in between
   uint16_t w = f->wr_idx, r = f->rd_idx;
 
   uint16_t cnt = _tu_fifo_count(f, w, r);
 
-  // Check overflow and correct if required
+  // Check overflow and correct if required - may happen in case a DMA wrote too fast
   if (cnt > f->depth)
   {
     _ff_lock(f->mutex_rd);
@@ -922,36 +916,40 @@ tu_fifo_linear_wr_info tu_fifo_get_linear_read_info(tu_fifo_t *f, uint16_t offse
     cnt = f->depth;
   }
 
-  tu_fifo_linear_wr_info info = {0,0,NULL,NULL};
-
   // Skip beginning of buffer
-  if (cnt == 0 || offset >= cnt) return info;
+  if (cnt == 0)
+  {
+    info->len_lin = 0;
+    info->len_wrap = 0;
+    info->ptr_lin = NULL;
+    info->ptr_wrap = NULL;
+    return;
+  }
 
-  // Check if we can read something at and after offset - if too less is available we read what remains
-  cnt -= offset;
   if (cnt < n) n = cnt;
 
   // Get relative pointers
-  w = get_relative_pointer(f, w, 0);
-  r = get_relative_pointer(f, r, offset);
+  w = get_relative_pointer(f, w,0);
+  r = get_relative_pointer(f, r,0);
 
   // Copy pointer to buffer to start reading from
-  info.ptr_lin = &f->buffer[r];
-  info.ptr_wrap = f->buffer;
+  info->ptr_lin = &f->buffer[r];
 
   // Check if there is a wrap around necessary
   if (w > r) {
     // Non wrapping case
-    info.len_lin = tu_min16(n, w - r);            // Limit to required length
-    info.len_wrap = 0;
+    info->len_lin = tu_min16(n, w - r);            // Limit to required length
+    info->len_wrap = 0;
+    info->ptr_wrap = NULL;
   }
   else
   {
-    info.len_lin = tu_min16(n, f->depth - r);     // Also the case if FIFO was full
-    info.len_wrap = n-info.len_lin;               // n was already limited to what is available
+    info->len_lin = tu_min16(n, f->depth - r);   // Also the case if FIFO was full
+    info->len_wrap = n-info->len_lin;
+    info->ptr_wrap = f->buffer;
   }
 
-  return info;
+  return;
 }
 
 /******************************************************************************/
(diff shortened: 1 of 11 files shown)


HiFiPhile:
> Change get_linear_read/write_info() to return linear and wrapped part at once

Oh thanks you are quick :sweat_smile:, I think I would write it this weekend...

src/common/tusb_fifo.c
```diff
@@ -933,23 +933,26 @@ uint16_t tu_fifo_get_linear_read_info(tu_fifo_t *f, uint16_t offset, void **ptr,
   w = get_relative_pointer(f, w, 0);
   r = get_relative_pointer(f, r, offset);
 
+  // Copy pointer to buffer to start reading from
+  *ptr_lin = &f->buffer[r];
+  *ptr_wrap = f->buffer;
```
HiFiPhile, line 938: You can add a zero check of `len_wrap` and `ptr_wrap` before assignment if you like, so they can leave to 0 if not used.

PanRe: I thought it is faster if we simply avoid this check. Anyway, if `len_wrap=0`, then ptr_wrap is not valid.

PanRe:
@hathach Could you have a look into this include hassle with `pico.h`? I don't know why this error comes up...

HiFiPhile:
Seems there is a depenency loop:
osal -> osal_pico -> tusb_fifo -> osal

PanRe:
Ah yes again. This already was once an issue, but @hathach took care of it. I don't know how exactly. I guess this came up again somehow as i needed to include `tusb_fifo.h` within `audio_device.h`.

src/common/tusb_fifo.h
```diff
@@ -80,6 +80,14 @@ typedef struct
 
 } tu_fifo_t;
 
+typedef struct
+{
+  uint16_t len_lin                      ; ///< linear length in item size
+  uint16_t len_wrap                     ; ///< wrapped length in item size
+  void * ptr_lin                        ; ///< linear part start pointer
+  void * ptr_wrap                       ; ///< wrapped part start pointer
+} tu_fifo_linear_wr_info;
```
hathach, line 89: nice idea, It could simplify the application usage for these direct buffer access. Though the name is not quite right since it is used for both read/write. Please change it to `tu_fifo_buffer_info_t` . I think the member name should be changed as well to simply buffer1 and buffer2, the lin/wrap will cause user question for those that aren't familiar with fifo implementation. 

```C
typedef struct
{
  uint16_t len1                      ; ///< length in item size of first buffer
  uint16_t len2                     ; ///< length in item size of second buffer (wrapped)
  void * buffer1                        ; ///< buffer1 pointer
  void * buffer2                       ; ///< buffer2 pointer (wrapped)
} tu_fifo_buffer_info_t;
```

src/common/tusb_fifo.h
```diff
@@ -134,8 +142,10 @@ void     tu_fifo_advance_read_pointer   (tu_fifo_t *f, uint16_t n);
 // This functions deliver a pointer to start reading/writing from/to and a valid linear length along which no wrap occurs.
 // In case not all of your data is available within one read/write, update the read/write pointer by
 // tu_fifo_advance_read_pointer()/tu_fifo_advance_write_pointer and conduct a second read/write operation
-uint16_t tu_fifo_get_linear_read_info   (tu_fifo_t *f, uint16_t offset, void **ptr, uint16_t n);
-uint16_t tu_fifo_get_linear_write_info  (tu_fifo_t *f, uint16_t offset, void **ptr, uint16_t n);
+// TODO - update comments
+
+tu_fifo_linear_wr_info tu_fifo_get_linear_read_info(tu_fifo_t *f, uint16_t offset, uint16_t n);
```
hathach, line 147: 1. Do we ever need to use `offset`. If not I think we can drop that parameter for a simpler code.
2. I think we could just drop `n` as well, just return len1 and len2 as it is available from buffer. This API would be like query the current buffer state of the fifo. And we could rename function to  `tu_fifo_get_read_info()` or `tu_fifo_get_read_buffer_info()`
3. As general of thumb, we shouldn't return a struct object (> 4 bytes), it should be passed from application to fill out data. Function return can be `bool`. Return false when there is not 

Similar comment apply to `tu_fifo_get_linear_write_info`

hathach (commented):
thanks @PanRe for another great PR. Sorry the in cross include between pico and fifo is getting out of control. I temporarily fixed it by alternating the include order a bit and push the update. Will definitely fix it more properly later on with an overall include revise. 

Except from that, I have an a couple of feedback, that I think we could simplify the API for the read/write buffer info. Let me know if that makes sense to you.

src/common/tusb_fifo.h
PanRe: Mh, the name `buffer_info` is a bit too non-speaking don't you mean? `linear_wr_info` stands for linear write/read info.

PanRe: Ah i get your idea from below! Ok i rename it!

src/common/tusb_fifo.h
PanRe: I would go with `tu_fifo_get_read_info()`.
To 3: Where does this restriction come from? And return `false` when there is no place to write to or nothing to read from?

PanRe: I don't know if offset will be needed or when the offset came up the first time... we can remove it completely from the `fifo.c`, i.e. also from the write and read functions.

HiFiPhile: > To 3: Where does this restriction come from?

Depends on the compiler, sometimes it will make a copy of `tu_fifo_buffer_info_t info`.

hathach: > I would go with `tu_fifo_get_read_info()`.
> To 3: Where does this restriction come from? And return `false` when there is no place to write to or nothing to read from?

Yeah right, I was mixing thought, I think we can have it to return void if both offset and n are dropped from the argument as 1 and 2

hathach: > I don't know if offset will be needed or when the offset came up the first time... we can remove it completely from the `fifo.c`, i.e. also from the write and read functions.

Yeah, I think offset is is kind of needed for peek_at() initialy, which is trivial with initial implementation. Though If needed, we can just removed the peek_at(offset), and just implement peek() to look at current index only. peek_at random offset is not used by any class driver/application so far. So we don't really need it :) and clean API is much better needed.

PanRe: Ok i have but one question, the first time i implemented the `get_read_info()` i did it without struct but with poiners `get_read_info(* uint16_t len_1, * uint16_t len_2, ...)`. However, the CI compilers complained, that `len_1` etc. were potentially uninitialized and aborted with an error. To fix this, i came up with the struct at the very end. Now, if i change it to `get_read_info(* buffer_info into)`, then i bet the compiler complains again. This would force the user to always init the struct with zeros or so. Is there a way to tell the compiler that the struct will hold valid values in any case?

hathach: hmm, that is weird, did you have any branch in the function that doesn't assign value for the struct memeber. Like invalid short return, it would cause gcc to trigger possible path for uninitialized value.

PanRe: Me yes... in case there was nothing to read i set `len_1 = 0` and left the rest uninitialized to save time.
Nevermind, i change it to the pointer version to the struct if this is your desired design. I guess its not worth spending more time on this ^^

hathach: Yes, please do. I do prefer it that way .

PanRe:
OK all set up as desired!

src/common/tusb_fifo.c
```diff
@@ -923,103 +911,90 @@ uint16_t tu_fifo_get_linear_read_info(tu_fifo_t *f, uint16_t offset, void **ptr,
   }
 
   // Skip beginning of buffer
```
hathach, line 913: `Skip beginning of buffer` does not seem to fit anymore. It should be `Check If the fifo is empty`.

src/common/tusb_fifo.c
```diff
@@ -923,103 +911,90 @@ uint16_t tu_fifo_get_linear_read_info(tu_fifo_t *f, uint16_t offset, void **ptr,
...
+   TAKE CARE TO NOT OVERFLOW THE BUFFER MORE THAN TWO TIMES THE FIFO DEPTH - IT CAN NOT RECOVERE OTHERWISE!
    @param[in]       f
                     Pointer to FIFO
-   @param[in]       offset
-                    Number of ITEMS to ignore before start writing
-   @param[out]      **ptr
-                    Pointer to start writing to
+   @param[out]      *info
+                    Pointer to struct which holds the desired infos
    @param[in]       n
```
hathach, line 960: comment for `n` should be removed

hathach (approved):
superb ! Thank you for your effort, great PR as usual. the removal of offset make the whole fifo API much cleaner any easier to read/understand. FIFO should be supported offset (random) in the first place. It is my bad to introduce such operation with peek (since I thought it could be handy). Should user need to peek randomly, the new get_read_info() can be used to do it. I have a couple of minor clean up, along with a couple of unit test for new read/write info will push and merge it asap.

PanRe:
Thanks a lot! :)

hathach:
> Thanks a lot! :)

I am the one to say thanks :+1:

PR #808, opened 2021-04-26
Lpc55 port1 hs

hathach:
**Describe the PR**
This PR implement Port1 Highspeed for lpc55 and lpc546xx/lpc540xx (not tested on actual hardware). There is a few gotcha 
- LPC55 highspeed usb can only write/access to USB RAM (0x40100000),  __attribute__((section("m_usb_global")))
- USB RAM does not support unaligned access despite it is cortex M4. Extra macro `TUP_MCU_STRICT_ALIGN`  is used to overwrite ARCH one, extra effort is also needed to prevent compiler to generate optimized code that result in unaligned access.
- Full and Highspeed command/status words for buffer offset & nbytes are different
- Lots of other works for multiple ports and mixed full/high speed support. 
- Tested worked with cdc_msc, hid_composite etc .. advanced class such as net and audio is not tested. ISO transfer is not yet supported.

@nxf58843 @gsteiert @UweBonnes may be interested.


UweBonnes:
" make BOARD=mcu_link  flash" enumerates successfull. Thanks!
"make BOARD=mcu_link SPEED=full" gets overwritten by hw/bsp/lpc55/family.mk.

hathach:
> " make BOARD=mcu_link flash" enumerates successfull. Thanks!
> "make BOARD=mcu_link SPEED=full" gets overwritten by hw/bsp/lpc55/family.mk.

`make BOARD=mcu_link PORT=0` to select fullspeed port0, unlike stm32, the lpc55 port1 always has HS PHY so it is always operated at highspeed.

UweBonnes:
The mcu_link has only port 1 connected.

hathach:
then just use it with highspeed, why would you want to force it to operate at fullspeed.

UweBonnes:
So the line "SPEED ?= high" in board.mk is superfluous. Should I do another PR, will you correct or does it not matter?

hathach:
yeah, it is not needed, no worry, I will may a push for to clean it up.

PR #820, opened 2021-05-07
Update main.c

corvus-ossifragus:
Added new RTOS names to the name string array in print_greeting() in the host example program. 

Before adding the new names, the program seemed to hang when a new RTOS was indexed in CFG_TUSB_OS.


hathach (approved):
look good thanks for your PR

PR #821, opened 2021-05-07
Update hcd_rp2040.c

corvus-ossifragus:
Remove reference to the deprecated "num" and "in" members  of struct hw_endpoint which still existed in an assert statement that would break DEBUG builds.


hathach (approved):
thanks for the PR, look good

Issue #709, opened 2021-03-09
RP2040 Host Stack fails with HID devices


ristohelkio:
**Set up**
 - **Board**   : Raspberry Pi Pico with UART tracing via Raspberry Pi 4
 - **Firmware**: /examples/host/cdc_msc_hid
 - **Keyboard**: Generic wireless keypad (Exibel - Provision Technology, Inc)
 - **HUB-keyboard**: Apple

**Describe the bug**
When connecting a keyboard to Pico the process fails at the point where the stack is trying to read the full Device Descriptor. If the Device Descriptor fits into one single packet (bMaxPacketSize0 == 64) the reading is successful. When bMaxPacketSize0 == 8 the stack fails.

**To reproduce**
Connect a keyboard to pico

**Log**
Exibel keypad. When the stack is trying to read an 18 bytes long Device Descriptor it fails. Result is PANIC with 'ep 0 in was already available'
[trace-exibel-keypad.txt](https://github.com/hathach/tinyusb/files/6109326/trace-exibel-keypad.txt)

Additional logs added in file src/portable/raspberrypi/rp2040/rp2040_usb.c (pico_trace replaced with TU_LOG2). The stack seems to receive 8 + 8 + 2 bytes but results in PANIC with 'Data Seq Error ':
[trace-exibel-keypad-detailed.txt](https://github.com/hathach/tinyusb/files/6109385/trace-exibel-keypad-detailed.txt)

In comparison a trace with an Apple keyboard-hub. The hub is connecting properly but when trying to read the Device Descriptor of the keyboard behind the hub, the stack fails. Hub bMaxPacketSize0 is 64 which seems to do the trick.
[trace-apple-keyboard-hub.txt](https://github.com/hathach/tinyusb/files/6109426/trace-apple-keyboard-hub.txt)

hathach:
thanks for the issue, make sure you are running the example with latest master. Though I am not surprized that it doesn't work. It is not your setup or device, it is just the host stack isn't received much of developing time. So far I have reworked usbh and host msc, will do host hid next when I find time.

hathach:
probably be fixed by  raspberrypi/tinyusb#7 based approach.

PR #822, opened 2021-05-07
Issue #709 

corvus-ossifragus:
Revert a change that was made to rp2040_usb.c that causes slow speed devices to not get configured by the host. It seems likely it is the same as Issue #709.

A low-speed mouse that works with the old TinyUSB source code in the Pico SDK  does not work with the latest master from the main repo. This pull request reverts the change that causes this problem. After this line is changed, the mouse works with either version of code.

The reverted change limits the entire data transfer to only 8 bytes for low-speed devices, not just the data packets. So, when the higher level logic requests the 18 byte Device Descriptor, this line of code reduces it to 8 bytes. Lower-level code requests an 8 byte IN packet and signals Last Transfer to the USB controller. Higher-level code may then try to request the remaining 10 bytes, but whatever happens results in a STALL from the device and a DATA SEQ error from the USB controller.

### DIFF
--- a/src/portable/raspberrypi/rp2040/rp2040_usb.c
+++ b/src/portable/raspberrypi/rp2040/rp2040_usb.c
@@ -161,7 +161,10 @@ void _hw_endpoint_xfer_start(struct hw_endpoint *ep, uint8_t *buffer, uint16_t t
     // Fill in info now that we're kicking off the hw
     ep->total_len = total_len;
     ep->len = 0;
-    ep->transfer_size = tu_min16(total_len, ep->wMaxPacketSize);
+    // The following limits the TOTAL transfer size to 64 bytes.
+    // This is not the maximum size of the individual data packets,
+    // so it does not need to be limited to 8 for low-speed devices.
+    ep->transfer_size = total_len > 64 ? 64 : total_len;
     ep->active = true;
     ep->user_buf = buffer;
 #if TUSB_OPT_HOST_ENABLED


corvus-ossifragus:
I was thinking about this issue, and it seems like a low-speed device is very unlikely to be doing ISO transfers. So, it may not be pretty, but maybe something like this until a better solution is found:
 if (TUSB_SPEED_LOW==tuh_device_get_speed(ep->dev_addr))
 {
   ep->transfer_size = total_len > 64 ? 64 : total_len;
 }
 else
 {
   ep->transfer_size = tu_min16(total_len, ep->wMaxPacketSize);
 }
​

May 8, 2021, 4:19 AM by ***@***.***:

pigrew:
> I was thinking about this issue, and it seems like a low-speed device is very unlikely to be doing ISO transfers.

I think isochronous transfers are not allowed on low-speed busses:

USB 2.0 spec 5.12.6:
> These requirements limit isochronous transfers to full-speed and high-speed
devices only, because low-speed devices do not see SOFs on the bus.

hathach:
Low speed device only support interrupt transfer, however as mentioned above this is out of specs from rp2040 hcd. I would like to look at the datasheet and test it out before making the changes. The reason is doing 64 bytes transfer on 8-bytes buffer can cause buffer overrun and could corrupt other part of memory.

hathach (commented):
thanks for your PR, however, after giving a few tries, look like the proposal in https://github.com/raspberrypi/tinyusb/pull/7 does a correct job to fix the root issue with handling 3rd packet issue with transfer. Although that proposal may need some changes, e.g wrapping the walkround with host mode only since those can affect the device stack. However, I think it is indeed the right direction to look at.

hathach:
closed in prefer of  raspberrypi/tinyusb#7 based approach.

hathach:
UPDATE: I spend a bit more time testing, and make changes to add max(64, mps). It works but the issue with incorrect buf status bit still exists. There is also conflict with submitting multiple transferring as well in host mode. I will need to spend more time with rp2040 hcd. You could follow that branch for update.

https://github.com/hathach/tinyusb/commit/164778a71682559670cc2549565b4d8d5f021fa9

PR #824, opened 2021-05-09
Update main.c

corvus-ossifragus:
Fix code in host example where HID keyboard was being stuffed into the HID mouse buffer.

I haven't tested this, but it seems obviously wrong. The HID mouse report buffer was being used for HID keyboard reports.


hathach (approved):
thank for spotting the typo

PR #827, opened 2021-05-10
audio_device: Fix build error

kasjer:
ep_in_as_intf_num was incorrectly used to access out interface
which is defined for 'in' interface.
Code related to 'out' endpoint should use ep_out_as_intf_num instead.

**Describe the PR**
Speaker only devices would not build due to incorrect variable usage.
Probably in/out devices would use incorrect variable even if there is no build error.

**Additional context**


PanRe:
Yes thanks for the catch! This is a mistake! Thank you for fixing!!

PR #828, opened 2021-05-10
Fix ep struct trace statements

corvus-ossifragus:
Updated some pico_trace statements that were using deprecated elements of the hw_endpoint structure.


hathach (approved):
ah thanks again,  I don't use pico_trace much and totally miss this.

PR #829, opened 2021-05-11
Allow build for NRF5340 MCU

kasjer:
**Describe the PR**
There is not definition **NRF_USBD** in **NRF5340** headers.
There is however **NRF_USBD** definition in both cases
NRF52 has:
```c
#define NRF_USBD                    ((NRF_USBD_Type*)          NRF_USBD_BASE)
```
and NRF5340 has:
```c
#define NRF_USBD         NRF_USBD_S
#define NRF_USBD_S                  ((NRF_USBD_Type*)          NRF_USBD_S_BASE)
```
When common definition is used file can be build for both families.
**Additional context**

### DIFF
--- a/src/portable/nordic/nrf5x/dcd_nrf5x.c
+++ b/src/portable/nordic/nrf5x/dcd_nrf5x.c
@@ -833,9 +833,8 @@ void tusb_hal_nrf_power_event (uint32_t event)
         NRF_USBD->EVENTCAUSE = USBD_EVENTCAUSE_READY_Msk;
         __ISB(); __DSB(); // for sync
 
-        /* Enable the peripheral */
+#ifdef NRF52_SERIES
         // ERRATA 171, 187, 166
-
         if ( nrfx_usbd_errata_187() )
         {
           // CRITICAL_REGION_ENTER();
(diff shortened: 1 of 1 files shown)


kasjer:
> we need to cast the NRF_USBD to `uint8_t*` first before doing arithmetic operator on it. Though It probably cause gcc to throw warning when casting back to uint32_t* (cast that increases alignment). I think it probably better to have ifdef to unify the base name.
> 
> Note: I notice that the base address for secure and non-secure is different as well, one thing we need to keep it in mind later on.

Missing cast fixed.

For now errata 166 does not apply to NRF5340 so whatever you choose will work, PR just removes build error that popped up in NRF5340 build.

hathach:
yeah, just quickly check the 53 errata, not only 166 but all of other errata is not needed for 5340 as well. I think it would be better if we could wrap these errata walkaround with `#ifdef NRF52_SERIES`. I have push to update the PR. Please let me know if it work on your hardware.

I actually have an PDK 5340 here, but too bad, it doesn't have USB connector populated on the board (!!). I will try to solder one and test it out later on. Though It seems like 5340 revA doesn't have working USB, which is probably the reason for that.

kasjer:
Erratas that you ifdefed were not applied before and USB worked on 5430 with your changes it will still work provided that code builds.

PR #831, opened 2021-05-12
Add interface set protocol function

corvus-ossifragus:
Adds a function so an application can set the mouse or keyboard report format to the standardized Boot format. HID devices that support the Boot format do not always use it by default.


hathach (commented):
Thank for your PR, but host hid is being reworked, this PR need to wait to see if it still fit after that is done.

hathach:
close since it is implemented by  #834

PR #837, opened 2021-05-18
Fix 835 hid cleanup

### LINKED ISSUE #835: Insistent signature of tud_hid_boot_mode_cb()
The signature of `tud_hid_boot_mode_cb()` is

```c
void tud_hid_boot_mode_cb(uint8_t itf, uint8_t boot_mode);
```
However, the `boot_mode` argument is a `bool` taken from:
https://github.com/hathach/tinyusb/blob/2d15e1183064a35cc646a46bddf2b82915114cc8/src/class/hid/hid_device.c#L41-L55
`boot_mode` is passed using this code:
https://github.com/hathach/tinyusb/blob/2d15e1183064a35cc646a46bddf2b82915114cc8/src/class/hid/hid_device.c#L338
So should the signature instead be this?
```c
void tud_hid_boot_mode_cb(uint8_t itf, bool boot_mode);
```
(I was confused for a while due to the `1-` code in the previous routines which flip the sense of the value.)

Obviously this is very minor.

hathach:
**Describe the PR**
Fix #835  and more
- add tud_hid_n_interface_protocol()
- rename tud_hid_n_boot_mode() to  tud_hid_n_get_protocol()
- rename tud_hid_boot_mode_cb() to tud_hid_set_protocol_cb()
- use HID spec protocol value for get/set_protocol() instead of bool
- add HID_PROTOCOL_BOOT/REPORT to avoid magic number 0,1
- rename HID_PROTOCOL_NONE/KEYBOARD/MOUSE to HID_ITF_PROTOCOL_ to avoid
confusion
- rename `itf` to `instance` in multiple API to have a better meaning.

### DIFF
--- a/examples/device/hid_composite/src/usb_descriptors.c
+++ b/examples/device/hid_composite/src/usb_descriptors.c
@@ -108,7 +108,7 @@ uint8_t const desc_configuration[] =
   TUD_CONFIG_DESCRIPTOR(1, ITF_NUM_TOTAL, 0, CONFIG_TOTAL_LEN, TUSB_DESC_CONFIG_ATT_REMOTE_WAKEUP, 100),
 
   // Interface number, string index, protocol, report descriptor len, EP In address, size & polling interval
-  TUD_HID_DESCRIPTOR(ITF_NUM_HID, 0, HID_PROTOCOL_NONE, sizeof(desc_hid_report), EPNUM_HID, CFG_TUD_HID_EP_BUFSIZE, 5)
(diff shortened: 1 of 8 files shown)


hathach:
@dhalbert the value return by API get_protocol and in set_protocol is now follow HID spec, either as HID_PROTOCOL_BOOT/REPORT. Although the boot_mode() is easy for new user, the invert logic cause confusion for people that are familiar with HID

image

dhalbert:
Thanks! This makes it consistent internally and externally, and having the names instead of a bool is great.

PR #841, opened 2021-05-20
fix uac2_headset example

noodlefighter:
**Describe the PR**

type of bRequest should be uint8_t, not enumerate...

size of a enum is dynamic, it will break the program in some complier(such as xtensa-esp32s2-elf-gcc), error logging like: 

```
Get request not handled, entity = 0, selector = 8, request = 131330
```


hathach (approved):
yup, you are spot on, Thank you very much for the PR.

PR #849, opened 2021-05-26
fix uac2_headset example

noodlefighter:
**Describe the PR**
fix uac2_headset example

tud_audio_rx_done_cb() is departed, replace with
tud_audio_rx_done_pre_read_cb()

**Additional context**
test OK on esp32s2(with patch #552 , and run `tud_task()` in a separate task)

test method:
generate sine wave to uac device with python script:

```
# refer: https://pythonaudiosynthesisbasics.com/
import numpy as np
import sounddevice as sd

sample_rate = 44100
duration = 120.0
frequency = 440.0

x = np.linspace(0, duration * 2 * np.pi, int(duration * sample_rate))
sinewave_data = np.sin(frequency * x)

# best to attenuate it before playing, they can get very loud
sinewave_data = sinewave_data * 0.3

while True:
    sd.play(sinewave_data, sample_rate)
    sd.wait()
```

in audacity, record:

image

### DIFF
--- a/examples/device/uac2_headset/src/main.c
+++ b/examples/device/uac2_headset/src/main.c
@@ -332,13 +332,14 @@ bool tud_audio_set_itf_cb(uint8_t rhport, tusb_control_request_t const * p_reque
   return true;
 }
 
-bool tud_audio_rx_done_cb(uint8_t rhport, uint8_t *buffer, uint16_t buf_size)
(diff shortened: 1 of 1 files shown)


hathach (approved):
superb! thank you very much. I think the audio got changes and this example does not make use of latest API.

Milili-Tech:
@noodlefighter I didn't follow up tinyusb for long time, it's a great work you've done! I didn't test a real sine wave when tried esp32-s2, can you also raise a PR to submit uac2_headset_freertos? So it will be easy to test for everyone.

PR #859, opened 2021-05-29
Adaptations for Renesas CCRX toolchain and Rx72N controller performed

Wini-Buh:
As discussed in #785  this a draft PR  for the port of tinyUSB to the CCRX toolchain and the RX72N device from Renesas.

It is still not complete and only the CDC device class is fully ported. But I guess it is good enough for a discussion.

The following points had to be overcome for porting it to the CCRX toolchain (and the Rx72N used in my project):
- lack of the __attribute__ directive
     - overcome the missing weak directive
     - overcome of the missing pack directive (used a pragma directive instead)
     - overcome of a missing alignment directive
- endian order (use of big endian order instead of little endian order)
- order of the bit fields (left or right ordering)

Because the FreeRTOS implementation only  support the "...Static" API and I use the "non-Static" API and a slightly modified task switching mechanism, I had to add/change some code. Please ignore that.

BTW: I have setup a development environment to build a sample project (with the Rx72N) with following configurations:
  - CCRX toolchain in little endian order
  - CCRX toolchain in big endian order
  - GCC toolchain in little endian order

With this environment, I can easily check if all configuration produce the same results

### DIFF
--- a/src/class/cdc/cdc.h
+++ b/src/class/cdc/cdc.h
@@ -245,7 +245,7 @@ typedef struct TU_ATTR_PACKED
   uint8_t bDescriptorSubType              ;\
   uint8_t bControlInterface               ;\
   uint8_t bSubordinateInterface[no_slave] ;\
-}                                          \
+}
 
 /// Country Selection Functional Descriptor (Communication Interface)
 typedef struct TU_ATTR_PACKED
@@ -264,7 +264,7 @@ typedef struct TU_ATTR_PACKED
   uint8_t bDescriptorSubType        ;\
   uint8_t iCountryCodeRelDate       ;\
   uint16_t wCountryCode[no_country] ;\
-}                                    \
+}
 
 //--------------------------------------------------------------------+
 // PUBLIC SWITCHED TELEPHONE NETWORK (PSTN) SUBCLASS
--- a/src/class/cdc/cdc.h
+++ b/src/class/cdc/cdc.h
@@ -215,7 +215,9 @@ typedef enum
 // Class Specific Functional Descriptor (Communication Interface)
 //--------------------------------------------------------------------+
 
-TU_ATTR_PACKED_BEGIN  // Start of definition of packed structs (used by the CCRX toolchain)
+// Start of all packed definitions for compiler without per-type packed
+TU_ATTR_PACKED_BEGIN
+TU_ATTR_BIT_FIELD_ORDER_BEGIN
 
 /// Header Functional Descriptor (Communication Interface)
 typedef struct TU_ATTR_PACKED
@@ -236,10 +238,7 @@ typedef struct TU_ATTR_PACKED
   uint8_t bSubordinateInterface    ; ///< Array of Interface number of Data Interface
 }cdc_desc_func_union_t;
 
-TU_ATTR_PACKED_END  // End of definition of packed structs (used by the CCRX toolchain)
-
 #define cdc_desc_func_union_n_t(no_slave)\
- TU_ATTR_PACKED_BEGIN                      \
  struct TU_ATTR_PACKED {                   \
   uint8_t bLength                         ;\
   uint8_t bDescriptorType                 ;\
@@ -247,10 +246,6 @@ TU_ATTR_PACKED_END  // End of definition of packed structs (used by the CCRX too
   uint8_t bControlInterface               ;\
   uint8_t bSubordinateInterface[no_slave] ;\
 }                                          \
-TU_ATTR_PACKED_END
-
-
-TU_ATTR_PACKED_BEGIN  // Start of definition of packed structs (used by the CCRX toolchain)
 
 /// Country Selection Functional Descriptor (Communication Interface)
 typedef struct TU_ATTR_PACKED
@@ -262,10 +257,7 @@ typedef struct TU_ATTR_PACKED
   uint16_t wCountryCode       ; ///< Country code in the format as defined in [ISO3166], release date as specified inoffset 3 for the first supported country.
 }cdc_desc_func_country_selection_t;
 
-TU_ATTR_PACKED_END  // End of definition of packed structs (used by the CCRX toolchain)
-
 #define cdc_desc_func_country_selection_n_t(no_country) \
-  TU_ATTR_PACKED_BEGIN               \
   struct TU_ATTR_PACKED {            \
   uint8_t bLength                   ;\
   uint8_t bDescriptorType           ;\
@@ -273,17 +265,13 @@ TU_ATTR_PACKED_END  // End of definition of packed structs (used by the CCRX too
   uint8_t iCountryCodeRelDate       ;\
   uint16_t wCountryCode[no_country] ;\
 }                                    \
-TU_ATTR_PACKED_END
 
 //--------------------------------------------------------------------+
 // PUBLIC SWITCHED TELEPHONE NETWORK (PSTN) SUBCLASS
 //--------------------------------------------------------------------+
 
-TU_ATTR_PACKED_BEGIN  // Start of definition of packed structs (used by the CCRX toolchain)
-
 /// \brief Call Management Functional Descriptor
 /// \details This functional descriptor describes the processing of calls for the Communications Class interface.
-TU_BIT_FIELD_ORDER_BEGIN
 typedef struct TU_ATTR_PACKED
 {
   uint8_t bLength            ; ///< Size of this descriptor in bytes.
@@ -298,10 +286,7 @@ typedef struct TU_ATTR_PACKED
 
   uint8_t bDataInterface;
 }cdc_desc_func_call_management_t;
-TU_BIT_FIELD_ORDER_END
 
-
-TU_BIT_FIELD_ORDER_BEGIN
 typedef struct TU_ATTR_PACKED
 {
   uint8_t support_comm_request                    : 1; ///< Device supports the request combination of Set_Comm_Feature, Clear_Comm_Feature, and Get_Comm_Feature.
@@ -310,12 +295,11 @@ typedef struct TU_ATTR_PACKED
   uint8_t support_notification_network_connection : 1; ///< Device supports the notification Network_Connection.
   uint8_t TU_RESERVED                             : 4;
 }cdc_acm_capability_t;
-TU_BIT_FIELD_ORDER_END
 
 TU_VERIFY_STATIC(sizeof(cdc_acm_capability_t) == 1, "mostly problem with compiler");
 
-/// \brief Abstract Control Management Functional Descriptor
-/// \details This functional descriptor describes the commands supported by by the Communications Class interface with SubClass code of \ref CDC_COMM_SUBCLASS_ABSTRACT_CONTROL_MODEL
+/// Abstract Control Management Functional Descriptor
+/// This functional descriptor describes the commands supported by by the Communications Class interface with SubClass code of \ref CDC_COMM_SUBCLASS_ABSTRACT_CONTROL_MODEL
 typedef struct TU_ATTR_PACKED
 {
   uint8_t bLength                  ; ///< Size of this descriptor in bytes.
@@ -326,7 +310,6 @@ typedef struct TU_ATTR_PACKED
 
 /// \brief Direct Line Management Functional Descriptor
 /// \details This functional descriptor describes the commands supported by the Communications Class interface with SubClass code of \ref CDC_FUNC_DESC_DIRECT_LINE_MANAGEMENT
-TU_BIT_FIELD_ORDER_BEGIN
 typedef struct TU_ATTR_PACKED
 {
   uint8_t bLength            ; ///< Size of this descriptor in bytes.
@@ -339,7 +322,6 @@ typedef struct TU_ATTR_PACKED
     uint8_t TU_RESERVED           : 5;
   } bmCapabilities;
 }cdc_desc_func_direct_line_management_t;
-TU_BIT_FIELD_ORDER_END
 
 /// \brief Telephone Ringer Functional Descriptor
 /// \details The Telephone Ringer functional descriptor describes the ringer capabilities supported by the Communications Class interface,
@@ -356,7 +338,6 @@ typedef struct TU_ATTR_PACKED
 /// \brief Telephone Operational Modes Functional Descriptor
 /// \details The Telephone Operational Modes functional descriptor describes the operational modes supported by
 /// the Communications Class interface, with the SubClass code of \ref CDC_COMM_SUBCLASS_TELEPHONE_CONTROL_MODEL
-TU_BIT_FIELD_ORDER_BEGIN
 typedef struct TU_ATTR_PACKED
 {
   uint8_t bLength            ; ///< Size of this descriptor in bytes.
@@ -369,12 +350,10 @@ typedef struct TU_ATTR_PACKED
     uint8_t TU_RESERVED           : 5;
   } bmCapabilities;
 }cdc_desc_func_telephone_operational_modes_t;
-TU_BIT_FIELD_ORDER_END
 
 /// \brief Telephone Call and Line State Reporting Capabilities Descriptor
 /// \details The Telephone Call and Line State Reporting Capabilities functional descriptor describes the abilities of a
 /// telephone device to report optional call and line states.
-TU_BIT_FIELD_ORDER_BEGIN
 typedef struct TU_ATTR_PACKED
 {
   uint8_t bLength            ; ///< Size of this descriptor in bytes.
@@ -390,8 +369,8 @@ typedef struct TU_ATTR_PACKED
     uint32_t TU_RESERVED            : 26;
   } bmCapabilities;
 }cdc_desc_func_telephone_call_state_reporting_capabilities_t;
-TU_BIT_FIELD_ORDER_END
 
+// TODO remove
 static inline uint8_t cdc_functional_desc_typeof(uint8_t const * p_desc)
 {
   return p_desc[2];
@@ -410,21 +389,17 @@ typedef struct TU_ATTR_PACKED
 
 TU_VERIFY_STATIC(sizeof(cdc_line_coding_t) == 7, "size is not correct");
 
-TU_BIT_FIELD_ORDER_BEGIN
 typedef struct TU_ATTR_PACKED
 {
   uint16_t dte_is_present : 1; ///< Indicates to DCE if DTE is presentor not. This signal corresponds to V.24 signal 108/2 and RS-232 signal DTR.
   uint16_t half_duplex_carrier_control : 1;
   uint16_t : 14;
 } cdc_line_control_state_t;
-TU_BIT_FIELD_ORDER_END
-
-TU_ATTR_PACKED_END  // End of definition of packed structs (used by the CCRX toolchain)
-
 
 TU_VERIFY_STATIC(sizeof(cdc_line_control_state_t) == 2, "size is not correct");
 
-/** @} */
+TU_ATTR_PACKED_END  // End of all packed definitions
+TU_ATTR_BIT_FIELD_ORDER_END
 
 #ifdef __cplusplus
  }
--- a/src/class/cdc/cdc.h
+++ b/src/class/cdc/cdc.h
@@ -215,7 +215,7 @@ typedef enum
 // Class Specific Functional Descriptor (Communication Interface)
 //--------------------------------------------------------------------+
 
-TU_PACK_STRUCT_BEGIN  // Start of definition of packed structs (used by the CCRX toolchain)
+TU_ATTR_PACKED_BEGIN  // Start of definition of packed structs (used by the CCRX toolchain)
 
 /// Header Functional Descriptor (Communication Interface)
 typedef struct TU_ATTR_PACKED
@@ -236,10 +236,10 @@ typedef struct TU_ATTR_PACKED
   uint8_t bSubordinateInterface    ; ///< Array of Interface number of Data Interface
 }cdc_desc_func_union_t;
 
-TU_PACK_STRUCT_END  // End of definition of packed structs (used by the CCRX toolchain)
+TU_ATTR_PACKED_END  // End of definition of packed structs (used by the CCRX toolchain)
 
 #define cdc_desc_func_union_n_t(no_slave)\
- TU_PACK_STRUCT_BEGIN                      \
+ TU_ATTR_PACKED_BEGIN                      \
  struct TU_ATTR_PACKED {                   \
   uint8_t bLength                         ;\
   uint8_t bDescriptorType                 ;\
@@ -247,10 +247,10 @@ TU_PACK_STRUCT_END  // End of definition of packed structs (used by the CCRX too
   uint8_t bControlInterface               ;\
   uint8_t bSubordinateInterface[no_slave] ;\
 }                                          \
-TU_PACK_STRUCT_END
+TU_ATTR_PACKED_END
 
 
-TU_PACK_STRUCT_BEGIN  // Start of definition of packed structs (used by the CCRX toolchain)
+TU_ATTR_PACKED_BEGIN  // Start of definition of packed structs (used by the CCRX toolchain)
 
 /// Country Selection Functional Descriptor (Communication Interface)
 typedef struct TU_ATTR_PACKED
@@ -262,10 +262,10 @@ typedef struct TU_ATTR_PACKED
   uint16_t wCountryCode       ; ///< Country code in the format as defined in [ISO3166], release date as specified inoffset 3 for the first supported country.
 }cdc_desc_func_country_selection_t;
 
-TU_PACK_STRUCT_END  // End of definition of packed structs (used by the CCRX toolchain)
+TU_ATTR_PACKED_END  // End of definition of packed structs (used by the CCRX toolchain)
 
 #define cdc_desc_func_country_selection_n_t(no_country) \
-  TU_PACK_STRUCT_BEGIN               \
+  TU_ATTR_PACKED_BEGIN               \
   struct TU_ATTR_PACKED {            \
   uint8_t bLength                   ;\
   uint8_t bDescriptorType           ;\
@@ -273,13 +273,13 @@ TU_PACK_STRUCT_END  // End of definition of packed structs (used by the CCRX too
   uint8_t iCountryCodeRelDate       ;\
   uint16_t wCountryCode[no_country] ;\
 }                                    \
-TU_PACK_STRUCT_END
+TU_ATTR_PACKED_END
 
 //--------------------------------------------------------------------+
 // PUBLIC SWITCHED TELEPHONE NETWORK (PSTN) SUBCLASS
 //--------------------------------------------------------------------+
 
-TU_PACK_STRUCT_BEGIN  // Start of definition of packed structs (used by the CCRX toolchain)
+TU_ATTR_PACKED_BEGIN  // Start of definition of packed structs (used by the CCRX toolchain)
 
 /// \brief Call Management Functional Descriptor
 /// \details This functional descriptor describes the processing of calls for the Communications Class interface.
@@ -419,7 +419,7 @@ typedef struct TU_ATTR_PACKED
 } cdc_line_control_state_t;
 TU_BIT_FIELD_ORDER_END
 
-TU_PACK_STRUCT_END  // End of definition of packed structs (used by the CCRX toolchain)
+TU_ATTR_PACKED_END  // End of definition of packed structs (used by the CCRX toolchain)
 
 
 TU_VERIFY_STATIC(sizeof(cdc_line_control_state_t) == 2, "size is not correct");
--- a/src/class/cdc/cdc.h
+++ b/src/class/cdc/cdc.h
@@ -215,6 +215,8 @@ typedef enum
 // Class Specific Functional Descriptor (Communication Interface)
 //--------------------------------------------------------------------+
 
+TU_PACK_STRUCT_BEGIN  // Start of definition of packed structs (used by the CCRX toolchain)
+
 /// Header Functional Descriptor (Communication Interface)
 typedef struct TU_ATTR_PACKED
 {
@@ -234,14 +236,21 @@ typedef struct TU_ATTR_PACKED
   uint8_t bSubordinateInterface    ; ///< Array of Interface number of Data Interface
 }cdc_desc_func_union_t;
 
+TU_PACK_STRUCT_END  // End of definition of packed structs (used by the CCRX toolchain)
+
 #define cdc_desc_func_union_n_t(no_slave)\
- struct TU_ATTR_PACKED { \
+ TU_PACK_STRUCT_BEGIN                      \
+ struct TU_ATTR_PACKED {                   \
   uint8_t bLength                         ;\
   uint8_t bDescriptorType                 ;\
   uint8_t bDescriptorSubType              ;\
   uint8_t bControlInterface               ;\
   uint8_t bSubordinateInterface[no_slave] ;\
-}
+}                                          \
+TU_PACK_STRUCT_END
+
+
+TU_PACK_STRUCT_BEGIN  // Start of definition of packed structs (used by the CCRX toolchain)
 
 /// Country Selection Functional Descriptor (Communication Interface)
 typedef struct TU_ATTR_PACKED
@@ -253,21 +262,28 @@ typedef struct TU_ATTR_PACKED
   uint16_t wCountryCode       ; ///< Country code in the format as defined in [ISO3166], release date as specified inoffset 3 for the first supported country.
 }cdc_desc_func_country_selection_t;
 
+TU_PACK_STRUCT_END  // End of definition of packed structs (used by the CCRX toolchain)
+
 #define cdc_desc_func_country_selection_n_t(no_country) \
-  struct TU_ATTR_PACKED {\
+  TU_PACK_STRUCT_BEGIN               \
+  struct TU_ATTR_PACKED {            \
   uint8_t bLength                   ;\
   uint8_t bDescriptorType           ;\
   uint8_t bDescriptorSubType        ;\
   uint8_t iCountryCodeRelDate       ;\
   uint16_t wCountryCode[no_country] ;\
-}
+}                                    \
+TU_PACK_STRUCT_END
 
 //--------------------------------------------------------------------+
 // PUBLIC SWITCHED TELEPHONE NETWORK (PSTN) SUBCLASS
 //--------------------------------------------------------------------+
 
+TU_PACK_STRUCT_BEGIN  // Start of definition of packed structs (used by the CCRX toolchain)
+
 /// \brief Call Management Functional Descriptor
 /// \details This functional descriptor describes the processing of calls for the Communications Class interface.
+TU_BIT_FIELD_ORDER_BEGIN
 typedef struct TU_ATTR_PACKED
 {
   uint8_t bLength            ; ///< Size of this descriptor in bytes.
@@ -282,8 +298,10 @@ typedef struct TU_ATTR_PACKED
 
   uint8_t bDataInterface;
 }cdc_desc_func_call_management_t;
+TU_BIT_FIELD_ORDER_END
 
 
+TU_BIT_FIELD_ORDER_BEGIN
 typedef struct TU_ATTR_PACKED
 {
   uint8_t support_comm_request                    : 1; ///< Device supports the request combination of Set_Comm_Feature, Clear_Comm_Feature, and Get_Comm_Feature.
@@ -292,6 +310,7 @@ typedef struct TU_ATTR_PACKED
   uint8_t support_notification_network_connection : 1; ///< Device supports the notification Network_Connection.
   uint8_t TU_RESERVED                             : 4;
 }cdc_acm_capability_t;
+TU_BIT_FIELD_ORDER_END
 
 TU_VERIFY_STATIC(sizeof(cdc_acm_capability_t) == 1, "mostly problem with compiler");
 
@@ -307,6 +326,7 @@ typedef struct TU_ATTR_PACKED
 
 /// \brief Direct Line Management Functional Descriptor
 /// \details This functional descriptor describes the commands supported by the Communications Class interface with SubClass code of \ref CDC_FUNC_DESC_DIRECT_LINE_MANAGEMENT
+TU_BIT_FIELD_ORDER_BEGIN
 typedef struct TU_ATTR_PACKED
 {
   uint8_t bLength            ; ///< Size of this descriptor in bytes.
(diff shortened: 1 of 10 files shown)


hathach:
Thank you very much for your effort to port the ccrx toolchain. I am not familiar with renesas toolchain, does it has trial with limited functionality ( I don't need all the features), only enough so that I could test this out hand-on before merging and possibly help with tweaking the implementation.

PS: this PR come at pretty good time, my rx65n cloud kit board just come, it will come handy, I still need to get the board ported those. But it is probably similar to existing rx72n.

Wini-Buh:
Yes there is an evaluation version of the toolchain available. The trial period is the 60 days from the day the first build is executed after installing an evaluation edition. After the trial period exceeded the linkage is limit to 128 KB or smaller. 

Here is the link to toolchain webside : [CC-RX](https://www.renesas.com/eu/en/software-tool/cc-compiler-package-rx-family)
I use the version without IDE. 

The USB HW implementation of the Rx65N is equal to the one found on the Rx72N.

BTW: I also have an evaluation board with the Rx65N, the Envision Kit RPBRX65N.

hathach:
thanks, that is great to hear, most example is probably less than 128 KB limit. I will give it a try, yeah, the envision seems to be a good kit, though I think it doesn't route the USB to connector and requires additional add-on board from what I could understand.

Wini-Buh:
The Envision Kit HW only offers the host controller functionality (on a USB-A connector). On my Rx72N Envision Kit (which is equal as the Rx65N Envision Kit around the USB HW) I had to modify my board slightly, so it could be used as a function controller (remove the ISL6186USB Port Power Supply Controller and add a resistor). 

Unfortunately there is no CCRX toolchain version for Linux (I use a Windows VM on my Linux host). Only a Linux version of their e2studio IDE (this is an eclipse based IDE) is available, of course the GCC toolchain could be used with.

src/portable/renesas/usba/dcd_usba.c
```diff
@@ -91,39 +92,47 @@
...
     uint32_t     : 0;
   };
+  TU_BIT_FIELD_ORDER_END
 } pipe_state_t;
+TU_PACK_STRUCT_END
 
 typedef struct
 {
-  pipe_state_t pipe[9];
+  pipe_state_t pipe[10];
```
kkitayam, line 135: good catch:thumbsup:

hathach:
> The Envision Kit HW only offers the host controller functionality (on a USB-A connector). On my Rx72N Envision Kit (which is equal as the Rx65N Envision Kit around the USB HW) I had to modify my board slightly, so it could be used as a function controller (remove the ISL6186USB Port Power Supply Controller and add a resistor).
> 
> Unfortunately there is no CCRX toolchain version for Linux (I use a Windows VM on my Linux host). Only a Linux version of their e2studio IDE (this is an eclipse based IDE) is available, of course the GCC toolchain could be used with.

That is a pain, I will try to run it on vm as well, could take a bit for set up since my VM is not well installed for developing. Will get back to this asap.

src/portable/renesas/usba/dcd_usba.c
```diff
@@ -513,7 +566,13 @@ void dcd_init(uint8_t rhport)
   USB0.SYSCFG.BIT.DCFM = 0;
   USB0.SYSCFG.BIT.USBE = 1;
 
+  USB0.PHYSLEW.LONG = 0x5;
```
kkitayam, line 569: The `PHYSLEW` register is not present on RX63N. It should be wrapped with ifdef.

hathach:
still on my radar, just too busy with paid work. Please be patient.

src/common/tusb_compiler.h
```diff
@@ -123,6 +136,77 @@
...
+  #define ADD_WEAK_FUNC_TUD_MOUNT_CB                        0
+  #define ADD_WEAK_FUNC_TUD_UMOUNT_CB                       0
+  #define ADD_WEAK_FUNC_TUD_SUSPEND_CB                      0
+  #define ADD_WEAK_FUNC_TUD_RESUME_CB                       0
+  #define ADD_WEAK_FUNC_TUD_VENDOR_CONTROL_XFER_CB          0
+
+  // "Weak function" emulation defined in dcd.h
+  #define ADD_WEAK_FUNC_DCD_EDPT0_STATUS_COMPLETE  0
+  #define ADD_WEAK_FUNC_DCD_EDPT_CLOSE             0
+  #define ADD_WEAK_FUNC_DCD_EDPT_XFER_FIFO         0
```
hathach, line 207: This is troublesome, I don't think we could replicate this for all the stack, nor I could keep it consistent after this PR si merge. And explicitly requires user to manually enable each callback is too much. This is deal breaker,

src/portable/renesas/usba/dcd_usba.c
```diff
@@ -292,6 +310,13 @@ static void process_setup_packet(uint8_t rhport)
   setup_packet[1] = USB0.USBVAL;
   setup_packet[2] = USB0.USBINDX;
   setup_packet[3] = USB0.USBLENG;
+#if TU_BYTE_ORDER==TU_BIG_ENDIAN
+ #if defined(__CCRX__)
```
hathach, line 314: there is no need to check for compiler, if there is other one, it will be handled by the tu_le16toh().

hathach (commented):
Thank you very much for your effort putting into this PR. This is lots of work and testing, there is also fixes to existing rx driver as well. However, the weak function walk-around is not good enough and is a total deal breaker. It is totally not your fault but rather mine for choosing the weak function for user convenience and assume that most toolchain would support it.

I think we either have to come up with another walkaround for weak function or have to merge other changes first. And then think of way to address weak function as separate/follow up PR.

PS: I have a question, if instead of having weak as optional, how about we force all project compile with ccrx to define all the function. e.g even if it is not used/implement, application still have to declare the function. Then they are all "strong" and normal function, and that will work as others without having to have an walkround. It is like an strict requirement for ccrx, but it is not too difficult to do. Does this walkaround work in your opinion ?

src/class/cdc/cdc.h
```diff
@@ -252,38 +259,47 @@ typedef struct TU_ATTR_PACKED
...
 /// \brief Call Management Functional Descriptor
 /// \details This functional descriptor describes the processing of calls for the Communications Class interface.
+TU_PACK_STRUCT_BEGIN
 typedef struct TU_ATTR_PACKED
 {
   uint8_t bLength            ; ///< Size of this descriptor in bytes.
   uint8_t bDescriptorType    ; ///< Descriptor Type, must be Class-Specific
   uint8_t bDescriptorSubType ; ///< Descriptor SubType one of above CDC_FUCN_DESC_
 
+  TU_BIT_FIELD_ORDER_BEGIN
```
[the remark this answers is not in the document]
Wini-Buh: Putting the bit field ordering at the beginning and the end of a structure declaration is fine for me.
BTW: I'm not sure if the bit field order is really important on every bit field. I did not haven't a deeper look on it, if it is a local used bit field. I guess it is only important, if the bit field is transferred over USB or if it is mapped over a RAM/HW region  and not if it is used as a local status variable.

src/class/cdc/cdc.h
```diff
@@ -216,15 +216,18 @@ typedef enum
 //--------------------------------------------------------------------+
 
 /// Header Functional Descriptor (Communication Interface)
+TU_PACK_STRUCT_BEGIN
```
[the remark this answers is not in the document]
Wini-Buh: Yes, that is possible. The only exception, which comes to my mind at the moment, is that it could be a negative speed impact if accessing packed structures. Maybe we should check if the packed attribute is really required on all structs here ?

src/portable/renesas/usba/dcd_usba.c
Wini-Buh: Yes, I agree with you

src/class/cdc/cdc.h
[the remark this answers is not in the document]
hathach: we only group these packed structure that are consecutive to each other in tusb_types.h i.e we only group them together instead of declaring for each struct. So it is literally the same.

src/class/cdc/cdc.h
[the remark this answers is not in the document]
hathach: This only need to apply to those in usb specs, local vaiable should have no problem. However, the default right is the same as the default of gcc. So I think we are good to start with.

src/common/tusb_compiler.h
Wini-Buh: Ok, I understand that. The advantage of this would be, that the compiler does not generate code if the callback is defined as NULL.

hathach: It is not convenience to force user to manually enable this one by one. Also the modification is too much to support and maintain. There are dozens more used by other class driver and host stack in the future. For ccrx, without weak support, I guess this is good compromise.

Wini-Buh:
Of course, the easiest way would be to leave the weak macro for the CCRX port undefined. It would the be required by the user to implement all the functions which have the weak attribute in the GCC port. But I can not estimate does this have any impact on the execution speed (means calling empty callback function all the time) ?

If I see a negative impact on the project I'm using tinyUSB, then I might make a local change to overcome the problem for the time being.

hathach:
> Of course, the easiest way would be to leave the weak macro for the CCRX port undefined. It would the be required by the user to implement all the functions which have the weak attribute in the GCC port. But I can not estimate does this have any impact on the execution speed (means calling empty callback function all the time) ?
> 
> If I see a negative impact on the project I'm using tinyUSB, then I might make a local change to overcome the problem for the time being.

It should be that much, if that becomes an issue, you can actually can add `#define tud_cdc_rx_cb(x)  (*weak_func)` or something like that, could change the function prototype to function declaration, and you could has the `weak_func = NULL;` in your application. Though this is a hack, and may need some actual hand-on and testing to pull it off. If that all sound OK, please update the PR if possible. For now, I think just defined all the function for simplicity.

kkitayam:
If the compiler does not support a weak feature, why not call the default inline callback function specified by the compiler argument? I tried that the following experiment.

```c
#include <stdio.h>
static inline void default_cb(void){}
int main(void) {
    if (hello) {
        printf("ok\n");
        hello();
    }
    return 0;
}
```

I compiled the above file which name is hello.c with following command.
`ccrx.exe -change_message=information -output=src -isa=rxv2 -inline -define="hello=default_cb" hello.c`
Then I get the following result. It indicates the needless callback functions can be eliminated.
```
;RX Family C/C++ Compiler (V3.03.00 [20 Nov 2020])  08-Jun-2021 00:43:27

;*** CPU TYPE ***

;-ISA=RXV2

;*** COMMAND PARAMETER ***

;-output=src=hello.src
;-change_message=information
;-isa=rxv2
;-inline
;-define=hello=default_cb
;hello.c

		.glb	_main
		.glb	_printf
		.SECTION	P,CODE
_main:
		.STACK	_main=8
		SUB #04H, R0
		MOV.L #_L10, R14
		MOV.L R14, [R0]
		BSR _printf
		ADD #04H, R0
		MOV.L #00000000H, R1
		RTS
		.SECTION	L,ROMDATA,ALIGN=4
_L10:
		.byte	"ok"
		.byte	0AH,00H
		.END
```

If an application no need some call back function, a macro should be defined as default inline callback function specified by the compiler argument in the build script (Makefile) .

Wini-Buh:
@kkitayam It could be a possible solution, but I guess @hathach has to answer that. In the mean time I can leave with the solution to leave the weak macro empty.

hathach:
inline function may not work with

```C
if (weak_func) weak_func()
```

Since it does not emit symbol. Also the static inline require function must be included by the class driver, and which means all the weak function must be implemented in header file of `tusb_config.h` (or one of the file it includes). Which can grow very unpleasantly. Therefore I still prefer to force the strong symbol on CCRX, eventually when they support weak symbol, we can easily flip the TU_ATTR_WEAK to support that. 

For slow execution, ccrx probably smart enough with -Ofast to know that the function is empty and just omit that, not entirely sure though.

kkitayam:
I understood @hathach's opinion, and I agree with to force the strong symbol on CCRX.

If application developers are concerned about calling empty functions on CCRX, they can achieve the trick I suggested by defining a static inline function in the application's `tusb_config.h`. Maybe.

Wini-Buh:
@hathach Some note about the missing weak attribute on CCRX. I tried to cleanup the code and removing my solution of the weak functionality. While doing so I found the following code situation:

https://github.com/hathach/tinyusb/blob/1b8473902fb0e808c32183f306938accc9ce0d5f/src/device/usbd.c#L998-L1008

If I implement an empty `tud_descriptor_device_qualifier_cb`  what would the return value be ? I think it should be NULL, but this is maybe a problem because the return value is actually not checked. I guess I need to modify also the code here, so if NULL is returned by `tud_descriptor_device_qualifier_cb` then `process_get_descriptor` should return false here. 

BTW: I have not checked yet if there are other similar situations in the code.

hathach:
just return NULL `TU_ASSERT(desc_qualifier);` already take care of that and return if it is NULL.

Wini-Buh:
@hathach Removed the missing weak attribute work around and merged again from current master

src/common/tusb_types.h
```diff
@@ -342,13 +343,19 @@ typedef struct TU_ATTR_PACKED
   } bmAttributes     ; ///< This field describes the endpoint's attributes when it is configured using the bConfigurationValue. \n Bits 1..0: Transfer Type \n- 00 = Control \n- 01 = Isochronous \n- 10 = Bulk \n- 11 = Interrupt \n If not an isochronous endpoint, bits 5..2 are reserved and must be set to zero. If isochronous, they are defined as follows: \n Bits 3..2: Synchronization Type \n- 00 = No Synchronization \n- 01 = Asynchronous \n- 10 = Adaptive \n- 11 = Synchronous \n Bits 5..4: Usage Type \n- 00 = Data endpoint \n- 01 = Feedback endpoint \n- 10 = Implicit feedback Data endpoint \n- 11 = Reserved \n Refer to Chapter 5 of USB 2.0 specification for more information. \n All other bits are reserved and must be reset to zero. Reserved bits must be ignored by the host.
 
   struct TU_ATTR_PACKED {
+#if defined(__CCRX__)
+    //FIXME the original defined bit field has a problem with the CCRX toolchain, so only a size field is defined
+    uint16_t size;
```
hathach, line 348: I am curious what would be the problem with this bitfield, the bmAttributes also defined something similar but you didn't change it.

src/portable/renesas/usba/dcd_usba.c
```diff
@@ -485,13 +518,22 @@ static void process_set_address(uint8_t rhport)
 {
   const uint32_t addr = USB0.USBADDR.BIT.USBADDR;
   if (!addr) return;
+#if defined(__CCRX__)
+  tusb_control_request_t setup_packet;
```
hathach, line 522: oh, ccrx does not support this style of initialization ? Hmm, there is a lots more of this within codebase.

hathach (commented):
Superb !! without the weak macros, thing is much cleaner. I think we are very closed to the merge. I only have a few questions. Will try to setup my board to test this soon enough.

src/portable/renesas/usba/dcd_usba.c
kkitayam: This is the style of the C99 standard. CCRX supports the C99 standard by specifying `-lang=c99` as a command line argument. Since TinyUSB implicitly requires c99, I think CCRX should also have this option.

kkitayam:
@Wini-Buh 
If you don't mind, could you add your copyright notice at `dcd_usba.c`.

src/portable/renesas/usba/dcd_usba.c
hathach: yeah c99 is really old now, it should be enabled to build with.

src/device/usbd_control.c
```diff
@@ -55,8 +55,17 @@ typedef struct
 
 static usbd_control_xfer_t _ctrl_xfer;
 
+#if defined(TU_HAS_NO_ATTR_ALIGNED)
```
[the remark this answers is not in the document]
kkitayam: As I understand it, there is no need for alignment if DMA is not used. And the current implementation does not use DMA.

If using DMA, the buffer must be 2 bytes aligned.

src/portable/renesas/usba/dcd_usba.c
Wini-Buh: Don't worry. I found the solution that also works with GCC, so the check which toolchain is used is no longer required. The problem with this struct definition and CCRX is, that one of the struct member is defined as an union which includes some bit fields (that could be an error of the CCRX toolchain ? ). 

BTW: I already use C99 language option.

src/device/usbd_control.c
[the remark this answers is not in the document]
Wini-Buh: RX MCU does not need an alignment on this. It was introduced, because there was an error in `dcd_usba.c`, which is corrected. On my test code the `_usbd_ctrl_buf` buffer started on an odd address and with the error in `dcd_usba.c` the transfer over an endpoint was faulty. So this fix is no longer required (as long as no DMA transfer is supported).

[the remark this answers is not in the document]
hathach: It is surprised  that  ccrx has align pragma for function but not variable. Anyway there is several other places that may requires alignment especially endpoint buffer in each class driver. Therefore, maybe we revert this changes for now, and will revisit these later when doing DMA. SInce doing this alone is not enough anyway.

hathach:
I want to help to speed up the merge of this PR (has been going for a while), so that #874 can be merged, and also help testing with the net lwip example. 

However, setting up the [rn65n cloud kit](https://www.renesas.com/us/en/products/microcontrollers-microprocessors/rx-32-bit-performance-efficiency-mcus/rx65n-cloud-kit-renesas-rx65n-cloud-kit) (picked because it come with add-on board with usb connection) is a bit difficult to me since this is first time I worked with renesas mcu. Renesas driver is either too bare metal or too complicated with FIT module (I don't really like generated code sigh). Will give it more try, any advice is welcome :)

src/osal/osal_freertos.h
```diff
@@ -131,14 +138,31 @@ typedef QueueHandle_t osal_queue_t;
 
 static inline osal_queue_t osal_queue_create(osal_queue_def_t* qdef)
 {
+#if defined(__Tx36V5_Maincard__)
```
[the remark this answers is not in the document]
Wini-Buh: Switching from `CFG_TUSB_OS=OPT_OS_FREERTOS` to `CFG_TUSB_OS=OPT_OS_CUSTOM` to use my own RTOS API integration file to overcome this issue

src/osal/osal_freertos.h
```diff
@@ -51,6 +51,7 @@ static inline void osal_task_delay(uint32_t msec)
 typedef StaticSemaphore_t osal_semaphore_def_t;
 typedef SemaphoreHandle_t osal_semaphore_t;
 
+#if (configSUPPORT_STATIC_ALLOCATION == 1)  //FIXME Only static API supported
```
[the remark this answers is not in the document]
Wini-Buh: Switching from `CFG_TUSB_OS=OPT_OS_FREERTOS` to `CFG_TUSB_OS=OPT_OS_CUSTOM` to use my own RTOS API integration file to overcome this issue

src/common/tusb_types.h
Wini-Buh: I need to have deeper look at that. It seems to be a problem between the endian macros found in descriptor macro and the bit field definition in the `tusb_desc_endpoint_t` struct. Especially in the big endian configuration (in little endian configuration the code seems to work).
BTW: Sorry for the time delay, but at the moment I'm very busy at work. I hope to get it done this weekend

Wini-Buh:
@hathach Maybe I could support you a little bit. I don't also like the FIT modules (I had to much negative experience with it, like with other similar technique from other vendors). Is there a board support package available for your board ? It would at least help you to speed up the clock programming for the first time.

BTW: Sorry for the time delay about this PR, but at the moment I'm very busy at work.

src/osal/osal_freertos.h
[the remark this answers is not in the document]
hathach: FreeRTOS can have both dynamic and static co-exist, I am curious why don't you just enable the static allocation. Sure I could update the osal to support dynamic api from freeRTOS as well, but yeah, it should be on its own PR.

src/common/tusb_types.h
hathach: That is problematic with endian, I would love to have some hand-on testing with this as well. I am ok to have it as it is for now.

hathach:
> @hathach Maybe I could support you a little bit. I don't also like the FIT modules (I had to much negative experience with it, like with other similar technique from other vendors). Is there a board support package available for your board ? It would at least help you to speed up the clock programming for the first time.

That would be great to have more help, here is link to my board from renesas site https://www.renesas.com/us/en/products/microcontrollers-microprocessors/rx-32-bit-performance-efficiency-mcus/rx65n-cloud-kit-renesas-rx65n-cloud-kit
I try to follow @kkitayam gr_citrus (rx63n) board 
- hw/bsp/ https://github.com/hathach/tinyusb/tree/master/hw/bsp/rx63n/boards/gr_citrus
- hw/mc/renesas https://github.com/kkitayam/rx_device/tree/4a51dfe6ecdf936d2d89f223f069e24a2d109207/rx63n

Though It is a bit low level with register value for hardware/clock init (maybe it is more apparent to renesas user). My goal is coyping generated file from e2studio for rx65n and then copy/modify from gr_cirtrus board. But it is a bit tough so far. 

> BTW: Sorry for the time delay about this PR, but at the moment I'm very busy at work.

You have nothing to sorry about, we are busy with work. That is why I am hoping to jump in to help with testing and verifying changes with actual hardware.

kkitayam:
@hathach
I might be help you. I have a [RX65N target board](https://www.renesas.com/jp/en/products/microcontrollers-microprocessors/rx-32-bit-performance-efficiency-mcus/rx-family-target-board-target-board-rx-family).

I had bought it before gr_citrus in order to study RX MCU. Actually, I had tried to implement DCD on this board, but I failed. There were many transaction errors. I don't know if the lack of external XTALs is to blame, or if my wiring for the USB connector is to blame.

hathach:
> @hathach
> I might be help you. I have a [RX65N target board](https://www.renesas.com/jp/en/products/microcontrollers-microprocessors/rx-32-bit-performance-efficiency-mcus/rx-family-target-board-target-board-rx-family).
> 
> I had bought it before gr_citrus in order to study RX MCU. Actually, I had tried to implement DCD on this board, but I failed. There were many transaction errors. I don't know if the lack of external XTALs is to blame, or if my wiring for the USB connector is to blame.

That would be great, I only manage to get it compiled with board_test for testing LED and timer (not yet any other usb example) by copying and commenting code that I have no idea about :D. Here is what I have done so far.
- https://github.com/kkitayam/rx_device/pull/1
- https://github.com/hathach/tinyusb/tree/add-rx65n-board

Though I am still struggle to flash the mcu using E2 Lite debugger 
 
```
Renesas Flash Programmer CLI V1.01
Module Version: V3.08.02.000
Load: "/home/hathach/Dropbox/tinyusb/tinyusb/examples/device/board_test/_build/rx65n_cloud_kit/board_test.mot" (Size=6894, CRC=806324C4)

Connecting the tool (E2 emulator Lite)
Tool: E2 emulator Lite (OBE110001)
Interface: 2 wire UART
Tool Firmware Version: V3.02.00.000

Emulator power supply: OFF
Connecting the target device

[Error] E4000004: A framing error occurred while receiving data. (BFW: 0354)
Disconnecting the tool
```

kkitayam:
I had used [Renesas Flash Programmer GUI](https://www.renesas.com/tw/en/software-tool/renesas-flash-programmer-programming-gui) on Windows. I've never used the CUI. And, I am using JLink.

When using the JLink, the following pin connections can be used. 
- short EJ2 jumper header.
- short EMLE pin (P03) and 3V3(VCC)

| Function  | J1 pin header | RX65N pin name  | JLink pin number |
|:---------:|:---------:|:---------------:|:-------------:|
| 3V3       |   14      | VCC             |  1(VTref) |
|TRST       |   16      | P34             |  3 |
|GND        |   12      | VSS             |  4 |
|TDI        |   20      | P30             | 5  |
|TMS        |   19      | P31             |         7|
|TCK/FINEC  |   21      | P27             |        9|
|TDO        |   22      | P26             |     13|
|nRES       |   10      | RES#            |       15     |

hathach:
Thanks @kkitayam to be honest, I would expect it should just work out of the box, They are all tools from renesas and it would make it easier for user that want to test out the board as well, I will give a few more try and then back to jlink if it doesn't work out.

Wini-Buh:
@hathach  I had a look at the documentation of your board. Here are some hints (maybe more or less useful):

- The onboard E2 Lite debugger can only use the FINE interface for debugging/programming, JTAG is not supported
- For the board there is board description file available, that should lower the entry level to get the board working

I would suggest to use the board description file and the smart configurator to setup the clock system, because this part is sometimes tricky and the graphical view of the clock system shows you any wrong configuration clearly.

hathach:
@Wini-Buh thanks, I will try to use the E2 studio to generate some clock, though I include bunch of other FIT files. One of the issue is E2 studio on Linux does not support RX at all. VM is ok, but it is a bit annoying. I will give it more try this week.

kkitayam:
@hathach 
I pushed some commits into [my branch](https://github.com/kkitayam/tinyusb/tree/add-rx65n-board).  I have confirmed that the commits make `board_led_write()` and `board_button_read()` work fine on `board_test` example. But, `board_uart_read()` doesn't work.

And, I will try `Renesas Flash Programmer CLI`.

hathach:
@kkitayam wow thank you, I am current switched to other works, but will test it out when possible.

kkitayam:
> And, I will try Renesas Flash Programmer CLI.

I was able to confirm the writing with rpi-cli by using following steps on Windows 10. I think that it works on Linux.
```
make BOARD=rx65n_cloud_kit
rx-elf-objcopy -O ihex _build\rx65n_cloud_kit\board_test.elf board_test.hex
rfp-cli -d RX65x -t e2l -if fine -fo id FFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFF -auth id FFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFF -a board_test.hex
```

`rfp-cli` had output following logs.
```
Renesas Flash Programmer CLI V1.00
Module Version: V3.08.01.000
Load: "C:\Users\kkitayam\tinyusb\examples\device\board_test\board_test.hex" (Size=7265, CRC=EED13467)

Connecting the tool (E2 emulator Lite)
Tool: E2 emulator Lite (OBE110001)
Interface: FINE
Tool Firmware Version: V3.02.00.000

Emulator power supply: OFF
Connecting the target device
Main Clock: 16 MHz
Speed: 1,000,000 bps
Connected to RX Group

Erasing the target device
  [Data Flash 1]       00100000 - 00107FFF
  [Code Flash 1]       FFE00000 - FFFFFFFF

  [Code Flash 1]       FFE00000 - FFE07FFF
  [Code Flash 1]       FFFFE000 - FFFFFFFF
Writing data to the target device
  [Code Flash 1]       FFE00000 - FFE0097F
  [Code Flash 1]       FFFFFF80 - FFFFFFFF
Verifying data on the target device
  [Code Flash 1]       FFE00000 - FFE0097F
  [Code Flash 1]       FFFFFF80 - FFFFFFFF
Writing data to the target device
  [Config Area]        FE7F5D00 - FE7F5D2F
  [Config Area]        FE7F5D40 - FE7F5D7F
Verifying data on the target device
  [Config Area]        FE7F5D00 - FE7F5D2F
  [Config Area]        FE7F5D40 - FE7F5D7F


Disconnecting the tool

Operation successful
```

hathach:
@kkitayam bravos, that is a great progress :+1: :+1:

src/common/tusb_types.h
Wini-Buh: I did some test about this issue here. First I setup a new build variant with GCC in big endian order to check if the same problem occurs as with the CCRX. This was not successfully, build my test project with GCC in big endian order shows a  problem in the tinyUSB sources with bitfields. 

First the order of the bitfields is changed and second there is no possibility to change the ordering of the bitfields in GCC (at least I found no possibility). Every thing I tried (for example `#pragma reverse_bitfields` ) did not work.

BTW: My conclusion about this investigation is to avoid bitfields and access bit coded stuff trough macros. That maybe sounds very hard so I suggest to discuss this in a new discussion thread about how to get highly portable code. What do you think about it @hathach ?

Wini-Buh:
The compilation error with GCC should be solved. I did not realize that the GCC throws an error on this. So there were only two solutions. Use a toolchain dependent compilation (like I did it now) or try to temporary disable the error  while compiling (don't know how with GCC... only know it on CCRX).

src/common/tusb_types.h
hathach: @Wini-Buh  unfortunately bitfield is used widely throughout the stack, it is probably too much of hassle for now to change it. I think we should still get bit endian correctly with pragma.

hathach:
@Wini-Buhhow I am still catching up with this PR. Would you compile rxcc on windows with stock example, I tried to use make command, however, since rxcc use different compile and linker option from command line than rxgcc and I don't understand it enough to make change to makefile to get it compiled. I guess you are having an e2studio project file, It would be great if you could share the file so that I could compiler and test it out.

Wini-Buh:
@hathach I will try to build a "stock example" with the e2studio and give you the project file. But I need to do that at the weekend and I can't test the example because I do not own the board used for the RX examples. In the meantime I show you the arguments given to the compiler, assembler and linker so you could see the how the command line options look like. You should find the description of each argument in the manual of the toolchain.
[commandline_options_examples.txt](https://github.com/hathach/tinyusb/files/6780180/commandline_options_examples.txt)

src/common/tusb_types.h
Wini-Buh: @hathach Ok, if will try to investigate more on this topic. I'm not sure if it is a problem of the toolchain (my hope was to reproduce the issue with the GCC, but because it is not possible to get the same settings with the bit field ordering the try was hopeless).

hathach:
thanks, a project file to compile with stock example is very useful, if you could make one for cdc_msc (for testing) then the net_lwip_server as well so that I could help. The example for compile option txt is useful, I wish I know these well enough to put that into makefile to easily switch between rxgcc and rxcc. But I don't think I could pull it off, and could be error prone. It is better to address one thing at a time and stick with things that work first.

Wini-Buh:
@hathach Here are the two e2studio project files of the cdc_msc stock example. I setup it up for a RX65N Cloud-Kit, which unfortunately I didn't own. That means I could not test it. I could only compile it with the CCRX. After all I'm not sure if it is very useful for you, but please have look at it.
[project_files.zip](https://github.com/hathach/tinyusb/files/6833390/project_files.zip)

hathach:
@Wini-Buh late response, I tried to import your project files, though look like it expect some certains file structure, and also missing some of your local files to build. This is why I don't like from IDE, it takes lots time to configure and setting up for an port/board and may not portable to other users. 
image

Anyway, since I got the gccrx working well now, I will try to update PR to my best effort  so that we could merge this soon. Since the pr has been pending for so long, and have some good addition. For rxcc I will reply on you to do the actual testing.

hathach:
> @hathach I will try to build a "stock example" with the e2studio and give you the project file. But I need to do that at the weekend and I can't test the example because I do not own the board used for the RX examples. In the meantime I show you the arguments given to the compiler, assembler and linker so you could see the how the command line options look like. You should find the description of each argument in the manual of the toolchain.
> [commandline_options_examples.txt](https://github.com/hathach/tinyusb/files/6780180/commandline_options_examples.txt)

I notice from you have `bit_order=left` in your command line option, the default for rxcc is right and that is also the order used by gcc. That could cause issue with the code, is there any reason you compile with bit-order = left ?

Wini-Buh:
> @Wini-Buh late response, I tried to import your project files, though look like it expect some certains file structure, and also missing some of your local files to build. This is why I don't like from IDE, it takes lots time to configure and setting up for an port/board and may not portable to other users.
>
> Anyway, since I got the gccrx working well now, I will try to update PR to my best effort so that we could merge this soon. Since the pr has been pending for so long, and have some good addition. For rxcc I will reply on you to do the actual testing.

Ok. Great to hear you got CCRX working well now. I mentioned that the project file alone maybe is not very useful for you. 
If required I could of course build a complete e2studio project (with all sources) and send it to you.

> I notice from you have `bit_order=left` in your command line option, the default for rxcc is right and that is also the order used by gcc. That could cause issue with the code, is there any reason you compile with bit-order = left ?

That is because I used one of my projects based on the project I'm working on (and that is using tinyUSB). This project uses a lot of existing code (based on the SH2A controller from Renesas) and there the default bit order is left. That's the only reason.

hathach:
> Ok. Great to hear you got CCRX working well now. I mentioned that the project file alone maybe is not very useful for you.
> If required I could of course build a complete e2studio project (with all sources) and send it to you.

Actually not rxcc, the gccrx though, sorry for not being clear enough. That allows to make changes, mostly rename compiler attribute and other minor clean up with confidence. I have push update to your fork to update PRs. Please pull first if you want to do further work with it.

> That is because I used one of my projects based on the project I'm working on (and that is using tinyUSB). This project uses a lot of existing code (based on the SH2A controller from Renesas) and there the default bit order is left. That's the only reason.

Ah, i see, that makes sense. I am just curious why would you need that.

src/common/tusb_types.h
```diff
@@ -342,9 +350,14 @@ typedef struct TU_ATTR_PACKED
...
+#else
```
hathach, line 356: I am still not convinced there is issue with bit-field within a struct. The bitfield order should give us the correct layout. Maybe it is endian issue. We could just enable the LOG=2 to see if the size of openning endpoit is correct or not. If you are busy, maybe we could revert this for now to get PR to merge first (I could do the revert).

src/portable/renesas/usba/dcd_usba.c
```diff
@@ -245,12 +264,6 @@ static int fifo_write(volatile void *fifo, pipe_state_t* pipe, unsigned mps)
 
   hw_fifo_t *reg = (hw_fifo_t*)fifo;
   uintptr_t addr = pipe->addr + pipe->length - rem;
-  if (addr & 1u) {
```
hathach, line 248: seems like a change in logic, I don't understand the mcu enough to decide if modification is correct. Maybe if you could, explain/confirm this a bit.

src/portable/renesas/usba/dcd_usba.c
```diff
@@ -396,11 +409,12 @@ static bool process_pipe_xfer(uint8_t rhport, uint8_t ep_addr, uint8_t* buffer,
   USB0.PIPESEL.WORD  = num;
   const unsigned mps = USB0.PIPEMAXP.WORD;
   if (dir) { /* IN */
-    USB0.D0FIFOSEL.WORD = num | USB_FIFOSEL_MBW_16;
-    while (!(USB0.D0FIFOSEL.BIT.CURPIPE != num)) ;
-    int r = fifo_write(&USB0.D0FIFO.WORD, pipe, mps);
+    USB0.D0FIFOSEL.WORD = num | USB_FIFOSEL_MBW_16 | (TU_BYTE_ORDER == TU_BIG_ENDIAN ? USB_FIFOSEL_BIGEND : 0);
+    while (USB0.D0FIFOSEL.BIT.CURPIPE != num) ;
```
hathach, line 413: like above for confirmation.

src/portable/renesas/usba/dcd_usba.c
```diff
@@ -230,7 +249,7 @@ static unsigned select_pipe(unsigned num, unsigned attr)
 {
   USB0.PIPESEL.WORD  = num;
   USB0.D0FIFOSEL.WORD = num | attr;
-  while (!(USB0.D0FIFOSEL.BIT.CURPIPE != num)) ;
+  while (USB0.D0FIFOSEL.BIT.CURPIPE != num) ;
```
hathach, line 252: this is an 180 degree logic changes, though I think it is correct one. I understand that we are reading and waiting to confirm out num is written. But again, I am not entirely, though it is worth a confirmation.

hathach (commented):
It has been pending for awhile, although I couldn't get tested with ccrx, the modification look good. It also contains bug fix when using the stack with big-endian mcu (which isn't tested yet). I think it is best to merge it for now, and resolve the `net_lwip_webserver` example with ccrx as follow up PR later on. That example is complicated, there may need some config tweak with lwip (much like tinyusb) to work with rxcc as intended as well. 

I did a bit of rename and cleanup (please do pull if you plan to make more changes to the PR). There is only a few feedback as follows:
-  revert change to wMaxPacketSize field until tested if you don't have time to do it for now.
- confirmation for removal of odd address check in fifo_write() of dcd
- confirmation for inverted logic of  statement `while (USB0.D0FIFOSEL.BIT.CURPIPE != num)`

@kkitayam if you have some time, would you mind help to confirm the changes, it is your great work, you understand this much better than me.

src/portable/renesas/usba/dcd_usba.c
kkitayam: Yes, the modification is correct. The original code was wrong logic.

kkitayam: The user manual describes followings.
image

src/portable/renesas/usba/dcd_usba.c
kkitayam: I think this change is correct.

The codes are regarding load from misaligned address. At that time I thought that RX MCU has no support for load from misaligned address. But, it was my misunderstanding. In fact, it has support for load from misaligned address. So, the part is unnecessary.

hathach: thank you very much for confirmation :+1: :+1:

src/portable/renesas/usba/dcd_usba.c
hathach: thanks, it is much clearer now. I think the PR is almost ready to merge :)

src/common/tusb_types.h
hathach: @Wini-Buh is it OK if I revert this change for now to get this PR merged first. I am ok for this change in future PR, I only need confirmation that it does indeed needed for ccrx to work.

Wini-Buh: For me it's OK to revert it. It is definitively required in the project I'm using tinyUSB, which is build in `big endian` and `left bitfield order`. I have already setup test code to investigate the problem deeper (for me it looks like a problem of the CCRX toolchain if the configuration I mentioned is used... and if so, there is maybe another workaround for it). 
I keep you informed about the result of the investigation about this problem.

hathach: Thanks for confirmation, I will leave it on for now since you actually need it for ccrx. We could update it later as follow up PR. However, the bmAttributes (just right above) also use bitfield within struct, you may also need to check it out as well. Too bad, I couldn't test ccrx myself, they make it difficult to run on Linux.

Anyway, when you have time please try to run an very simply example using tusb_desc_endpoint_t struct (copy it to helloworld example), then write an value to the wMaxPacketSize like 0x1234 then print out its field out to see how it is layout etc ... I tried to run it within my Windows VM, but got issue connecting to E2 debugger `GDB action 'connect to target', has failed with error code, 0xffffffffd`. I just gave up on trying it out for now.

hathach (approved):
Thank you very much for your PR. Let merge this as it is now, since it is good enough. For more issue with ccrx and net driver, we could resolve it later on as folllow up PR when having more time to work on it.

PR #862, opened 2021-05-30
Fix host rp2040 with low speed device

### LINKED ISSUE #709: RP2040 Host Stack fails with HID devices

hathach:
**Describe the PR**
Fix host rp2040 when mounting low speed device (e.g keyboard) with max packet size = 8
- Correct Data Toggle by number of packet: e.g  64 bytes = 8 x 8 bytes --> Data Toggle should not be flipped in the next buffer.  
- Improve RP2040-E4 (host buf sel) fix to prevent panic with already available.

Fix #709, also replace https://github.com/raspberrypi/tinyusb/pull/7 as the fix

### DIFF
--- a/src/portable/raspberrypi/rp2040/rp2040_usb.c
+++ b/src/portable/raspberrypi/rp2040/rp2040_usb.c
@@ -161,7 +161,10 @@ void _hw_endpoint_xfer_start(struct hw_endpoint *ep, uint8_t *buffer, uint16_t t
     // Fill in info now that we're kicking off the hw
     ep->total_len = total_len;
     ep->len = 0;
-    ep->transfer_size = tu_min16(total_len, ep->wMaxPacketSize);
+
+    // Limit by packet size but not less 64 (i.e low speed 8 bytes EP0)
+    ep->transfer_size = tu_min16(total_len, tu_max16(64, ep->wMaxPacketSize));
+
     ep->active = true;
     ep->user_buf = buffer;
 #if TUSB_OPT_HOST_ENABLED
@@ -240,8 +243,9 @@ bool _hw_endpoint_xfer_continue(struct hw_endpoint *ep)
     _hw_endpoint_xfer_sync(ep);
 
     // Now we have synced our state with the hardware. Is there more data to transfer?
+    // Limit by packet size but not less 64 (i.e low speed 8 bytes EP0)
     uint16_t remaining_bytes = ep->total_len - ep->len;
-    ep->transfer_size = tu_min16(remaining_bytes, ep->wMaxPacketSize);
+    ep->transfer_size = tu_min16(remaining_bytes, tu_max16(64, ep->wMaxPacketSize));
 #if TUSB_OPT_HOST_ENABLED
     _hw_endpoint_update_last_buf(ep);
 #endif
--- a/src/portable/raspberrypi/rp2040/rp2040_usb.c
+++ b/src/portable/raspberrypi/rp2040/rp2040_usb.c
@@ -46,7 +46,7 @@ static inline void _hw_endpoint_lock_update(struct hw_endpoint *ep, int delta) {
 #if TUSB_OPT_HOST_ENABLED
 static inline void _hw_endpoint_update_last_buf(struct hw_endpoint *ep)
 {
-    ep->last_buf = ep->len + ep->transfer_size == ep->total_len;
+    ep->last_buf = (ep->len + ep->transfer_size == ep->total_len);
 }
 #endif
 
@@ -126,7 +126,28 @@ void _hw_endpoint_start_next_buffer(struct hw_endpoint *ep)
 
     // PID
     val |= ep->next_pid ? USB_BUF_CTRL_DATA1_PID : USB_BUF_CTRL_DATA0_PID;
+
+#if TUSB_OPT_DEVICE_ENABLED
     ep->next_pid ^= 1u;
+
+#else
+    // For Host (also device but since we dictate the endpoint size, following scenario does not occur)
+    // Next PID depends on the number of packet in case wMaxPacketSize < 64 (e.g Interrupt Endpoint 8, or 12)
+    // Special case with control status stage where PID is always DATA1
+    if ( ep->transfer_size == 0 )
+    {
+      ep->next_pid ^= 1u;
+    }else
+    {
+      uint32_t packet_count = 1 + ((ep->transfer_size - 1) / ep->wMaxPacketSize);
+
+      if ( packet_count & 0x01 )
+      {
+        ep->next_pid ^= 1u;
+      }
+    }
+#endif
+
 
 #if TUSB_OPT_HOST_ENABLED
     // Is this the last buffer? Only really matters for host mode. Will trigger
@@ -143,6 +164,7 @@ void _hw_endpoint_start_next_buffer(struct hw_endpoint *ep)
     // the next time the controller polls this dpram address
     _hw_endpoint_buffer_control_set_value32(ep, val);
     pico_trace("buffer control (0x%p) <- 0x%x\n", ep->buffer_control, val);
+    //print_bufctrl16(val);
 }
 
 
@@ -189,10 +211,14 @@ void _hw_endpoint_xfer_sync(struct hw_endpoint *ep)
 
 #if TUSB_OPT_HOST_ENABLED
     // tag::host_buf_sel_fix[]
+    // TODO need changes to support double buffering
     if (ep->buf_sel == 1)
     {
         // Host can erroneously write status to top half of buf_ctrl register
         buf_ctrl = buf_ctrl >> 16;
+
+        // update buf1 -> buf0 to prevent panic with "already available"
+        *ep->buffer_control = buf_ctrl;
     }
     // Flip buf sel for host
     ep->buf_sel ^= 1u;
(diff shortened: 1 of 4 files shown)


src/portable/raspberrypi/rp2040/rp2040_usb.c
```diff
@@ -126,8 +126,30 @@ void _hw_endpoint_start_next_buffer(struct hw_endpoint *ep)
...
+    // For Host (also device but since we dictate the endpoint size, following scenario does not occur)
+    // Next PID depends on the number of packet in case wMaxPacketSize < 64 (e.g Interrupt Endpoint 8, or 12)
+    // Special case with control status stage where PID is always DATA1
+    if ( ep->transfer_size == 0 )
+    {
+      // ZLP also toggle data
+      ep->next_pid ^= 1u;
+    }else
+    {
+      uint32_t packet_count = 1 + ((ep->transfer_size - 1) / ep->wMaxPacketSize);
```
kilograham, line 143: haven't read the spec, but this just smells wrong... toggle the PID IF the transfer is an odd number number of packets total?

kilograham (commented):
not sure i you meant to leave all the extra logging stuff in there?

src/portable/raspberrypi/rp2040/rp2040_usb.c
hathach: Yup, the Data Toggle (PID) is toggled every packet. Following is the scheduling 108 bytes with endpoint with max packet size = 8. 
- Buffer 0 will start with 64 then 44 as last buf
- 64 is 8 packets starting with DATA1, each packet, DT toggled which mean after 8 packet, it is still DATA1. i.e at the 9th packet  (start of 44 bytes, transaction 87 in capture), the DT should remain the same (DATA1).

image

hathach: PS: The log is there is on purpose, since I want to work more with rp2040 hcd/dcd, what I have in my mind is:
- Double buffering starting with host
- Make rp2040 common usb function more primitive, currently it handles also host and device specific code, making it more difficult to fix specific issue with either host/device (like above). Also it makes it difficult to follow in my taste.
- Host and device mode should be able to both enabled, and dynamically run depending on what the software want e.g using ID pin and/or VBUS sense to detect whether we attached to peer as host/device to initialize the correct mode.

kilograham: my question was whether the code was correct - as far as i can see the packet_count is only dependent on the total transfer size, and therefore does change), or is ep->transfer_size the size per packet?

kilograham:
note i'd like to pull this into the pico-0.10.0 branch anyway which we'll ship SDK1.2.0 with most likely

src/portable/raspberrypi/rp2040/rp2040_usb.c
hathach: I don't quite understand your question could you give an example with above total transfer size = 108 bytes above to elaborate.  Here is how the transfer size is determined https://github.com/hathach/tinyusb/blob/host-rp2040/src/portable/raspberrypi/rp2040/rp2040_usb.c#L189

hathach:
> note i'd like to pull this into the pico-0.10.0 branch anyway which we'll ship SDK1.2.0 with most likely

no problem, I could release this as 0.10.1 , there is also issue when using pico reading msc thumbdrive as well. I will try to look at when I got time.

src/portable/raspberrypi/rp2040/rp2040_usb.c
hathach: I will try to explain as much as I could, as above example total transfer = 108 on endpoint with size = 8, which is broken into 2 sub transfers, both on buffer0 since we don't support double buffer just yet. 
- xfer1: transfer size = 64, start with DATA1, packet count = 8 
- xfer2: transfer size = 44, also start with DATA1, packet count = 5

if there is ever another xfer on the endpoint, it will start with DATA0 since the packet count is odd. Total transfer is broken in to 2 sub transfer size and therefore already taken into account when computing the next PID after this total transfer is complete. Hope that make sense, else I am happy to hear from your example.

src/portable/raspberrypi/rp2040/rp2040_usb.c
```diff
@@ -126,8 +126,30 @@ void _hw_endpoint_start_next_buffer(struct hw_endpoint *ep)
...
+
+      if ( packet_count & 0x01 )
```
kilograham, line 145: I guess to be clear, this was the line i was actually complaining about (based on the above).... for the the packets where packet_count is 8 the pid won't toggle, for all the ones packet count is 5 it will.

hathach: I see what you mean there, you suspect that all the 8 packets of first 64 bytes has the same PID right ?

kilograham: correct

liamfraser: Looking at the trace above the DATA PID in low speed mode needs to be toggled every 8 bytes. It depends on the approach. You can either limit the ep->transfer_size to do several 8 byte transfers. OR you can start a 64 byte transfer in low speed mode and the hardware will flip the PID for you every 8 bytes.

hathach: > correct

That isn't the case here, the controller is very smart to toggle the PID every packet as shown in the captured which is actual capture to develop this PR. We only need to write the correct initial PID. 

> You can either limit the ep->transfer_size to do several 8 byte transfers. OR you can start a 64 byte transfer in low speed mode and the hardware will flip the PID for you every 8 bytes.

I tried both way, both works. Actually the code used to limit transfer_size at 8 (max packet size), but having it with as 64 make it much more faster, this will also help to enable the double buffering later on as well.

liamfraser: @kilograham do you agree this is resolved if the hardware toggles the pid every 8 bytes in LS mode?

liamfraser (approved):
Looks good to me assuming @kilograham agrees the Low Speed DATA PID conversation is resolved.

hathach:
> Looks good to me assuming @kilograham agrees the Low Speed DATA PID conversation is resolved.

The PID is toggled by hardware as captured in analyzer. without the fix, it will trigger the incorrect PID `USB_INTS_ERROR_DATA_SEQ_BITS` error. Thank for the review, I will merge this since I have another PR pending since yesterday.

src/portable/raspberrypi/rp2040/rp2040_usb.c
hathach: actually this is only tested with IN transfer, therefore the PID is toggled from the device side. I haven't tested the OUT transfer where host controller is the one responsible for PID toggle. Though there is not many control transfer that has large payload > 64. We can keep this in mind, should it be an issue in the future.

kilograham (commented):
ah, ok; "the code may be wrong, but fixes the problem at hand" :-) I'm cool with it for now!

src/portable/raspberrypi/rp2040/rp2040_usb.c
hathach: > well i'm still extremely confused by the code... at least a comment is needed... i'm still not clear why you toggle the packet every buffer (and indeed whether that is a 8 byte buffer or not) when the total packet length is odd. what if there are 4 packets?

I have explained it very clearly as above. PID flipped every packet per USB specs, 4 wont flip it

hathach:
> ah, ok; "the code may be wrong, but fixes the problem at hand" :-) I'm cool with it for now!

The code is spot on and there is no maybe afaik. Please provide example or hands on test case if you are still in doubt.

PR #866, opened 2021-05-31
Enable non-fifo based custom vendor device

jgressmann:
Define CFG_TUD_VENDOR_CUSTOM to non-zero to use your own callbacks.

### DIFF
--- a/src/class/vendor/vendor_device.c
+++ b/src/class/vendor/vendor_device.c
@@ -33,6 +33,10 @@
 
 #include "vendor_device.h"
 
+#if CFG_TUD_VENDOR_CUSTOM
+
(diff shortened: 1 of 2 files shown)


hathach (commented):
The better approach is not eanble the builtin `CFG_TUD_VENDOR` at all, and write your own driver from application using `usbd_app_driver_get_cb()` . Here are 2 examples
- how rpi team add their raspberrypi/pico-sdk#197
- https://github.com/fluffymadness/tinyusb-xinput/blob/master/main.c

jgressmann:
I'll look into that. Thanks for pointing me into the right direction.

PR #873, opened 2021-06-03
Add __unused to variables that are only used if TU_LOG does something

kilograham:
Fixes unused variable compiler warnings


hathach (approved):
thank you, it is my bad, it is temporarily for debug only. Will definitely enable all warning on rpi.

PR #882, opened 2021-06-07
Add support for 32 button gamepads

mmosca:
**Describe the PR**
More buttons = more better 

This pull request implements a configuration option to switch between the current 16 buttons and 32, with the current 16 buttons being the default value.

Adding "#define CFG_TUD_MAX_BUTTONS       32" to tusb_config.h enables the extra 16 buttons.

**Additional context**

While trying to map a real life RC transmitter to a controller that can easily integrate with flight sims I quickly run out of buttons to map all my 3 position switches and 32 seems to be the magic number that can allows me to map the 16 proportional channels and 2 digital channels as 6 axis, 10 tri state switches and 2 buttons for the digital channels.


src/class/hid/hid.h
```diff
@@ -222,7 +234,25 @@ typedef enum
   GAMEPAD_BUTTON_MODE   = TU_BIT(12), ///< Mode button
   GAMEPAD_BUTTON_THUMBL = TU_BIT(13), ///< L3 button
   GAMEPAD_BUTTON_THUMBR = TU_BIT(14), ///< R3 button
-//GAMEPAD_BUTTON_       = TU_BIT(15), ///< Undefined button
+  GAMEPAD_BUTTON_15     = TU_BIT(15), ///< Button 15
+#if (CFG_TUD_MAX_BUTTONS > 16)
+  GAMEPAD_BUTTON_17     = TU_BIT(16), ///< Button 17
```
hathach, line 239: off by one, also incorrect Button 16 in comment for typo

hathach (changes requested):
nice, I thought 16 is already enough, apparently not. To simply thing, we should just update the number of buttons to 32 entirely without the need of CFG_TUD_MAX_BUTTONS. Would you mind updating the PR to just use 32 buttons to replace existing 16 buttons. More is better !!!

src/class/hid/hid.h
mmosca: I have no problems making 32 buttons the only option. 

The reason I went the long way around was that I though people might be bothered by the increase in  size of the joystick report.

I also updated the GAMEPAD_BUTTON_ entries and comments. GAMEPAD_BUTTON_15 is now GHAMEPAD_BUTTON_16 to match how windows and most game reference the buttons.

hathach: > I have no problems making 32 buttons the only option.
> 
> The reason I went the long way around was that I though people might be bothered by the increase in size of the joystick report.

Yeah, but don't worry, people shouldn't be affected if using the helper API. Those that uses the raw API should know what to do :)

> I also updated the GAMEPAD_BUTTON_ entries and comments. GAMEPAD_BUTTON_15 is now GHAMEPAD_BUTTON_16 to match how windows and most game reference the buttons.

I am not sure if this is a correct way to do it, @nekuneko what do you think. Should we start the button from 0 or from 1. I would love to hear your thought since you are the original author that add enhance support for gamepad.

mmosca: @hathach I considered starting it at 0, but since the other buttons have symbolic names (A, B, SELECT, START, etc) I think it makes more sense to match how they are displayed to the end user.

hathach: thanks, that makes sense for sure. Though I still like to hear from another user on another platform (Linux) as well. Let just wait for his input for a day or two.

mmosca: No problem.

At least one app in Linux does 0-31 :) I don't know any actual games to test on the pi though. 

image

hathach:
I have updated the gamepad helper with uint32_t for buttons, also do a rebase since mater has some changes.

src/class/hid/hid.h
hathach: I just did a quick check, and Steam controller setting seems to start from 0, and is spot-on with Button South. Now I am more lean toward starting from 0.
image

mmosca: Should I go ahead and change them to be 0 based, or do you feel like we should wait a bit more? I really don't have a strong opinion on this one and would be ok with either.

mmosca:
> I have updated the gamepad helper with uint32_t for buttons, also do a rebase since mater has some changes (push to your fork). Please pull first if you want to make more changes.

Good catch. I had missed those.

src/class/hid/hid.h
hathach: > Should I go ahead and change them to be 0 based, or do you feel like we should wait a bit more? I really don't have a strong opinion on this one and would be ok with either.

Yes, please do, also please remove the "Note: button number from 1 on host OS" as well. I think we will just merge it starting with Button0.

mmosca: Done.

hathach (approved):
Perfect, I actually have a final tweak to have all button number to make it consistency. The std naming for button (x,y,a,b) is now defined as macro alias.

PR #889, opened 2021-06-10
Make examples standalone buildable via CMake

kilograham:
A lot of users of Raspberry Pi Pico SDK1.2.0 tried to build the TinyUSB examples directly from within the example directories in the TinyUSB submodule, which failed without a helpful error message

This change changes the builds to include `family_support.cmake`  and have that include the `<FAMILY>/family.camke` so that it can produce a useful error message (this is cleaner anyway).


hathach:
seem like it has an issue with esp32s2, give me a bit of time, I will check it out and try to fix the build.

hathach (approved):
Thank you for your PRs, sorry for late response, have been busy testing the double buffered and other works. I made the push to set TOP for an example to pass build with esp32sx. Will merge when ci passed.

kilograham:
> Thank you for your PRs, sorry for late response, have been busy testing the double buffered and other works. I made the push to set TOP for an example to pass build with esp32sx. Will merge when ci passed.

no worries; i figured that TOP was the problem, but hadn't got as far as fixing it either... i would have set TOP in the family.cmake for the esp32x instead though

hathach:
Yeah, that is a better place to set TOP. Though I just tried to it fixed for not, will have to update/write the cmake for esp32sx anyway later on.

PR #891, opened 2021-06-11
RP2040 double buffer

hathach:
**Describe the PR**
This PR implements double buffered for RP2040, rework/refactor the existing quite a log
- we don't have to deal with RP2040-E4 hardware our of synce
- faster transfer for control and bulk 
- Should help https://github.com/raspberrypi/pico-sdk/issues/442 with logitech universal receiver (8-byte control in Fullspeed) and wired keyboard with 8byte control in Lowspeed
- Lots of internal rp2040 usb renaming
- Testing with device stack, cdc_msc, hid_composite running fine as normal. 
- There is an issue with net_lwip_webserver example when there is a short packet on buffer0 (i.e < 64 bytes) while we expect data on both buffer. At the hanlding time (currently triggered per 2 buffer), the buffer1 is probably filled with data from the next transfer (not current one). For now we disable double buffered for device OUT.  

PR is tested on both host and device mode and should be ready for review and merge. @kilograham @liamfraser please let me know if changes look good to you and it would be great if you could have some time to test it out.

### DIFF
--- a/examples/host/cdc_msc_hid/src/hid_app.c
+++ b/examples/host/cdc_msc_hid/src/hid_app.c
@@ -30,7 +30,8 @@
 // MACRO TYPEDEF CONSTANT ENUM DECLARATION
 //--------------------------------------------------------------------+
 
-// If your host terminal support ansi escape code, it can be use to simulate mouse cursor
+// If your host terminal support ansi escape code such as TeraTerm
+// it can be use to simulate mouse cursor movement within terminal
 #define USE_ANSI_ESCAPE   0
 
 #define MAX_REPORT  4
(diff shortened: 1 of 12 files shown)


liamfraser:
Sorry for the delay here. The code looks good to me. I don't have much time to test it at the moment because I am working on a different project. @kilograham do you have any thoughts?

hathach:
> Sorry for the delay here. The code looks good to me. I don't have much time to test it at the moment because I am working on a different project. @kilograham do you have any thoughts?

No problem at all, I am busy with other works anyway. Note currently the double buffered is not fully operated (skip with device IN), but we could address this later. For now, anyone that could help with testing would be great.

hathach:
I have double checked again with all of device examples: cdc_msc, hid_composite, net_lwip_webserver, webusb, midi_test. Everything seems working perfectly. It is good for a merge. Thank you very much for reviewing the changes. Should there be any issue, it could be fixed as followup pr.

kilograham:
Awesome!

PR #899, opened 2021-06-18
Changed order of struct initializer to avoid compiler error

ned-pcs:
**Describe the PR**
This re-orders the order of two of the designators in a struct initializer to match the struct field declaration order to avoid compiler errors.

**Additional context**
I'm compiling for an Arm Cortex-M0, and my compiler threw an error because the designator order was different in the initializer than in the struct declaration.


hathach (approved):
ah right, this should be how it be initialized. Thank you for the PR

PR #905, opened 2021-06-20
UAC fixes

HiFiPhile:
**Describe the PR**
- Add bracket to switch case, fix warning `switch transfer of control bypasses initialization of xxx`

- Put `tu_desc_subtype()` in `#if CFG_TUD_AUDIO_ENABLE_ENCODING || CFG_TUD_AUDIO_ENABLE_DECODING`, fix unused function.

- Fix IAR compile error on incomplete pointer type. 

Seems Diff is a little bugged...


HiFiPhile:
@PanRe  Could you take a look if it's ok for you ?

PanRe:
The diff is a bit rugged yes, but as far as i see it you only added the brakets and changed some pointer definitions. It is ok from my side i guess, i did not conduct any tests. I am a bit puzzeled why IAR complains about pointers here while gcc does not... :D

hathach (approved):
Thank @PanRe for reviewing the PR, I also did a quick check, the diff is a bit out of place, but actual change is very minimal. Should be no problem to merge.

PR #911, opened 2021-06-22
Enable empty transfers for tud_vendor_n_write()

thierer:
Extend `tud_vendor_n_write()` to make a zero length transfer when `buffer==NULL && bufsize==0`, as discussed in #903.

I consider implementing this via a flag to `maybe_transmit()` the best solution:
- `tud_vendor_n_write()` doesn't have access to neither `buffer` nor `bufsize`, so it can't make the decision.
- Implementing it completely in `tud_vendor_n_write()` would mean duplicating some functionality of `maybe_transmit()` (but judging by your latest comment maybe that's what you prefer?).


hathach:
I have a second thought though, I think ZLP should only be sent when there is no data and the last packet is multiple of endpoint size) much like cdc https://github.com/hathach/tinyusb/blob/master/src/class/cdc/cdc_device.c#L469 .

thierer:
Something like this? (Compiles, but not tested).

It does not exactly what you wrote in that it doesn't check for a **multiple**, but the **equivalent** of the endpoint size, because the actual transfer can't really be bigger than the EP size? Or were you referring to the last `bufsize` parameter to `tud_vendor_n_write()`? If so, I don't understand what you want to achieve.

Even so I don't fully get the rationale behind that requirement. On the protocol level, the device can still send packages that are smaller than the EP size and in that case there would still be no way to send a ZLP? (I don't think change would help for the problem that lead me to open this thread).

src/class/vendor/vendor_device.c
```diff
@@ -112,10 +114,11 @@ static bool maybe_transmit(vendord_interface_t* p_itf)
   TU_VERIFY( !usbd_edpt_busy(TUD_OPT_RHPORT, p_itf->ep_in) );
 
   uint16_t count = tu_fifo_read_n(&p_itf->tx_ff, p_itf->epin_buf, CFG_TUD_VENDOR_EPSIZE);
-  if (count > 0)
+  if (count > 0 || last_in_transfer_was_epsize)
   {
     TU_ASSERT( usbd_edpt_xfer(TUD_OPT_RHPORT, p_itf->ep_in, p_itf->epin_buf, count) );
   }
+  last_in_transfer_was_epsize = count && (count == CFG_TUD_VENDOR_EPSIZE);
```
kasjer, line 121: condition seems to be overkill (only right side of && makes a difference) or wrong (&& where || should be)

thierer: `&&` ist correct, but you're right, it's redundant.

HiFiPhile:
Sorry for the extra long delay, I've merged latest ZLP handling from CDC class.

HiFiPhile (approved):

PR #918, opened 2021-06-24
Add fifo get function to CDC class for DMA transfer

HiFiPhile:
**Describe the PR**
Like audio class, CDC need DMA support in order to support high speed transfer.

**Additional context**

### DIFF
--- a/src/class/cdc/cdc_device.c
+++ b/src/class/cdc/cdc_device.c
@@ -483,4 +483,18 @@ bool cdcd_xfer_cb(uint8_t rhport, uint8_t ep_addr, xfer_result_t result, uint32_
   return true;
 }
 
+// Get the Receive FIFO (for DMA transfer)
+tu_fifo_t* tud_cdc_n_get_rx_ff (uint8_t itf)
+{
+  TU_ASSERT(itf < CFG_TUD_CDC);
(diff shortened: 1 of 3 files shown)


HiFiPhile:
It's not enough to just export the fifo, need a mechanism to call `_prep_out_transaction`.

hathach (commented):
I don't think exposing fifos is good idea. I don't have a specific suggestion for now, but we will need to spend more time to look at this later on.

HiFiPhile:
In fact CDC streaming is needed by my company's project which restarted recently, following my tests last year.

For the short time expose the fifo fit my need, of cause we can make something better for example expose only fifo_info even add edpt_xfer_fifo support.

hathach:
> In fact CDC streaming is needed by my company's project which restarted recently, following my tests last year.
> 
> For the short time expose the fifo fit my need, of cause we can make something better for example expose only fifo_info even add edpt_xfer_fifo support.

It is easy to expose an API, but it is very hard to drop one. I would prefer to figure it all out first. I think for now, you could make an "hack" for your fork e.g dropping static in the function/variable in the cdc_device and have extern in your app. It is not ideal, but I would be more flexible

HiFiPhile:
I've opened a new issue to discuss #920 .

PR #926, opened 2021-06-27
CDC class dcd_transdimension improvements

HiFiPhile:
**Describe the PR**
Various improvements in order to improve CDC efficiency.

Test result can be found in https://github.com/hathach/tinyusb/issues/920, which shows a great increment in throughput and 
and reduce execution time.

- Add xfer_fifo support for CDC and dcd_transdimension
- Add functions to support direct (DMA) fifo transfer for CDC class
- `tud_cdc_rx_wanted_cb` is still WIP since can't be done without accessing underlying fifo struct members
- Added explicite xfer_fifo option for UAC class

**Additional context**
Linear buffer is still used by default since more RAM maybe needed (at least 4K for dcd_transdimension), and no `tud_cdc_rx_wanted_cb` for the moment.

Instead of export the fifo it's also possible to export only `tu_fifo_buffer_info_t`, but it's inconsistent with audio class.

PR #938, opened 2021-06-29
Bug fix and Enhancements of UAC2

HiFiPhile:
**Describe the PR**

- Fix audiod_get_AS_interface_index in audio class.
- Enhance uac2_headset example with multiple sample rates.
- Add macro to calculate EP size.


src/device/usbd.h
```diff
@@ -542,6 +542,11 @@ TU_ATTR_WEAK bool tud_vendor_control_xfer_cb(uint8_t rhport, uint8_t stage, tusb
   /* Standard AS Isochronous Feedback Endpoint Descriptor(4.10.2.1) */\
   TUD_AUDIO_DESC_STD_AS_ISO_FB_EP(/*_ep*/ _epfb, /*_interval*/ 1)\
 
+//   Calculate wMaxPacketSize of Endpoints
+#define TUD_AUDIO_EP_SIZE(_maxFrequency, _nBytesPerSample, _nChannels) \
```
PanRe, line 546: Nice helper!

examples/device/uac2_headset/src/usb_descriptors.h
```diff
@@ -99,21 +99,21 @@ enum
     /* Type I Format Type Descriptor(2.3.1.6 - Audio Formats) */\
     TUD_AUDIO_DESC_TYPE_I_FORMAT(_nBytesPerSample, _nBitsUsedPerSample),\
     /* Standard AS Isochronous Audio Data Endpoint Descriptor(4.10.1.1) */\
-    TUD_AUDIO_DESC_STD_AS_ISO_EP(/*_ep*/ _epout, /*_attr*/ (TUSB_XFER_ISOCHRONOUS | TUSB_ISO_EP_ATT_ADAPTIVE | TUSB_ISO_EP_ATT_DATA), /*_maxEPsize*/ _epoutsize, /*_interval*/ (CFG_TUSB_RHPORT0_MODE & OPT_MODE_HIGH_SPEED) ? 0x04 : 0x01),\
+    TUD_AUDIO_DESC_STD_AS_ISO_EP(/*_ep*/ _epout, /*_attr*/ (TUSB_XFER_ISOCHRONOUS | TUSB_ISO_EP_ATT_DATA), /*_maxEPsize*/ _epoutsize, /*_interval*/ 0x01),\
```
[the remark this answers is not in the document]
HiFiPhile: Oops... But in UAC ISOCHRONOUS only is also valid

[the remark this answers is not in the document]
PanRe: Ah ok. Well, i consulted the UAC2 spec again to be sure and there it is stated that you have to select one of the three types.

PanRe (approved):

hathach (commented):
Thanks @HiFiPhile  @PanRe for yet another P & review for audio. I am pretty much clueless for UAC2 at the moment. I push some example skip for mcu that does not have enough sram to run the example. Will merge when ci is passed.

HiFiPhile:
> Will merge when ci is passed.

Wait a little, I'll add handling when alt settings changes :)

hathach:
> > Will merge when ci is passed.
> 
> Wait a little, I'll add handling when alt settings changes :)

no problem, please pull first to prevent any conflict, let me know when you are done.

src/class/audio/audio_device.c
```diff
@@ -1502,6 +1507,15 @@ static bool audiod_set_interface(uint8_t rhport, tusb_control_request_t const *
   {
     audio->ep_out_as_intf_num = 0;
     usbd_edpt_close(rhport, audio->ep_out);
+
+#if !CFG_TUD_AUDIO_ENABLE_ENCODING
+    // Clear FIFOs, since data is no longer valid
+    tu_fifo_clear(&audio->ep_out_ff);
```
HiFiPhile, line 1513: @PanRe I've added a fifo clear on set interface.

HiFiPhile:
>  does not have enough sram to run the example

I've reduced the ram needed, maybe they can run now.

HiFiPhile:
@hathach I think it's ready to be merged.

examples/device/uac2_headset/src/main.c
```diff
@@ -218,14 +257,14 @@ static bool tud_audio_feature_unit_get_request(uint8_t rhport, audio_control_req
...
-      TU_LOG2("Get channel %u volume range (%d, %d, %u) dB\r\n", request->bChannelNumber,
+      TU_LOG1("Get channel %u volume range (%d, %d, %u) dB\r\n", request->bChannelNumber,
               range_vol.subrange[0].bMin / 256, range_vol.subrange[0].bMax / 256, range_vol.subrange[0].bRes / 256);
       return tud_audio_buffer_and_schedule_control_xfer(rhport, (tusb_control_request_t const *)request, &range_vol, sizeof(range_vol));
     }
     else if (request->bRequest == AUDIO_CS_REQ_CUR)
     {
       audio_control_cur_2_t cur_vol = { .bCur = tu_htole16(volume[request->bChannelNumber]) };
-      TU_LOG2("Get channel %u volume %u dB\r\n", request->bChannelNumber, cur_vol.bCur);
+      TU_LOG1("Get channel %u volume %d dB\r\n", request->bChannelNumber, cur_vol.bCur / 256);
```
PanRe, line 267: Where does the divison by `256` come from?

src/class/audio/audio_device.c
PanRe: Well spotted thanks!

PanRe (approved):
Thanks a lot for rectifying the example and fixing the bugs in the audio driver and the fifo! This is a very nice piece of work you did here! Thanks a lot for the time! :)

examples/device/uac2_headset/src/main.c
HiFiPhile: I'm not sure in which page, but at least the result is correct.....

HiFiPhile:
@PanRe I've some problem in sending feedback. I want to call `tud_audio_n_fb_set` inside `tud_audio_set_itf_cb` to initialize feedback value to default value, but since `_audiod_fct[func_id].ep_fb`was still 0, `fb_val` was sent in EP0.

Actually there are multiples `tud_audio_set_itf_cb` inside `audiod_set_interface`, in the callback there is no way to identify them.

PanRe:
> 
> 
> @PanRe I've some problem in sending feedback. I want to call `tud_audio_n_fb_set` inside `tud_audio_set_itf_cb` to initialize feedback value to default value, but since `_audiod_fct[func_id].ep_fb`was still 0, `fb_val` was sent in EP0.
> 
> Actually there are multiples `tud_audio_set_itf_cb` inside `audiod_set_interface`, in the callback there is no way to identify them.

I see, this means we can not rely on the host to open the feedback EP first. Anyway, it should be possible to determine where the call comes from by parsing the request the call back function gives you `bool tud_audio_set_itf_cb(uint8_t rhport, tusb_control_request_t const * p_request)`. I admit that it would be easier for the user to introduce explicite callbacks for IN, OUT, and FB EPs but at the time writing i was not sure which infos from the request is of interest for the user. So at the time being i decided to give the user the complete request at his disposal.

HiFiPhile:
> it should be possible to determine where the call comes from by parsing the request the call back function gives you

Actually there is no difference between ep_out and ep_fb, all fields are the same.

One possible fix is 12747b6

PanRe:
> 
> 
> > it should be possible to determine where the call comes from by parsing the request the call back function gives you
> 
> Actually there is no difference between ep_out and ep_fb, all fields are the same.
> 
> One possible fix is [12747b6](https://github.com/hathach/tinyusb/commit/12747b61d5a8d4ecf51a93bbbcee09a9c71f183b)

Not sure if i can follow you... Shouldn't the direction of the FB EP be of type IN and the direction of the OUT EP of type OUT as checked in line 1642 within `audio_device.c`?

HiFiPhile:
> checked in line 1642 within audio_device.c

What was checked is the endpoint descriptor, with the EP descriptor it's easy to identify them.

But what passed to the callback is `tusb_control_request_t`, I printed all fields, for ep_out and ep_fb they are the same.

PanRe:
Ahhh man, i need more sleep ^^ Yes you are right! Mhhh... ok it seems passing the request pointer to the user is kind of useless then. How about splitting the CB into four parts: CB_EP_IN, CB_EP_OUT, CB_FB_IN, and CB_FB_OUT. I am currently not sure which parameters to pass for the CB functions but i guess those which are parsed within `set_interface()` are enough for the beginning! What do you think?

HiFiPhile:
> What do you think?

I think we can simply pass both ep_desc (or ep_addr) and p_request, actually p_request is useful since interface and alt number are inside.

hathach:
@HiFiPhile seem like some of mcus still failed to run, would you mind adding back those skip files to get ci passed. @PanRe I will reply on you to review this PR. Please let me know if you think all your questions for this PR is resolved then re-approve it again.

HiFiPhile:
> would you mind adding back those skip files to get ci passed

Now it's better.

PanRe (changes requested):

HiFiPhile:
Maybe it's better to add `TUD_AUDIO_PREFER_RING_BUFFER` after #926 ... Or even take this change into #926

PanRe (approved):

PanRe:
Mh i don't see your commit with the ring buffer on/off switch in the review section. Is this intended or is something out of sync here?

HiFiPhile:
> Mh i don't see your commit with the ring buffer on/off switch in the review section. Is this intended or is something out of sync here?

Since xfer_fifo of dcd_transdimension hasn't been merged yet and I added macro for lpc, I hesitate a little and put it in #926

PanRe:
Okay! Do you intend to add some more on this commit or is it finished?

HiFiPhile:
I think I'm finished.

I have a question about interleaved copy decode/encode, nBytesToCopy is nChannelsPerFifo*nBytesPerChannel, in case of 2ch/24bit & 2ch/32bit nChannelsPerFifo has to be 1 since it handle only 1-4 bytes copy ?

PanRe:
> 
> 
> I think I'm finished.
> 
> I have a question about interleaved copy decode/encode, nBytesToCopy is nChannelsPerFifo*nBytesPerChannel, in case of 2ch/24bit & 2ch/32bit nChannelsPerFifo has to be 1 since it handle only 1-4 bytes copy ?

Actually no, you have found another bug here! Instead of `nBytesToCopy` it should be `nBytesPerSample` i.e. `audio->n_bytes_per_sampe_tx` within `audiod_interleaved_copy_bytes_fast_encode` and likewise for the decode function. The amount of bytes to copy is controlled via start and end pointers.

PanRe:
> 
> 
> > I think I'm finished.
> > I have a question about interleaved copy decode/encode, nBytesToCopy is nChannelsPerFifo*nBytesPerChannel, in case of 2ch/24bit & 2ch/32bit nChannelsPerFifo has to be 1 since it handle only 1-4 bytes copy ?
> 
> Actually no, you have found another bug here! Instead of `nBytesToCopy` it should be `nBytesPerSample` i.e. `audio->n_bytes_per_sampe_tx` within `audiod_interleaved_copy_bytes_fast_encode` and likewise for the decode function. The amount of bytes to copy is controlled via start and end pointers.

Could you incorporate that bug fix here please? You already fixed a lot within this PR ;)

HiFiPhile:
> Could you incorporate that bug fix here please?

Like this ?

HiFiPhile:
On LPC performance is quite awesome. Total CPU usage is only about 12% while streaming 32bit/384k ;)
image

PanRe:
> 
> 
> > Could you incorporate that bug fix here please?
> 
> Like this ?

Jep, thanks!

PanRe:
A very big thanks to @HiFiPhile for all the effort and time he spent on improving the audio driver! @hathach, i am happy with all commits, PR is ready for merge!

hathach (approved):
Superb !! Huge thanks to both @HiFiPhile for the awesome PR and @PanRe for reviewing. I wished I could help better with audio driver, however, my experience and knowledge with UAC 2.0 is very limited at the moment.

PR #940, opened 2021-06-30
Fix fifo overflow correction.

HiFiPhile:
**Describe the PR**
Fix fifo overflow correction.

**Additional context**

### DIFF
--- a/src/common/tusb_fifo.c
+++ b/src/common/tusb_fifo.c
@@ -325,7 +325,7 @@ static uint16_t advance_pointer(tu_fifo_t* f, uint16_t p, uint16_t offset)
   // We are exploiting the wrap around to the correct index
 
   // TODO warning: assuming signed overflow does not occur when assuming that (X + c) < X is always false [-Wstrict-overflow]
-  if ((p > p + offset) || (p + offset > f->max_pointer_idx))
+  if ((p > (uint16_t)(p + offset)) || ((uint16_t)(p + offset) > f->max_pointer_idx))
   {
     p = (p + offset) + f->non_used_index_space;
   }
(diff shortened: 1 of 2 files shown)


PanRe (approved):

HiFiPhile:
For `advance_pointer`
```
  if ((p > p + offset) || (p + offset > f->max_pointer_idx))
  {
    p = (p + offset) + f->non_used_index_space;
  }
  else
  {
    p += offset;
  }
```
I feel like the check can simplified into `if (p + offset > f->max_pointer_idx))`, have I missed something ?

PanRe:
> 
> 
> For `advance_pointer`
> 
> ```
>   if ((p > p + offset) || (p + offset > f->max_pointer_idx))
>   {
>     p = (p + offset) + f->non_used_index_space;
>   }
>   else
>   {
>     p += offset;
>   }
> ```
> 
> I feel like the check can be simplified into `if (p + offset > f->max_pointer_idx))`, have I missed something ?

The first condition covers the case when `p` overflows solely by `offset`. It is possible that the second condition never hits i.e. if `offset` is big enough to "jump over" the non-used index space.

src/common/tusb_fifo.c
```diff
@@ -325,7 +325,7 @@ static uint16_t advance_pointer(tu_fifo_t* f, uint16_t p, uint16_t offset)
   // We are exploiting the wrap around to the correct index
 
   // TODO warning: assuming signed overflow does not occur when assuming that (X + c) < X is always false [-Wstrict-overflow]
```
hathach, line 327: The GCC seem to be right about this. I got the warning when compiling with gnu++17, haven't got time to trace it down. So I put the warning comment here.  Thanks for the PR.

PR #941, opened 2021-06-30
Add support for RTT logging by NEWLIB on Renesas RX family.

kkitayam:
**Describe the PR**
~~Add `rx_rtt_printf()` to work RTT logging with NEWLIB on Renesas RX family.
`rx_rtt_printf()` should be set to `CFG_TUSB_DEBUG_PRINTF`. The function just call `SEGGER_RTT_vprintf`.~~
Add support for RTT logging by newlib.
Remove support for OPTLIB.

**Additional context**
OPTLIB is removed from [GCC 8.3.0.202102-GNURX](https://llvm-gcc-renesas.com/release-notes/rx/latest/8.3.0.202102/release_notes.pdf). So, we need to make RTT work properly with NEWLIB.
I don't know why, but RTT didn't work via `_write` with NEWLIB. Using `SEGGER_RTT_vprintf`, the RTT worked.


hw/bsp/rx/family.mk
```diff
@@ -22,11 +22,16 @@ endif
...
+    CFLAGS += -DCFG_TUSB_DEBUG_PRINTF=rx_rtt_printf
+    SRC_C += $(RTT_SRC)/RTT/SEGGER_RTT_printf.c
+  endif
 else
   # setup for optlib
   CFLAGS += -nostdinc \
     -isystem $(OPTLIBINC) \
-    -DLWIP_NO_INTTYPES_H
+    -DLWIP_NO_INTTYPES_H \
+    -DOPTLIB
```
hathach, line 34: since optlib is deprecated now, can we  just remove it altogether and use NEWLIB since it is pretty standard on other ports.It is fine if you think we could still keep OPTLIB as an option.

kkitayam: You're right, I'll remove it. I was thinking it was a pain to test with older version gcc. :smile:

hathach:
Look great, thanks for the PR. Regarding the _write() hook, maybe renesas use an slightly differrent name for write such ash: `write()` or `sys_write()`  msp430 has something like that as well https://github.com/hathach/tinyusb/blob/master/hw/bsp/board.c#L28 . Let me know if you want to check it out, this is already good for merge.

kkitayam:
> Let me know if you want to check it out, this is already good for merge.

OK, I'll try the same way as msp430 and report back with the results.

kkitayam:
@hathach 
Thank for your advice. I confirmed that RTT works fine with the same way as msp430 on RX65N target board.

Following modifications were added.
- Extend stack areas, because there was not enough to work newlib's `printf`.
- Added some dummy functions, because some linker warnings occur when logging is enabled.

hathach (approved):
superb !! thank you, I guess it depends on the gcc version, seem like version 8 and before use write() instead of _write(). Glad that work out. I resolve the conflict and push update, will merge when ci passed.

PR #949, opened 2021-07-03
Add SUSPEND/RESUME handling for Renesas RX family.

kkitayam:
**Describe the PR**
Add SUSPEND/RESUME handling for Renesas RX family. 

**Additional context**
I have confirmed suspend/resume works with `hid_composite` example on `rx65n_target`.
But, `DCD_EVENT_UNPLUGGED` is not issued like before. I have no idea why VBINT(it is the VBUS changing interruption) is not occurred when the USB cable is removed.

### DIFF
--- a/src/portable/renesas/usba/dcd_usba.c
+++ b/src/portable/renesas/usba/dcd_usba.c
@@ -553,7 +553,6 @@ void dcd_remote_wakeup(uint8_t rhport)
 {
   (void)rhport;
   USB0.DVSTCTR0.BIT.WKUP = 1;
-  while (USB0.DVSTCTR0.BIT.WKUP) ;
 }
 
 void dcd_connect(uint8_t rhport)
@@ -701,6 +700,11 @@ void dcd_int_handler(uint8_t rhport)
   }
   if (is0 & USB_IS0_SOFR) {
     if (_dcd.suspended) {
+      /* When USB host resumes caused by `dcd_remote_wakeup()`,
+       * RESM interrupt does not rise.
+       * Therefore we need to manually send resume event.
+       * Of course, when USB host resumes on its own,
+       * RESM interrupt rise properly, then this statements are ignored. */
       dcd_event_bus_signal(rhport, DCD_EVENT_RESUME, true);
       _dcd.suspended = 0;
     }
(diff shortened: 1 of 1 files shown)


src/portable/renesas/usba/dcd_usba.c
```diff
@@ -689,6 +695,17 @@ void dcd_int_handler(uint8_t rhport)
       dcd_disconnect(rhport);
     }
   }
+  if (is0 & USB_IS0_RESM) {
+    dcd_event_bus_signal(rhport, DCD_EVENT_RESUME, true);
+    _dcd.suspended = 0;
+  }
+  if (is0 & USB_IS0_SOFR) {
+    if (_dcd.suspended) {
+      dcd_event_bus_signal(rhport, DCD_EVENT_RESUME, true);
```
hathach, line 704: hmm, this is a bit odd, SOF should be recognized by hardware as Resume signal. Do we really need to do it manual ?

kkitayam: I am agree your guessing. But unfortunately, RX does not signal resume interrupt caused by `dcd_remote_wakeup()` , in my experience. Of course, when the host resumes on its own, it occurs resume interrupt.

hathach: @kkitayam I see, that is a bit odd for rx65n controller for not raising Resume interrupt in this case. Could you put an comment explaining why we need to manually send resume event here. Just in case when going back to this and/or we could figure out a better way to detect resume.

src/portable/renesas/usba/dcd_usba.c
```diff
@@ -547,7 +552,8 @@ void dcd_set_address(uint8_t rhport, uint8_t dev_addr)
 void dcd_remote_wakeup(uint8_t rhport)
 {
   (void)rhport;
-  /* TODO */
+  USB0.DVSTCTR0.BIT.WKUP = 1;
+  while (USB0.DVSTCTR0.BIT.WKUP) ;
```
hathach, line 556: I think we don't really need to blocking wait here,which could take a few ms, just return. We will send resume event later on with SOF walkaround. The usbd has its own suspended status bit, and shouldn't do any transfer until RESUME event is reported by dcd.

kkitayam: OK, I will remove this statement.

src/portable/renesas/usba/dcd_usba.c
kkitayam: Comments are added at 1ff3b76.

src/portable/renesas/usba/dcd_usba.c
kkitayam: The statement is removed at 1ff3b76.

hathach (approved):
perfect ! Thank you very much for the PR.

PR #956, opened 2021-07-08
add settings for Renesas RX family

kkitayam:
**Describe the PR**
Add audio format settings for Renesas RX family.

**Additional context**
The maximum packet size limit in the standard is 1024 bytes as ISO EP, but the RX USB controller only supports up to 256 bytes. The current configuration of the master branch requires a minimum of 384 bytes and will not work with RX.


hathach (approved):
I haven't tested this yet, but It looks good to me. Some MCUs will have issue keeping up with high speed rate. When more mcus join the low-bitrate club we will try to make it look more elegant.

PR #959, opened 2021-07-10
[PORT] Add GD32VF103 support and Sipeed Longan Nano Board support

KarlK90:
**Describe the PR**

This is a rough first draft to support the GD32VF103 RISC-V MCU by GigaDevice. It uses the same Synopsys USB OTG full-speed peripheral that is found on the STMF105 and STM32F107. I have added the [nuclei-sdk](https://github.com/Nuclei-Software/nuclei-sdk) that adds support for the GD32VF103. I chose to extract register definitions for the OTG peripheral from the stm32f105.h headers and put them into synopsys_common.h in order to not rename all registers in the sources, which is hacky and another solutions should probably be found. So this needs clean up, atm this PR is meant to help debugging with adafruit/tinyuf2#131.

Additional files for development:

**Toolchain**

I can highly recommend the [pre-built toolchains from embecosm](https://www.embecosm.com/resources/tool-chain-downloads/#riscv-stable).

**Datasheets and SVD files**

[GD32VF103 Datasheet](https://github.com/hathach/tinyusb/files/6795031/GD32VF103_Datasheet_Rev1.4.pdf)
[GD32VF103 User Manual](https://github.com/hathach/tinyusb/files/6795034/GD32VF103_User_Manual_Rev1.3.pdf)
[GD32VF103 SVD (remove .txt)](https://github.com/hathach/tinyusb/files/6795035/GD32VF103.svd.txt)

These files except the SVD were downloaded from [GigaDevice](http://www.gd32mcu.com/en/download/0?kw=GD32VF1). GigaDevice has given all registers different names for the USB peripheral but it is the same as found on the others mcus. So an SVD file for the STM32F105/STM32F107 should give the familiar STM32 names. Seggers J-Link and RTT support the GD32VF103 out of the box.

The Longan Nano has a broken out JTAG Debug Header, I have enabled the USART peripheral on this header as well it is the T0 and R0 pins.


hathach:
sorry for late review, I have been late with other works. will try this out asap.

KarlK90:
No Problem and thank you for helping with the debugging! 🙂 

I hope you and your families are safe and healthy.

src/portable/st/synopsys/synopsys_common.h
```diff
@@ -0,0 +1,1448 @@
+#include "stdint.h"
```
hathach, line 1: ah right, this is quite a pain that GD32V name register differently despite they are pretty much the same thing. I am ok adding this for gd32, some other non-stm may need this as well since synopsys dwc controller is quite popular. There is only one issue though, I think most of the definition is from ST driver, we should keep their license in this file, luckily it is compatible with MIT so we are all good.

hathach (changes requested):
Sorry for keeping you waiting, I have been overwhelmed with other paid work and non-work (real life) issue that we are all facing. Anyway, I finally manged to review this PR, it is much easier than I thought, you are doing a really great work !! Though there are only couple of minor feedbacks
1. nuclei-sdk is overkill, please use the simpler firmware library from gigadevice for low level mcu. Since the stack doesn't need anything else
2. please add the ST license to the `synopsys_common.h`

hathach:
> No Problem and thank you for helping with the debugging! 
> 
> I hope you and your families are safe and healthy.

Thank you, same to you. Sorry for late response again.

src/portable/st/synopsys/synopsys_common.h
KarlK90: Perfect, I wasn't sure if you would be fine with this "hack". I have added the stm32 license header.

.gitmodules
```diff
@@ -124,3 +124,6 @@
 [submodule "hw/mcu/mindmotion/mm32sdk"]
 	path = hw/mcu/mindmotion/mm32sdk
 	url = https://github.com/zhangslice/mm32sdk.git
+[submodule "hw/mcu/gd/nuclei-sdk"]
+	path = hw/mcu/gd/nuclei-sdk
+	url = https://github.com/Nuclei-Software/nuclei-sdk.git
```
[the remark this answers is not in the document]
KarlK90: I completely agree that the nuclei-sdk is overkill for this port, but I had chosen it over the multiple other repositories of the GD32VF firmware on github because their repository is the most up to date one and actively maintained by nucleisys which provided the RV32 core IP. Furthermore there is a bug in the startup code that throws an compilation error if `-Wundef` and `-Werror` is used with GCC, for which I had [opened a PR.](https://github.com/Nuclei-Software/nuclei-sdk/pull/30) Therefore I would vote to keep this despite being more then necessary, what do you think?

The other repositories:

https://github.com/riscv-mcu/GD32VF103_Firmware_Library (Semi-official, contains a note to use nuclei-sdk, last activity in October 2020)

https://github.com/GigaDevice-Semiconductor/GD32VF103_Firmware_Library (Is this official?, last real activity in November 2020)

KarlK90:
I have rebased on the latest master and pushed my changes! Thanks for your review and no need to apologize, I hope everything is well :-). 

If you find the time, then I'm curios what you find out about the [strange errors](https://github.com/adafruit/tinyuf2/issues/131#issuecomment-877617268) I had with the demos and tinyuf2 bootloader.

.gitmodules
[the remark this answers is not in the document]
hathach: I pull the nuclei and did a quick check, the https://github.com/GigaDevice-Semiconductor/GD32VF103_Firmware_Library is more updated in term of the mcu driver. The last activities you said in 2020 is update the mcu lib to v1.1. while nuclei still uses the v1.0 version. nuclei is more maintained by it is mostly for their platform, tinyusb is only interested in low level mcu driver (clock, register def). Therefore I think it make more sense to move to the said repo.

src/portable/st/synopsys/dcd_synopsys.c
```diff
@@ -92,6 +93,18 @@
 #define EP_MAX_FS       6
 #define EP_FIFO_SIZE_FS 1280
 
+#elif CFG_TUSB_MCU == OPT_MCU_GD32VF103
+#define STM32F1_SYNOPSYS
+#include "gd32vf103.h"
+#include "nmsis_core.h"
```
hathach, line 99: I am not familiar with riscv and I am not sure if nuclei nmsis is as universal used as arm cmsis. If we didn't use it much e.g only enable/disable IRQ, maybe it is best to just use the low level code from giga firmware. That will allow user to run tinyusb with bare metal without the need for nuclei sdk.

hathach:
> I have rebased on the latest master and pushed my changes! Thanks for your review and no need to apologize, I hope everything is well 
> 
> If you find the time, then I'm curios what you find out about the [strange errors](https://github.com/adafruit/tinyuf2/issues/131#issuecomment-877617268) I had with the demos and tinyuf2 bootloader.

I will look at that, though I think firstly we need to make sure all stock examples in this repo run well with vf103 first. Since it is new port, it may has bug/issue. Therefore it is best to test against known-good example first. Once it is confirmed working  well with stock example, we will move on with external one such as tinyuf2.

.gitmodules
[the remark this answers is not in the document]
KarlK90: I'm a bit confused the latest firmware library from GigaDevice is 1.1.2 from June 2021 but can only be found on http://www.gd32mcu.com/en/download/0?kw=GD32VF1 so all the repos are outdated. Judging from the activity in repositories all but nuclei-sdk seem to be unmaintained, as no pull request are merged or answered.

I hope that I don't bother you with this discussion, but I think that nuclei-sdk is the better choice from a maintenance perspective and I would happily open a PR to update the supplied firmware in nuclei-sdk. If that is not an option at all, I'll switch to the suggested https://github.com/GigaDevice-Semiconductor/GD32VF103_Firmware_Library repository.

src/portable/st/synopsys/dcd_synopsys.c
KarlK90: The whole risc-v ecosystem is really young and because of the decentralized and open nature of the ISA I doubt that something universal as CMSIS is going to emerge. So NMSIS is largely Nuclei specific as the processor core have vendor specific extensions but very well designed in my opinion. It also provides a CMSIS compatibility layer. Although for this port very few parts are actually used.

.gitmodules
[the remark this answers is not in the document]
hathach: I am not bothered at all, I have looked a bit more closer. To be honest, at first I thought Nuclei is a 3rd party, but look like they are IP creator similar to ARM. And Giga put Nuclei's N200 risc-v core into their silicon. Furthermore, Nuclei seems to spend much more effort on the software/firmware side than Giga does. And I could see why you prefer to go with their sdk. Feel free to correct me if I get it wrong. 

https://github.com/Nuclei-Software/NMSIS is indeed a really great repo which is set to do what CMSIS does for ARM. I think we could actually depend on this nmsis. However, I still don't want to force any other users to use nuclei sdk just to run the usb stack. They may already be on another platform/framework, that could cause an conflict. Unless there is something that we would really need from sdk that we couldn't write it ourself.

To sum up, do you think it is good enough to just add 2 submodules
- semi official https://github.com/riscv-mcu/GD32VF103_Firmware_Library for low level mcu, I don't mind that it is old v1.0, mcu driver doesn't change much. And USB does not use much of them. 
- https://github.com/Nuclei-Software/NMSIS for providing abstraction for risc-v core for eclic enabled if needed.

This is actually very similar to stm32 with an cmisis and driver combo, https://github.com/hathach/tinyusb/blob/master/hw/bsp/stm32f103bluepill/board.mk#L2

Note: when integrating tinyusb to an application, you can freely use nuclei, and have your own board setup, just changes the include path to the corresponding. The idea is if we use an smaller set of files (mcu lib), it would be more portable.

src/portable/st/synopsys/dcd_synopsys.c
hathach: looking more at nmsis, I think it is a great repo. I will leave to you to decide whether we should use nmsis here. Note that, anything used in this dcd file will be considered as tinyusb dependency.

hw/bsp/gd32vf103_longan_nano/system_gd32vf103.c
```diff
@@ -0,0 +1,1256 @@
+/*!
```
hathach, line 1: I just notice this file in the bsp, is there any reason that you don't use the `system_gd32vf103.c` included by the firmware library instead since this file is mostly "fixed" per mcu.

KarlK90: There is a [define block](https://github.com/hathach/tinyusb/blob/3eb54d878af0492a85fab2811b5f84187c2f703c/hw/bsp/gd32vf103_longan_nano/system_gd32vf103.c#L46-L60) at the beginning of the file that where the system clock is selected by commenting and uncommenting the correct define. Which is a design flaw in my opinion because the define could be set outside. The current design is meant to copy the file into your project and ship it with your sources. While this is not optimal I would trim down the file quite a bit.

.gitmodules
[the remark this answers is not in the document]
KarlK90: > I am not bothered at all, I have looked a bit more closer. 

That is good to hear!

You are totally right about Nuclei, they provided the N200/BumbleBee Core for the GD32VF103 to GigaDevice and kind of co-developed the firmware library. I'm not sure why GigaDevice doesn't show a bit more effort with their firmware, it is a bit frustrating.

> I still don't want to force any other users to use nuclei sdk just to run the usb stack. They may already be on another platform/framework, that could cause an conflict. Unless there is something that we would really need from sdk that we couldn't write it ourself.

I'm totally with you that we don't want conflicts in other projects. I think you are referring to namespace conflicts like multiple definitions of the same symbol but correct me here. My point is that the parts that we include and actively use in the compilation of tinyusb from `nuclei-sdk` is really just the NMSIS library and the gd32vf103 firmware library with modifications from nuclei to play along nicely with NMSIS (see below). The other header and source files related to different socs are not included in the build, it is just the bare minimum. So the conflicts should not increase by using `nuclei-sdk` and should be the same as using the separate NMSIS and GD32VF103 firmware repositories.

On the other hand I had started to change the current port to use the two independent repositories as suggested and ran into trouble. The NMSIS library wants specific [defines](https://github.com/Nuclei-Software/NMSIS/blob/master/Device/Nuclei/NUCLEI_N/Include/NUCLEI_N.h#L120-L142) that describe the capabilities of the GD32VF103. Unfortunately we can't just use this file because the GD32VF103 doesn't have a FPU and has other IRQs defined. So we would have to provide this block ourselfs in a separate header file before [including nmsis_core.h in the dcd driver](https://github.com/hathach/tinyusb/blob/3eb54d878af0492a85fab2811b5f84187c2f703c/src/portable/st/synopsys/dcd_synopsys.c#L98-L99). 

But by doing so we have a board specific header-file in a rather generic driver, which is not good. The `nuclei-sdk` provides these defines in [gd32vf103.h](https://github.com/Nuclei-Software/nuclei-sdk/blob/master/SoC/gd32vf103/Common/Include/gd32vf103.h) which is different from the [gd32vf103.h](https://github.com/riscv-mcu/GD32VF103_Firmware_Library/blob/master/Firmware/GD32VF103_standard_peripheral/gd32vf103.h) that is supplied with the firmware library. And they actually modified the firmware library in other parts to play a long nicely with NMSIS framework. The GD32VF103 firmware library is really just meant to be used on its own. There maybe even more conflicts that but I haven't explored that way further. 

Given the above reasons and addressing your concerns about the conflicts could a dependency on just `nuclei-sdk` be the better option? What do you think?

src/portable/st/synopsys/dcd_synopsys.c
KarlK90: That sounds good! See my other comment for more rationale.

.gitmodules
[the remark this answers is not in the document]
hathach: > That is good to hear!

You know this much better than me, there is obvious reason why you prefer the nuclei sdk
 
> You are totally right about Nuclei, they provided the N200/BumbleBee Core for the GD32VF103 to GigaDevice and kind of co-developed the firmware library. I'm not sure why GigaDevice doesn't show a bit more effort with their firmware, it is a bit frustrating.

Yeah, I figured it out, Giga doesn't spend much effort like ST do with their stm32

> But by doing so we have a board specific header-file in a rather generic driver, which is not good. The `nuclei-sdk` provides these defines in [gd32vf103.h](https://github.com/Nuclei-Software/nuclei-sdk/blob/master/SoC/gd32vf103/Common/Include/gd32vf103.h) which is different from the [gd32vf103.h](https://github.com/riscv-mcu/GD32VF103_Firmware_Library/blob/master/Firmware/GD32VF103_standard_peripheral/gd32vf103.h) that is supplied with the firmware library. And they actually modified the firmware library in other parts to play a long nicely with NMSIS framework. The GD32VF103 firmware library is really just meant to be used on its own. There maybe even more conflicts that but I haven't explored that way further.
> 
> Given the above reasons and addressing your concerns about the conflicts could a dependency on just `nuclei-sdk` be the better option? What do you think?

My ultimate goal is not having usb stack to depend on nuclei-sdk to run. User should just compile it with any GD32VF103_Firmware_Library variant, especially one downloaded from the Giga site since that is what user will do. With that in mind, in the fact the nuclei has to modify the library to get it work with nmsis, I think we actually have to change the approach a bit.
1. We can keep using nuclei-sdk, and use it for running example only ie. bsp folder.
2. However we shouldn't use any of nuclei-sdk and/or nmsis within the `src` folder, i.e the `dcd_synopsys.c`. The reason is it will force user to use the modified firmware lib from nuclei and not be able to use **official** lib downloaded from Giga website. Even it is not well maintained, it is still what most user would use (unfortunately).

src/portable/st/synopsys/dcd_synopsys.c
hathach: I actually take this back, given the fact that nmsis requires firmware lib modification and couldn't work with official lib downloaded from Giga. Since it is simple enough, probably only ECLIC_EnableIRQ/DisableIRQ(), do you think we could do it with register level, I actually have to go this low for some of mcu ports to stay independent enough.

hw/bsp/gd32vf103_longan_nano/system_gd32vf103.c
hathach: I agree, having manually def/undef clock selection is a really bad design, they can just wrap them around `#ifndef` or simply parameterize it will make thing more portable any easier to upgrade. So yeah, I totally agree with your choice here.

src/portable/st/synopsys/dcd_synopsys.c
KarlK90: This is absolutely possible! Enabling and Disabeling the IRQs is a simple setting and clearing the MIE bit in the mstatus register. I'll figure out if we go with the gd32vf103 firmware library provided functions or some inline asm.

.gitmodules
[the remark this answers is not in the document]
KarlK90: That sounds good to me! I'll change the implementation to meet these constraints, maybe the core code in src can even be implemented without any reliance on any gd32vf103 firmawre library or nmsis/nuclei-sdk.

src/portable/st/synopsys/dcd_synopsys.c
KarlK90: Oh except for actually enabling and disabling a specific IRQ, but this is rather easy to do.

.gitmodules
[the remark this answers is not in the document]
hathach: Thanks for patiently following my suggestion, yeah, the example bsp is mostly for demonstrating/testing and maintaining. If nuclei-sdk make it easier for us, no problem at all, I wasn't sure before and kind of worry it will become an dependency due to the nmsis. If we could get it dcd_synopsys neutral then that would be no problem.

src/portable/st/synopsys/dcd_synopsys.c
hathach: Thank you, we only need to specifically enable/disable the USB IRQ only, and don't need a generic function like NVIC. And bit set/clear with correct mask and register would do the trick hopefully

KarlK90: Please have a look at https://github.com/hathach/tinyusb/pull/959/commits/c6d495d6436c797b4efd029bd16ee8f19ff0a9f0 this is the commit that removes any external dependency for the core code. Are you fine with helper functions being directly in the driver code?

.gitmodules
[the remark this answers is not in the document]
KarlK90: Perfect! I have cleaned up and fixed a bug with the systick timer not beeing reloaded. I decided to go with `nuclei-sdk` inside bsp for obvious reasons ;-). Let me know what you think!

hathach:
@KarlK90 everything look great with latest commit, however, when I tried to flash your pr to my longan nano board, it does not seems to run at all. Maybe I haven't done it correctly, 
- I tried with both your recommend toolchain from embecosm and nuclei toolchain from https://nucleisys.com/download.php .
- use dfu-util (`flash` target) to download firmware, and press reset button to reboot it.

```
$ make BOARD=gd32vf103_longan_nano all

  15884    8492    3540   27916    6d0c _build/gd32vf103_longan_nano/cdc_msc.elf

$ make BOARD=gd32vf103_longan_nano flash

Opening DFU capable USB device...
ID 28e9:0189
Run-time device DFU version 011a
Claiming USB DFU Interface...
Setting Alternate Setting #0 ...
Determining device status: state = dfuIDLE, status = 0
dfuIDLE, continuing
DFU mode device DFU version 011a
Device returned transfer size 2048
DfuSe interface name: "Internal Flash  "
Downloading to address = 0x08000000, size = 25980
Download        [=========================] 100%        25980 bytes
Download done.
File downloaded successfully
Resetting USB to switch back to runtime mode
```

However with both `cdc_msc` example and the `board_test` the on-board LED does not blink, there is no usb connection. hmm, maybe I forgot something. Here is my longan nano removed the lcd since it is kind of getting the button (hopefully it doesn't impact anything). 

IMG_1507

KarlK90:
> @KarlK90 everything look great with latest commit, however, when I tried to flash your pr to my longan nano board, it does not seems to run at all. Maybe I haven't done it correctly

Great that my changes look good.

I think the flashing problem is connected to the buggy GigaDevice DFU bootloader that reports a transfer size of 2048 bytes, `dfu-util` versions 0.10 or the current master branch use the correct size of 1024 bytes. My suggestion is to confirm that the dfu-util version does not contain the fix or use a jlink and the broken out jtag connector. Removing the LCD is fine and does not interfere with the function at all.

hathach:
Ah you are right, upgrading to dfu-util 0.10 and it successfully flash the device (even tried to force the upload-size with 1024 on 0.9 just to test, but it doesn't help). Thanks for suggestion, I will wire up to use it with jlink for faster developement.

The cdc_msc example have the cdc echo fine, however, the msc does not, I hook it into analyzer. Look like there is some issue when host read  4k of data. Notice that the data is **corrupted** with 0x55AA repeatedly somehow. Host give up and decide device isn't properly formatted. It is probably buffer/endpoints config/overflow issue etc ..  which is typical for a new port based on existing dcd driver. I will spend sometime this week (or so) to poke at GDVF103 manual to see if I could spot any issue. Hopefully it would be only an minor change.

PS: It is not important but the blink timer is a bit slow, it should toggle 1000ms when connected, 250ms when not. We could look into this later on  just to make it consistent with other ports. 

image

KarlK90:
Good to hear that the upgrade solved the problem with the flashing.

You are right about the timer, I set the wrong reload value. The systick timer is driven by the AHB clock divided by 4. On the current implementation the AHB clock should equal the system clock (72Mhz by default). I quickly pushed a fix, hopefully that corrects the blinking frequency.

Thank you for the debugging effort, you clearly have tools and knowledge that I lack :muscle:. The tinyusb HID examples worked fine from what I can remember. I had ported the stm32/synopsis otg driver in chibios to the GD32VF103 (just renaming registers) but it worked fine with HID device applications as well. I don't know if this is of any help but [here it is anyways](https://github.com/ChibiOS/ChibiOS-Contrib/blob/chibios-20.3.x/os/hal/ports/GD/GD32VF103/OTG/hal_usb_lld.c).

hathach:
yeah, both HID and CDC with example config only send up to 64 bytes per fifo, this issue probably only happens when doing large buffer transferring back to back, may involve race condition. Will troubleshoot it later on.

hathach (approved):
@KarlK90  Thank you very much for the PR and your patient. I have spent an hour or so trying to see if I could spot any obvious issue. Unfortunately, I couldn't, it may need a overall revise of the dcd. which I don't have the bandwidth to do at the moment. It is relatively normal for an new port, I have to do something similar with esp32 (also using synopsys IP). Therefore I think it is best to merge this to master and then file an issue so that other people can get on helping to resolve this. I also made a couple of changes to this PR mostly to get passed CI build
- Move the bsp to its own familay, also rename board to `sipeed_longan_nano`, that would help to add more boards with similar mcu.
- Change default CROSS_COMPILE to `riscv-none-embed-` installed with xpack, mainly is for CI, since that is toolchain used by fomu (another existing risc-v port). there is also comment for nuclei and embecosm as well in the family.mk if user want to change.

KarlK90:
@hathach Thank you for you continued effort and merging this PR, the changes look good!

Hopefully the bug with the dcd and this particular peripheral is not to complicated so it can be resolved in the future. I'll try to poke at it a bit, maybe I can find something.

hathach:
I am the one to say thanks, will definitely try to solve it in the future.

PR #967, opened 2021-07-16
Add openinput to the 'Uses' section (aka projects that use Tusb)

perigoso:
Signed-off-by: Rafael Silva <perigoso@riseup.net>

**Describe the PR**
The project is young, but we already rely heavily on Tusb and we are planning on working on it in foreseeable future as well as develop a community around it, so i think it makes sense to add here.

[Project documentation.](https://openinput.readthedocs.io/en/latest/)


hathach (approved):
thank you for the addition

Issue #968, opened 2021-07-18
dcd_synopsis driver fails to flush data of length 256 bytes when in HS mode


BennyEvans:
### Operating System

Windows 10

### Board

STM32H7

### Firmware

examples/device/dcd_msc (tinyUSB 10.1)

### What happened ?

The dcd_synopsys.c driver fails to flush when the data to flush is exactly 256 bytes in length in HS mode. Writing more data will eventually cause a flush but calling to flush with 256 bytes of data does not actually perform the transfer.

The line that I believe causes the issue is:
https://github.com/hathach/tinyusb/blob/21bfd11683c4bb6e0996130bd1061a97d3049121/src/portable/st/synopsys/dcd_synopsys.c#L659

When total_bytes is 256 and xfer->max_size is 512, this line will evaluate to 256, but as the result type is a uin8_t, the result is 0.  Further on in the code, this prevents num_packets from being incremented.

I'll do more testing but I believe the soultion to this might simply be to change the line to:
`uint16_t const short_packet_size = total_bytes % xfer->max_size;`

Please let me know if you require anymore information. I'll try to perform some more testing and put together a pull request over the next week or so when I have some free time.

**UPDATE:**
There are two instances of this by the look of it:
https://github.com/hathach/tinyusb/blob/21bfd11683c4bb6e0996130bd1061a97d3049121/src/portable/st/synopsys/dcd_synopsys.c#L659

https://github.com/hathach/tinyusb/blob/21bfd11683c4bb6e0996130bd1061a97d3049121/src/portable/st/synopsys/dcd_synopsys.c#L690

### How to reproduce ?

1) Load the examples/device/dcd_msc on a board using the dcd_synopsys.c driver in HS mode.
2) From your PC, send a chunk of data with length 256, expecting it to be echoed back.
3) You will not receive a response until you send more data. Other lengths will echo back immediately.

### Debug Log

Logs after sending data of length 256 bytes and 257 bytes below. Note that the 256 byte transfer is not completed.

256 byte transfer:

> USBD Setup Received A1 21 00 00 00 00 07 00 
>   CDC control request
>   Get Line Coding
>   Queue EP 80 with 7 bytes ...
> USBD Xfer Complete on EP 80 with 7 bytes
>   CDC control complete
>   Queue EP 00 with 0 bytes ...
> USBD Xfer Complete on EP 00 with 0 bytes
> USBD Xfer Complete on EP 02 with 256 bytes
>   CDC xfer callback
[...]

PR #969, opened 2021-07-19
Fix for dcd_synopsys driver integer overflow in HS mode (issue #968).

BennyEvans:
Fixes an issue (#968) in the dcd_synopsys.c driver where it fails to flush because of an integer overflow in HS mode.


hathach (approved):
superb ! thank you for your PR, you really got an eagle eye to spot this issue.

PR #980, opened 2021-07-27
Waveshare OpenH743I-C BSP

BennyEvans:
Waveshare OpenH743I-C Board Support Package.

### DIFF
--- a/hw/bsp/waveshare_openh743i/board.c
+++ b/hw/bsp/waveshare_openh743i/board.c
@@ -1,0 +1,483 @@
+/*
+ * The MIT License (MIT)
+ *
+ * Copyright (c) 2019
+ *    Benjamin Evans
+ *    William D. Jones (thor0505@comcast.net),
+ *    Ha Thach (tinyusb.org)
+ *    Uwe Bonnes (bon@elektron.ikp.physik.tu-darmstadt.de
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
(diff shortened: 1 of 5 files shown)


hw/bsp/waveshare_openh743i/board.c
```diff
@@ -0,0 +1,495 @@
...
+  USB_OTG_HS->GOTGCTL |= USB_OTG_GOTGCTL_BVALOEN;
+  USB_OTG_HS->GOTGCTL |= USB_OTG_GOTGCTL_BVALOVAL;
+#endif
+
+  // Force device mode
+  USB_OTG_HS->GUSBCFG &= ~USB_OTG_GUSBCFG_FHMOD;
+  USB_OTG_HS->GUSBCFG |= USB_OTG_GUSBCFG_FDMOD;
+
+  //Reset the PHY - Change the delays as you see fit
+  timer_board_delay(5U); //Delay 5ms
```
hathach, line 418: since we don't need a strict timing, we could actually use nop or better use DWT from debug core to delay https://github.com/adafruit/Adafruit_nRF52_Arduino/blob/master/cores/nRF5/delay.h#L103 . Since I think this is primary reason for not integrating this board into the existing stm32h7 family folder.

hathach (approved):
superb! thank you very much for your PR and effort to put the board hw issue note for user with same hardware. I understand there is a few special config for this board that prevent it to be part of the stm32h7 family here https://github.com/hathach/tinyusb/tree/master/hw/bsp/stm32h7/boards. e.g
- Different ULPI DIR, NXT pin selection than the reset (this is easy)
- Require an way to delay resetting PHY, but it is doable with nop and/or dwt I think

However, I still prefer to integrate it there since it reduce the maintenance effort later on. I will merge this PR as it is now and will try to do follow up pr to integrate it later on. Please help me to test with follow up PR since I don't own the hw to test with. Thank you

BennyEvans:
@hathach happy to help where I can and happy to test on my hardware if you integrate it later on. I agree that integrating it will be a better solution and less maintenance.

If you don't get around to integrating it within the next few weeks then I can have a go at it if you'd like? But I then run into the same problem of not having the other hardware to test on and ensure I haven't broken anything on those boards.

hathach:
Thanks, I will try to put up an PR (if I could) and tag you for review.

PR #992, opened 2021-08-03
net_lwip_webserver: utilize pbuf_copy_partial()

majbthrd:
The existing example code duplicates functionality that it turns out can be achieved with an existing function in lwIP (pbuf_copy_partial).  Using pbuf_copy_partial() saves about 24 bytes in the net_lwip_webserver executable size and results in ever so slightly less code to maintain.

### DIFF
--- a/examples/device/net_lwip_webserver/src/main.c
+++ b/examples/device/net_lwip_webserver/src/main.c
@@ -181,9 +181,7 @@ uint16_t tud_network_xmit_cb(uint8_t *dst, void *ref, uint16_t arg)
 
   (void)arg; /* unused for this example */
 
-  pbuf_copy_partial(p, dst, p->tot_len, 0);
-
-  return p->tot_len;
+  return pbuf_copy_partial(p, dst, p->tot_len, 0);
 }
 
 static void service_traffic(void)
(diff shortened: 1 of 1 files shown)


examples/device/net_lwip_webserver/src/main.c
```diff
@@ -178,21 +178,12 @@ bool tud_network_recv_cb(const uint8_t *src, uint16_t size)
...
-  {
-    memcpy(dst, (char *)q->payload, q->len);
-    dst += q->len;
-    len += q->len;
-    if (q->len == q->tot_len) break;
-  }
+  pbuf_copy_partial(p, dst, p->tot_len, 0);
 
-  return len;
+  return p->tot_len;
```
duempel, line 186: Since `pbuf_copy_partial` returns 'the number of bytes copied, or 0 on failure' it might be more logical to return this value instead of always `p->tot_len`.

duempel (commented):
Just tested your changes and it works fine as expected. Maybe the return value could be changed: `return pbuf_copy_partial(p, dst, p->tot_len, 0);`
What do you think about it?

examples/device/net_lwip_webserver/src/main.c
majbthrd: That makes sense; I have added a commit that does just that.  Thank you for the review and the suggestion!

hathach (approved):
thank @majbthrd  for the PR and @duempel for reviewing. It looks good.

PR #1017, opened 2021-08-11
da1469x: Handle VBUS interrupt

kasjer:
**Describe the PR**
For salf-powered code should handle VBUS changes.
Function tusb_vbus_chaneged() was present but not used in bsp;

dcd_connect() was called only at startup which is fine for
bus-powered devices.
If device was self-power and started without VBUS present
some registers would not be correctly set.

Now tusb_vbus_changed() is called whenever actual change on VBUS
is detected starting and stopping USB state machine.

**Additional context**
Tested only with da1469x_dk-pro (not USB stick)

### DIFF
--- a/hw/bsp/da14695_dk_usb/da14695_dk_usb.c
+++ b/hw/bsp/da14695_dk_usb/da14695_dk_usb.c
@@ -46,6 +46,14 @@ void USB_IRQHandler(void)
 
 #define BUTTON_PIN      6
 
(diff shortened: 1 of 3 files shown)


kasjer:
After some user input I received, I'm going to modify this so VBUS detection will be an option and not necessity.

kasjer:
After inspecting history I decided to abandon in this form.
I will prepare another PR with VBUS handling that is more in sync with all other drivers.

Issue #179, opened 2019-09-23
Repeated disconnections causes the cdc_msc demo to stop functioning.


cr1901:
**Describe the bug**
While testing whether the Synopsys driver survives a USB disconnect and reset, I decided to test repeated disconnects. After enough disconnects and reconnects, the `cdc_msc` demo will stop working on my laptop; both the Mass Storage device fails to appear, and the serial port will not send or echo data. This is despite me hearing the "device connected" jingle.

**Set up (please complete the following information):**
 - OS: Windows 7
 - Board: stm32f407disco
 - Firmware Code: e.g examples/device/cdc_msc

**To Reproduce**
1. Compile the `cdc_msc` demo. I used the stm32f407disco board. Make sure to use [my branch](https://github.com/cr1901/tinyusb/tree/syn-unplug)- I thought the bug was related, but I'm not sure anymore.
2. Repeatedly plug and unplug the USB cable into the host/device (_this only works for boards which don't rely on the user USB cable for power_).
3. Eventually, the Mass Storage component will fail to mount in Explorer. Additionally, while the serial port will appear according to Windows, and opening it will succeed, the port will never send data (banner or otherwise).

**Expected behavior**
Repeated connections and disconnections should consistently bring back both the CDC and MSC interfaces. The MSC interface should always be responding to IN packets.

**Screenshots**
Not applicable.

**Additional context**
I have attached a zip file of the bus activity from my Beagle: [disconnect.zip](https://github.com/hathach/tinyusb/files/3640986/disconnect.zip)

The MSC interface eventually stops responding to IN packets after enough disconnects/reconnects. Eventually Windows gives up trying to talk to the device completely; this naturally also breaks the CDC portion of the demo. Unfortunately at present, I don't know enough about MSC or SCSI to debug this myself.

The CDC portion of the demo works fine, even after the MSC portion of the demo stops responding, up until Windows stops talking to the device completely.

_I am not able to reproduce this in `hid_composite`._

pigrew:
I have been noticing this issue (or similar), too, using USBTMC. 

Does the USBD queue need to be cleared when queueing a reset event? What if transfers were queued, reset received, transfers sent, then core reset happens?

jeremyherbert:
I think I have tripped on this bug as well. I have the tinyusb device (STM32F042) connected to a powered hub, and the hub is connected to my laptop. With enough sleep/wakeup cycles on the laptop, it seems to be disconnected.

[...]

PR #1021, opened 2021-08-13
Fix dcd_set_address bug when called more than once

tringis:
**Describe the PR**
dcd_set_address() is broken when called more than once, e.g. when plugging in the device into a different port. The problem is that it only sets new bits in the register, it doesn't clear any. Setting addres 1 followed by address 2 therefore causes the address to be set to (1|2) = 3.

### DIFF
--- a/src/portable/st/synopsys/dcd_synopsys.c
+++ b/src/portable/st/synopsys/dcd_synopsys.c
@@ -192,6 +192,10 @@ static void bus_reset(uint8_t rhport)
     out_ep[n].DOEPCTL |= USB_OTG_DOEPCTL_SNAK;
   }
 
+  // clear device address
+  dev->DCFG &= ~USB_OTG_DCFG_DAD_Msk;
+
+  // TODO should probably assign value when reset rather than OR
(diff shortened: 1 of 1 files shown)


hathach (approved):
Yeah, right, this is spot-on, it may be related to several other issues with unplug/re-plug stm32 (may solve issue #179). Though I think we should also clear the DAD within the bus_reset()  as well to make it consistent. Since it is an easy change, I have already push the commit  to your fork. Thank you for the fix.

tringis:
Thanks for the merge! I did notice several more places that looked very suspicious. I plan to go though it from top to bottom, but it may take a while before I get to it as I have a big deadline coming up.

hathach:
I am the one to say thank, looking forward to your PRs

cr1901:
Oops, this was my bad. It's been there since [the beginning](https://github.com/hathach/tinyusb/pull/38/files#diff-414c74b8f427abd9dae3f0ec318375f305b47911c744c59b8dc8ee1e853cdb61R184). Thanks @tringis for finding this :)!

hathach:
> Oops, this was my bad. It's been there since [the beginning](https://github.com/hathach/tinyusb/pull/38/files#diff-414c74b8f427abd9dae3f0ec318375f305b47911c744c59b8dc8ee1e853cdb61R184). Thanks @tringis for finding this :)!

No worry, I have tons of bug all over the places as well 😅😅

PR #1024, opened 2021-08-16
Fix nuc126 buffer copy with IAR

HiFiPhile:
**Describe the PR**
DCD buffer copy doesn't work correctly with IAR, after investigation IAR use optimized STM/LDM instructions in `memcpy` but there must be some access restrictions on USB buffer just like STM32 fsdev.

Use byte copy just like StdDriver.

And fix a warning.

### DIFF
--- a/src/portable/nuvoton/nuc121/dcd_nuc121.c
+++ b/src/portable/nuvoton/nuc121/dcd_nuc121.c
@@ -101,6 +101,11 @@ static void usb_detach(void)
   USBD->SE0 |= USBD_SE0_SE0_Msk;
 }
 
+static void usb_memcpy(uint8_t *dest, uint8_t *src, uint16_t size)
+{
+  while(size--) *dest++ = *src++;
+}
+
 static void usb_control_send_zlp(void)
 {
   USBD->EP[PERIPH_EP0].CFG |= USBD_CFG_DSQSYNC_Msk;
@@ -153,7 +158,7 @@ static void dcd_in_xfer(struct xfer_ctl_t *xfer, USBD_EP_T *ep)
   else
 #endif
   {
-    memcpy((uint8_t *)(USBD_BUF_BASE + ep->BUFSEG), xfer->data_ptr, bytes_now);
+    usb_memcpy((uint8_t *)(USBD_BUF_BASE + ep->BUFSEG), xfer->data_ptr, bytes_now);
   }
 
   ep->MXPLD = bytes_now;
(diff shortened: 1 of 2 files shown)


src/portable/nuvoton/nuc121/dcd_nuc121.c
```diff
@@ -153,7 +158,7 @@ static void dcd_in_xfer(struct xfer_ctl_t *xfer, USBD_EP_T *ep)
   else
 #endif
   {
-    memcpy((uint8_t *)(USBD_BUF_BASE + ep->BUFSEG), xfer->data_ptr, bytes_now);
+    usb_memcpy((uint8_t *)(USBD_BUF_BASE + ep->BUFSEG), xfer->data_ptr, bytes_now);
```
hathach, line 161: to be honest, I feel a bit odd about this change, would it work if we cast the buffer with `volatile uint8_t*` instead of uint8_t*

HiFiPhile: `volatile` doesn't work as ` Error[Pe167]: argument of type "uint8_t volatile *" is incompatible with parameter of type "void *"` 

In fact read setup packet from buffer with `memcpy` (LDM instruction) works, only write doesn't work.

I read the manual, it didn't mention the access type.

hathach: I checked the manual as well, it is not mentioned. But It is very likely that USB SRAM is byte access. So yeah, this PR is spot on and needed. 
https://github.com/majbthrd/nuc_driver/blob/master/nuc126/StdDriver/inc/usbd.h#L524

Only 1 request, would you mind adding an comment at this line and usb_memcpy below to explain the reason something like `although it is not mentioned in manual, but the USB SRAM seems to only support byte access and memcpy could possibly do it by words etc ...`

src/portable/nuvoton/nuc121/dcd_nuc121.c
```diff
@@ -101,6 +101,11 @@ static void usb_detach(void)
   USBD->SE0 |= USBD_SE0_SE0_Msk;
 }
 
+static void usb_memcpy(uint8_t *dest, uint8_t *src, uint16_t size)
```
hathach, line 104: minor but maybe make this function inline

hathach (approved):
I realized that dcd nuc120 is very similar to nuc121, and therefore also apply the same changes to it as well. This should be ready to merged once ci passed.

PS: due to the dcd similarity, I open #1029 as reminder to merge them together to make it easier to maintain in the future.

HiFiPhile:
I've spotted another issue of device stopped functioning after PC reboot with set address failure.
Clear `assigned_address` in dcd_reset() maybe fixed it but not well tested.

hathach:
`assigned_address` can actually changed to use the dcd_edpt0_status_complete(). other port also use this for set_address(). I think the API is added after the nuc port. I could do this but I need to find the nuc board in my drawer first :D

PR #1025, opened 2021-08-17
add hid_boot_interface example

hathach:
**Describe the PR**
- Add new hid boot interface example to demonstrate using boot interface with boot keyboard and boot mouse. 

Note: I don't have an true bios to test with, if anyone could try it out, that would be great.


ndt1896:
Re: "I don't have an true bios to test with, if anyone could try it out, that would be great."

-> I tested on different hardware and confirmed keyboard and mouse both worked in a CSM BIOS.

hathach:
> -> I tested on different hardware and confirmed keyboard and mouse both worked in a CSM BIOS.

Thank you for testing it out on that many platforms. getting merged now

PR #1027, opened 2021-08-17
Remove redundant SOF processing from the Renesas RX family.

kkitayam:
**Describe the PR**
The same logic regarding the resume signal was implemented by usbd.
See also: #1023

**Additional context**
I have confirmed the patch works `hid_composite` example on rx65n_taget.

### DIFF
--- a/src/portable/renesas/usba/dcd_usba.c
+++ b/src/portable/renesas/usba/dcd_usba.c
@@ -139,7 +139,6 @@ typedef struct
 {
   pipe_state_t pipe[10];
   uint8_t ep[2][16];   /* a lookup table for a pipe index from an endpoint address */
-  uint8_t suspended;
 } dcd_data_t;
 
 //--------------------------------------------------------------------+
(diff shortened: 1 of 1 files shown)


hathach:
great thanks, I was wanting to do this as well as follow up, since #1023 is focused on nrf. Though I think we should only enable SOF when suspended and disable it afterwards. The reason is that SOF is currently not used at all by USBD https://github.com/hathach/tinyusb/blob/master/src/device/usbd.c#L1082 didn't submit the SOF event to usbd task queue. The reason is that I am not entirely sure if SOF is really needed for driver, mostly only audio/video where it has strict timing constraint with the frame and would like to send it as soon as usb frame start or just have an synch point etc ... but to be honest, I am not entirely sure. 

SOF is triggered very often 1 per 1ms and could waste cpu load while usbd does not make any use of it. Maybe having and define USE_SOF as synopsys https://github.com/hathach/tinyusb/blob/master/src/portable/st/synopsys/dcd_synopsys.c#L34 to make it easier to switch on and off later on when needed. Sorry it is not too consistent because we haven't figured everything out yet.

kkitayam:
OK. I understood that in the current usbd, SOF events are only used for resuming. I will try if SOF events can be enabled during suspended only.

hathach (approved):
superb !! thank you very much for the PR with SOF disabled.

PR #1033, opened 2021-08-19
merge waveshare to h7 family

hathach:
**Describe the PR**
Follow up to #980, added as `waveshare_openh743i_2` under h7. Will rename and remove existing one once verified working. @BennyEvans would you mind testing it out on your board, take your time, there is no hurry.

```
make BOARD=waveshare_openh743i_2
```


BennyEvans:
@hathach I'll take a look and test on the weekend.

hathach:
> @hathach I'll take a look and test on the weekend.

thanks, take your time :)

BennyEvans:
@hathach I've just tested and most seems to be working except for the FreeRTOS example. I believe you still need `SysTick->CTRL &= ~1U;` at the beginning of the initialisation to ensure you don't enter the systick ISR. I'm not sure if you want to add that in a broader sense or just for this board.

I haven't tested every example but have tested a few in both FS and HS. UART logging all seems to be working well. Once the FreeRTOS examples are fixed, I'll perform some more testing. My board is currently set up for some other functionality so I'll disconnect everything and test from there.

hathach:
Thank you @BennyEvans for testing this out, yeah you are right, it is best to explicitly disabled systick to prevent its ISR triggered before scheduler start. It should be done for other family as well, but it is less of an issue since most of them board_init() finishes fast enough before the tick expire. I have expanded this to other stm families, should do for all, but  I will do it gradually later on.

From your testing feedback, I think the PR is good for merged, no need for further testing especially you are hooking it up for other work. I have tested with h743eval, it is mostly for the PHY reset on the waveshare. which is confirmed to work.

BennyEvans:
@hathach good to hear. I'll be jumping back into some USB development soon so will be able to give it a thorough test then. In the meantime, if there are any discussions related to this board that you need me to help out with, don't hesitate to mention me (so that I get an email) and I can help out. Thanks for all your help.

Issue #1040, opened 2021-08-22
Potential fifo racing on self powered device


HiFiPhile:
### Operating System

Windows 10

### Board

Custom NUC126

### Firmware

In loop:
```
   uint8_t data[64];
    if(tud_vendor_write_available() >= sizeof(data))
    {
        tud_vendor_write(data, sizeof(data));
    }     
```

### What happened ?

This could happen on CDC & Vendor class which use fifo when USB is re-plugged.

Take Vendor class as example, fifo is cleared in `vendord_reset` when re-plugged, if fifo is fully written between `vendord_reset` and `vendord_open`, it will enter a deadlock condition that fifo is full but no transfer is scheduled.

Either `maybe_transmit` or `tu_fifo_clear` need to be called in `vendord_open`.

### How to reproduce ?

1. Run a minimum vendor class on self-powered device
2. Write data in a loop
3. Plug & Unplug multiple times
4. Tx fifo is written but no data is transfered

hathach:
closed by #1041

HiFiPhile:
Do you want to add `tud_cdc_n_write_flush`  in `cdcd_open` ?

Issue #900, opened 2021-06-18
Only first bulk OUT EP write works


diggit:
###  Set Up

- Archlinux, distro kernel 5.12.8, 
- custom board with STM32H730 and USB PHY USB3343 (may provide schematics on request)
- **TinyUSB version** b34724215bea28ef93fd80ed9c2e4ddc87874093 with [synopsys hotfix ](https://github.com/diggit/tinyusb/commit/69e9397b8ac5201db2a79041593e76b31563c9d9) #496
#### Firmware
- based on RTOS ([modified Argon](https://github.com/diggit/argon-rtos))
- osal implemented and queue passes events to tud_task
##### EPs
```c
uint8_t const desc_hs_configuration[] = {
	// Config number, interface count, string index, total length, attribute, power in mA
	TUD_CONFIG_DESCRIPTOR(1, ITF_NUM_TOTAL, 0, CONFIG_TOTAL_LEN, TUSB_DESC_CONFIG_ATT_REMOTE_WAKEUP, 500),

	TUD_VENDOR_DESCRIPTOR(ITF_NUM_VENDOR_0, 4, EPNUM_VENDOR_0, 0x80 | EPNUM_VENDOR_0, 512)
};
```

### Describe The Bug
Driver callback is called only after first bulk OUT xfer.

### To Reproduce

1. Write more than once from master to bulk OUT EP.
3. Observe, that vendor callback is called only in response to first xfer.

### Logs
I've added logging of `int_status` in the beginning of `dcd_int_handler` in `dcd_synopsys.c` (`TU_LOG2("(IS %08X)", int_status);`)
This log is enclosed in parentheses for easier reading due to its async nature.
There is less interrupts during second xfer.

#### First try
```
(IS C4808428)USBD Resume 
(IS 44808438)(IS 44808428)
USBD Setup Received 80 00 00 00 00 00 02 00 
  Get Status
  Queue EP 80 with 2 bytes ... (IS 44848428)(IS 44848428)OK
USBD Xfer Complete on EP 80 with 2 bytes
  Queue EP 00 with 0 bytes ... (IS 44808438)(IS 44808428)OK

USBD Setup Received 00 09 01 00 00 00 00 00 
  Set Configuration
  Queue EP 80 with 0 bytes ... O(IS 44848428)K
USBD Xfer Co(IS 44808438)(IS 44808428)mplete on EP 00 with 0 bytes
USBD Xfer Complete on EP 80 with 0 bytes
[...]

PR #1041, opened 2021-08-22
Fix vendor fifo deadlock, add tud_vendor_n_read_flush

HiFiPhile:
**Describe the PR**

- Fix vendor fifo deadlock in #1040
- Add tud_vendor_n_read_flush


hathach (approved):
Thank you look good, note vendor driver is lacking behind and is not thread safe if you are using rtos (mainly due to lack of time) . https://github.com/hathach/tinyusb/issues/900#issuecomment-870261839

PR #1044, opened 2021-08-23
Feature k32l2b

nxf58843:
Adding support for K32L2B

This builds on the KL25Z and RT10xx support and uses the existing mcux-sdk submodule.


hathach (approved):
thank you for the PR, I don't have this board to test with, but it looks good for the merge

PR #1062, opened 2021-08-31
Improvements for dcd_transdimension

HiFiPhile:
**Describe the PR**
1st part of #926.

- Add xfer_fifo support, fifo must aligned to 4k boundary and size must be multiple of 4k.
- Add dcd_edpt_close and iso xfer.

### DIFF
--- a/src/portable/nxp/transdimension/dcd_transdimension.c
+++ b/src/portable/nxp/transdimension/dcd_transdimension.c
@@ -516,7 +516,8 @@ bool dcd_edpt_xfer_fifo (uint8_t rhport, uint8_t ep_addr, tu_fifo_t * ff, uint16
     }
     else
     {
-      // TODO we need to carry the wrapped length after the linear part complete
+      // TODO we may need to carry the wrapped length after the linear part complete
+      // for now only transfer up to linear part
     }
   }
 
@@ -560,7 +561,6 @@ static void process_edpt_complete_isr(uint8_t rhport, uint8_t epnum, uint8_t dir
   }
   
   // only number of bytes in the IOC qtd
-  // TODO there is still a case with xfer_fifo with additional wrapped buffer to fullfil the requested length
   dcd_event_xfer_complete(rhport, tu_edpt_addr(epnum, dir), xferred_bytes, result, true);
 }
 
--- a/src/portable/nxp/transdimension/dcd_transdimension.c
+++ b/src/portable/nxp/transdimension/dcd_transdimension.c
@@ -298,51 +298,30 @@ void dcd_disconnect(uint8_t rhport)
 
 static void qtd_init(dcd_qtd_t* p_qtd, void * data_ptr, uint16_t total_bytes)
 {
+  // Force the CPU to flush the buffer. We increase the size by 31 because the call aligns the
+  // address to 32-byte boundaries. Buffer must be word aligned
+  CleanInvalidateDCache_by_Addr((uint32_t*) tu_align((uint32_t) data_ptr, 4), total_bytes + 31);
+
   tu_memclr(p_qtd, sizeof(dcd_qtd_t));
 
-  p_qtd->next        = QTD_NEXT_INVALID;
-  p_qtd->active      = 1;
-  p_qtd->total_bytes = p_qtd->expected_bytes = total_bytes;
+  p_qtd->next            = QTD_NEXT_INVALID;
+  p_qtd->active          = 1;
+  p_qtd->total_bytes     = p_qtd->expected_bytes = total_bytes;
+  p_qtd->int_on_complete = true;
 
   if (data_ptr != NULL)
   {
-    p_qtd->buffer[0]   = (uint32_t) data_ptr;
+    p_qtd->buffer[0] = (uint32_t) data_ptr;
+
+    uint32_t const bufend = p_qtd->buffer[0] + total_bytes;
     for(uint8_t i=1; i<5; i++)
     {
-      p_qtd->buffer[i] |= tu_align4k( p_qtd->buffer[i-1] ) + 4096;
-    }
-  }
-}
+      uint32_t const next_page = tu_align4k( p_qtd->buffer[i-1] ) + 4096;
+      if ( bufend <= next_page ) break;
 
-static void qtd_init_fifo(dcd_qtd_t* p_qtd, tu_fifo_buffer_info_t *info, uint16_t total_bytes)
-{
-  tu_memclr(p_qtd, sizeof(dcd_qtd_t));
+      p_qtd->buffer[i] = next_page;
 
-  p_qtd->next        = QTD_NEXT_INVALID;
-  p_qtd->active      = 1;
-  p_qtd->total_bytes = p_qtd->expected_bytes = total_bytes;
-
-  // Fifo length has been trimmed to total_bytes
-  int16_t len_lin  = info->len_lin;
-
-  if (len_lin != 0)
-  {
-    p_qtd->buffer[0]   = (uint32_t) info->ptr_lin;
-
-    len_lin -= 4096 - ((uint32_t) info->ptr_lin - tu_align4k((uint32_t) info->ptr_lin));
-
-    // Set linear part
-    uint8_t i = 1;
-    for(; i<5; i++)
-    {
-      if (len_lin <= 0) break;
-      p_qtd->buffer[i] |= tu_align4k( p_qtd->buffer[i-1] ) + 4096;
-      len_lin -= 4096;
-    }
-    // Set wrapped part
-    for(uint8_t page = 0; i<5; i++, page++)
-    {
-      p_qtd->buffer[i] |= (uint32_t) info->ptr_wrap + 4096 * page;
+      // TODO page[1] FRAME_N for ISO transfer
     }
   }
 }
@@ -445,11 +424,17 @@ void dcd_edpt_close(uint8_t rhport, uint8_t ep_addr)
   dcd_reg->ENDPTCTRL[epnum] &=~(ENDPTCTRL_ENABLE << (dir ? 16 : 0));
 }
 
-bool dcd_edpt_xfer(uint8_t rhport, uint8_t ep_addr, uint8_t * buffer, uint16_t total_bytes)
+static void qhd_start_xfer(uint8_t rhport, uint8_t epnum, uint8_t dir)
 {
   dcd_registers_t* dcd_reg = _dcd_controller[rhport].regs;
-  uint8_t const epnum = tu_edpt_number(ep_addr);
-  uint8_t const dir   = tu_edpt_dir(ep_addr);
+  dcd_qhd_t* p_qhd = &_dcd_data.qhd[epnum][dir];
+  dcd_qtd_t* p_qtd = &_dcd_data.qtd[epnum][dir];
+
+  p_qhd->qtd_overlay.halted = false;            // clear any previous error
+  p_qhd->qtd_overlay.next   = (uint32_t) p_qtd; // link qtd to qhd
+
+  // flush cache
+  CleanInvalidateDCache_by_Addr((uint32_t*) &_dcd_data, sizeof(dcd_data_t));
 
   if ( epnum == 0 )
   {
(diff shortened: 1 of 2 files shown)


src/portable/nxp/transdimension/dcd_transdimension.c
```diff
@@ -314,6 +316,39 @@ static void qtd_init(dcd_qtd_t* p_qtd, void * data_ptr, uint16_t total_bytes)
...
+    {
+      if (len_lin <= 0) break;
+      p_qtd->buffer[i] |= tu_align4k( p_qtd->buffer[i-1] ) + 4096;
+      len_lin -= 4096;
+    }
+    // Set wrapped part
+    for(uint8_t page = 0; i<5; i++, page++)
+    {
+      p_qtd->buffer[i] |= (uint32_t) info->ptr_wrap + 4096 * page;
+    }
```
hathach, line 348: I don't quite understand these lines, are you trying to queue both linear and wrap buffer into one TD ?  Which I don't think is possible, we probably need 2 TDs for this.

src/portable/nxp/transdimension/dcd_transdimension.c
```diff
@@ -353,14 +385,25 @@ bool dcd_edpt_open(uint8_t rhport, tusb_desc_endpoint_t const * p_endpoint_desc)
...
+    p_qhd->zero_length_termination = 1;
+  }
+
   CleanInvalidateDCache_by_Addr((uint32_t*) &_dcd_data, sizeof(dcd_data_t));
 
   // Enable EP Control
   dcd_registers_t* dcd_reg = _dcd_controller[rhport].regs;
+
+  // Clear EP type
+  dcd_reg->ENDPTCTRL[epnum] &=~(0x03 << (dir ? 18 : 2));
```
hathach, line 405: nice catch.

hathach (commented):
thank you for your great PR as usual. Though can you explain a bit for the usage of ptr wrap in the qtd_init_fifo()

src/portable/nxp/transdimension/dcd_transdimension.c
HiFiPhile: Yes I'm trying to queue both parts in 5*4kB TDs. Fifo need to be 4k aligned and has a size of multiple of 4K, for a max of 20K.
TD[0] can start at a arbitrary address and TD[1]-TD[4] must start at 4k boundary.

Take an example of a 8k fifo filled between 2000-100:
image

TD[0] is set to offset of `fifo->ptr_lin`, then each TD is aligned to 4k until len_lin is finished.

Then it wrap to top and start at offset 0.

hathach: I think you mis-understand the TD specs, within this function, it only preps 1 TD. The buffer[] page array is only used when the total bytes is large enough and memory cross the 4k boundary e.g when total bytes is 20 KB.

HiFiPhile: > within this function, it only preps 1 TD

You're right. 

In fact I mean 5 pointers within the same TD, sorry for the confusion.
image

> we probably need 2 TDs for this.

I've tried 2 TDs, one for linear and one for wrapped part. In this case it cut the data into 2 separate transfers, for example cdc will receive half data.

hathach: we can transfer 1 TD after another instead of using 2 TDs at once. Can you update the above code to initialize the buffer page correctly. It should only affected by ptr_lin, and is mostly identical to existing qtd_init(). In fact, I think we should just use the qtd_init(ptr_lin, len) instead of having additional qtd_init_fifo()

HiFiPhile: It won't work when you need continuous data, for example in UAC2.

And host will think the transfer is finished since len_lin can be a short length packet.

At the beginning I've already tried using 2 TDs since it's easier to implement, doesn't work at all.

hathach: The current TD buffer setup is not correct in general, the requirement with 4K aligned is too specific to make its way into the buffer page array set up. And this does not work with other MCU dcd. 

For ISO, you have to understand this limit and need to declare the fifo depth to be multiple of ISO frame. Re-config the fifo length to even smaller depth than its buffer when needed when the ISO packet size is changed is better implementation. E.g fifo buffer is 4x1024, but should be changed to 40x100 when iso packet changed from 1024 to 100.

HiFiPhile: > the requirement with 4K aligned is too specific to make its way into the buffer page array set up. And this does not work with other MCU dcd.

Yes it is specific, so I've introduced a on/off switch in #1063 to not enable fifo_xfer by default, except explicitly.

> For ISO, you have to understand this limit and need to declare the fifo depth to be multiple of ISO frame. 

That's won't work since sample length could be +-1, wrapped part has to be always considered.

hathach: > Yes it is specific, so I've introduced a on/off switch in #1063 to not enable fifo_xfer by default, except explicitly.

I don't like the `TUD_AUDIO_PREFER_RING_BUFFER`, it is only temporarily solution, we will eventual remove that.

> That's won't work since sample length could be +-1, wrapped part has to be always considered.

Apparently, I am not aware of this issue. In this case, can you modify to make it more generic. xfer_fifo() will make use of qtd_init() with the linear part first. Then update the buffer pagep[] with wrapped part if following conditions is met
1. Linear len < total _bytes
2. wrap pointer is 4k alignment.

HiFiPhile: > Then update the buffer pagep[] with wrapped

you mean modify `buffer pagep[]` inside ` xfer_fifo()` and remove `qtd_init_fifo()` ?

hathach: it is mostly to reduce code duplication, for now, just keep it as it is if it is too confused. Only set up the buffer page only above 2 condition are met. currently wrapped part is always set up.

hathach: @HiFiPhile to clarify: not all edpt_xfer_fifo() need to combine linear + wrapped part, e.g CDC should have no issue breaking these into 2 sub transfers (one after another when USB complete ISR). Therefore not all fifo buffer need to be 4K. It is mostly for ISO applcation/driver, so having a 4K check would fit both.

src/portable/nxp/transdimension/dcd_transdimension.c
```diff
@@ -508,12 +508,20 @@ bool dcd_edpt_xfer_fifo (uint8_t rhport, uint8_t ep_addr, tu_fifo_t * ff, uint16
   // address to 32-byte boundaries.
   // void* cast to suppress cast-align warning, buffer must be
   CleanInvalidateDCache_by_Addr((uint32_t*) tu_align((uint32_t) fifo_info.ptr_lin, 4), fifo_info.len_lin + 31);
-  if (fifo_info.len_wrap > 0)
+
+  //------------- Prepare qtd -------------//
+
+  // In case of : wrapped part is present & buffer is aligned to 4k & buffer size is multiple of 4k
+  if (total_bytes > fifo_info.len_lin && (uint32_t)fifo_info.ptr_wrap == tu_align4k((uint32_t)fifo_info.ptr_wrap) && tu_fifo_depth(ff) == tu_align4k(tu_fifo_depth(ff)))
```
HiFiPhile, line 515: I've add the switch, should be good but haven't tested. (better to have a macro `bool is_aligned_4k(x)`)

hathach: yeah, the condition look good, you can use the `!tu_offset4k()` as is_aligned_4k(), which make it easier to read.

hathach:
@HiFiPhile I am doing resolving conflict and refactor this pr just now. Please hold on a bit if you plan to work more on this

src/portable/nxp/transdimension/dcd_transdimension.c
```diff
@@ -394,25 +443,86 @@ bool dcd_edpt_xfer(uint8_t rhport, uint8_t ep_addr, uint8_t * buffer, uint16_t t
...
+          p_qtd->buffer[i] = (uint32_t) fifo_info.ptr_wrap + 4096 * page;
+          page++;
+        }
+      }
+
+      CleanInvalidateDCache_by_Addr((uint32_t*) tu_align((uint32_t) fifo_info.ptr_wrap, 4), total_bytes - fifo_info.len_wrap + 31);
+    }
+    else
+    {
+      // TODO we need to carry the wrapped length after the linear part complete
```
hathach, line 519: @HiFiPhile I do some refactoring and happy with the refactored code. However, the xfer_fifo is not complete yet. In case where total_bytes > than linear length and the buffer is not 4K. DCD must 
1. make an td with linear length
2. when td is complete, continue to submit another TD with the rest.

In short, The dcd must transfer all requested bytes by usbd/class driver, depending on the driver, breaking it into 2 sub transfer may still be acceptable e.g with stream/buffered endpoint such as cdc/midi.

hathach: @HiFiPhile I have changed my mind again, since xfer fifo isn't used by other driver just yet (though there is a plan for it to be used more in other drivers). It is best to not write code that aren't run yet. I just added an TODO note, we can merge this now once you verified that the refactored code work well with your ISO setup.

hathach (approved):
Thank you for your effort adding the xfer_fifo. I am happy with the refactored code now, please verify if it works with your ISO setup. Once it is confirmed, we could merge it and resolve other case (total > len and buffer is not 4k aligned) later on when needed.

HiFiPhile:
Now UAC2 is broken with no transfer, maybe forget to reset EP busy flag in `usbd_edpt_close` ?

```
USBD Setup Received 01 0B 03 00 01 00 00 00
  Set Interface
  AUDIO control request
  Set itf: 1 - alt: 3
  Open EP 01 with Size = 776
  Queue EP 01 with 776 bytes ...
  Open EP 81 with Size = 4
Set interface 1 alt 3
  Queue EP 81 with 4 bytes ...
  Queue EP 80 with 0 bytes ...

USBD Setup Received 01 0B 00 00 01 00 00 00
  Set Interface
  AUDIO control request
  Set itf: 1 - alt: 0
  CLOSING Endpoint: 0x01
  CLOSING Endpoint: 0x81
  Queue EP 80 with 0 bytes ...

USBD Setup Received 21 01 00 01 00 04 04 00
  AUDIO control request
  Queue EP 00 with 4 bytes ...
USBD Xfer Complete on EP 00 with 4 bytes
  0000:  00 77 01 00                                      |.w..|
  AUDIO control complete
Clock set current freq: 96000
  Queue EP 80 with 0 bytes ...

USBD Setup Received 01 0B 03 00 01 00 00 00
  Set Interface
  AUDIO control request
  Set itf: 1 - alt: 3
  Open EP 01 with Size = 776
  Queue EP 01 with 776 bytes ...
usbd_edpt_xfer 1267: ASSERT FAILED
  Queue EP 80 with 0 bytes ...
```

hathach:
@HiFiPhile which audio example you are running

HiFiPhile:
> which audio example you are running

I've tested both `uac2_headset` and my own config.

Somehow EP flags are not reset when closed.

With xfer_fifo enabled it fails for the same reason:
```
USBD Setup Received A1 01 01 01 00 02 01 00
  AUDIO control request
Get channel 1 mute 0
  Queue EP 80 with 1 bytes ...
USBD Xfer Complete on EP 80 with 1 bytes
  AUDIO control complete
  Queue EP
USBD Setup Received 01 0B 01 00 01 00 00 00
  Set Interface
  AUDIO control request
  Set itf: 1 - alt: 1
  Open EP 01 with Size = 52
Set interface 1 alt 1
  Queue ISO EP 01 with 52 bytes ... OK
  Queue EP 80 with 0 bytes ...

USBD Setup Received 01 0B 00 00 01 00 00 00
  Set Interface
  AUDIO control request
  Set itf: 1 - alt: 0
  CLOSING Endpoint: 0x01
  Queue EP 80 with 0 bytes ...

USBD Setup Received 01 0B 01 00 01 00 00 00
  Set Interface
  AUDIO control request
  Set itf: 1 - alt: 1
  Open EP 01 with Size = 52
Set interface 1 alt 1
  Queue ISO EP 01 with 52 bytes ... usbd_edpt
USBD Setup Received 01 0B 00 00 01 00 00 00
  Set Interface
  AUDIO control request
  Set itf: 1 - alt: 0
  CLOSING Endpoint: 0x01
  Queue EP 80 with 0 bytes ...

USBD Setup Received 01 0B 01 00 01 00 00 00
  Set Interface
  AUDIO control request
  Set itf: 1 - alt: 1
  Open EP 01 with Size = 52
Set interface 1 alt 1
  Queue ISO EP 01 with 52 bytes ... usbd_edpt_xfer_fifo 1299: ASSERT FAILED
  Queue EP 80 with 0 bytes ...
```

Could be fixed by adding:
```
void usbd_edpt_close(uint8_t rhport, uint8_t ep_addr)
{
  TU_ASSERT(dcd_edpt_close, /**/);
  TU_LOG2("  CLOSING Endpoint: 0x%02X\r\n", ep_addr);

  uint8_t const epnum = tu_edpt_number(ep_addr);

  dcd_edpt_close(rhport, ep_addr);
  + _usbd_dev.ep_status[epnum][dir].stalled = false;
  + _usbd_dev.ep_status[epnum][dir].busy = false;
  + _usbd_dev.ep_status[epnum][dir].claimed = false;

  return;
}
```

hathach:
@HiFiPhile yeah, I think this is the right fix, closing endpoint should also reset its state. Please make the commit if possible 

```C
void usbd_edpt_close(uint8_t rhport, uint8_t ep_addr)
{
  TU_ASSERT(dcd_edpt_close, /**/);
  TU_LOG2("  CLOSING Endpoint: 0x%02X\r\n", ep_addr);

  uint8_t const epnum = tu_edpt_number(ep_addr);

  dcd_edpt_close(rhport, ep_addr);
  + _usbd_dev.ep_status[epnum][dir].stalled = false;
  + _usbd_dev.ep_status[epnum][dir].busy = false;
  + _usbd_dev.ep_status[epnum][dir].claimed = false;

  return;
}
```

hathach:
@HiFiPhile let me know if the rest of your testing is going well enough

HiFiPhile:
It goes well in my side, you can merge it.

hathach:
> It goes well in my side, you can merge it.

great, thank you very much for the PR and testing

PR #1063, opened 2021-08-31
UAC2: Add xfer_fifo support for dcd_transdimension.

HiFiPhile:
**Describe the PR**
#2nd part of #926

- UAC2: Add ring buffer support for dcd_transdimension.
- Add ring buffer ON/OFF switch.

Maybe we can use a better name than `TUD_AUDIO_PREFER_RING_BUFFER`.

### DIFF
--- a/src/class/audio/audio_device.c
+++ b/src/class/audio/audio_device.c
@@ -64,10 +64,19 @@
 // MACRO CONSTANT TYPEDEF
 //--------------------------------------------------------------------+
 
+// Use ring buffer if it's available, some MCUs need extra RAM requirements
+#ifndef TUD_AUDIO_PREFER_RING_BUFFER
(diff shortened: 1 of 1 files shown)


hathach (commented):
Thank you for your PR, to be honest I don't entirely like the solution. Since I want to refactor the iso and its buffer usage. Though I could merge this as it is to make it easier for you (without keeping local changes in your fork), but keep in mind that `TUD_AUDIO_PREFER_RING_BUFFER` is subject to changed/refactored in the future.

HiFiPhile:
It depends on your planning. 
If you are too busy I prefer to let it merged and revisit later with iso refactor, or if you are going to refactor iso soon then no need to merge.

hathach (approved):
Since I don't know when I could work on this, will merge it as its for now, will probably refactor out later on. Thank you for your PR.

PR #1070, opened 2021-09-03
Fix for MSC during stage STATUS

kamtom480:
**Describe the PR**
After https://github.com/hathach/tinyusb/commit/be98cd56c7ac3bbd8b8762e42f752c9cec7c92f0, MSC on Spresense stopped working. This was because the if  condition (`p_cbw->total_bytes > p_msc->xferred_len`) was true: https://github.com/hathach/tinyusb/blob/831a45f14bcc833d536cab39bef61cc67533fa73/src/class/msc/msc_device.c#L604

I think `p_msc->total_len` should be here, not `p_cbw->total_bytes`.

### DIFF
--- a/src/class/msc/msc_device.c
+++ b/src/class/msc/msc_device.c
@@ -601,10 +601,10 @@ bool mscd_xfer_cb(uint8_t rhport, uint8_t ep_addr, xfer_result_t event, uint32_t
     // skip status if epin is currently stalled, will do it when received Clear Stall request
     if ( !usbd_edpt_stalled(rhport,  p_msc->ep_in) )
     {
-      if ( (p_cbw->total_bytes > p_msc->xferred_len) && is_data_in(p_cbw->dir) )
+      if ( (p_msc->total_len > p_msc->xferred_len) && is_data_in(p_cbw->dir) )
       {
         // 6.7 The 13 Cases: case 5 (Hi > Di): STALL before status
-        TU_LOG(MSC_DEBUG, "  SCSI case 5 (Hi > Di): %lu > %lu\r\n", p_cbw->total_bytes, p_msc->xferred_len);
+        TU_LOG(MSC_DEBUG, "  SCSI case 5 (Hi > Di): %lu > %lu\r\n", p_msc->total_len, p_msc->xferred_len);
         usbd_edpt_stall(rhport, p_msc->ep_in);
       }else
       {


hathach (approved):
yeah, spot-on, thank you, the naming is too similar, I really need to rename these

hathach:
@kamtom480 hmm, I take this back, seem like i really meant to use total_bytes according to comment there. Please give me time to pull out Spresense to test.

hathach:
@kamtom480 the stall condition (before PR logic) is indeed correct here, according to the MSC BOT 13 cases. this particular if is for case 5 when Host expect more data than Device could send. In this particular case is Read Format Capacity command, where host expect 252 bytes, and we only send 12 bytes in response. Device really need to send an STALL first, then resume to send SCSI status afterwards when received clearing stall request https://github.com/hathach/tinyusb/blob/master/src/class/msc/msc_device.c#L318.  Doing differently by this PR will cause MSC compliance test suite to fail.

image

image

The issue here is probably the stall and clear stall handling of cxd56 port. In general, when stalling the endpoint must be stalled, then all the queuing (not complete) transfer of the endpoint must be aborted and removed. Other ports also has this problem, but is fixed while doing compliance testing. Since stall is less seen on non-control endpoints, it is easily to overlook. It is a nice chance to test them out.

Please give me a bit of time to see if I could find a better fix with stall/clear stall.

hathach:
@kamtom480 hmm seem like the dcd didn't notice the stack when setup packet of the next CLEAR STALL request. Therefore the msc couldn't continue.  I haven't looked too closed at dcd_cxd56.c so far, I will try to see if I could get any clues

Here is the log level 2 from cp210x from `make BOARD=spresense LOG=2`

```
USBD Xfer Complete on EP 05 with 31 bytes                                                                                               
  MSC xfer callback                                                                                                                     
  SCSI Command: Read Format Capacity                                                                                                    
  Queue EP 84 with 12 bytes ...                                                                                                         
USBD Xfer Complete on EP 84 with 12 bytes                                                                                               
  MSC xfer callback                                                                                                                     
  SCSI Data                                                                                                                             
  SCSI case 5 (Hi > Di): 252 > 12                                                                                                       
    Stall EP 84
```

hathach:
I see spresense has its own usb stack under nuttx system, and it **Doesn't** forward the Clear Feature (Endpoint) request toward tinyusb. https://github.com/sonydevworld/spresense-nuttx/blob/2ffe58b6dcb8ca7e83a4d2acec5622276f7c6f5d/arch/arm/src/cxd56xx/cxd56_usbdev.c#L1367

It is quite troublesome, since most of the msc recover is based on the Clear Stall request. I will try to figure an way. In general the setup seem to be conditionally forwards (most but not all), and Status stage can also be auto ACKed. Which could cause race condition where next setup arrive while tinyusb task doesn't fully process previous packet hence the setup queue.

PR #1071, opened 2021-09-03
More IAR support

HiFiPhile:
**Describe the PR**

Add IAR project connection files, which can be used to automatically import tusb stack & examples into IAR project.

Since SDKs in `hw\mcu` only have GCC's cstartup file (and there are too many MCUs for me...), a buildable IAR project need to be created before import tusb.

Example's source files (`main.c`, `tusb_config.h`, etc.) will be copied to project's directory, os they can be edited safely.

`tools/iar.ipcf` only contains the stack, `*.ipcf` inside examples contain stack & example.

`iar_gen.py` is used to generate `*.ipcf` inside examples from `tools/iar.ipcf`.

**Additional context**
I'm going to add a few words in `docs/reference/getting_started.rst`, for image I create a folder `assert` in `docs/reference` ?


hathach (commented):
This is great and I am sure IAR user will find it helpful. However, as you mentioned `ipcf` only add source files as they appear in the actual file system and not buildable. Therefore it is not too useful IMHO.  

To be honest, I have a mixed feeling about adding these file. I don't use IAR at all therefore cannot and won't be able to maintain these project file in the future. My experience with IDE is that they are super nice with specific project on specific MCUs with all of its great debug tool. However, adding cross-platform factor such as mcus, it becomes too time consuming to maintain (too many clicks to config/change). 

Anyway, It is still useful to have `tools/iar_gen.py and iar.icpf` (maybe rename the later to iar_template.icpf). For the example ipcf, maybe just let user to generate those on their own.  That way we don't have to maintain those files.

HiFiPhile:
> For the example ipcf, maybe just let user to generate those on their own. That way we don't have to maintain those files.

Yes it's easier to maintain less files... I've removed examples and added doc.

hathach (approved):
superb ! thank you very much for your PR. It look perfect now

PR #1075, opened 2021-09-07
Fix cxd56 msc

hathach:
**Describe the PR**

Revert changes in #1070 , fix the root cause of the issue with cxd56. Following is the explanation (copied from other pr)

@kamtom480 the stall condition (before PR logic) is indeed correct here, according to the MSC BOT 13 cases. this particular if is for case 5 when Host expect more data than Device could send. In this particular case is Read Format Capacity command, where host expect 252 bytes, and we only send 12 bytes in response. Device really need to send an STALL first, then resume to send SCSI status afterwards when received clearing stall request https://github.com/hathach/tinyusb/blob/master/src/class/msc/msc_device.c#L318.  Doing differently by this PR will cause MSC compliance test suite to fail.

image

image

The issue is spresense has its own usb stack under nuttx system, and it **Doesn't** forward the Clear Feature (Endpoint) request toward tinyusb. https://github.com/sonydevworld/spresense-nuttx/blob/2ffe58b6dcb8ca7e83a4d2acec5622276f7c6f5d/arch/arm/src/cxd56xx/cxd56_usbdev.c#L1367

It is quite troublesome, since most of the msc recover is based on the Clear Stall request. I will try to figure an way. In general the setup seem to be conditionally forwards (most but not all), and Status stage can also be auto ACKed. Which could cause race condition where next setup arrive while tinyusb task doesn't fully process previous packet hence the setup queue.

**WALKAROUND**

The workaround is unconditionally unstall the status endpoints for cxd56 and hope for the best, which won't pass compliance test suite, but will work for most of the case with sensible host driver.

### DIFF
--- a/hw/bsp/spresense/board.mk
+++ b/hw/bsp/spresense/board.mk
@@ -70,4 +70,5 @@ $(BUILD)/$(PROJECT).spk: $(MKSPK)
 
 # flash
 flash: $(BUILD)/$(PROJECT).spk
-	@$(TOP)/hw/mcu/sony/cxd56/tools/flash_writer.py -s -c $(SERIAL) -d -b 115200 -n $<
+	@echo FLASH $<
+	@$(PYTHON) $(TOP)/hw/mcu/sony/cxd56/tools/flash_writer.py -s -c $(SERIAL) -d -b 115200 -n $<
--- a/examples/make.mk
+++ b/examples/make.mk
@@ -61,15 +61,16 @@ GDB = $(CROSS_COMPILE)gdb
 OBJCOPY = $(CROSS_COMPILE)objcopy
 SIZE = $(CROSS_COMPILE)size
 MKDIR = mkdir
-PYTHON = python
 
 ifeq ($(CMDEXE),1)
   CP = copy
   RM = del
+  PYTHON = python
 else
   SED = sed
   CP = cp
   RM = rm
+  PYTHON = python3
 endif
 
 #-------------- Source files and compiler flags --------------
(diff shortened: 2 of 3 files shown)


hw/bsp/spresense/board.mk
```diff
@@ -70,4 +70,5 @@ $(BUILD)/$(PROJECT).spk: $(MKSPK)
 
 # flash
 flash: $(BUILD)/$(PROJECT).spk
-	@$(TOP)/hw/mcu/sony/cxd56/tools/flash_writer.py -s -c $(SERIAL) -d -b 115200 -n $<
+	@echo FLASH $<
+	@$(PYTHON) $(TOP)/hw/mcu/sony/cxd56/tools/flash_writer.py -s -c $(SERIAL) -d -b 115200 -n $<
```
kamtom480, line 74: Is adding $(PYTHON) necessary here?

kamtom480 (commented):
@hathach Thank you for this PR! I tested it and it works on Spresense.

hw/bsp/spresense/board.mk
hathach: yeah I got following issue without explicitly invoked with python under ubuntu 20.04

```
/usr/bin/env: ‘python3\r’: No such file or directory
```

kamtom480: Hmm.. I am also using Ubuntu 20.04 and I don't have this issue. Have you upgraded your Ubuntu from a previous version or have you installed a new Ubuntu 20.04?

If you change `PYTHON = python` to `PYTHON = python3` will it work for you?

kamtom480 (approved):

hathach:
thank you for your review

kamtom480 (commented):
Why not use `python3` for Windows too?

hathach:
> Why not use `python3` for Windows too?

Windows only has python.exe there is no python3.exe even though you install 3.9x

Issue #1059, opened 2021-08-30
USB Compliance Verification Test Suite (USBCV)


hathach:
**What is this issue for**

The test suite include many corner cases which require both update of usbd and dcd to pass. This issue is used to keep track of  current status of compliance test of tinyusb **device stack** on each DCD ports. Once it is all passed, we could close this.

**What is USB Compliance Verification (USBCV)**

USBCV is test suite tool used by usb-if to verify all usb hardware/software/controller which is required to passed in order to have USB logo on the product. https://www.usb.org/compliancetools 

**How to run USBCV**

usbcv tool only runs on windows , depending on usb controller type on your PC you will need to either download the USB3CV (xHCI) or USB2CV  (EHCI) from https://www.usb.org/compliancetools . Note: even your mcu is only full speed, but testing on xHCI host, you would need to run USB3CV.

**Which tests suite to run**

Within the scope of tinyusb device stack, we only need to run follow test suite
- Chapter 9 Test [USB 2 devices]
- HID Tests
- MSC Tests

<img width="480" alt="usbcv" src="https://user-images.githubusercontent.com/249515/131323710-ad07e8e0-4922-41e8-8b85-3f7c9408ef29.PNG">

**Compliance Status**

Example/Application can have large impact on the test result, the follow result is tested mostly for **cdc_msc** and **hid_composite**

| MCU \ Test Suite | Chapter9 | HID  | MSC  |
| :--------------- | :------: | :--: | :--: |
| nRF52840         |    :heavy_check_mark:    |  :heavy_check_mark:  |  :heavy_check_mark:  |
| SAMD21, SAMD51   |    :question:     |  :question:   |  :question:   |
| RP2040           |   :question:    |  :question:  |  :question:   |

HiFiPhile:
I'm trying to test my project based on STM32L071.

SetConfiguration passed even without #1058
```
TD 9.13 - SetConfiguration Test (Configuration Index 0)Passed
INFO
Start time: Aug 30, 2021 - 14:23:20
INFO
SetConfiguration with configuration value : 0x1
INFO
Unconfigured the device
INFO
SetConfiguration with configuration value : 0x1
INFO

Stop time: Aug 30, 2021 - 14:23:21
INFO
Duration:  1 second.
INFO
Stopping Test [ TD 9.13 - SetConfiguration Test (Configuration Index 0):
     Number of: Fails (0); Aborts (0); Warnings (0) ]

[...]

PR #1077, opened 2021-09-07
St synopsy compliance

hathach:
**Describe the PR**
make synopsys usbcv compliance
- implement dcd_edpt_close_all()
- reset dtoggle when open endpoint
- correct dev interrupt mask on bus reset
- also implement remote wakeup for stm synopsys with nop delay() based on SystemCoreClock

This is tested on stm32f411disco which passed all the compliance test (chapter 9, hid, msc) more detail in #1059


src/portable/st/synopsys/dcd_synopsys.c
```diff
@@ -96,6 +96,9 @@
 #elif CFG_TUSB_MCU == OPT_MCU_GD32VF103
 #include "synopsys_common.h"
 
+// for remote wakeup delay
+#define __NOP()   __asm volatile ("nop")
```
HiFiPhile, line 100: `__NOP()` is defined in CMSIS, redefine might cause error.

PR #1087, opened 2021-09-10
NUC126: fix remote_wakeup

HiFiPhile:
**Describe the PR**
* Fix remote-wakeup on NUC126.

Tested OK.
[Chapter 9 Tests - USB 2 - Passed - DEBUG - 2021-09-10 13-00-45.zip](https://github.com/hathach/tinyusb/files/7143356/Chapter.9.Tests.-.USB.2.-.Passed.-.DEBUG.-.2021-09-10.13-00-45.zip)

### DIFF
--- a/src/portable/nuvoton/nuc121/dcd_nuc121.c
+++ b/src/portable/nuvoton/nuc121/dcd_nuc121.c
@@ -200,11 +200,10 @@ static void bus_reset(void)
 }
 
 /* centralized location for USBD interrupt enable bit mask */
-#if USE_SOF
-static const uint32_t enabled_irqs = USBD_INTSTS_VBDETIF_Msk | USBD_INTSTS_BUSIF_Msk | USBD_INTSTS_SETUP_Msk | USBD_INTSTS_USBIF_Msk | USBD_INTSTS_SOFIF_Msk;
-#else
-static const uint32_t enabled_irqs = USBD_INTSTS_VBDETIF_Msk | USBD_INTSTS_BUSIF_Msk | USBD_INTSTS_SETUP_Msk | USBD_INTSTS_USBIF_Msk;
-#endif
+enum {
+  ENABLED_IRQS = USBD_INTSTS_VBDETIF_Msk | USBD_INTSTS_BUSIF_Msk | USBD_INTSTS_SETUP_Msk |
+                 USBD_INTSTS_USBIF_Msk   | (USE_SOF ? USBD_INTSTS_SOFIF_Msk : 0)
+};
 
 /*
   NUC121/NUC125/NUC126 TinyUSB API driver implementation
@@ -226,8 +225,8 @@ void dcd_init(uint8_t rhport)
 
   usb_attach();
 
-  USBD->INTSTS = enabled_irqs;
-  USBD->INTEN  = enabled_irqs;
+  USBD->INTSTS = ENABLED_IRQS;
+  USBD->INTEN  = ENABLED_IRQS;
 }
 
 void dcd_int_enable(uint8_t rhport)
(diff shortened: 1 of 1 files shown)


HiFiPhile:
Without 22571ec with `cdc_msc` disk won't show up, but with this commit it fails strangely:
```
USBD Xfer Complete on EP 83 with 13 bytes
  MSC xfer callback
  SCSI Status = 0
  Queue EP 03 with 31 bytes ...
USBD Xfer Complete on EP 03 with 31 bytes
  MSC xfer callback
  SCSI Command: Read10
  Queue EP 83 with 512 bytes ...
USBD Xfer Complete on EP 83 with 512 bytes
  CDC xfer callback
cdcd_xfer_cb 431: ASSERT FAILED
```

src/portable/nuvoton/nuc121/dcd_nuc121.c
```diff
@@ -368,13 +381,15 @@ void dcd_edpt_clear_stall(uint8_t rhport, uint8_t ep_addr)
   (void) rhport;
   USBD_EP_T *ep = ep_entry(ep_addr, false);
   ep->CFG |= USBD_CFG_CSTALL_Msk;
+  ep->CFG &=~USBD_CFG_DSQSYNC_Msk;
```
hathach, line 384: we should clear the Data PID first before clear stall to prevent possible race condition. Could you move this before the clear stall.

hathach (commented):
Thank you for another great PR, look good,  The NUC manual doesn't explicitly say how to use the remote wake-up bit, but this seem to be the case.  Though I think we should clear the PID before clear stall to prevent possible race condition.

src/portable/nuvoton/nuc121/dcd_nuc121.c
```diff
@@ -368,13 +381,15 @@ void dcd_edpt_clear_stall(uint8_t rhport, uint8_t ep_addr)
...
 }
 
 void dcd_int_handler(uint8_t rhport)
 {
   (void) rhport;
 
-  uint32_t status = USBD->INTSTS;
+  // Mask non-enabled irqs, ex. SOF
+  uint32_t status = USBD->INTSTS & (enabled_irqs | 0xffffff00);
```
hathach, line 392: could you explain why should we OR  enabled_irqs with 0xffffff00 here

src/portable/nuvoton/nuc121/dcd_nuc121.c
hathach: I have pushed update to correct this.

src/portable/nuvoton/nuc121/dcd_nuc121.c
HiFiPhile: `enabled_irqs` cover only DCD itself's irqs, 0xffffff00 allows EP's irqs to occur.

hathach: in that case, we should use this instead

```C
uint32_t status = USBD->INTSTS &  USBD->INTEN
```

HiFiPhile: In `USBD->INTEN` there is no EPEVTx
image
image

hathach: you are right, the nuc inten isn't same mapping as intsts as other mcu. Though, the value `0xffffff00` isn't quite correct, 0xffff0000 would be more precise (to also take `wken` into account, innaken is never used).

hathach: update: ah I see, 0xffffff00 since wken and innaken is not present on the INTSTS :facepalm:

hathach (approved):
thank you very much for your PR. everything looks great.

Issue #880, opened 2021-06-07
Multiple out of bound read Vulnerabilities In  tinyusb


hac425xxx:
These bugs are caused by the failure to check `request->wLength`.

## hidd_control_xfer_cb

case 1

if `request->wLength > the size of  p_hid->epin_buf` it could lead buffer out of bound read.

```
      case HID_REQ_CONTROL_GET_REPORT:
        if ( stage == CONTROL_STAGE_SETUP )
        {
          uint8_t const report_type = tu_u16_high(request->wValue);
          uint8_t const report_id   = tu_u16_low(request->wValue);

          uint16_t xferlen = tud_hid_get_report_cb(hid_itf, report_id, (hid_report_type_t) report_type, p_hid->epin_buf, request->wLength);
          TU_ASSERT( xferlen > 0 );

          tud_control_xfer(rhport, request, p_hid->epin_buf, xferlen);
        }
```

case 2

```

        else if ( stage == CONTROL_STAGE_ACK )
        {
          uint8_t const report_type = tu_u16_high(request->wValue);
          uint8_t const report_id   = tu_u16_low(request->wValue);
          
          //  bug !!!
          tud_hid_set_report_cb(hid_itf, report_id, (hid_report_type_t) report_type, p_hid->epout_buf, request->wLength);
        }
```

## dfu_req_dnload_setup

```
static void dfu_req_dnload_setup(uint8_t rhport, tusb_control_request_t const * request)
{
 
  // setup for data phase
  tud_control_xfer(rhport, request, _dfu_state_ctx.transfer_buf, request->wLength);  // bug !!!
}
```

## btd_control_xfer_cb

```
  if ( stage == CONTROL_STAGE_SETUP )
  {
  
    .....................
    .....................

    // bug!!!
    return tud_control_xfer(rhport, request, &_btd_itf.hci_cmd, request->wLength);
  }
```

hathach:
have you tried it, and do you have any examples to prove it ?

hac425xxx:
no, I found these by reading the code

hathach:
Ah thanks, I thought you would have a running examples so that we could quickly test it out and fix it. Regardless, these are valid issue. Ultimately we shouldn't just host for its length. 
- The 1) can be fixed by limited the requestlength to the endpoint size. 
- For 2 DFU, it is newly added, I have an plan to enhance it later on, will keep this in mind. 
- For 3) same as 1.

[...]

PR #1090, opened 2021-09-12
Prevent buffer overflow in bth_device.c

szymonh:
Address possible buffer overflow in bth_device.c as described in issue #880 (btd_control_xfer_cb).

Passing size of _btd_itf.hci_cmd as len parameter to tud_control_xfer so that actual _ctrl_xfer.data_len will be set to minimum of buffer size and request wLength.

### DIFF
--- a/src/class/bth/bth_device.c
+++ b/src/class/bth/bth_device.c
@@ -214,7 +214,7 @@ bool btd_control_xfer_cb(uint8_t rhport, uint8_t stage, tusb_control_request_t c
     }
     else return false;
 
-    return tud_control_xfer(rhport, request, &_btd_itf.hci_cmd, request->wLength);
+    return tud_control_xfer(rhport, request, &_btd_itf.hci_cmd, sizeof(_btd_itf.hci_cmd));
   }
   else if ( stage == CONTROL_STAGE_DATA )
   {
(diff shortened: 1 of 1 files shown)


hathach (commented):
thank you for your PR, look good to me. Though let's wait for @kasjer feedback since he is original author of bth driver.

src/class/bth/bth_device.c
```diff
@@ -214,7 +214,7 @@ bool btd_control_xfer_cb(uint8_t rhport, uint8_t stage, tusb_control_request_t c
     }
     else return false;
 
-    return tud_control_xfer(rhport, request, &_btd_itf.hci_cmd, request->wLength);
+    return tud_control_xfer(rhport, request, &_btd_itf.hci_cmd, sizeof(_btd_itf.hci_cmd));
```
kasjer, line 217: I would also consider adding similar size check in call to `tud_bt_hci_cmd_cb()` in line 224, as not to deceive BT stack about size of command buffer if wLength can not be trusted.
```c
if (tud_bt_hci_cmd_cb) tud_bt_hci_cmd_cb(&_btd_itf.hci_cmd,
              tu_min16(sizeof(_btd_itf.hci_cmd), request->wLength));
```

kasjer (commented):
Looks good.

hathach (approved):
Thank you very much for your PR and @kasjer for helping with the review. Will merge when ci passed.

PR #1093, opened 2021-09-13
Enforced buffer boundaries for hid devices

szymonh:
Addressed two issues in hidd_control_xfer_cb as reported in #880.


hathach (approved):
Look great, thank you for your PR.

PR #1096, opened 2021-09-20
dcd_da146xx: Implement dcd_edpt_close_all()

kasjer:
**Describe the PR**
Unconditionally disables all endpoints except EP0.
Related to #1059

**Additional context**
~~Not tested yet with **USBCV**~~

### DIFF
--- a/src/portable/dialog/da146xx/dcd_da146xx.c
+++ b/src/portable/dialog/da146xx/dcd_da146xx.c
@@ -885,7 +885,12 @@ bool dcd_edpt_open(uint8_t rhport, tusb_desc_endpoint_t const * desc_edpt)
 void dcd_edpt_close_all (uint8_t rhport)
 {
   (void) rhport;
-  // TODO implement dcd_edpt_close_all()
+
+  for (int epnum = 1; epnum < EP_MAX; ++epnum)
+  {
+    dcd_edpt_close(0, epnum | TUSB_DIR_OUT);
+    dcd_edpt_close(0, epnum | TUSB_DIR_IN);
+  }
 }
 
 void dcd_edpt_close(uint8_t rhport, uint8_t ep_addr)


hathach (commented):
we should also reset the `xfer_ctl_t xfer_status[EP_MAX][2];` (except EP0) as well. Since the new configure may have less endpoint than the one we are closing.

Note: for multiple configuration , SET_CONFIG can be set without an bus_reset.

kasjer:
@hathach thanks for input, I will update PR.
But before this I will make another PR that will remake power state handling. Bus reset and sleep related code was only testes on live systems. Now when I've used **USB3CV**, I see that more changes are needed in that area.

But since you are on this, let me ask about something that prevents **USB3CV** tests to pass on this board.
EP0 size is 8 for this chip family, when code below is NOT removed tests fail.
```
      // Only send up to EP0 Packet Size if not addressed
      // This only happens with the very first get device descriptor and EP0 size = 8 or 16.
      if ((CFG_TUD_ENDPOINT0_SIZE < sizeof(tusb_desc_device_t)) && !_usbd_dev.addressed)
      {
        len = CFG_TUD_ENDPOINT0_SIZE;

        // Hack here: we modify the request length to prevent usbd_control response with zlp
        ((tusb_control_request_t*) p_request)->wLength = CFG_TUD_ENDPOINT0_SIZE;
      }
```
It looks like **USB3CV** during test queries about device descriptors with request size 8 at the beginning,
then when address is set is asks with request size 18 as expected, but then it resets and sends again
device descriptor requests with size 18 (with no address set up first) which **TinyUSB** cuts to 8 bytes only,
then **USB3CV** fails waiting for some more data that never comes.

Have you tried running different board (like STM32F4, that are now configured for EP0 size 64) to have EP0 size 8 to
see if this will also happen?

I'm not familiar with USB spec so I'm not sure if **USB3CV** is wrong or **TinyUSB**.
I only used **USB3CV** not **USB2CV** which did not see USB controller on PC I used for testing.

hathach:
@kasjer this is really an edge case. This particular code is added since it is needed when testing with actual host driver such linux/windows with 8-byte control endpoint e.g msp430. The host driver won't bother getting the 2nd packet EP0 and will send ZLP as status right after the 1st packet. Maybe the USBCV has different expectation, I will try to test it later on.

You can try to comment it out to run the USBCV, but you may have issue with normal enumeration.

kasjer:
@hathach I decided to make dcd_edpt_close_all() look very simple.
This PR will be followed by next change that will take into account what should be done when endpoint is closed so basically I will update dcd_edpt_close() as well.

kasjer:
@hathach about **USB3CV** and Windows/Linux enumeration difference.
For me behavior of **USB3CV** is more reasonable, but then I also understand what was done in Windows/Linux.
It could be fixed in **TinyUSB** if we had EP0 OUT endpoint ready not just after expected length of DATA STAGE, but (for example) after first packet or from the beginning. Then STATUS STAGE could just stop control request.

But there is simpler solution that can make **USB3CV** and Windows happy.
We could slightly modify condition for _hacked_ code:
```c
      if ((CFG_TUD_ENDPOINT0_SIZE < sizeof(tusb_desc_device_t)) && !_usbd_dev.addressed &&
         ((tusb_control_request_t*) p_request)->wLength > sizeof(tusb_desc_device_t)) // <---- New condition
      {
        len = CFG_TUD_ENDPOINT0_SIZE;

        // Hack here: we modify the request length to prevent usbd_control response with zlp
        ((tusb_control_request_t*) p_request)->wLength = CFG_TUD_ENDPOINT0_SIZE;
      }
```

**USB3VC** when requests device descriptor in non-addressed state asks for exactly 18 bytes (descriptor size).
Windows/Linux asks for 64 bytes which is far more then needed for valid descriptor size and they are happy
with whatever they get in single packet and send STATUS STAGE packet.

You may consider this new condition that will only affect boards with small EP0 size. I will not insist, but it was quite
annoying to comment this code for tests and restore for normal usage.

I did not checked Mac behavior though.

hathach:
> @hathach I decided to make dcd_edpt_close_all() look very simple.
> This PR will be followed by next change that will take into account what should be done when endpoint is closed so basically I will update dcd_edpt_close() as well.

The previous edpt_close_all() is actually simpler since it is used when expecting a full reconfiguration. The edpt_close() is currently meant for ISO only, in the future, I plan to retire/remove it to prefer and better API e.g edpt_resize() due to the fact that close/reopen does not work well with some MCUs. But I am ok if you still prefer to keep it this way though.

kasjer:
I looked simpler but it was missing stuff that you mentioned in previous comment.
dcd_edpt_close() is still missing that part but I already have code ready for PR that does xfer cleanup there.

hathach:
@kasjer the new condition look great. IMHO, It is really OS host driver fault, getting the 1st 8bytes to get EP0 size is common approach similar to getting first 9 bytes of configuration descriptor. I did this as well in the host stack https://github.com/hathach/tinyusb/blob/master/src/host/usbh.c#L712

Back to your suggestion, I think it is great, could you made an PR for it along with the comment to explain the different between the 2 cases 
- Host only want to receive only 1 packet with 64 bytes for EP0 size  
- Deliberately request 18 bytes (full descriptor) length, which it will try to get multiple packets

I will try to check with msp430 and macOS on my side. That is the port that requires this condition to work with. #184

hathach (approved):
> I looked simpler but it was missing stuff that you mentioned in previous comment.
> dcd_edpt_close() is still missing that part but I already have code ready for PR that does xfer cleanup there.

Look good, thank you for your PRs.

PS: esp32s2 failed due to unrelated issue, I will check it out.

PR #1099, opened 2021-09-23
da1469x_dk_pro: Fix button setup

kasjer:
**Describe the PR**
Button on board has 1k resistor to ground when pressed.
When not pressed pin 6 is floating.

This forces MCU pull-up for this pin for correct behavior.

**Additional context**


hathach (approved):
look good, will merge when ci is passed.

PR #1100, opened 2021-09-23
da14695_dk_usb: Fix button configuration

kasjer:
**Describe the PR**
Unlike da1469x_dk_pro, K1 pin (6) is connected to VCC when active.
This adds required pull down and reverses active logic state.

**Additional context**


hathach (approved):
Thanks, it will be useful for me doing the test later on

PR #1101, opened 2021-09-23
da146xx bus reset and sleep reworked

kasjer:
**Describe the PR**
So far bus reset was handled (with some holes in it).
Sleep and remote wake-up were not really tested and did not work.

With this change:
- bus reset is reworked
- remote wakeup works
- bus signal sleep is handled
- in some cases register modification (read/update/write) was change to just writes.
- FRAME interrupt is used for bus reset end condition and for delay needed in remote wakeup condition.

**Additional context**
While it work with Windows and Linux some change in usbd.c is needed
for **USB3CV** tests to pass due to EP0 size 8.
Following lines prevent tests but are good for normal usage:
```
      // Only send up to EP0 Packet Size if not addressed
      // This only happens with the very first get device descriptor and EP0 size = 8 or 16.
      if ((CFG_TUD_ENDPOINT0_SIZE < sizeof(tusb_desc_device_t)) && !_usbd_dev.addressed)
      {
        len = CFG_TUD_ENDPOINT0_SIZE;

        // Hack here: we modify the request length to prevent usbd_control response with zlp
        ((tusb_control_request_t*) p_request)->wLength = CFG_TUD_ENDPOINT0_SIZE;
      }
```

This was tested with **USB3CV** tool **Chapter 9 Tests [USB 2 devices]** and **HID Tests**

### DIFF
--- a/src/portable/dialog/da146xx/dcd_da146xx.c
+++ b/src/portable/dialog/da146xx/dcd_da146xx.c
@@ -60,10 +60,17 @@
 
 #define EP_MAX            4
 
+// Node functional states
 #define NFSR_NODE_RESET         0
 #define NFSR_NODE_RESUME        1
 #define NFSR_NODE_OPERATIONAL   2
 #define NFSR_NODE_SUSPEND       3
+// Those two following states are added to allow going out of sleep mode
+// using frame interrupt.  On remove wakeup RESUME state must be kept for
+// at least 1ms. It is accomplished by using FRAME interrupt that goes
+// through those two fake states before entering OPERATIONAL state.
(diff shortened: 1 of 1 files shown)


hathach:
I am in the middle of esp32s2 work. Give me a bit of time to pull out the da146xx usb to test and review this.

kasjer:
I will re-run test. I assume USCV has some log that can be send to you (or to #1059) or it's just screen shot?
If it is clear what needs to be sent in #1059 (which I have not read to carefully)  just ignore this comment I will follow what others did.

kasjer:
Ups, I could have added Chapter 9 test on HID instead of MSC, that did ask for wake up button activation. It work on pro-dk and usb-dk.
I will attach this file once I have MSC tests fully working.

PR #1106, opened 2021-09-25
Assure that block size is not zero in msc_device

szymonh:
Assure division by zero is not attempted (block_sz) and add lba validation to address msc issues in #880.

### DIFF
--- a/examples/device/cdc_msc_freertos/Makefile
+++ b/examples/device/cdc_msc_freertos/Makefile
@@ -7,12 +7,18 @@ FREERTOS_SRC = lib/FreeRTOS-Kernel
 
 INC += \
 	src \
+	src/FreeRTOSConfig \
 	$(TOP)/hw \
 	$(TOP)/$(FREERTOS_SRC)/include \
 	$(TOP)/$(FREERTOS_SRC)/portable/GCC/$(FREERTOS_PORT)
 	
 # Example source
-EXAMPLE_SOURCE += $(wildcard src/*.c)
+EXAMPLE_SOURCE = \
+	src/freertos_hook.c \
+	src/main.c \
+	src/msc_disk.c \
+	src/usb_descriptors.c
+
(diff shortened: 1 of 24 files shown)


hathach (commented):
Thank you for your PR, but all of these scenarios are already addressed. Actually it is part of the USB certification verification #1059 MSC test suite. The block_sz == 0 when either block_count = 0 and/or total_bytes = 0, both are  checked when command is received before proc_read10/write10_cmd() is invoked

https://github.com/hathach/tinyusb/blob/master/src/class/msc/msc_device.c#L158

szymonh:
Could you please verify that validation for the case when cbw->total_bytes = 0 and block_count = 0 is correct. From what I can see rdwr10_validate_cmd will return MSC_CSW_STATUS_PASSED thus proc_read10 will be called. This will lead to a division by zero in proc_read10_cmd.

hathach:
> Could you please verify that validation for the case when cbw->total_bytes = 0 and block_count = 0 is correct. From what I can see rdwr10_validate_cmd will return MSC_CSW_STATUS_PASSED thus proc_read10 will be called. This will lead to a division by zero in proc_read10_cmd.

status is passed, but the total_bytes = 0, will skip the data phase, go straight to status and didn't invoke the proc_read10 https://github.com/hathach/tinyusb/blob/master/src/class/msc/msc_device.c#L425

> Also rdwr10_validate_cmd does not handle cases where cbw->total_bytes and block_count are both greater than zero but cbw->total_bytes divided by block_count is less than one. For example with total_bytes = 1, block_count = 0xffff we will end up with block_sz equal to zero and rdwr10_validate_cmd will not prevent division by zero.

You are right about this case, This request should be check in rdwr10_validate_cmd() and rejected with `MSC_CSW_STATUS_FAILED`

szymonh:
I altered the PR to make sure that the total_bytes division by block_count is not zero in rdwr10_validate_cmd.
Thank you for your feedback and please let me know if any additional amendments are required.

PR #1107, opened 2021-09-25
add NCM driver in a compatible manner : hathach/tinyusb#550

majbthrd:
This adapt the Nov 2020 era CDC NCM driver from @j4cbo to be compatible with the TinyUSB network API.

The big difference between this PR and PR #550 is that instead of the user having to craft a new user application to use the NCM driver, the existing net_lwip_webserver example just works, regardless of which network driver is used. Similarly, anyone who is writing their own TinyUSB-based network application can switch between the two drivers, using the same API, to choose which works best for their usage.

To pick between CDC NCM and RNDIS / ECM is just a matter of setting CFG_TUD_NCM in tusb_config.h to 0 (RNDIS/ECM) or 1 (NCM).

The existing network API was tweaked so that:

1) tud_network_can_xmit() now takes a parameter as to the size of the would-be transmit packet
2) tud_network_link_state_cb() was added since the NCM driver wanted to inform the user code as to the link state


hathach:
Brilliant !! and great work. Thank you very much, I will check this out this week or sooner and  more test with other ports.  I have always wanted to do more testing with net driver as well as update the example to be more interactive (e.g control LED, display state of button etc ..) but haven't got the time to do so.

kripton:
So, I've been using the pre-rebased version of this (5a1a6ba5c3526e822e060745be8e643e24537c3c) for about a month now on several RPi Picos. I'm perfectly happy with it and didn't have any issues yet.
I'm running a DHCP server over it, an HTTP server and I have quite a bit UDP traffic (unicast and multicast) host -> board.

hathach:
> So, I've been using the pre-rebased version of this ([5a1a6ba](https://github.com/hathach/tinyusb/commit/5a1a6ba5c3526e822e060745be8e643e24537c3c)) for about a month now on several RPi Picos. I'm perfectly happy with it and didn't have any issues yet. I'm running a DHCP server over it, an HTTP server and I have quite a bit UDP traffic (unicast and multicast) host -> board.
> 
> _EDIT_: And I'm using the "Microsoft OS 2.0 USB descriptors" so Microsoft actually loads its NCM driver: https://github.com/kripton/rp2040-dongle/blob/7fae0860fe148ce11a2c0d502d086949c1d6d60c/src/usb_descriptors.c#L244. Silly that this is required at all but it makes the device just work on Linux, macOS and Windows partying_face

Thank you for your input, I will test and merge this soon enough. Since you already modify the MS OS 2.0 to get it auto binding with winncm, please take the time to make an PR to update example for it after the PR got merged.

hathach (approved):
Excellent work :+1: :tada: , I have tried it and it works perfectly on Linux. On windows it requires ncm driver (skip for now) and will test with windows later on when adding MS OS 2.0 descriptors. But I am sure it works as well. Thank you @majbthrd and @j4cbo for making this driver possible. 

@majbthrd I rename the `CFG_TUD_NET` to `CFG_TUD_ECM_RNDIS`, although they both  implement network, they are independent, therefore I think it make more sense for each to has its own macro feature. There is a check that prevent both `ECM_RNDIS` and `NCM` are enabled. We could them to  `NET_ECM_RNDIS` and `NET_NCM` but that are too verbose especially with other NTB macros. Let me know if you name it differently

@j4cbo I also merge your #550 so that your work is also integrated to commit log. Since you are original author of NCM drivers, we would like to receive your feedback (if you have time) on the PR. Any suggestion or improvement that we are missing. It doesn't need to be fully details since we can always enhance it later on.

hathach:
@majbthrd if you have time, please let me know if you are ok with the rename from `CFG_TUD_NET` to `CFG_TUD_ECM_RNDIS` I am open to all suggestion. Otherwise, I think the PR is ready to merge :tada: :+1:

hathach:
@majbthrd thank you very much for your feedback and effort. I did once more commit with ncm.h to hold magic number. It is perfect now, will merge when ci passed. Thank you again for this brilliant PR.

PR #1113, opened 2021-09-27
Assure msc device block size is not zero

szymonh:
In case total_bytes divided by block_count evaluates to zero a division by zero may happen in proc_read10_cmd leading to a crash.


hathach (approved):
Thank you very much for the PR. I take a look back at the MSC BOT specs, the case where `total bytes < block num` is actually counted as 
- READ10: case 7 Hi < Di host read less than device want to send or
- WRITE10 (we need this check for write as well): case 13 Ho < Do, host send less than device would like to receive

In both cases, the specs say we must response with phase error. Therefore, I made a bit of changes and push the update

image

Issue #908, opened 2021-06-21
Support USB Video Class (UVC)


hathach:
**Is your feature request related to a problem? Please describe.**
ISO transfer and Audio driver is supported (for some mcus), maybe the next thing is getting video class implemented.

PR #1118, opened 2021-09-29
Add Video class

kkitayam:
**Describe the PR**
Add Video class driver and a video capture example. #908
This version is experimental and has many limitations. I have only implemented a minimum number of features.

- Support only uncompressed format.
- Support only isochronous IN. No support isochronous OUT and bulk.
- Support only Full speed. So, the maximum bitrate is 1Mbps.
- Support only input-terminal and output-terminal. No support any other unit.
- No support dynamic streaming settings modification.
- No support still image capture.

The video capture example runs as a virtual analog video capture device. The device captures TinyUSB logo animation as 128x96@10fps NV12(YUV420 planar). The video source is hard-coded into the flash.

**Additional context**
The video capture example was tested by using FRDM-KL25Z with [VirtualDub](https://www.virtualdub.org/) for Windows.
CI reported some errors on other boards. I will resolve these errors to be ready for merge.


hathach:
AMAZING!!! brilliant PR :+1: :+1: , thank you very much for making this. UVC along with UAC are 2 most complicated classes. This awesome PR with testable examples will greatly help further development (iso out, highspeed), and I have no doubt that we could pull this off together in the next several PRs. 

Since the changes are huge, please give me a bit of time to digest it. I will also take the chance to test it with other ports as well. We will probably make an separated issue to keep track of which ports is tested with UCV (and UAC as well) similar to Compliance Verification in the future.

kkitayam:
Thank you for your comments.
If you find any issues or questions, please let me know.

kkitayam:
Fix CI errors, and I confirmed video capture example works fine on Raspberry Pi Pico and RX65N target board.

hathach:
> Fix CI errors, and I confirmed video capture example works fine on Raspberry Pi Pico and RX65N target board.

Superb ! thank you for more testing, I will check this out and do more testing on other ports within this week as well. It may be the good chance to enhance/update the ISO api as well. Current one has issue with certain port when switching the bandwidth.I have some idea, but haven't got time to do something with it. Thank you again :+1:

hathach:
@kkitayam I need help with the testing, I have tried this on both KL25z and pico with windows. However, I couldn't get the logo displayed. It could be me since I am not familiar with virtualdub. I am on windows 10 with virtualDub 1.10.4 amd64. What I have done:
- Launch virtualDub, File -> Capture AVI ...
- Device -> select TinyUSB UVC
- Didn't see anything
- Try to change  Video -> Set custom format, but vb doesn't allow me to set to anything else.

Screenshot 2021-10-06 183223
Screenshot 2021-10-06 183157


On Linux uvc is detected (yah !!!) but it complain that format is unknown maybe it does not support our device **128x96@10fps NV12(YUV420 planar)** and not valid video chain. Not a problem at all, just to mention my testing. I will try to do further testing with linux later on.

```
[45883.869206] usb 3-4: Product: TinyUSB Device
[45883.869208] usb 3-4: Manufacturer: TinyUSB
[45883.869209] usb 3-4: SerialNumber: 123456
[45883.877143] uvcvideo: Unknown video format 3231564e-0000-0010-0080-719b3800aa00
[45883.877147] uvcvideo: Found UVC 1.50 device TinyUSB Device (cafe:4020)
[45883.877150] uvcvideo: No valid video chain found.
```

kkitayam:
@hathach Could you try following steps. I had tested the followings. Please note that it is not possible to preview by VirtualDub with this patch. We can record a video stream into a file after that play the video file.

- Launch virtualDub, File -> Capture AVI ...
- File -> Set capture file...
image
Input a file path to save captured video stream.
- Device -> USB (Direct Show)
image
- Video -> No display
image
- Capture -> Capture Video
image
- A timer count is displayed to indicate status capturing a video stream from TinyUSB device.
image
Wait few seconds to save some video frames into the avi file specified at the previous step.
- Capture -> Stop Capture
image
- File -> Exit capture mode
image
- File -> Open video file... -> Choose the file specified at the previous step
image
- File -> Preview input
image
Video stream will be rendered on the pane on virtualdub.
image

> On Linux uvc is detected (yah !!!) but it complain that format is unknown maybe it does not support our device 128x96@10fps NV12(YUV420 planar) and not valid video chain. Not a problem at all, just to mention my testing. I will try to do further testing with linux later on.

Thank you for testing it! I am happy that UVC is detected on Linux! I agree that the video format may be not supported by Linux. I will try to switch the video format to YUY2. YUY2 is a typical video format for analog, so I guess it is supported. But, YUY2 needs more ROM area larger than NV12.

kkitayam:
@hathach Thank you for confirming this patch on the several boards. That's great it works on your boards! 

> if size is an issue, we could change to good old color bar pattern or pattern that could be generated without having a huge flash.

Yeah, I am agree with you on the size issue. This weekend I'm going to try to change the video to the generated pattern or somethings to reduce rom size.

Actually, I was going to implement a color bar display but I didn't. Because there were many things to consider like as followings, I got tired of doing.
How do I generate color bar patterns? Should I implement BT.2111 or other specification? How do I move color bars? Panning? Zooming? Etc.
And, streaming process(tud_task) must be called every 1 ms to keep bandwidth. Therefore each generation process(video_task) must be completed earlier than 1 ms. In no OS environments, if video_task exceeds 1ms, it must be kept within 1 ms using pipelining or multi-stage processing.

That's why I decided to use the logo at first implementation.

me-no-dev:
@kkitayam I really want to try this on ESP32-S2+camera. Unfortunately, outputting 12bit YUV is not an output option for the sensors. Could you link me to some docs to see how should I go about changing the format to MJPEG? It could actually make for a somewhat decent webcam :)

kkitayam:
@me-no-dev Thank you for your comment. I am happy that you are interested in this PR!

> Could you link me to some docs to see how should I go about changing the format to MJPEG?

Unfortunately, in current status this PR has no support for motion JPEG. To add support for motion JPEG, we need to modify at least `_update_streaming_parameters()` and `_negotiate_streaming_parameters()` to handle motion JPEG.

Motion JPEG is a frame-based format, so it should not be too difficult to add support. In the next step I will try to add support.

hathach:
> @hathach
> 
> I have changed the video data to scrolling [EBU color bars](https://en.wikipedia.org/wiki/SMPTE_color_bars) and stream format to YUY2 at [b6d09ca](https://github.com/hathach/tinyusb/commit/b6d09ca7619b2f624cc3ee92a795ba2a9b9d68e9).

Superb! I will do more tests with the new format (couldn't do in the weekend since there is lots of house work to be done :smile: )

qwertymodo:
I've been playing around with this code, and have run into some issues with customizing it for my application.  First of all, I'm unable to change the frame rate at all.  Anything other than 10 (higher OR lower) causes Windows to no longer recognize it as a connected video device.  Also, I'm trying to increase the resolution, and have hit a few walls there.  I know that there are going to be limits wrt frame buffer size, due to RAM and time constraints, but I thought I'd post my findings here in case I'm missing something.  I'm currently using the Raspberry Pi Pico, and if I increase the resolution above 128x128, the video device does show up, but all I get is a black screen, no color bars.  I'm able to make it work if I also increase CFG_TUD_VIDEO_STREAMING_EP_BUFSIZE in tusb_config.h, which seems to max out at 1023 (if I set it to 1024 or anything higher, Windows no longer sees the video device).  With the ep bufsize set to 1023, I can get the resolution up to about 256x190 before the color bars stop displaying and I just get a black screen again.  Am I hitting hardware limitations here, or is there something else I can try tweaking to get the resolution up further?

kkitayam:
@qwertymodo   Thank you for your comments!

> I'm able to make it work if I also increase CFG_TUD_VIDEO_STREAMING_EP_BUFSIZE in tusb_config.h, which seems to max out at 1023 (if I set it to 1024 or anything higher, Windows no longer sees the video device).

It is a limitation of the USB Full-Speed specification. Isochronous transfer of USB-FS up to 1023 bytes. Please find attached picture for USB 2.0 spec. Raspberry PI Pico supports only USB-FS in my understanding.
image

Current TinyUSB UVC class driver supports only isochronous transfer, but bulk transfer. The maximum bandwidth of the driver is 1 mega-bytes per second. It is a limitation for isochronous transfer of USB-FS.

> With the ep bufsize set to 1023, I can get the resolution up to about 256x190 before the color bars stop displaying and I just get a black screen again. Am I hitting hardware limitations here, or is there something else I can try tweaking to get the resolution up further?

I think the settings reaches almost USB-FS limit. A 256x190 image needs 97,280 bytes(=256 x 190 x 16bits) per a frame in YUY2 raw pixel format. So, 256x190 @ 10 FPS consumes 972,800 bytes per second bandwidth.

To increase resolution and frame rate, We need to add other video formats reducing bandwidth consumption like as motion-jpeg.

qwertymodo:
Good catch on the payload size limit, as well as the bandwidth discrepancy between isochronous and bulk transfers.  It's been awhile since I did my last deep dive into the arcane lore that is the USB-IF documentation.  Any ideas why I'm unable to change the framerate at all?  Even lowering it to e.g. 8fps causes Windows to no longer recognize it as a video device.

kkitayam:
> Any ideas why I'm unable to change the framerate at all? Even lowering it to e.g. 8fps causes Windows to no longer recognize it as a video device.

I had tested with 10FPS and 5FPS. If you change to 5FPS instead of 8FPS, it works with Windows.

I have reproduced that `video_capture` example fails enumeration process with `#define FRAME_RATE    8` in `usb_descriptor.h`
And [USB Device Tree Viewer](https://www.uwe-sieber.de/usbtreeview_e.html) shows warning messages below. 
```
*!*WARNING:  dwMaxFrameInterval minus dwMinFrameInterval is not evenly divisible by dwFrameIntervalStep, this could cause problems
```
I think that the following statements causes this warning.
 https://github.com/hathach/tinyusb/blob/3b09b82123a50bef6b18cf90c2734ae7581da4a3/examples/device/video_capture/src/usb_descriptors.h#L106

dwMinFrameInterval and dwFrameIntervalStep are calculated by `10000000/_fps`. 10000000 is not evenly divisible by 8. Then the warning message occurs. So, in current `video_capture` implementation, FRAME_RATE must be a divisor of 10000000.
Of course, I think it is better to be able to specify any FPS. I will attempt to fix it later.

qwertymodo:
> dwMinFrameInterval and dwFrameIntervalStep are calculated by `10000000/_fps`. 10000000 is not evenly divisible by 8. Then the warning message occurs. So, in current `video_capture` implementation, FRAME_RATE must be a divisor of 10000000. Of course, I think it is better to be able to specify any FPS. I will attempt to fix it later.

10000000 / 8 = 1250000, so it's not that, but from that warning it's probably something in the interaction between dwMaxFrameInterval, dwMinFrameInterval, and dwFrameIntervalStep

qwertymodo:
Ok, I think I might have found the solution, but as I said, it's been awhile since I've read through the docs, so this is just a "seems to work" fix, without really being sure it's correct.  I think the only change that needs to happen is that dwFrameIntervalStep needs to include the _fps factor.  Changing this

https://github.com/hathach/tinyusb/blob/3b09b82123a50bef6b18cf90c2734ae7581da4a3/examples/device/video_capture/src/usb_descriptors.h#L106

to

`(10000000/_fps), (10000000/_fps), 10000000, (1000000/_fps)), \ `

seems to work.  You may want to add or remove a zero or two in there, depending on the dwFrameIntervalStep granularity you actually want.

kkitayam:
@qwertymodo  Thank you for the solution! I have confirmed the issue is fixed.

I will create a PR for your solution with minor modification. It is shown below. I had confirmed it works with 6FPS and 8FPS.

`            (10000000/_fps), (10000000/_fps), (10000000/_fps)*_fps, (10000000/_fps)), \`

qwertymodo:
What's the point of the 3rd argument being divided by and then multiplied by _fps, doesn't that cancel out?  Or does that achieve some kind of compile-time rounding?

kkitayam:
> What's the point of the 3rd argument being divided by and then multiplied by _fps, doesn't that cancel out? Or does that achieve some kind of compile-time rounding?

Yes, the value of `(10000000/_fps)*_fps` will be truncated to an integer.

When `_fps` is an integer, `(10000000/_fps)` is evaluated an integer and the decimal part of the result is truncated. So, when `_fps` is not a divisor of 10000000, `(10000000/_fps)*_fps` will be an integer which is not equal to 10000000.

PR #1122, opened 2021-10-03
WiP : MIDI Host

csBlueChip:
MIDI Host - work in progress

Includes descriptor decoding routines (I'm learning USB as I go)

Some commits can be included immediately as they are generic "bug fixes" ...these commits are flagged "**" because I have no idea how to use git properly (yet)

The MIDI Engine code (~3kloc) [see MIDI folder at root level] cannot be properly integrated, tested, and debugged until I get the basic driver going - which is not going well :( ...but it at least compiles in a test harness :)


src/class/audio/audio.h
```diff
@@ -663,15 +663,16 @@ typedef struct TU_ATTR_PACKED {
...
 
 /// AUDIO Class-Specific AC Interface Header Descriptor (4.7.2)
+// https://www.usb.org/sites/default/files/midi10.pdf (B.3.2) ...tested against a real device
 typedef struct TU_ATTR_PACKED
 {
-  uint8_t bLength            ; ///< Size of this descriptor in bytes: 9.
-  uint8_t bDescriptorType    ; ///< Descriptor Type. Value: TUSB_DESC_CS_INTERFACE.
-  uint8_t bDescriptorSubType ; ///< Descriptor SubType. Value: AUDIO_CS_AC_INTERFACE_HEADER.
-  uint16_t bcdADC            ; ///< Audio Device Class Specification Release Number in Binary-Coded Decimal. Value: U16_TO_U8S_LE(0x0200).
-  uint8_t bCategory          ; ///< Constant, indicating the primary use of this audio function, as intended by the manufacturer. See: audio_function_t.
```
kasjer, line 672: This changes definition from Audio 2 specification section 4.7.2 to version 1.0 section 4.3.2 with limitation to 1 interface.

Maybe better way to handle this is to have 1.0 version separated and not change just one structure to different specification version.

csBlueChip: Ah. Yes. I didn't realise that was the root of the problem! 

I _would_ suggest the v1 be called `audio_desc_cs_ac_interface_t` and v2 `audio_desc_cs_ac_interface_v2_t` ...but I guess that might be seen as an undesirable breaking-change :/

Do you have any ideas for the best/right way to do this?

hathach:
> ...I'm a n00b to "the git(hub) way", and mostly just wanted to make sure I exposed everything here - in case someone else wants to help, or 'steal' some code. [I also need to add MIT headers to all my files]

No problem at all, I am not git expert either, just use some GUI, that make adding file, commit and push easier. I uses git-cola on Linux, but any should do the work.

timonsku:
Is the work on this still active? MIDI host support would be quite useful for a project I'm working on

rppicomidi:
@timonsku I wrote a USB MIDI host application driver for TinyUSB in the [usb_midi_host](https://github.com/rppicomidi/usb_midi_host) project. You might find that helpful. I have only tested it on the RP2040 on a Raspberry Pi Pico board and a Pico W board.

hathach:
midi host is implemented in https://github.com/hathach/tinyusb/pull/1627

PR #637, opened 2021-02-07
add basic support for rt-thread

tfx2001:
**Describe the PR**
Add a basic support for RT-Thread RTOS. More about the [RT-Thread](https://www.rt-thread.io/).

**Additional context**
None.

PR #1149, opened 2021-10-16
add rt-thread software package support

tfx2001:
Hi,

This PR based on #637 by supporting RT-Thread RTOS. TinyUSB has been registered to the [RT-Thread software packages center](http://packages.rt-thread.org/en/search.html?search=tinyusb). Developers can use TinyUSB more easily by installing the TinyUSB package in package manager center.

This PR includes two `SConscript` files. RT-Thread uses Scons to organize files rather than CMake or Makefile, so it will not cause any conflicts with existing Makefiles. Meanwhile, this PR also contains some drivers for interfacing with the RT-Thread device framework.

However, there is a problem that the added files contain headers (those will be used in RT-Thread side) that do not exist in this repository, which will cause errors at the CI. Is it possible to exclude the `osal/rt-thread` folder?

I would be grateful if you consider about this PR. Thanks ahead.

Frank

[https://www.rt-thread.io](https://www.rt-thread.io/)

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -88,6 +88,7 @@ TinyUSB is completely thread-safe by pushing all Interrupt Service Request (ISR)
 - **No OS**
 - **FreeRTOS**
 - **Mynewt** Due to the newt package build system, Mynewt examples are better to be on its [own repo](https://github.com/hathach/mynewt-tinyusb-example)
+- **RT-Thread**
 
 Local Docs
 ==========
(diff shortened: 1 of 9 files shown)


tfx2001:
RT-Thread is not just a kernel. RT-Thread contents a device driver framework layer which will provide a uniform API for users. These `*.c` files port for device driver framework rather than a specific application.

PR #1164, opened 2021-10-25
update dcd_esp32sx with correct include list

suda-morris:
**Describe the PR**
A tiny minor update. `dcd_esp32sx.c` doesn't use any APIs in periph_ctrl.h but the USB interrupt source number listed in `soc/periph_defs.h`.


suda-morris:
Hi @hathach Will you consider this quick MR? Some API like `periph_ctrl.h` is going to be private, thus the including path will slightly change in the future. This MR can make the DCD driver stay work.

suda-morris:
Hi @hathach Will you consider this quick MR? Some API like `periph_ctrl.h` is going to be private, thus the including path will slightly change in the future. This MR can make the DCD driver stay work.

hathach (approved):
look good. thank you for the pr.

PR #1171, opened 2021-10-29
Add Adafruit QT Py board

mkj:
This just adds a simple config for the board, I've tested that it's usable with a fork of https://github.com/greatscottgadgets/apollo

Since this board doesn't have a standard LED or BUTTON I've made those optional


hathach (approved):
superb! thank you very much for your pr.

PR #1172, opened 2021-10-30
Adds support for Adafruit's Trinket M0

jgressmann:
Adds support for [Adafruit's Trinket M0 board](https://learn.adafruit.com/adafruit-trinket-m0-circuitpython-arduino).

Also adds initial support for UART for the board family.

### DIFF
--- a/hw/bsp/samd21/family.c
+++ b/hw/bsp/samd21/family.c
@@ -45,6 +45,11 @@ void USB_Handler(void)
 }
 
 //--------------------------------------------------------------------+
+// UART support
+//--------------------------------------------------------------------+
+void uart_init(void);
+
+//--------------------------------------------------------------------+
 // MACRO TYPEDEF CONSTANT ENUM DECLARATION
 //--------------------------------------------------------------------+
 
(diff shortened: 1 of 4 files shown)


hw/bsp/samd21/family.c
```diff
@@ -44,6 +44,11 @@ void USB_Handler(void)
   tud_int_handler(0);
 }
 
+//--------------------------------------------------------------------+
+// UART support
+//--------------------------------------------------------------------+
+TU_ATTR_WEAK void uart_init(void);
```
hathach, line 50: since we always defined `uart_init()`, why would we use weak function here. Are you planning to allow application to overwrite this ? Otherwise I think we could just simply remove the weak attribute

jgressmann: Yes, exactly. The Trinket M0 only has very few pins / pads available for external hardware components. 
In fact, I would argue it would make sense to make `board_init()` a weak function so users can easily replace it if need to.

hathach: it only makes sense if this repo providing board abstraction for application. Unfortunately, I have no intention to do so. The whole bsp and examples in this repo is only used to demonstrate how tinyusb API is used, and prove and test that it works on hardware. These can be changed completely without any notices.

Sum up application and user should write their own bsp init code for their own board. Therefore please remove the weak attribute in the function.

hathach (approved):
superb! Thank you very much for the PR.

PR #1176, opened 2021-11-01
samd21: enable SysTick only if running w/o OS

jgressmann:
When using FreeRTOS, with the `SysTick` being active, there is a race condition between the elapsing of the first tick and execution of `vTaskStartScheduler()`.


hathach (approved):
Yeah, this should be done for all ports. I have done to most but still forget to do it all. Thank you for the pr

PR #1178, opened 2021-11-04
broadcom/dcd: Fix mynewt build

kasjer:
**Describe the PR**
dcd_synopsys.c included **broadcom/interrupts.h** before
preprocessor check whether file should be used.

Since mynewt build does not include all folders it resulted in build
error.

Now offending include is moved few lines after preprocessor check.

**Additional context**
No change in logic.
No build should be negatively affected by this change.

### DIFF
--- a/src/portable/broadcom/synopsys/dcd_synopsys.c
+++ b/src/portable/broadcom/synopsys/dcd_synopsys.c
@@ -29,6 +29,10 @@
 
 #include "tusb_option.h"
 
+#if TUSB_OPT_DEVICE_ENABLED &&         \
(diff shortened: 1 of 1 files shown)


tannewt (commented):
Why also guard the common include and the USE_SOF define?

hathach (approved):
thank you for the PR, this is the reason why my mynewt example failed the ci

hathach:
> Why also guard the common include and the USE_SOF define?

newt build system and other build system with auto-generate makefile such as arduino include all (dcd) sources files by default. When BCM2711 is not its target mcu, the include path of broadcom peripherals is not set and not even checkout. It is typical for dcd driver to check for mcu supported before including its peripheral header.

PR #1181, opened 2021-11-05
Fix invalid assert in RP2040 host mode when compiled in debug mode

liamfraser:
RP2040 HCD: Move invalid ep->active assert in hw_trans_complete. The ep->active should only happen if a setup packet was just sent. Otherwise the transaction is handled in hw_handle_buff_status. Fixes https://github.com/raspberrypi/pico-sdk/issues/649


hathach (approved):
thank you for your PR, apparently I never run with DEBUG build with pico-sdk. I mostly just run `make LOG=2 all` as I did with other mcu. Will try to set the build to debug when LOG > 0 in the cmake later on.

PR #1182, opened 2021-11-05
Update bth for IAD support

kasjer:
**Describe the PR**
**IAD** was missing from BT interface descriptor macro.
This in itself did not matter much but it is in the specification so descriptor is updated now.

With missing **IAD** BT driver handled open separately for both interfaces.
For some time now BT is assumed to have 2 associated interfaces even when IAD is missing.
This inferred association resulted in only one call to **btd_open()** and second interface was no
handled at all.
This in turn resulted in assertion failure in **process_set_config()**.

Now single call to **btd_open()** checks for presence of both interfaces as expected by
**process_set_config()**.

This only changes conditions and how **itf_desc** is modified in the middle of **btd_open()**,
rest is just indentation change due to if () removal

### DIFF
--- a/src/device/usbd.h
+++ b/src/device/usbd.h
@@ -763,8 +763,8 @@ TU_ATTR_WEAK bool tud_vendor_control_xfer_cb(uint8_t rhport, uint8_t stage, tusb
 #define CFG_TUD_BTH_ISO_ALT_COUNT 0
 #endif
 
-// Length of template descriptor: 30 bytes + number of ISO alternatives * 23
-#define TUD_BTH_DESC_LEN (9 + 7 + 7 + 7 + (CFG_TUD_BTH_ISO_ALT_COUNT) * (9 + 7 + 7))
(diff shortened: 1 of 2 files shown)


src/device/usbd.h
```diff
@@ -806,6 +806,8 @@ TU_ATTR_WEAK bool tud_vendor_control_xfer_cb(uint8_t rhport, uint8_t stage, tusb
 // Interface number, string index, attributes, event endpoint, event endpoint size, interval, data in, data out, data endpoint size, iso endpoint sizes
 // TODO BTH should also use IAD like CDC for composite device
```
hathach, line 807: I did put a TODO for IAD here as well :smile:

hathach (approved):
Thank you, I notice this as well and put the TODO note for this :+1: . IAD in general is preferred since it allow device to be **composite* and able to add more class such as DFU runtime etc ...

PR #1210, opened 2021-11-25
Example with webusb, cdc and hid

TheStaticTurtle:
**Describe the PR**
This PR add an example for the use of multiple interface (CDC, Vendor and two HIDs). It also helps to clarify a bit about the difference between the interface number and the instance number (See https://github.com/hathach/tinyusb/discussions/1202)


TheStaticTurtle:
Please verify that I didn't make a mistake while numbering ep nums for the LPC and SAM mcus as I don't have one to test it with

hathach (commented):
Thank you for the PR, however we can't have all class combination as example. Most of this example is already covered by other such as webusb, cdc, hid respectively.

TheStaticTurtle:
OK, no worries, I think it would be a good idea to have a section in the docs about the difference between interface number, and instance numbers as it wasn't really clear reading the existing examples, the discussion clarified it tho.

PR #1216, opened 2021-11-29
Update doc

hathach:
**Describe the PR**
update supported list


perigoso:
good job, looks great 👍

hathach:
Thanks, it takes me too much time to edit the table with retext app. Which app do you use to edit rst file ?

HiFiPhile:
In the table LPC43xx missed highspeed support :)

perigoso:
> Thanks, it takes me too much time to edit the table with retext app. Which app do you use to edit rst file ?

I have some restructuredtext extensions on vscode, i made the table in csv and one of the extensions formated it to rst, i honestly dont know which did what
To edit Is a bit harder, never had to deal with a big table like this so never really looked into it, usually just do it by hand (this one is too big to be dealing with that though)

perigoso:
There are other supported table formats, maybe CSV would be good here, very simple.
I chose this because it allowed some fancier things like joined rows for the same manufacturer (which i think works great)

hathach:
> There are other supported table formats, maybe CSV would be good here, very simple. I chose this because it allowed some fancier things like joined rows for the same manufacturer (which i think works great)

yeah, I agree, this format look great, though editing it is a bit of a pain. though I get used to it already. 

> In the table LPC43xx missed highspeed support :)

right, will correct it next

kbladewht:
why this commit is the default tinyusb? I just see the submodoule default url

PR #1220, opened 2021-12-01
Add support for Allwinner F1C100s family

t123yh:
Allwinner F1Cx00s family is a series of SoCs with Mentor MUSB controller and HS phy. It comes with a slightly different register space layout, and some quirks, so it's not multiplexed with the existing musb support library.

This porting currently requires to be compiled with https://github.com/hongxuyao/F1C100s_with_Keil_RTX4_emWin5 . Support for Keil RTX4 is also added in this pull request.

There is a working CDC loopback demo project at https://github.com/t123yh/F1C100s_RTX4_USB .


hathach:
Superb !! Thank you very muhc, this is great PR. This would be great option to run TinyUSB on another popular SOC besides rpi broadcom. As required for maintaining the driver, could you please:
- Tell which board you are testing with, I am planning to order and **Sipeed Lichee Nano** to test with your PR. 
- Add mcu and bsp files so that existing example could run on the actual hardware. This is actually very important, it not only provides proof of concept but also ease the maintaining/developing in the future. You don't have to organize things as other ports/boards. You could just bundle all of the files into your board. Once it is compiled and tested, I will move them into separated mcu submodule later on. 

PS: I am ok with Allwinner F1C100s with is own musb driver for now. The existing musb is also newly added. We could later on merge them together in the future (much like synopsys dwc2 driver).

t123yh:
Hi hathach,

Thanks very much for your response!

I'm indeed testing with a Lichee Pi Nano, but it's more convenient to use a Widora V200 (Now called MangoPi-R3c). MangoPi-R3c have two USB ports for OTG and UART and comes with dedicated RESET and BOOT button, which brings a lot convenience to development.

Now a basic example is provided here: https://github.com/t123yh/F1C100s_RTX4_USB, which requires Keil to be compiled. I'll try to port it to the BSPs in TinyMCU later.

I forgot to mention that there's currently a bug in the current code: TX and RX cannot share an Endpoint Number. I don't know whether it's a hardware limitation or software bug, though from existing examples the hardware likely supports sharing Endpoint Numbers. This needs to be digged deeper.

hathach:
@t123yh thank you for the tip, I just placed an order for the MangoPi R3 from aliexpress instead of Lichee Pi nano. It probably takes a couple of weeks for it to deliver. Meanwhile we could try to get familiar with the chip and the PR finalized :+1: :+1:

t123yh:
BTW, currently I have only tried Keil RTX on F1C100s, but I believe RT-Thread will also work. I'll try later.

t123yh:
BSP support is now added. No RTOS is required.

Please see [README.md](https://github.com/t123yh/tinyusb/blob/master/hw/bsp/f1c100s/README.md) for details.

src/portable/sunxi/dcd_sunxi_musb.c
```diff
@@ -416,6 +423,8 @@ static unsigned find_free_memory(uint_fast16_t size_in_log2_minus3)
   }
   print_block_list(free_blocks, num_blocks);
 
+  USBC_SelectActiveEp(backup_ep);
```
kkitayam, line 432: Please let me know why this operation is needed.  I would like to know if there is a bug in `dcd_musb.c`

I had designed `dcd_msub.c` that EPIDX register must be written when accessing the registers for FIFO settings(TXFIFOSZ, RXFIFOSZ, TXFIFOADD, and RXFIFOADD). My understanding is EPIDX effects only these registers. So, there is no need to keep the EPIDX value for a long time. If the EPIDX value needs to save and restore, I missed something condition.

t123yh: Hi Koji,

In my design, I didn't use the addressed endpoint registers. EPIDX is used in all places. In `dcd_edpt_open`, after calling `find_free_memory`, if current EP index is not restored, the following adjustments to FIFO addr and size will go to the wrong endpoints. In your code however, you adjusted EPIDX after `find_free_memory`, so you are not affected here.

kkitayam: Thank you for your reply.

Ah, I see, you moved the statement of assigning a value to EPIDX register to before calling find_free_memory.

t123yh:
After fixing a stupid mistake, the cdc_msc example is now working.

t123yh:
Just played with the msc. It seems that there're a few STALLs before the disk is properly detected. Seems that the cause is the code below. I'm not familiar with msc, so I don't know what's happening.
```c
if ( p_msc->stage == MSC_STAGE_STATUS )
  {
    // skip status if epin is currently stalled, will do it when received Clear Stall request
    if ( !usbd_edpt_stalled(rhport,  p_msc->ep_in) )
    {
      if ( (p_cbw->total_bytes > p_msc->xferred_len) && is_data_in(p_cbw->dir) )
      {
        // 6.7 The 13 Cases: case 5 (Hi > Di): STALL before status
        // TU_LOG(MSC_DEBUG, "  SCSI case 5 (Hi > Di): %lu > %lu\r\n", p_cbw->total_bytes, p_msc->xferred_len);
        usbd_edpt_stall(rhport, p_msc->ep_in);
      }else
      {
        TU_ASSERT( send_csw(rhport, p_msc) );
      }
    }
```

Also I tried to do a cdc benchmark with dd (loopback code commented out). The result is not very impressive (as a High Speed device). Also, after repeating a few times of benchmark, the device just dies. No further benchmark can be made.
```
[I] root@t123yh-XPS15-9570 ~/t/e/d/cdc_msc (master)# dd if=/dev/zero of=/dev/ttyACM0 count=10000
10000+0 records in
10000+0 records out
5120000 bytes (5.1 MB, 4.9 MiB) copied, 2.37053 s, 2.2 MB/s
```

hathach:
> BSP support is now added. No RTOS is required.
> 
> Please see [README.md](https://github.com/t123yh/tinyusb/blob/master/hw/bsp/f1c100s/README.md) for details.
> 
> Currently tested with `audio_test` and `hid_generic_inout` and `cdc_msc`. `audio_test` is working. `cdc_msc` is not working, only echoing the first character.

Superb !! That is great, too bad, my tiny200 haven't shipped to me just yet for trying this out. This will make things super easy to support and maintain F1C in the future.

> Just played with the msc. It seems that there're a few STALLs before the disk is properly detected. Seems that the cause is the code below. I'm not familiar with msc, so I don't know what's happening.

MSC device may need to stall unsupported command and request from host. It is fairly typical, and required to pass USB compliance test for MSC device #1059 . We may have issue if we haven't fully implement stall/clear stall properly for this port, which is very typical for new chip. 

> Also I tried to do a cdc benchmark with dd (loopback code commented out). The result is not very impressive (as a High Speed device). Also, after repeating a few times of benchmark, the device just dies. No further benchmark can be made.

It is new port and may need some tweak, I have tested fullspeed device with cdc that could reach 80-90% of the bandwidth. E.g you could try to increase the `CFG_TUD_MSC_EP_BUFSIZE to 4096` to see if that help
- https://github.com/hathach/tinyusb/blob/master/examples/device/cdc_msc/src/tusb_config.h#L113

It is a great start, I am looking forward to trying this out.

t123yh:
Now there's one thing that's to be solved: if we want to build a flashable bin file to run on SPI Flash, we must use [mksunxi](https://github.com/VeiLiang/EasyUSB/blob/master/tools/f1c100s/mksunxi/mksunxi.c) tool to process the output bin file. I'm not sure where can I insert this into our build flow, could you please help me?

hathach:
@t123yh in general, tool should be installed separately by user and added to PATH similarly to gcc toolchain or openocd. You could also make it as makefile variable to pass through command line e.g we can put it in the board/family readme doc.

```
# default to mksunxi is in PATH
MKSUNXI ?= mksunxi
```

if user need to, it can be passed to makefile as `make MKSUNXI=path-to-tool`. Let me know if that make sense to you.

hathach:
> I've written a Python version to avoid the need of installing mksunxi. Now it will build a flashable image that you can write to SPI. README.md is updated to reflect the changes.

great works, do you have a need to merge it asap ? my board from aliexpress could take 2 more weeks to arrive. I would like to have hand-on testing and could make some cosmetic changes if needed. But we could merge it earlier if this blocks your work.

t123yh:
> great works, do you have a need to merge it asap ?

No need, it's better to merge after you have tested the whole thing. I don't think it's currently stable enough to be merged. I can just work on my own repo before merging.

BTW, after fixing a bug (commit 527036b) that causes MMU and cache not initialized when executing from FEL, serial performance test reaches 15.7MB/s, which is a lot better than previous tests.

hathach:
> No need, it's better to merge after you have tested the whole thing. I don't think it's currently stable enough to be merged. I can just work on my own repo before merging.

great, I prefer to have some local testing as well.

> BTW, after fixing a bug (commit [527036b](https://github.com/hathach/tinyusb/commit/527036b1f5a3f0aa3994f62e3ab18fa990a40cf8)) that causes MMU and cache not initialized when executing from FEL, serial performance test reaches 15.7MB/s, which is a lot better than previous tests.

Great to hear that, indeed caching can have an great impact on the performance, thank you for your update :+1: :+1:

t123yh:
@kkitayam Hi,

I am playing with [gud](https://github.com/notro/gud-pico) on F1C100s today. I doubt if there's a bug in the logic of musb dcd driver around `TU_LOG1("Drop CONTROL_STAGE_ACK\n");`. 

Consider the following situation:

1. Host sent a CONTROL SETUP IN (req1) with wLength > 0;
2. Device responds with wLength = 0 DATA OUT in respond to req1;
3. Host responds with STATUS OUT (musb handles this automatically);
3. Another CONTROL SETUP IN (req2) is sent immediately by host, so both events are processed in the same interrupt (this means CONTROL_STAGE_ACK is dropped);
4. Device stack calls `edpt0_xfer` to respond to req2 with some data;
5. musb driver now enters `if (req == REQUEST_TYPE_INVALID || _dcd.status_out) {` logic. **This will cause the response to req2 be dropped!**

I'm not sure if the logic flow above is valid or not. If it's valid, I would propose two possible solutions to the problem:

* Make some adjustments to `process_ep0` so that if `_dcd.status_out` is 1 when entering `process_ep0` -> `USB_CSRL0_RXRDY`, send a `dcd_event_xfer_complete` immediately to notify the stack;
* Or, simply move `return true;` to the if block of 'req == REQUEST_TYPE_INVALID'

kkitayam:
@t123yh

I have understood that the logic flow you said.  Please find attached sequence diagram. I illustrated this problem. 

sequence-3

In addition, I think we need to consider following conditions.
- A) Normal transfer
normal
- B) STATUS OUT request comes before TUD calls dcd_xfer()
status
- C) The next SETUP request comes before TUD calls dcd_xfer() for current STATUS OUT
delayed_task

> Make some adjustments to process_ep0 so that if _dcd.status_out is 1 when entering process_ep0 -> USB_CSRL0_RXRDY, send a dcd_event_xfer_complete immediately to notify the stack;

I guess this way is good. I would consider whether the way works for other conditions.

> Or, simply move return true; to the if block of 'req == REQUEST_TYPE_INVALID'

I think this way is not good for `C) The next SETUP request comes before TUD calls dcd_xfer() for current STATUS OUT`. In this case, DCD should ignore `dcd_edpt_xfer` called by tud_task.

kkitayam:
> > Make some adjustments to process_ep0 so that if _dcd.status_out is 1 when entering process_ep0 -> USB_CSRL0_RXRDY, send a dcd_event_xfer_complete immediately to notify the stack;
> 
> I guess this way is good. I would consider whether the way works for other conditions.

If this adjustments are applied, after the step 24 of `C) The next SETUP request comes before TUD calls `dcd_xfer()` for current STATUS OUT` , DCD will enqueue `DCD_EVENT_XFER_COMPLETE` to event queue. Therefore DCD generates event before TUD task calls dcd_xfer(). It is strange behavior, however, it might not be a problem as TUD task will call `dcd_xfer()` for STATUS OUT later.

@hathach 
Is it acceptable that DCD generates `DCD_EVENT_XFER_COMPLETE` before TUD task calls `dcd_edpts_xfer()` for STATUS OUT?

If this is not acceptable, I would consider whether this condition can be avoided by adding a new flag.
Or, if this is acceptable, I would modify that the step 20 of `Condition B)` and the step 20 of `Condition C)` generate `DCD_EVENT_XFER_COMPLETE` for consistency.

hathach:
@kkitayam the diagram is super useful, though I haven't tried to look closely at musb so far. Please give me a bit of time to catch up with this.

t123yh:
> @kkitayam the diagram is super useful, though I haven't tried to look closely at musb so far. Please give me a bit of time to catch up with this.

Hi hathach,

Just a friendly reminder on this issue.  Could you please take some time to figure this out? This has been a blocking bug for me for a while. BTW, have you received your Tiny200 board?

Thanks!

hathach:
@t123yh thanks for the reminder, I did indeed receive tiny200 board a few days ago. I have been busy with other works. Please give me a bit more time to get it setup and trying this out.

hathach (commented):
PR looks great, thank you very much for your effort. I only have one question regarding the keil warning suppression of "statement is unreachable"

src/device/usbd.c
```diff
@@ -44,6 +44,11 @@
   #define CFG_TUD_TASK_QUEUE_SZ   16
 #endif
 
+#ifdef __ARMCC_VERSION
+// Supress "statement is unreachable" warning
+#pragma diag_suppress 111
```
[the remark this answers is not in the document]
t123yh: The warnings are for the `switch(desc_type)` inside `process_get_descriptor` in which there're several `break;` that is unreachable because are `return` statements in both `if` and `else` block. Line 1016, 1033, 1059, 1072, 1086.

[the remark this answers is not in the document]
hathach: thank you I see, we could just remove the `break;` they are mostly are there for case/break pair.

hathach (approved):
I have tried it on my tiny200, and it works great. Thank you very much for the excellent PR. I did a bit of cosmetic change e.g adding license term to dcd file, move hw/mcu/allwinner to its [own submodule](https://github.com/hathach/allwinner_driver) etc.. Once the ci is passed, we could merge this

t123yh:
Hi @hathach ,

Thank you for testing. The bug mentioned above still exists, and I believe it needs to be resolved before merging. I have an example to reproduce this issue. I'll upload it to a branch soon.

hathach:
@t123yh no problem at all, I am downloading keil trial version on my VM windows, will try it out if you have any ready-to-build example

t123yh:
> @t123yh no problem at all, I am downloading keil trial version on my VM windows, will try it out if you have any ready-to-build example

No you don't have to download keil. You can run my example totally within the tinyusb example framework. The code to reproduce the problem is based on [gud-pico](https://github.com/notro/gud-pico) which is part of the [gud](https://github.com/notro/gud) project. You need to have RPi with the latest RPi OS (which backported gud to 5.10 kernel) or PC with Arch Linux to run my example.

t123yh:
@hathach The example is ready. You can downloaded it from [gud-test branch](https://github.com/t123yh/tinyusb/tree/gud-test/examples/device/gud). This is a dirty quick version based on earlier version of my port, but this shouldn't matter. It has my [workaround commit for the problem discussed above with kkitayam](https://github.com/t123yh/tinyusb/commit/8656b3bbe1282fc2a8452538d400f66d435cae8d) applied. You have to prepare a Linux host running kernel with [gud kernel module](https://github.com/torvalds/linux/tree/master/drivers/gpu/drm/gud). Latest RPi OS have this module backported.

To test the example, you can connect the USB device port to the host Linux device, and the UART port to your serial monitor on PC (note that the serial port is changed to UART1 to adapt to Tiny200v2), then compile and flash the gud code just like other examples. If you run my code directly, it should be recognized immediately on RPi as a USB screen (although it's sorry that you probably doesn't have a screen on your Tiny200v2, but it shouldn't matter), with the following logs on your serial monitor:

```
Boot to SPI mode
Timer INIT done
tud_mount_cb:
gud_req_get: request=0x1 index=0 size=30
XFER 30-30!!
gud_req_get: request=0x40 index=0 size=32
XFER 32-1!!
gud_req_get: request=0x41 index=0 size=128
gud_req_get: request=0x50 index=0 size=128
DONE5
XFER 128-5!!
Drop CONTROL_STAGE_ACK
gud_req_get: request=0x51 index=0 size=128
XFER 128-10!!
gud_req_get: request=0x54 index=0 size=1
XFER 1-1!!
gud_req_get: request=0x56 index=0 size=128
XFER 128-128!!
gud_req_get: request=0x55 index=0 size=128
XFER 128-24!!
Drop CONTROL_STAGE_ACK
controller_enable: enable=1
state_commit: mode=320x480 format=0x80 connector=0 num_properties=1
  prop=12 val=100
display_enable: enable=1
Set backlight: 100
```

Note the `CONTROL_STAGE_ACK` is where the problem occurs, but workarounded. If you revert my workaround commit (8656b3b), the whole thing will stop on the first `Drop CONTROL_STAGE_ACK`, and RPi will report an error in its dmesg.

Please tell me if you can reproduce this problem.

hathach:
Thank you for the detail info, I will grab an Pi4 and try to reproduce and look into this issue. To be honest, I haven't read the musb specs/code yet, thankfully, kkitayam already does an great job illustrated it.

t123yh:
musb doc can be found by Google `mentor usb sunxi`.

hathach:
yeah, I mean I haven't got time to really look at musb yet. So far all the great work is done by @kkitayam :)

hathach:
@t123yh sorry I am lacking behind PRs/issues. I still haven't managed time to look at musb and test the above problem just yet. This PR is already pending for a while and is relatively large. I think we should merge this as it is for now, since it has good running example already. And then do another PR to fix the problem, we could create an github ticket/issue for discussion. That would also minimize effort for other contributors to fix the issue as well. What do you say ?

t123yh:
> @t123yh sorry I am lacking behind PRs/issues. I still haven't managed time to look at musb and test the above problem just yet. This PR is already pending for a while and is relatively large. I think we should merge this as it is for now, since it has good running example already. And then do another PR to fix the problem, we could create an github ticket/issue for discussion. That would also minimize effort for other contributors to fix the issue as well. What do you say ?

Yes, I agree with you. Thanks for your time and effort.

hathach:
> Yes, I agree with you. Thanks for your time and effort.

Thanks, I wish I could be more involved with this port, but the Lunar New Year is coming and there is lots of other works/project. I will look at this later on when I have time. Thank you @t123yh for your contribution. I will make an new issue as reminder for us after the merge

t123yh:
> the Lunar New Year is coming and there is lots of other works/project

I totally understand this as a Chinese. Happy New Year!

hathach:
thanks, Happy New Year to you as well !!! :tada:

PR #1222, opened 2021-12-01
Remove unused-parameter errors when LOGGER=swo

scoudreau:
**Describe the PR**
Remove errors when compiling with LOGGER=swo.
All arguments of sys_read() must be used in function, otherwise it triggers "unused-parameter" errors.

**Additional context**
Before this patch, error was:

hw/bsp/board.c: In function '_read':
hw/bsp/board.c:126:44: error: unused parameter 'buf' [-Werror=unused-parameter]
  126 | TU_ATTR_USED int sys_read (int fhdl, char *buf, size_t count)
      |                                      ~~~~~~^~~
compilation terminated due to -Wfatal-errors.

### DIFF
--- a/hw/bsp/board.c
+++ b/hw/bsp/board.c
@@ -126,6 +126,8 @@ TU_ATTR_USED int sys_write (int fhdl, const void *buf, size_t count)
 TU_ATTR_USED int sys_read (int fhdl, char *buf, size_t count)
 {
   (void) fhdl;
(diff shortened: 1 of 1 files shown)


hathach (approved):
ah right, thank you for your PR. I was planning to add the SWO read, but then I realize swo is output only :D and kind of forgot to fix this.

PR #1225, opened 2021-12-02
after commit: eae4132 this dependency is no longer required

perigoso:
Signed-off-by: Rafael Silva <perigoso@riseup.net>

commit: eae4132


hathach (approved):
look good, thank you for your PR. Unicode is much easier to maintain the table with text editor. It is also displayed nicely with github web viewer. The only downside is it is not displayed as nicely on the generated doc. But easy to maintain take the priority.

PR #1227, opened 2021-12-03
Fix buffer overrun in dcd_musb driver

kkitayam:
**Describe the PR**
Fix the bug of buffer overrun at pipe_read_packet() in dcd_musb.c.

**Additional context**
Cast operations for the left side symbols were wrong. So, when packet length is not 4 bytes unit, it occurs buffer overrun and corrupts some buffers.


hathach (approved):
Great, thank you for your PR.

Issue #1232, opened 2021-12-04
Endpoints are not being closed on RP2040


vmilea:
### Operating System

Others

### Board

Raspberry Pi Pico

### Firmware

-

### What happened ?

The `dcd_edpt_close()` implementation in `dcd_rp2040.c` is incomplete. It was probably meant to call `hw_endpoint_close`, but that code remains commented out.

With USB audio playback it's common for the endpoints to close and reopen as needed. During silence, the driver selects alternate setting 0 to conserve bandwidth and the endpoint is closed. Then it's reopened when playback starts.

Because the endpoints are not properly closed on RP2040, there are warnings like:

`WARN: starting new transfer on already active ep 1 out`

and eventually it runs out of buffer space with failed assert `hw_data_offset(next_buffer_ptr) <= USB_DPRAM_MAX`.

### How to reproduce ?

Close and reopen endpoints repeatedly.

### Debug Log

-

### Screenshots

-

vmilea:
Pull request submitted. A couple of notes: 
* To keep is simple, buffer space is only reclaimed after _all_ non-control endpoints have closed.
* If a transfer event gets delivered late, after closing the endpoint, it will be ignored. This never happened in my testing.

geurtv:
Why this (more complicated) partial fix that actually seems to introduce a new issue (see below) instead of simply not assigning endpoint buffers multiple times?

Assuming `reset_non_control_endpoints()` will be called at a safe time: in `_hw_endpoint_alloc()` an `if (!ep->hw_data_buf)` around the assignment from `next_buffer_ptr` would fix it for all cases, right? That is, the warning will remain but no panics anymore and everything keeps functioning as expected. Note that `hw_endpoint_close()` must be `#if 0`'ed again, but do keep the `hard_assert`.

```c
  if (!ep->hw_data_buf) {
    ep->hw_data_buf = next_buffer_ptr;
    next_buffer_ptr += size;
  }
```
---

With the current fix and just an audio device I think I have the 'transfer event delivered late' bit reproducible, resulting in a `"Can't continue xfer on inactive ep 0 out"` panic. On a Raspberry Pi 3 I have a test application that transfers audio for 2 seconds then shuts down and is restarted after 1 second. The RP2040 is an audio device that simply echos audio received. The panic occurs mostly within one minute.

PR #1233, opened 2021-12-04
Handle the closing of endpoints on RP2040

vmilea:
**Describe the PR**
Close endpoints and reclaim buffer space on RP2040.

**Additional context**
Proposed fix for https://github.com/hathach/tinyusb/issues/1232.

### DIFF
--- a/src/portable/raspberrypi/rp2040/dcd_rp2040.c
+++ b/src/portable/raspberrypi/rp2040/dcd_rp2040.c
@@ -83,7 +83,7 @@ static void _hw_endpoint_alloc(struct hw_endpoint *ep, uint8_t transfer_type)
 
   assert(((uintptr_t )next_buffer_ptr & 0b111111u) == 0);
   uint dpram_offset = hw_data_offset(ep->hw_data_buf);
-  assert(hw_data_offset(next_buffer_ptr) <= USB_DPRAM_MAX);
+  hard_assert(hw_data_offset(next_buffer_ptr) <= USB_DPRAM_MAX);
 
   pico_info("  Alloced %d bytes at offset 0x%x (0x%p)\r\n", size, dpram_offset, ep->hw_data_buf);
 
@@ -93,7 +93,6 @@ static void _hw_endpoint_alloc(struct hw_endpoint *ep, uint8_t transfer_type)
   *ep->endpoint_control = reg;
 }
 
-#if 0 // todo unused
 static void _hw_endpoint_close(struct hw_endpoint *ep)
 {
     // Clear hardware registers and then zero the struct
@@ -103,6 +102,22 @@ static void _hw_endpoint_close(struct hw_endpoint *ep)
     *ep->buffer_control = 0;
     // Clear any endpoint state
     memset(ep, 0, sizeof(struct hw_endpoint));
+
+    // Reclaim buffer space if all endpoints are closed
+    bool reclaim_buffers = true;
+    for ( uint8_t i = 1; i < USB_MAX_ENDPOINTS; i++ )
+    {
+        if (hw_endpoint_get_by_num(i, TUSB_DIR_OUT)->hw_data_buf != NULL || hw_endpoint_get_by_num(i, TUSB_DIR_IN)->hw_data_buf != NULL)
+        {
+            reclaim_buffers = false;
+            break;
+        }
+    }
+    if (reclaim_buffers)
+    {
+        pico_info("  reclaim buffer space\n");
+        next_buffer_ptr = &usb_dpram->epx_data[0];
+    }
 }
 
 static void hw_endpoint_close(uint8_t ep_addr)
@@ -110,7 +125,6 @@ static void hw_endpoint_close(uint8_t ep_addr)
     struct hw_endpoint *ep = hw_endpoint_get_by_addr(ep_addr);
     _hw_endpoint_close(ep);
 }
-#endif
 
 static void hw_endpoint_init(uint8_t ep_addr, uint16_t wMaxPacketSize, uint8_t transfer_type)
 {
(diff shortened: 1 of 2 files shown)


src/portable/raspberrypi/rp2040/dcd_rp2040.c
```diff
@@ -224,6 +238,8 @@ static void reset_non_control_endpoints(void)
 
   // clear non-control hw endpoints
   tu_memclr(hw_endpoints[1], sizeof(hw_endpoints) - 2*sizeof(hw_endpoint_t));
+
+  pico_info("  reclaim buffer space\n");
```
hathach, line 242: this should be an comment than an info

vmilea: ok, I pushed a patch.

hathach (commented):
Thank you for your PR, all changes look great, except for the one with hw_endpoint_xfer_continue(). I think we should return `true` instead to notify actual happened event previously, even though endpoint is closed now (it probably doesn't make any difference with ISO endpoint). Though I am open to all suggestion.

PS: On a side note, I actually have an plan to remove dcd_edpt_close() since it causes quite a bit of trouble with other dcd. It is better with just reserving the highest bandwidth then disable/enable endpoint bit for SET_INTERFACE rather than going through the full close/open config cycle. But that could take time to test on all ports though.

src/portable/raspberrypi/rp2040/dcd_rp2040.c
hathach: thank you

src/portable/raspberrypi/rp2040/rp2040_usb.c
```diff
@@ -294,7 +294,9 @@ bool hw_endpoint_xfer_continue(struct hw_endpoint *ep)
   // Part way through a transfer
   if (!ep->active)
   {
-    panic("Can't continue xfer on inactive ep %d %s", tu_edpt_number(ep->ep_addr), ep_dir_string);
+    pico_info("Ignore xfer on inactive ep %d %s", tu_edpt_number(ep->ep_addr), ep_dir_string[tu_edpt_dir(ep->ep_addr)]);
+    _hw_endpoint_lock_update(ep, -1);
+    return false;
```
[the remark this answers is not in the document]
vmilea: Returning true here would call `driver->xfer_cb()` with the closed EP, not sure if driver supports that.

My understanding is endpoints are closed in response to a control transfer on EP0, by which time all transfers on the endpoint being closed should have completed. So handling `USB_INTS_BUFF_STATUS` before `USB_INTS_SETUP_REQ` in the IRQ should prevent the transfer-after-close situation.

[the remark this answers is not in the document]
hathach: If data is **actually transferred via bus**, it should be submitted to higher layer usbd/class driver. It would be their choice to whether process or ignore data. dcd shouldn't assume that.

> My understanding is endpoints are closed in response to a control transfer on EP0, by which time all transfers on the endpoint being closed should have completed. So handling USB_INTS_BUFF_STATUS before USB_INTS_SETUP_REQ in the IRQ should prevent the transfer-after-close situation.

Yes, but since dcd_edpt_close() is called within tud_task() and controller may be already in the middle of transferring by the time we process the SET_INTERFACE. It is a race, there is always chance. To be honest, I don't actually see any issue with that. 

Btw did you experience this often enough, and which callback it triggers ?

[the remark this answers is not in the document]
vmilea: It's a difficult topic. I've never seen a transfer event being delivered after the EP was closed. This change was a precaution based on the comment from `usb.c`. Do you agree to roll back to original `panic()`? If it comes up in practice, then it will be loud and clear and easier to deal with.

I propose also updating IRQ to handle buff_status before setup_req. This should take care of transfer completing immediately before EP close, even if there's still a chance for races like you said.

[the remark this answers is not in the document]
hathach: > It's a difficult topic. I've never seen a transfer event being delivered after the EP was closed. This change was a precaution based on the comment from usb.c. Do you agree to roll back to original panic()? If it comes up in practice, then it will be loud and clear and easier to deal with.

yeah sure, if you haven't run into this scenario. It is best to leave it as it is. I have an tendency to not write code for event that isn't happened yet (to avoid having code that isn't run).

> I propose also updating IRQ to handle buff_status before setup_req. This should take care of transfer completing immediately before EP close, even if there's still a chance for races like you said.

If you could make an PR, I will be happy to review.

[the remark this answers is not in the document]
vmilea: Patch submitted: ae970ba2e2cc83881c52e66cd79fad5947e5e59e

src/portable/raspberrypi/rp2040/dcd_rp2040.c
```diff
@@ -103,14 +102,28 @@ static void _hw_endpoint_close(struct hw_endpoint *ep)
...
+        if (hw_endpoint_get_by_num(i, TUSB_DIR_OUT)->hw_data_buf != NULL || hw_endpoint_get_by_num(i, TUSB_DIR_IN)->hw_data_buf != NULL)
+        {
+            reclaim_buffers = false;
+            break;
+        }
+    }
+    if (reclaim_buffers)
+    {
+        next_buffer_ptr = &usb_dpram->epx_data[0];
+    }
```
hathach, line 119: I have an question, We only reclaim buffer when all endpoints are closed, this does help with audio only class. But the issue still exists if we are using composite with additional endpoints right ?

vmilea: That's true. A better solution would be to track and recycle individual chunks of memory. Or simply not deallocate when endpoints are closed.

hathach: thank you for confirmation, it is still better than current state. It could buy me some time since I don't know when I could start refactoring.

hathach (approved):
perfect !! thank you for your PR. Look good although there is still work with composite device where there are more endpoints than just iso.

Issue #1234, opened 2021-12-04
Allow passing feedback value unchanged through tud_audio_n_fb_set


vmilea:
The Windows driver for USB Audio 2.0 seems to disregard the spec and expects 16.16 format instead of 10.14 on my full-speed device. Interestingly, on Linux the ALSA driver supports either format. So for now I'm calling `tud_audio_n_fb_set(feedback << 2)` to force 16.16 format everywhere. This works but is a bit hacky.

Could we have a config flag to disable the format conversion in `tud_audio_n_fb_set()`?

vmilea:
For future reference: this [blog](https://matthewvaneerde.wordpress.com/2017/10/23/how-to-gather-and-read-logs-for-microsofts-usb-audio-2-0-class-driver) describes how to get logs from the Windows driver. The driver complains after the feedback value being out of range when in 10.14 format.

HiFiPhile:
That's interesting, though I don't remember I had issue testing full-speed.

vmilea:
> That's interesting, though I don't remember I had issue testing full-speed.

It happens on the Raspberry Pi Pico. I'd be interested to learn if it's specific to this device.

HiFiPhile:
@PanRe haven't see you for awhile, do you remember something about this ?

PanRe:
@HiFiPhile yeah it has been a while true! How are you all good?

@vmilea The driver can not know in advance if it is operated in full or high speed mode. Hence, the driver was designed such that the feedback value is always formatted in 16.16 format and in case of full speed mode, the driver takes care of reforamting the value to the specified 10.14 format. This way, the user does not need to care about the format with respect to the speed setting, just always set the value in 16.16 format. Until now, there were no issues with this design. 
Does this answer your question?

vmilea:
@PanRe The API design is fine, it's just that the reformatting to 10.14 sometimes gets in the way. I'm proposing something like `CFG_TUD_AUDIO_ENABLE_FEEDBACK_FORMAT_CORRECTION` so it can be disabled if needed.

This is to work around a bug in the actual Windows driver. When passing 10.14 format on my full-speed device it refuses the value:

```
[USBAudio2]feedback value 0x000b0651 is out of range [0x002c0000,0x002d0000], ignoring value
```

The driver is expecting instead a 16.16 value between 0x2c.0-0x2d.0 (44.0-45.0).

PanRe:
[...]

PR #1235, opened 2021-12-06
Support disabling feedback format correction #1234

vmilea:
As proposed in #1234.


hathach (approved):
thank you very much, sorry for the late response. I make an quick update to get actual link speed. We are good to merge

PR #1238, opened 2021-12-08
Port ft90x

ftdigdm:
**Describe the PR**
Add support for the Bridgetek FT90x and FT93x devices. 

The FT90x and FT93x series are complete System-On-Chip 32-bit RISC microcontrollers for embedded applications featuring a high level of integration and low power consumption.

**Additional context**
Documentation is available at https://brtchip.com/ft90x/ and https://brtchip.com/ft93x/
The port requires to be compiled by the GNU ft32-elf- Toolchain which downloaded from http://brtchip.com/ft9xx-toolchain/

Currently the DCD code is implemented although these devices have a HCD as well. 
The BSP for an MM900EVxB is provided. This covers the MM900EV1B, MM900EV2B and MM900EV3B modules. Note that there are no LEDs or buttons on the modules these can be added using the CN3 connector.


perigoso:
surprised to see someone porting these MCUs, I use them at work for some legacy projects, annoying little things

perigoso:
I assume you're working with FTDI/Bridgetek, so i'll ask you if you don't mind.

Afaik there is no toolchain for linux, I tried to compile gcc for it and had some luck, but the images were larger than expected, and they showed some erroneous behaviours.
Are the compilation options used by the original toolchain available somewhere? I can only assume that's what cause the differences.

ftdigdm:
Correct, we don't currently have linux tools available. The [FT9xx Toolchain](https://brtchip.com/ft9xx-toolchain/) has the Windows gcc compiler (prefix ft32-elf-).

perigoso:
given that gcc is open source, and the ft32 architecture seems to be upstream, could you provide the build options that are used to compile the windows version of the toolchain?

hathach (approved):
sorry for late response. I don't have any ft90x board to verify the PR. We will merge this as it is and make any correct/update later on. If possible please add ft90x to ci build to prevent any future changes would break the build

PR #1252, opened 2021-12-20
samd21: make uart_init() static

mkj:
Avoids a linker conflict if programs are using that same function name.

Otherwise building a samd21 fork of [apollo](https://github.com/greatscottgadgets/apollo) we get

```
/usr/lib/gcc/arm-none-eabi/9.2.1/../../../arm-none-eabi/bin/ld: _build/qtpy/obj/../../firmware/src/boards/qtpy/uart.o (symbol from plugin): in function `uart_configure_pinmux':
(.text+0x0): multiple definition of `uart_init'; _build/qtpy/obj/hw/bsp/samd21/family.o (symbol from plugin):(.text+0x0): first defined here
/home/matt/3rd/fpga/apollo/lib/tinyusb/../../firmware/src/uart.h:20:6: error: type of 'uart_init' does not match original declaration [-Werror=lto-type-mismatch]
   20 | void uart_init(bool configure_pinmux, unsigned long baudrate);
```

### DIFF
--- a/hw/bsp/samd21/family.c
+++ b/hw/bsp/samd21/family.c
@@ -47,7 +47,7 @@ void USB_Handler(void)
 //--------------------------------------------------------------------+
 // UART support
 //--------------------------------------------------------------------+
-void uart_init(void);
+static void uart_init(void);
 
 //--------------------------------------------------------------------+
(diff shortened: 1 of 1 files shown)


hathach (approved):
Look good, thank you

PR #1258, opened 2021-12-21
nucleo-f439zi support

leptun:
**Describe the PR**
This PR implements the board definition for the ST nucleo f439zi development board.
I tested the uac2_headset and cdc_msc examples and both worked as they should. The uart config also works since I got an assert error with cdc_dual_ports (not enough endpoints, 5 needed, 4 available).

Limitations:
- The USB HS peripheral is not implemented. So only OTG_FS can be used. Nucleo boards have only one downstream USB port connected to OTG_FS.
- The host mode power switch pin doesn't seem to be defined/initialized anywhere. This is also the case for the f412zg nucleo board definition, so I suspect this isn't working there as well.


hathach (approved):
look great, thank you for your PR. Will merge when ci passed.

hathach:
> * The USB HS peripheral is not implemented. So only OTG_FS can be used. Nucleo boards have only one downstream USB port connected to OTG_FS.

OTG HS required external PHY, which is not available on nucleo platform. So it is the board limitation not your PR. HS work on other boards like f723 or h743.

> * The host mode power switch pin doesn't seem to be defined/initialized anywhere. This is also the case for the f412zg nucleo board definition, so I suspect this isn't working there as well.

host driver isn't implemented for synopsys dwc2 (used by ST MCU) just yet. We will worry about this later on.

PR #1264, opened 2021-12-27
build system: Changes for xc32 compiler

kasjer:
**Describe the PR**
Three changes are needed to accommodate **xc32** compiler build:
- optimized build flag other than -Os
   added **CFLAGS_OPTIMIZED** that defaults to `-Os` but can be overridden in boards
- build without `-lnosys`
   added **LIBS_GCC** with default libraries that can be changed in boards
- build without **LD_FILE** specification
   if **LD_FILE** is empty `-Wl,-T` options is not added to **LDFLAGS**

**Additional context**
xc32 compiler for Microchip mips chips in free edition have some functional limitations like
lack of `-Os` and `-flto`.
It also provides linker script automatically so it does not have to be specified on command line.
It does not have **nosys** library that is not needed for bulild.

### DIFF
--- a/examples/make.mk
+++ b/examples/make.mk
@@ -54,6 +54,8 @@ endif
 #-------------- Cross Compiler  ------------
 # Can be set by board, default to ARM GCC
 CROSS_COMPILE ?= arm-none-eabi-
+# Allow for -Os to be changed by board makefiles in case -Os is not allowed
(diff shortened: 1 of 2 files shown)


hathach (approved):
look great, thank you very much for pr

PR #1265, opened 2021-12-27
Add HCD driver for Renesas RX

kkitayam:
**Describe the PR**
Add a HCD driver for Renesas RX

**Additional context**

I have checked `cdc_msc_hid` works on [Renesas RX65N Target Board](https://www.renesas.com/us/en/products/microcontrollers-microprocessors/rx-32-bit-performance-efficiency-mcus/rtk5rx65n0c00000br-target-board-rx65n). It is tested with a low speed mouse and a Raspberry Pi Pico running as Mass Storage. If you check the example works on the board, short-circuit between `JP1-3` and VBUS to supply 5V power.

Because [Renesas RX65N Target Board](https://www.renesas.com/us/en/products/microcontrollers-microprocessors/rx-32-bit-performance-efficiency-mcus/rtk5rx65n0c00000br-target-board-rx65n) has no an external power supply IC like as described user's manual shown below. So, we need to supply 5V power to VBUS from a something source. 
image

`hcd_port_reset()` has a busy loop to assert USB reset signaling during 10-20ms. I think it is not a good way. I guess after `hcd_port_reset_end()` is implemented, refactor these implementation.


hathach:
amazing, thank you again for another PR. I will check it out as soon as I could.

hathach (approved):
Thank you very much for this excellent PR and sorry for the huge delay in response. I have been lagging behind due to overwhelming work. The PR look great

> hcd_port_reset() has a busy loop to assert USB reset signaling during 10-20ms. I think it is not a good way. I guess after hcd_port_reset_end() is implemented, refactor these implementation.

indeed the port reset with blocking delay is not ideal. It is better with RTOS thanks to scheduler but totally not great with non-OS. We will find another way to implement this later on. Host stack is still lagging behind device one.

PR #1271, opened 2022-01-05
Add Raspberry Pi Zero W and Zero 2 W

tannewt:
These are different Broadcom chips. The peripherals are essentially
the same. The main differences are:
* The CPU(s)
* The interrupt controller
* The peripheral base address (but not the peripherals that we use)


tools/build_utils.py
```diff
@@ -0,0 +1,61 @@
+import pathlib
```
hathach, line 1: great idea, thank you

hathach (approved):
Superb! the `skip.txt` is great idea. Every thing look perfect, thank you very much

PR #1272, opened 2022-01-06
Enable breakpoints for ARM8M (e.g. cortex-m33)

EmergReanimator:
**Describe the PR**
Breakpoints are useful while debugging the tinyusb. TU_ASSERT uses TU_BREAKPOINT which is now effective on LPC55S28.


hathach (approved):
great, thank you

PR #1274, opened 2022-01-11
nrf5x: Request HFXO via OS

kasjer:
**Describe the PR**
**Mynewt** (similar to Soft Device) has its own reference counting for
HFXO oscillator.
So far TinyUSB requested HFXO when VBUS was detected and stopped when
VBUS was removed.
But with Mynewt running HFXO can be stopped when other interested parties
don't require HFXO anymore. This results in very difficult to track
USB transmission errors when internal RC is used for USB clock source.

This change enables Mynewt specific HFXO management in Soft Device fashion.

**Additional context**
This PR should be merged AFTER matching mynewt PR is merged
https://github.com/apache/mynewt-core/pull/2748

### DIFF
--- a/src/portable/nordic/nrf5x/dcd_nrf5x.c
+++ b/src/portable/nordic/nrf5x/dcd_nrf5x.c
@@ -38,6 +38,10 @@
 #include "device/usbd.h"
 #include "device/usbd_pvt.h" // to use defer function helper
 
+#if CFG_TUSB_OS == OPT_OS_MYNEWT
+#include "mcu/mcu.h"
+#endif
+
 /*------------------------------------------------------------------*/
 /* MACRO TYPEDEF CONSTANT ENUM
  *------------------------------------------------------------------*/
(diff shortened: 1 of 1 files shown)


src/portable/nordic/nrf5x/dcd_nrf5x.c
```diff
@@ -891,6 +895,11 @@ static bool hfclk_running(void)
 
 static void hfclk_enable(void)
 {
+#if CFG_TUSB_OS == OPT_OS_MYNEWT
+  usb_clock_request();
+  return;
+#endif
```
hathach, line 901: we should do #ifelse for the whole hfclk_enable() / hfclk_disable() to prevent gcc to emit `unreachable statement warning`

hathach (changes requested):
look great, though I think we should wrap it all to prevent gcc warnings (if any).

hathach (approved):
perfect, thank you for the PR

PR #1276, opened 2022-01-12
Fix video_capture example fails enumeration when 8FPS

kkitayam:
**Describe the PR**
If `FRAME_RATE` was changed to 8, enumeration process had been failed. `FRAME_RATE` is defined followings.
https://github.com/hathach/tinyusb/blob/3b09b82123a50bef6b18cf90c2734ae7581da4a3/examples/device/video_capture/src/usb_descriptors.h#L38

See also: https://github.com/hathach/tinyusb/pull/1118#issuecomment-1010401051

**Additional context**
I had confirmed this patch on Raspberry Pi Pico.


hathach (approved):
Superb !! thank you for the fix. sorry for late response, I have been busy and still catching up with PRs :)

PR #1277, opened 2022-01-13
Fix family_support.cmake to use new skip.txt and only.txt files

liamfraser:
Commit 7b27b8f498a185d8a7c1bea5dac8f43e7030ebd7 which swaps out the skip / only files for a text file did not include an update to the cmake builds. This PR fixes the issue.


hathach (approved):
thank you very much for the fix.

PR #1280, opened 2022-01-14
nrf5x: Fix DMA access race condition

kasjer:
**Describe the PR**
In multi-thread mode starting DMA in thread mode was
prone to race condition resulting in infinite loop.
It may happen on single core CPU with strict priority based
tasks scheduler where ready high prio task never yields to
ready low prio task (Mynewt).

Sequence that failed (T1 - low priority task, T2 - high priority task)
- T1 called start_dma()
- T1 set _dcd.dma_running (DMA not started yet, context switch happens)
- T2 took CPU and saw that _dcd.dma_running is set, so waits for _dcd.dma_running to be 0
- T1 never gets CPU again, DMA is not started T2 waits forever

OSAL mutex resolves problem of DMA starting from thread-context.

**Additional context**
Happened while stress testing **Nimble** stack on **Mynewt** with USB transport.

### DIFF
--- a/src/portable/nordic/nrf5x/dcd_nrf5x.c
+++ b/src/portable/nordic/nrf5x/dcd_nrf5x.c
@@ -79,6 +79,9 @@ typedef struct
 
 } xfer_td_t;
 
+static osal_mutex_def_t dcd_mutex_def;
+static osal_mutex_t dcd_mutex;
+
 // Data for managing dcd
 static struct
 {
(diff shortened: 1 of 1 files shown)


kasjer:
Additional mutex lock added during transfer setup to prevent premature interrupt enable that could happen if two tasks started two separate transfers.

hathach (approved):
Thank your for the PR and sorry for late response, this somehow falled off my radar. The current implementation of nrf5x indeed has issue with race condition in preempted RTOS. I was hoping to use LDREX/STREX to have mutex, but couldn't get those to work. In the future, I think we could make use of semaphore as resource management instead of mutex which make a little bit more sense.

PR #1283, opened 2022-01-16
olimex pic32 board fixes

kasjer:
**Describe the PR**
Buttons for olimex_hmz144 and olimex_emz64 were incorrectly set to be active high.

Remote wakeup requires that RESUME bit be active for 10ms.
It was covered for OS build for non-OS build it is now handled by
utilizing board_millis() function.

**Additional context**
Changes needed for USB3V suite tests for remote wakeup functionality.
It was originally tested with Mynewt where remote wakeup was triggered
other way.

### DIFF
--- a/src/portable/microchip/pic32mz/dcd_pic32mz.c
+++ b/src/portable/microchip/pic32mz/dcd_pic32mz.c
@@ -163,6 +163,10 @@ void dcd_remote_wakeup(uint8_t rhport)
   USB_REGS->POWERbits.RESUME = 1;
 #if CFG_TUSB_OS != OPT_OS_NONE
   osal_task_delay(10);
+#else
+  // TODO: Wait in non blocking mode
+  unsigned cnt = 2000;
+  while (cnt--) __asm__("nop");
 #endif
   USB_REGS->POWERbits.RESUME = 0;
 }
--- a/hw/bsp/pic32mz/boards/olimex_emz64/olimex_emz64.c
+++ b/hw/bsp/pic32mz/boards/olimex_emz64/olimex_emz64.c
@@ -63,7 +63,7 @@ void led_init(void)
   // RB8 - LED
   // ANASELB RB8 not analog
   ANSELBCLR = TU_BIT(8);
-  // TRISH RH2 input
+  // TRISH RH2 output
   TRISBCLR = TU_BIT(8);
   // Initial value 0, LED off
   LATBCLR = TU_BIT(8);
@@ -128,7 +128,7 @@ void board_led_write(bool state)
 
 uint32_t board_button_read(void)
 {
-  return (PORTB >> 12) & 1;
+  return ((PORTB >> 12) & 1) == 0;
 }
 
 int board_uart_write(void const * buf, int len)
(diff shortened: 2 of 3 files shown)


kasjer:
Test suite evidence run on cdc_msc and hid_composit examples.
[USB3CV.zip](https://github.com/hathach/tinyusb/files/7877242/USB3CV.zip)

hathach (changes requested):
bsp and `board_millis()` are for example only. Please put a TODO or an blocking loop there instead.

src/portable/microchip/pic32mz/dcd_pic32mz.c
```diff
@@ -163,6 +163,10 @@ void dcd_remote_wakeup(uint8_t rhport)
   USB_REGS->POWERbits.RESUME = 1;
 #if CFG_TUSB_OS != OPT_OS_NONE
   osal_task_delay(10);
+#else
+  // TODO: Wait in non blocking mode
+  unsigned cnt = 2000;
```
hathach, line 168: I am not entirely sure if 2000+ nops is enough for at least 1ms delay. It could have issue with fast mcu. Does pic32 has something like SystemCoreClock. 

```C
  // try to delay for 1 ms
  uint32_t count = SystemCoreClock / 1000;
  while ( count-- )
  {
    __NOP();
  }
```


It is not really a problem for now, we could come back to this later

src/portable/microchip/pic32mz/dcd_pic32mz.c
```diff
@@ -163,6 +163,10 @@ void dcd_remote_wakeup(uint8_t rhport)
   USB_REGS->POWERbits.RESUME = 1;
 #if CFG_TUSB_OS != OPT_OS_NONE
   osal_task_delay(10);
```
hathach, line 165: The usb specs state the delay must be within 1-15 ms. Other mcu is doing well with 1 ms, maybe we should reduce it to 1-2 ms only, 10ms is a bit overkill, would it pass the cv test with 1ms of delay ? 

image

hathach (commented):
look great, I just have 2 minor questions, nothing important, but would be nice if they could be clarified.

src/portable/microchip/pic32mz/dcd_pic32mz.c
kasjer: There is no standard SystemCoreClock as provided by CMSIS
There is CPU0 register COUNTER that is kind like Cortex SysTick, I can use it if you insist.
In fact 2000 NOPs at 200MHz is much faster then time you mentioned from datasheet but in my case it was enough to trigger remote wake up every time in contrast to where there is no delay between toggling bit and remote wake up is not activated.

src/portable/microchip/pic32mz/dcd_pic32mz.c
hathach: just do a quick check with musb spec, it does indeed say 10ms as optimum value (2-15 ms). So I think 10m here make sense 
image

src/portable/microchip/pic32mz/dcd_pic32mz.c
hathach: if it works, then we will leave it as it is now. And come back with a more accurate delay later on. No problem at all. 

Note since the timing is out of specs, this maybe specific to the host driver (in this case the usbcv), other host driver (linux/windows/macos) may decide this signal is too short to register as remote wakeup.

hathach (approved):
perfect !! Thank you, we will come back later on with blocking delay should there is any timing issue with different host drivers.

PR #1284, opened 2022-01-17
ft9xx: Fix Mynewt build

kasjer:
**Describe the PR**
Includes were moved few lines down to restore build with Mynewt build system.


hathach (approved):
oops, sorry for should have been more careful with this

PR #1285, opened 2022-01-17
Rudimentary bulk endpoint support for RP2040 host mode.

eskimo-software:
Allows bulk endpoints to use entries in ep_pool, and forces only control endpoints to share ep_pool[0].

*THIS IS A HACK.*

It's only tested to the extent that the MSC example gets as far as invoking inquiry_complete_cb().
This exercises an input & an output bulk endpoints.
Coexistence of interrupt & bulk EPs is entirely untested and is sure to fail.

### DIFF
--- a/src/portable/raspberrypi/rp2040/hcd_rp2040.c
+++ b/src/portable/raspberrypi/rp2040/hcd_rp2040.c
@@ -72,10 +72,19 @@ static struct hw_endpoint *get_dev_ep(uint8_t dev_addr, uint8_t ep_addr)
   for ( uint32_t i = 1; i < TU_ARRAY_SIZE(ep_pool); i++ )
   {
     struct hw_endpoint *ep = &ep_pool[i];
-    if ( ep->configured && (ep->dev_addr == dev_addr) && (ep->ep_addr == ep_addr) ) return ep;
+    if ( ep->configured && (ep->dev_addr == dev_addr) && (tu_edpt_number(ep->ep_addr) == num) ) return ep;
   }
 
   return NULL;
+}
+
+static struct hw_endpoint *get_epx_ep(void)
+{
+    uint8_t dev_addr = (usb_hw->dev_addr_ctrl & USB_ADDR_ENDP_ADDRESS_BITS) >> USB_ADDR_ENDP_ADDRESS_LSB;
+    uint8_t ep_addr = (usb_hw->dev_addr_ctrl & USB_ADDR_ENDP_ENDPOINT_BITS) >> USB_ADDR_ENDP_ENDPOINT_LSB;
+    pico_trace("  epx dev %d ep %d\n", dev_addr, ep_addr);
+
+    return get_dev_ep(dev_addr, ep_addr);
 }
 
 static inline uint8_t dev_speed(void)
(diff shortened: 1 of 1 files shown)


cjhilder:
I have picked up this pull request in my own fork, and successfully used it to read and write to a USB flash drive using a RP2040 in host mode. There is a full working example here: [examples/host/msc_with_fatfs](https://github.com/cjhilder/tinyusb/tree/master/examples/host/msc_with_fatfs). However, it has an outstanding problem. If this could be resolved then we would **have the ability to use standard USB flash drives from RP2040 boards**.

The problem is that the USB hardware randomly fails to respond to an apparently valid usb_hw->sie_ctrl setting with an interrupt. I have found that I can improve the chances of success by inserting a delay immediately prior to setting usb_hw->sie_ctrl in line 569 of [hcd_rp2040.c](https://github.com/cjhilder/tinyusb/blob/master/src/portable/raspberrypi/rp2040/hcd_rp2040.c). I don't understand what is going on here. I presume there is some state that needs to be waited for, but I don't know what that state could be. Any help would be much appreciated, and we could maybe get this code into a state that it is ready for a final pull request (or at least useable).

rppicomidi:
@cjhilder The [hcd_rp2040.c](https://github.com/cjhilder/tinyusb/blob/master/src/portable/raspberrypi/rp2040/hcd_rp2040.c) code still shares EPx except for interrupt transfers. FWIW, I tried this [hcd_rp2040.c](https://github.com/cjhilder/tinyusb/blob/master/src/portable/raspberrypi/rp2040/hcd_rp2040.c) in my [fork](https://github.com/rppicomidi/tinyusb) with the MIDI host example and it works as long as I implement the fix I described above. The issues I described in discussion #1261 still apply.

cjhilder:
> @cjhilder I encountered a similar issue with transactions not completing when I implemented bulk endpoints for MIDI Host using a shared EPx endpoint for both bulk endpoints and the control endpoint. EPx sharing works but is not great (see discussion #1261). My workaround for the endpoint not completing bulk transfers looks like the code below. Did you try this fix already? I would dearly love to get rid of EPx sharing in my code.
> 
> ```
>         uint32_t flags = SIE_CTRL_BASE |
>                  (ep_dir ? USB_SIE_CTRL_RECEIVE_DATA_BITS : USB_SIE_CTRL_SEND_DATA_BITS);
>         // Set pre if we are a low speed device on full speed hub
>         flags |= need_pre(dev_addr) ? USB_SIE_CTRL_PREAMBLE_EN_BITS : 0;
>         // Set up the hardware with all flags except the start bit
>         usb_hw->sie_ctrl = flags;
>         // Now start the transaction; if you don't do this in two parts,
>         // and the host has switched direction, sometimes the transaction
>         // does not start
>         flags |= USB_SIE_CTRL_START_TRANS_BITS;
>         usb_hw->sie_ctrl = flags;
> ```

This might be the problem I was encountering. I'm pretty interested in trying it out, but have now re-designed the project I was working on to use serial communication in device mode, so it's difficult for me to find the time to test it. Anyone else who is trying to access flash drives from RP2040 might like to pick this up and test it out, using my FATFs files: [https://github.com/cjhilder/tinyusb/tree/master/examples/host/msc_with_fatfs/FATFs](url). I will get to it when I have a chance!

FCare:
I am using a CD ROm drive as device with a rp2040 as host. I can confirm that this patch prevent transfer to not start sometime.
Some added a delay but with this patch, I can read blocks on my CDROM without any interruption.

Thanks for finding it :-)

hathach:
#3561 does a better implementation. Thank you for everyone

PR #1287, opened 2022-01-19
fix idfgh-6508: return type in tu_fifo_peek_n()

alisitsyn:
**Describe the PR**
This simple PR is submitted to fix the IDFGH-6508 issue as described here: https://github.com/espressif/esp-idf/issues/8161
The return type in the line /src/common/tusb_fifo.c:719 should be changed to `uint16_t` instead of bool.
The [esp-idf fork](https://github.com/espressif/tinyusb/blob/master/src/common/tusb_fifo.c#L488) then can be rebased on upstream master.


hathach (approved):
spot-on, thank you for the bug fix. I was probably writing this while sleepy

alisitsyn:
Thank you for merging!

PR #1289, opened 2022-01-19
vendor: Write improvements

kasjer:
**Describe the PR**
Other drivers already have notification about data sent.
It allows batter control in application on vendor protocol level.

 So far `tud_vendor_n_write()` always flushed data.
It requires to have whole vendor packed constructed before in one buffer.

With this change data do get flushed when endpoint size
is filled on write, when there is no enough data to
fill endpoint data is not sent and subsequent calls to
write functions can add more bytes.

Vendor code needs to call `tud_vendor_n_flush()` when packet is
ready to be sent.

**Additional context**


hathach (approved):
Thank you for your PR. Look great

PR #1305, opened 2022-01-28
OHCI: Fix array out of bounds issue

Ryzee119:
**Describe the PR**
If using a USB hub with OHCI host backend will result in a array index out of bounds error.

**Additional context**
Original line:
https://github.com/hathach/tinyusb/blob/ffb257ac17f162bc5a4c26596d7a1e954db98aa5/src/portable/ohci/ohci.h#L162

It can happen in a few places but one such example is here:
https://github.com/hathach/tinyusb/blob/ffb257ac17f162bc5a4c26596d7a1e954db98aa5/src/portable/ohci/ohci.c#L460

ie. As HUB addresses are >=5, this results in a array index out of bounds on control endpoints as `CFG_TUH_DEVICE_MAX+1` is only 5 with HUBs enabled..

This fix just includes num of hubs in the reserved array size.
Changing this locally fixed my issue issue.

### DIFF
--- a/src/portable/ohci/ohci.h
+++ b/src/portable/ohci/ohci.h
@@ -159,7 +159,7 @@ typedef struct TU_ATTR_ALIGNED(256)
   struct {
     ohci_ed_t ed;
     ohci_gtd_t gtd;
-  }control[CFG_TUH_DEVICE_MAX+1];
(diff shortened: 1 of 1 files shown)


hathach (approved):
Thanks, this code used to work but ohci is getting behind host updates/refactoring

PR #1311, opened 2022-02-02
Solve problems with CCRX toolchain

Wini-Buh:
This PR solves two issues that happens with the CCRX toolchain.

The first issue was found in the `TU_ARGS_NUM` macro. Without this modification it is not possible to build the tinyUSB successfully. 

The second issue belong to local pointer variable definition of the type `reg_pipetre_t`, which is mapped over the `PIPExTRE` register of the RX USB hardware. To ensure access in the correct size the keyword `__evenaccess` needs to be used. 
Note: Without this modifier, it seems that receiving of data from a host is working, but only as long as the number of bytes is less then 65 bytes.

The modification was successful tested with the CCRX toolchain in little and big endian mode and also with the GCC toolchain in little endian mode.

### DIFF
--- a/src/common/tusb_compiler.h
+++ b/src/common/tusb_compiler.h
@@ -75,7 +75,11 @@
  * Nth position is the same as the number of arguments
  * - ##__VA_ARGS__ is used to deal with 0 paramerter (swallows comma)
  *------------------------------------------------------------------*/
-#define TU_ARGS_NUM(...) 	 _TU_NARG(_0, ##__VA_ARGS__,_RSEQ_N())
+#if !defined(__CCRX__)
+#define TU_ARGS_NUM(...)   _TU_NARG(_0, ##__VA_ARGS__,_RSEQ_N())
+#else
+#define TU_ARGS_NUM(...)   _TU_NARG(_0, __VA_ARGS__,_RSEQ_N())
+#endif
 
 #define _TU_NARG(...)      _GET_NTH_ARG(__VA_ARGS__)
 #define _GET_NTH_ARG( \
--- a/src/portable/renesas/usba/dcd_usba.c
+++ b/src/portable/renesas/usba/dcd_usba.c
@@ -494,7 +494,11 @@ static bool process_pipe_xfer(int buffer_type, uint8_t ep_addr, void* buffer, ui
       while (USB0.D0FIFOSEL.BIT.CURPIPE) ; /* if CURPIPE bits changes, check written value */
     }
   } else {
+#if defined(__CCRX__)
+    __evenaccess volatile reg_pipetre_t *pt = get_pipetre(num);
+#else
     volatile reg_pipetre_t *pt = get_pipetre(num);
+#endif
     if (pt) {
       const unsigned     mps = edpt_max_packet_size(num);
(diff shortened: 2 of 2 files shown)


src/common/tusb_compiler.h
```diff
@@ -75,7 +75,11 @@
  * Nth position is the same as the number of arguments
  * - ##__VA_ARGS__ is used to deal with 0 paramerter (swallows comma)
  *------------------------------------------------------------------*/
-#define TU_ARGS_NUM(...) 	 _TU_NARG(_0, ##__VA_ARGS__,_RSEQ_N())
+#if !defined(__CCRX__)
+#define TU_ARGS_NUM(...)   _TU_NARG(_0, ##__VA_ARGS__,_RSEQ_N())
+#else
+#define TU_ARGS_NUM(...)   _TU_NARG(_0, __VA_ARGS__,_RSEQ_N())
```
hathach, line 81: just for clarification, could ccrx automatically swallow the extra comma when `__VA_ARGS__` expanded as no arguments ?

hathach (commented):
thank you for your PR, and sorry for late response. I have an question regarding swallowing no arguments for ccrx. Could you confirm its behavior.

src/common/tusb_compiler.h
Wini-Buh: No, it generates a compiler error if no argument is passed. Is it intended to call this macro without an argument ?

hathach: Thanks for clarification, it is not quite intended to use without arguments, but the macro is for counting number of arguments. So it is more like it will work from 0 to n. Luckily, currently there is no usage for counting zero arguments, but it may (or not) be needed in the future. 

If the CCRX does not support this extension, we could just merge this as it is, no problem at all.

hathach: @Wini-Buh would you confirm that `##__VA_ARGS__` as `swallowing semicolon with zero-argument` is not supported on CCRX

Wini-Buh: Calling this macro with an argument `TU_CHECK_MCU(OPT_MCU_RX72N)` :

with `_TU_NARG(_0, ##__VA_ARGS__,_RSEQ_N())` sub-macro
 -> compiles with an error with the CCRX toolchain

with `_TU_NARG(_0, __VA_ARGS__,_RSEQ_N())` sub-macro
 -> compiles **successful** with the CCRX toolchain
 
 
Calling this macro without an argument TU_CHECK_MCU() 

with `_TU_NARG(_0, ##__VA_ARGS__,_RSEQ_N())` sub-macro
 -> compiles with an error with the CCRX toolchain
 
with `_TU_NARG(_0, __VA_ARGS__,_RSEQ_N())` sub-macro
 -> compiles with an error with the CCRX toolchain

@hathach  I hope this helps you further.

hathach: I see, apparently `##__VA_ARGS__` is not supported by ccrx

hathach (approved):
seems like `##__VA_ARGS__` is not supported on CCRX, luckily so far we haven't tried to count zero arguments

PR #1340, opened 2022-02-23
Add concurrent host and device example

tannewt:
This reads HID devices over host and then translates that to ASCII
and sends it over CDC device.


hathach (approved):
Superb! This is a great example to show that the stack is able to run both host and device at the same time. I have tested and did a bit of example update. Will merge when ci passed.

PR #1343, opened 2022-02-23
Open OUT endpoint for HID host

Daft-Freak:
**Describe the PR**
Updates the host HID code to open the OUT endpoint (if it exists). Also adds handling for interrupt OUT to the RP2040 host code. (at least, that seemed to be enough)

**Additional context**
I've been trying to connect a Wii U GameCube adapter, which requires some data to be sent to it before it starts sending reports. This PR doesn't add any code for actually doing that, but having the endpoint open helps.

(I've only tested this as far as sending the one byte I need for that)


hathach (approved):
thank you, I haven't tested it out, but look good and changes is spot-on with `//TODO` ,

Issue #1344, opened 2022-02-23
Compilation fails with IAR C++ compiler


alexandre-perrin:
### Operating System

Windows 11

### Board

Any

### Firmware

Compile a simple cpp file including `tusb.h`

```C
// main.cpp
#include <tusb.h>

int main(void)
{
  return 0;
}
```

### What happened ?

Cpp compiler complains about `TU_VERIFY_STATIC` with the following errors:
- `this declaration has no storage class or type specifier`
- `too many initializer values`
- `variable "_Static_assert" has already been defined`

### How to reproduce ?

Create an empty IAR project with including tinyUSB/src in search paths.
Add provided `main.cpp`
Try to compile.

### Debug Log

```
 Error[Pe077]: this declaration has no storage class or type specifier tinyusb\src\common\tusb_types.h 303 
 Error[Pe146]: too many initializer values tinyusb\src\common\tusb_types.h 303 
 Error[Pe077]: this declaration has no storage class or type specifier tinyusb\src\common\tusb_types.h 314 
 Error[Pe247]: variable "_Static_assert" has already been defined (at line 303) tinyusb\src\common\tusb_types.h 314 
 Error[Pe146]: too many initializer values tinyusb\src\common\tusb_types.h 314 
 Error[Pe077]: this declaration has no storage class or type specifier tinyusb\src\common\tusb_types.h 330 
 Error[Pe247]: variable "_Static_assert" has already been defined (at line 303) tinyusb\src\common\tusb_types.h 330 
 Error[Pe146]: too many initializer values tinyusb\src\common\tusb_types.h 330 
 Error[Pe077]: this declaration has no storage class or type specifier tinyusb\src\common\tusb_types.h 347 
 Error[Pe247]: variable "_Static_assert" has already been defined (at line 303) tinyusb\src\common\tusb_types.h 347 
 Error[Pe146]: too many initializer values tinyusb\src\common\tusb_types.h 347 
 Error[Pe077]: this declaration has no storage class or type specifier tinyusb\src\common\tusb_types.h 368 
 Error[Pe247]: variable "_Static_assert" has already been defined (at line 303) tinyusb\src\common\tusb_types.h 368 
 Error[Pe146]: too many initializer values tinyusb\src\common\tusb_types.h 368 
 Error[Pe077]: this declaration has no storage class or type specifier tinyusb\src\common\tusb_types.h 400 
[...]

PR #1345, opened 2022-02-23
Fix CPP compilation with IAR

alexandre-perrin:
IAR defines __STDC_VERSION__ in cpp as well.
Which causes TU_VERIFY_STATIC to be defined as _Static_assert
instead of cpp's static_assert.

This reorders __cplusplus__ to be first, then to fallback to
__STDC_VERSION__ if not CPP.

### DIFF
--- a/src/common/tusb_compiler.h
+++ b/src/common/tusb_compiler.h
@@ -51,10 +51,10 @@
 #endif
 
 // Compile-time Assert
-#if defined (__STDC_VERSION__) && __STDC_VERSION__ >= 201112L
-  #define TU_VERIFY_STATIC   _Static_assert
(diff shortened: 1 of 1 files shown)


yoyopod:
I experienced this same problem with an IAR build, and fixed it in the exact manner proposed.

I have subsequently learned that C11 also defines a macro, 'static_assert', in assert.h, that maps to C11's '_Static_assert' keyword.  Relying on the macro would allow the C and C++ cases both to use static_assert(), with no need to differentiate.

HiFiPhile (approved):

alexandre-perrin:
Fixes #1344

PR #1360, opened 2022-03-02
in TUD_AUDIO_HEADSET_STEREO_DESCRIPTOR, _nitfs should be 3.

ccrome:
I believe that there is a mistake in the definition of TUD_AUDIO_HEADSET_STEREO_DESCRIPTOR.  The nitfs in the audio descriptor is always 3 regardless of ITF_NUM_TOTAL.

**Describe the PR**
I change the number of interfaces to 3, so that I could add a CDC descriptor after, and it worked without complaints from Linux.


kasjer (approved):
@ccrome You are right, it should be 3 as this examples clearly has 3 interfaces (control/speaker/microphone). My bad.

hathach:
great work, thank you for your fix

ccrome:
You're welcome.... took me a bit to track down when I added a CDC device to my audio device.  Thanks for the awesome library and examples!

PR #1369, opened 2022-03-04
Add host string descriptor functions

tannewt:
Plus typo fixes, GCC11 array bounds fix, snprintf for malloc-free
debug and pragmas for alignment checks.


hathach (approved):
thank you for PR, I make some changes to the API 
- skip get index, and straight to get manufacturer/product/serial string (you may need to update the cpythong accordingly)
- add more API for all descriptor: device, configuration, string
- add configuration_set() not required to be called by application, added just it is there
- There is other unrelated name change for max endpoint macros. Just do it since I am already there.

I also add an stub example `bare_api` to demonstrate the api usage without class driver (direct endpoint read/write). Since this PR is getting big, we will do it in separate one. The goal would be somewhat similar to libusb API which is used by pyUSB and pretty much all *nix os. However, ours will be quite minimal.

PR #1372, opened 2022-03-04
Fix string descriptor fetch

tannewt:
Setting the indices was commented out. Return false if the index is 0 because that indicates there is no string descriptor for it.


hathach (approved):
Ah right, I forgot to revert the debug changes. Will do more test with new bare api example later

PR #1374, opened 2022-03-05
stm32wb55 support

kasjer:
**Describe the PR**
This PR adds support for **STM32WB55** MCU.
**nucleo-wb55rg** board is also added.

This MCU has **stm32_fsdev** FS device USB block same as some others ST devices (F1, Gx)

**Additional context**
ST provided headers have incorrect value of **USB_PMAADDR** that is used in stm32_fsdev code.
Issue is filed on their github.
So for now this is fixed on TinyUSB side.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -97,6 +97,12 @@
 [submodule "hw/mcu/st/stm32l5xx_hal_driver"]
 	path = hw/mcu/st/stm32l5xx_hal_driver
 	url = https://github.com/STMicroelectronics/stm32l5xx_hal_driver.git
+[submodule "hw/mcu/st/cmsis_device_wb"]
+	path = hw/mcu/st/cmsis_device_wb
+	url = https://github.com/STMicroelectronics/cmsis_device_wb.git
+[submodule "hw/mcu/st/stm32wbxx_hal_driver"]
+	path = hw/mcu/st/stm32wbxx_hal_driver
+	url = https://github.com/STMicroelectronics/stm32wbxx_hal_driver.git
 [submodule "lib/sct_neopixel"]
 	path = lib/sct_neopixel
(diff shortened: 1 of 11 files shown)


hathach (commented):
It is very minor, I think the family name should be `stm32wb` 55 is too specific. Other than that, everything looks great.

kasjer:
> It is very minor, I think the family name should be `stm32wb` 55 is too specific. Other than that, everything looks great.

I agree, let me change it. I used 55 because this is the only part that has USB.

hathach:
> > It is very minor, I think the family name should be `stm32wb` 55 is too specific. Other than that, everything looks great.
> 
> I agree, let me change it. I used 55 because this is the only part that has USB.

oh I didn't know that, but I guess that could be changed in the future. It is still better to rename it

kasjer:
I have not problem with changing family it to WB only.
btw. while you are on it, maybe you can suggest how to set **UF2_FAMILY_ID =** in make file.
It's new to me, and I did not dwell on the subject really.

hathach:
It is the ID for DFU example with board pre-flashed with uf2 bootloader such as https://github.com/adafruit/tinyuf2 . The ID is managed by uf2 specs, ID for stm32wb is as follows. It is optional, for convenient purpose only. So far most of the board is devkit with dedicated jtag/swd debugger.

https://github.com/microsoft/uf2/blob/f0506988eaf24cc12646ed1a39799fa29b204eed/utils/uf2families.json#L128-L130

kasjer:
Family changed to STM32WB.
I also change board name to stm32wb55nucleo to match other nucloe boards naming convention.

hathach (approved):
perfect, thank you very much for your PR.

PR #1379, opened 2022-03-07
Modify the wrong macro definition code

Qianhao2579:
Modify the wrong macro definition code


hathach (approved):
thank you for the fix, will merge when ci passed.

PR #1381, opened 2022-03-07
Add SOF IRQ Handler 

### LINKED ISSUE #1231: Need accurate SOF event on RP2040 for explicit feedback
First of all, thanks for TinyUSB! It's an awesome little library.

I'm implementing a sound card using the RP2040. The device provides an asynchronous output endpoint and an explicit feedback endpoint to manage the data rate.

The feedback method is described in `5.12.4.2` of the USB 2.0 spec: 

> To generate the desired data rate Ff, the device must measure its actual sampling rate Fs, referenced to the USB notion of time, i.e., the USB (micro)frame frequency. [...] A sink can determine Ff by counting cycles of the master clock Fm for a period of 2^(K-P) (micro)frames. The counter is read into Ff and reset every 2^(K-P) (micro)frames. As long as no clock cycles are skipped, the count will be accurate over the long term.

So at each SOF we measure elapsed CPU clock cycles. The CPU clock frequency is fixed to a multiple of the sample rate, and by dividing we get the exact number of samples played each frame. This approach is very reliable and low-latency.

But currently there is no way to be notified on SOF.  This event would need to be delivered straight from IRQ handler to minimize jitter. When I tried dispatching on the `dcd_event_bus`, the extra 1-2us of jitter was causing audio glitches.

hathach:
**Describe the PR**
mostly implement and close #1231 , add new
- dcd_sof_enable(uint8_t rhport, bool en)
- void tud_sof_isr_set(tud_sof_isr_t sof_isr)
- implement dcd_sof_enable() for rp2040
- add dcd_sof_enable() stubs for other ports for now

Other unrelated clean up
- remove broadcom/synopsys (replaced by synopsys/dwc2)
- add CFG_TUD_INTERFACE_MAX for max number of interfaces, default to <s>8</s> 16.


geekbozu:
2 Questions to go with this as my memory is failing me. 
1: Queuing it with in_isr set to true means that the event fires in the ISR context? SoF entry timing tends to be quite critical so would like to confirm. 
2: This "breaks" the current Driver convention of specifying the callbacks/event handlers in the Driver Structs. Should the initializers be updated to support that instead of just this "new api" specific to SoF as it is now?


Great work as always happy to see this getting implemented!

hathach:
> 2 Questions to go with this as my memory is failing me. 1: Queuing it with in_isr set to true means that the event fires in the ISR context? SoF entry timing tends to be quite critical so would like to confirm. 2: This "breaks" the current Driver convention of specifying the callbacks/event handlers in the Driver Structs. Should the initializers be updated to support that instead of just this "new api" specific to SoF as it is now?
> 
> Great work as always happy to see this getting implemented!

1. yes, `in_isr=true` means ISR context, but is is set from dcd. For SOF, all ports should call this in ISR context.
2. This is great question, this API is for user space, `tud_sof_isr_set(sof_isr)` is called to enable or SOF interrupt when needed. Maybe we should have an class specific e.g audio sof set function, sof driver hook is not used now

https://github.com/hathach/tinyusb/blob/73896a3b71c525a3ee4cefa7e35ce3b3a93786ef/src/device/usbd.c#L134-L144
 
this will allow us to auto enable sof when audio/video alt interface is enabled and disabled it when not active. Though, to be honest, I haven't tested/done much with audio class driver.  
maybe @PanRe would have a comment on this.

geekbozu:
I would love to hear @PanRe 's opinion as well. This is also something that can be dealt with later. More of a "trying to keep everything unified" and not have a ton of specific extensions to things. 

SoF is not needed for every application, but those who need it tend to always need it. Being how expensive it is however I definitely think being able to disable it should be kept as part of the API.
 I think for the user land driver specifying that sof callback should populate the ISR callback but not enable it by default. I think that would be a fair and consistent compromise.

PanRe:
I think it is a great feature! The correct place to impment this in the audio driver is the `audiod_set_interface()` function. There, the interfaces are open and closed. I lookup the implementation details this week (i hope). Unfortunately, i miss the hardware to test the speaker functionality, @geekbozu  i guess you could test it?

geekbozu:
I will need to find time but yes I only have STM32 board for this however not RP2040 as of the moment.

hathach:
> I will need to find time but yes I only have STM32 board for this however not RP2040 as of the moment.

which stm32 board/mcu you are using, I could add sof implementation for it.

geekbozu:
STM32F411RCT
Its the synopsis driver, My implementation here 
https://github.com/geekbozu/tinyusb/commit/ce8499acb6a6268a42e4a4bc8b44a12e2e699ec7

Thanks!!

PanRe:
I am currently composing something, however, it would be nice to already have a working example of what should be done in the SOF isr. @geekbozu could you post a snippet here? @hathach i already implemented SOF for synopsis. I am aware that my implementation will not meet all your considerations but once i pushed it your are free to go to improve it! :)

PanRe:
I would like to update the feedback value within the audio class driver. For that, however, i need the `cpu_cycle_count` value at the time of the SOF interrupt. To be more precise, the cycly count of any master clock locked to the sample frequency. Most often, f_cpu is locked to f_s but it might be a different clock source as well. Is there a chance to introduce some callback function `tud_get_master_clock_cycle_count()` which need to be set from the user and i can call within the SOF interrupt routine?

geekbozu:
> I would like to update the feedback value within the audio class driver. For that, however, i need the `cpu_cycle_count` value at the time of the SOF interrupt. To be more precise, the cycly count of any master clock locked to the sample frequency. Most often, f_cpu is locked to f_s but it might be a different clock source as well. Is there a chance to introduce some callback function `tud_get_master_clock_cycle_count()` which need to be set from the user and i can call within the SOF interrupt routine?

To solve this on the STM32 I tied the master clock output to a Timer Input Clock pin and make use of that to count .

Below is my SOF interrupt. 

```c
volatile float bitrate;
void dcd_sof_cb(uint32_t frame) {
        // Get I2S MasterClock time
        uint32_t count = htim2.Instance->CNT;
        if ((frame % 4) == 0) { //Every 4 frames or 4ms
            //Set count to account for time it took to calculate
            //This is off in debug, But should be faster in -03 May need a padding 
            // Padding of +2 is a guess. 
            htim2.Instance->CNT = htim2.Instance->CNT - count + 2;
            //calculate sample rate. as MCLK/SampleCount/MCLKBinaryMultiplier
            bitrate = (float) count / 4 / 256;
            //Clamp to what ever windows decides is appropriate 
            if (bitrate > 45.0007){
                bitrate = 45.0007;
            }
            if ( bitrate < 44.0007) {
                bitrate = 44.0007;
            }
        }

}
```

This works really well at getting a close enough estimate of the bitrate, HOWEVER its not perfect. So you do still need to keep an eye on your buffer levels and tweak it accordingly sadly.

I don't have the code on hand but in my user context is where I update the actual feedback value, EG call tud_set_feedback(). I update it once every ms which is probably overkill but it seems to work well enough!

PanRe:
@geekbozu  Thanks for the snippet! I need a few days extra to finalize the code. Thanks for the pacients ;)!

hathach:
> I would like to update the feedback value within the audio class driver. For that, however, i need the `cpu_cycle_count` value at the time of the SOF interrupt. To be more precise, the cycly count of any master clock locked to the sample frequency. Most often, f_cpu is locked to f_s but it might be a different clock source as well. Is there a chance to introduce some callback function `tud_get_master_clock_cycle_count()` which need to be set from the user and i can call within the SOF interrupt routine?

`tud_get_master_clock_cycle_count()` API shouldn't be part of "usb stack", it should be got by application and if necessary passed as argument to audio class driver. I am still in middle of other work, will combine work to update this whenever I could.

geekbozu:
> > I would like to update the feedback value within the audio class driver. For that, however, i need the `cpu_cycle_count` value at the time of the SOF interrupt. To be more precise, the cycly count of any master clock locked to the sample frequency. Most often, f_cpu is locked to f_s but it might be a different clock source as well. Is there a chance to introduce some callback function `tud_get_master_clock_cycle_count()` which need to be set from the user and i can call within the SOF interrupt routine?
> 
> `tud_get_master_clock_cycle_count()` API shouldn't be part of "usb stack", it should be got by application and if necessary passed as argument to audio class driver. I am still in middle of other work, will combine work to update this whenever I could.

I agree here, its to setup dependent to be part of the "USB Stack"

```c
void dcd_sof_cb(uint32_t frame) {
		uint32_t count = htim2.Instance->CNT;
		if ((frame % 4) == 0) { //Every 4 frames or 4ms
			//Set count to account for time it took to calculate
			//This is off in debug, But should be faster in -03 May need a padding of 1 Count
			htim2.Instance->CNT = htim2.Instance->CNT - count + 2;
			bitrate = (float) count / 4 / 256; //calculate sample rate. as MCLK/SampleCount/MCLKBinaryMultiplier

			if(buff_count > FILL_THRESH){
				bitrate = bitrate - 0.01;
			}
			if(buff_count < LOW_THRESH ){
				bitrate = bitrate + 0.01;
			}
			if (bitrate > 45.0007){
				bitrate = 45.0007;
			}
			if ( bitrate < 44.0007) {
				bitrate = 44.0007;
			}

		}

}
```

Small adjustment to the above to help coax out any drift over time. its crude but its been working well, for like an hour now. 


Once the changes are made to this PR I will test them on the synopsis driver and move this over. Currently this is still running my modified variant.

PanRe:
What i pushed is work in progress and i wanted to share what i have so far. I added:

- SOF ISR routine in audio driver, in addition to the user SOF. The audio driver enables/disables SOF interrupt when needed. If no alternate setting is active, SOF is disabled. There is still some logic missing such that the disabling of the SOF interrupt is coordinated between the drivers (e.g. video and other iso related drivers) and the user. Currently, the driver would switch the ISR interrupt off although the user want to have it enabled.
- I added feeback value calcuation within the audio class SOF ISR in order to put away as much effort as possible from the user. The only thing the user would need to do is to provide some parameters like the master clock rate, the sample rate, and the current cycles counter value of the master clock. This information is gathered by use of a callback function, hence, the user is completely free to choose which master clock is used. Nevertheless, all the tweaks necessary to make the feedback calcuation work (limiting etc.) are currently not present and i hope these don't make this approach impossible. An additional advantage is, that with the feedback calculation within the audio driver i could also implement some checks e.g. if the chosen feedback value interval is not too small or not too big.
- The SOF feedback calculation approach is optional. The "old" way of letting the user set the feedback value completely on its own using `tud_audio_n_fb_set()` is still possible. It is implemented by a define in `audio_device.h`. This could be of interest if the extra CPU power of the SOF interrupt handling is not wanted and instead a bigger buffer is used to compensate for jitter.

Let me know what you think! :)

geekbozu:
I'll give this some attention soon! Looks way more robust then my current solution hah!

PanRe:
@geekbozu Thanks for the first remarks! Just for conformation: do you mean the calculation of the feedback value within the SOF ISR is too costly or the setting of the feedback value by use of the `tud_audio_n_fb_set()` function? I thought calculating and setting the feedback value in the SOF ISR was the very point of this pull request? This was the key question in #1231 in order to reduce the jitter. There are a number of possibilities now:

- The MCU is fast enough to conduct everything within the SOF ISR
- The MCU is not fast enough due to the required division - this can be mitigated by the user to ensure the divisor to be a power of two thus replacing the division by a shift. This is something i wanted to implement anyway. Another possibility would be to save the cycle count within the ISR and conduct the division in the event queue. This would be a bit more complicated for the usb stack logic but possible. However, i don't know if @hathach would like this additional amount of complexity. I just looked it up and the STMF4 needs 14 cylces for a single float division, integer division up to 12. In any case, this is not a dramatic impact. Nevertheless, it would be best to use floating point in case hardware support is given. We could make the SOF ISR and  `tud_audio_n_fb_set()` functions weak allowing the user to replace them with suitable code.
- The MCU is too slow because getting the cycle counter by use of the callback function `tud_audio_n_get_fm_n_cycles_cb()` takes too long. This would be a killer and prohibits all attempts to conduct this within the audio driver. However, since you and @vmilea already implemented such computations within the ISR tells me it should be possible.

If the MCU has too less performance for the SOF ISR approach, there is still the possibility to do it the old way with a bigger buffer as @HiFiPhile explained in #1381 i.e. using a bigger buffer and watch the fill status of the FIFO.
Anyway, i think we need to test it.

geekbozu:
No problem, 
Calculation is to be done in the ISR, that's not really that concern. I would not be shocked entering the ISR is more expensive then the actual math involved. :P
 
My concern is `tud_audio_n_fb_set()` which calls into the USB stack [and queues up packets and such, sometimes.](https://github.com/hathach/tinyusb/blob/master/src/class/audio/audio_device.c#L2274-L2278)
That particular call is what I'm afraid of. 
It needs testing. 
I know in my application the overhead of calling `tud_audio_n_fb_set()` every mS, in the interrupt context was way to high, more so when the Audio device I am using is a 1.1 compliant device, which means a FB update of 4mS not 1mS. On a faster/higher bandwidth device this might be more proper. I'm not sure.  


Honestly I may be off base here, Thanks for going back and forth with me! Reality I think says perf testing is the answer.

PanRe:
@geekbozu No worries, i think these a fruitful discussions here :). I had some thougth about the possible bottle necks. As for the math calcuations i found that in case of power of two shifts 1 cycle for computation will be needed and in case a FPU is available, 3 cycles will be needed. If non is present, then what i currently pushed can be used i.e. roughly 20 cycles. So i guess the computation should not be a concern. I will put the required logic into the next push.
Regarding `tud_audio_n_get_fm_n_cycles_cb()` i had the idea, that we do not really need it. Instead, we could save a pointer to the register of the cycle counter and simply read from there in the SOF ISR. I can't think of an example where this pointer approach would not work. So with respect to your implementation, there would be no loss in performance.
I am currently thinking if it would somehow be possible in the audio driver to disable the SOF interrupt in between the SOF updates e.g. if the update interval would be 4 frames and assuming every frame an audio packet gets send to use the audio packet interrupts to enable the SOF interrupt right before the 4th SOF arrives. But i guess this would be something we should test when everything else is already working ^^. What do think of it?

hathach:
@PanRe sorry for keeping you waiting, I have been working on host stack lately. will check this out in a couple of days or so :).

PanRe:
Ok, the pushed verions would be ready for testing. All that needs to be done is to enable SOF ISR feedback calculation and the desired calculation mode by setting (your parameter are suitable for a power of two shift, but you may try all three)
`#define CFG_TUD_AUDIO_ENABLE_FEEDBACK_DETERMINATION_WITHIN_SOF_ISR    1`
`#define CFG_TUD_AUDIO_ENABLE_FEEDBACK_DETERMINATION_MODE CFG_TUD_AUDIO_ENABLE_FEEDBACK_DETERMINATION_MODE_POWER_OF_TWO_SHIFT`

The parameters required for calculation of the feedback value need to be set by use of `tud_audio_set_fb_params()` within the callback function `tud_audio_set_itf_cb()`. In order to know which alternate setting was currently activated, `p_request` can be checked as
```
uint8_t const itf = tu_u16_low(p_request->wIndex);
uint8_t const alt = tu_u16_low(p_request->wValue);
```
I am apologizing in advance for any stupid error in this code as i was not able to test it live on hardware. The critical points would be

- Check if SOF interrupt is really firing
- The function `tud_audio_set_fb_params()`
- The function `audiod_sof()`

What remains to be done in any way is to define how the overall SOF interrupts shall be handled between the SOFs required by the user and the SOFs required by the individual drivers. Furthermore is it necessary to be able to handle the SOF interrupts in both, in ISR context and in queued context? This would require more logic in the usb stack and i don't know if this is really needed.

If you have any troubles or problems, please do not hesitate to ask ;) Thanks for testing!

hathach:
@PanRe sorry for keeping you waiting again, I have spent last 2 day with sleep-eat cycle in high fever due to covid-19. I feel much better now, but will take a couple more day to rest, I was in the middle of reviewing this PR before that, but It is kind of erased now. I will start to look at this again next week or so. Thanks.

PanRe:
No worries at all! I hope you are well soon! Take care :)!On 3/24/22, 19:23 Ha Thach ***@***.***> wrote:

   
  @PanRe sorry for keeping you waiting again, I have spent last 2 day with sleep-eat cycle in high fever due to covid-19. I feel much better now, but will take a couple more day to rest, I was in the middle of reviewing this PR before that, but It is kind of erased now. I will start to look at this again next week or so. Thanks.

hathach:
@PanRe reviewed the changes, everything look ok. To be honest, though I don't quite understand the feedback computation, I don't think that we could implement one-fit-all computation for feedback. also, required an pointer-to-cycle may not work on other platform.

In my opinion, it is just better to invoke user isr callback and let application to do the computation itself. Since this is most often got tweaked to specific application. We can provide some compute helper, with nominal value etc with your current code. But let the application decide the value. Let me know what you think

@geekbozu yeah, there is chain call especially jumper from one later to another DCD->USBD->Audio driver-> application. It is unfortunate price to pay for portable stack, but let try with this first, mcu is plenty fast nowsaday, if we got trouble with it, I will try to find some shortcut to reduce the call chain.

PS: I did a couple of minor clean up, commenting out the generic tud_sof_isr_set() since we should better use the sof with audio rather.

PanRe:
@hathach Thanks for the review! Well, the feedback value computation follows the official USB specification. IMHO, the specification is not written very clear and may cause a lot of frustration for the user to implement (especially for new users). What i came up with aims to support those poor souls by providing a framework which should be applicable in many cases. 

Boild down, the feedback value is n_samples per frame i.e. `fb = n_s / frame`. The feedback value must be accurate up to 1 sample per second, this required to measure at least 1 second (or 1000 frames for fullspeed, where the USB specification uses 1024). The feedback value therefore would be `fb = n_s / 1000`, where n_s is the number of samples received within 1000 frames. This, however, takes too long. We want the feedback value to be computable faster. The trick now is, that the sample frequency is almost always locked to a faster master clock. Now, instead of counting the number of samples over a periode of 1000 frames, we count the number of cycles of the master clock. Since those two frequencies are coherently locked by some division ratio and the master clock offers a "finer" resolution, we do not need to wait 1000 frames but much less. The feedback value therefore can also be computed as `fb = n_cycles / n_frames * f_s / f_m`. Now, very often the master frequency f_m is high enough that the required measurement period comes down to 1 frame. But better would be some average, thus usually 4 frames are chosen by the user. Finally, the feedback value must be calculated as 16.16 bit fixed point number.
I implemented 3 versions of calcuating the feedback value depending of the values of n_frames, f_s and f_m:
- If f_m * n_frames / f_s is a power of two, the feedback value can be computed fast by a simple shift
- If the above does not work, the next best option is to use float if hardware support is given
- If non is present some fixed point calcuation is performed

The only reason why the above computations look a bit weird is because at the end we need a 16.16 fixed point integer value. I implemented sanity checks to verify the computation works.

The user is free to choose which master clock is to be used (such a master clock is mandatory for I2S interfaces also, where this master clock must be coherently locked to the sample frequency). To obtain a cycle count, an ordinary hardware counter is required (which almost all MCUs should have). Hence, it should work in a lot of cases. One thing which is not very clear right now are those "magic" rules to tweak the feedback value. I am aware of that but i wanted to have this pull request tested before investing more time in handling these cases.
Anyway, in case this initial approach does not work because e.g. some more application specific "magic" rules are required, we can declare the two functions `void audiod_sof ()` and `bool tud_audio_set_fb_params()` as weak. Thus, the user may implement his own specific feedback value computation where he needs the SOF interrupt. If the SOF interrupt is not needed at all, the currently used version `tud_audio_n_fb_set()` is also available. Thus, the user has all possible options available.

@geekbozu Do you think you can squeeze in a test? Thanks in advance! ;)

HiFiPhile:
@PanRe I still think check I2S buffer count is the most reliable way for feedback calculation, due to MCLK measurement error with a timer and error accumulation over time.
Beside with a high frequency MCLK it's hard to get it's exact value. I believe the calculation method in UAC spec fit better the case of ASIC, where the MCLK/USB clock relationship can be calculated by hardware directly.

> One thing which is not very clear right now are those "magic" rules to tweak the feedback value. I am aware of that but i wanted to have this pull request tested before investing more time in handling these cases.

I gave you access to my full code, feel free to take anything useful.

hathach:
thank you @PanRe for a very in-depth explanation, For clarification, I just added the bracket for the formula (please correct if I get it wrong). 

```
fb = (n_cycles / n_frames) * (f_s / f_m)
``` 

I have no doubt that will enormously help lots of other audio newbie including me. My point is that, we don't have to compute it internally within the audio driver, making it difficult for user that want to adjust/tweak the feedback value since any correction will require PR and could affect others. My suggestion is something as follow

```C
uint32_t tud_audio_compute_feedback(n_cyle, n_frame, fs, fm); // which compute the value as exact way this pr is doing

// in user isr 'callback', they use the compute function and set the fb themself
__weak void tud_audio_sof_isr(uint8_t func_id, uint32_t frame)
{
  uint32_t new_fb = tud_audio_compute_feedback(get_cycle(), frame_diff, fs, fm);
  tud_audio_n_fb_set(func_id, new_fb);
} 
```

This way, if our computation need more tweaks, or user want to add magic to their computation, they could do it themselves. Let me know if this makes any senses to you. 

Btw: regarding the power-of-2, is that popular enough to be considered as a case. Normally, what would be the **most often used values** of n_cycle, frame, fs, fm in audio application. For most audio application, I think most MCU would be able to do float, so it would be the primary compute function, but we could also provide `tud_audio_compute_feedback_fixed()` verion as well etc..

PanRe:
@HiFiPhile Thanks a lot! I will have a look into it! I am completely aware, that you have the most experience with audio speakers of us and i value your opinion a lot. It is just that @vmilea and @geekbozu voted for the SOF option and since they made this approach work already i thought it would be a good idea to incorporate that into the audio driver. This would save us some PR in the future too. As for the precision (please tell me if i am wrong on this) i thought the only sensitive part is to obtain the master clock counter value at SOF interrupt time. Once this value is obtained, the feedback value can be computed in a relaxed way as this value must be ready once the host asks for it. If this counter value is read directly from the counter register within `audiod_sof()` then we would suffer from the ISR jitter and that might or might not work. My idea here would be to use a DMA which copies the counter register value to some variable from where `audiod_sof()` can read it. The DMA is triggered by a hardware event i.e. by the interrupt. I know this would work for STM MCUs but maybe not for others. This way, the jitter would be greatly reduced.
Your approach of measuring the sample count in the buffer still works and might be the best option in general. The SOF may be some feature for a stronger MCU.

@hathach Yes the brackets are corret. As for your idea, it appears a bit "split up" to me, leaving one half of the compuation in the driver and the other half in the user isr. Would it not be ok to declare `audiod_sof()` and `tud_audio_set_fb_params()` as weak and give the user the possiblity to copy and paste the code from the driver into his own files and change it to his needs? This way he can change everything if he wants.

Concerning the power-of-2 approach, yes i think it is the most standard case. It is the easiest approach to obtain a sample frequency (and the gold option in case you have an MCU which has no floating point support). For non-power of two ratios you would need a PLL but also this case may be present (i have such a setup). The last option is useful if everything else fails. It works but may need up to 40 cycles (whatever a uint64 division needs).

HiFiPhile:
>  It is just that @vmilea and @geekbozu voted for the SOF option and since they made this approach work already i thought it would be a good idea to incorporate that into the audio driver. 

Yes as you said it's a good idea to have something that should be work in most cases.

>  My idea here would be to use a DMA which copies the counter register value to some variable from where audiod_sof() can read it.

It's rather a hardware question, let's take an example of 72MHz STM32F1 & 12.288MHz MCLK. 

Because MCLK frequency is relatively high to system clock, timer capture & compare shouldn't be used to measure MCLK frequency otherwise the obtained result will be horrible (72/6=12MHz ~ 72/5=14MHz).

Only option is drive the timer with MCLK and count timer_cnt in sof_irq, in this case we have to "press the stop watch" precisely in the irq. DMA can't help as the transfer need to be triggered somehow, I don't aware any MCU which USB SOF can trigger a DMA transfer.

PanRe:
Actually true, after some research i also did not find the possiblity to trigger a DMA from a SOF interrupt signal... sorry i was too enthusiastic ^^. Regarding the timer, yes i had your second version the counter mode in my mind. A constant delay from the SOF interrupt to reading the counter value within the ISR would not be a problem iff it keeps constant from interrupt to interrupt. The SOF interrupt therefore should have a high priority such that the reading is "precise" as you said.

Well, i understand your concerns and i know this solution will not work on all MCUs. But it can work given the MCU and its developer/user is capable enough :)

hathach:
> Yes the brackets are corret. As for your idea, it appears a bit "split up" to me, leaving one half of the compuation in the driver and the other half in the user isr. Would it not be ok to declare audiod_sof() and tud_audio_set_fb_params() as weak and give the user the possiblity to copy and paste the code from the driver into his own files and change it to his needs? This way he can change everything if he wants.

@PanRe uh hmm, why would it result in "split up", the application could compute everything within its sof with compute helper, they could also compensate it +/- on the value or using any magic that works. Maybe I miss something !!!

PanRe:
@hathach Ah... sorry i guess i missunderstood your idea. Well, if you think this is better then of course. I had this idea in my head to avoid too many callback functions to keep the ISR as short as possible. But sure, would you commit such a PR?

hathach:
> @hathach Ah... sorry i guess i missunderstood your idea. Well, if you think this is better then of course. I had this idea in my head to avoid too many callback functions to keep the ISR as short as possible. But sure, would you commit such a PR?

will do, will make necessary changes soon enough.

PanRe:
I would like to finalize this PR so i have taken the liberty to introduce the desired changes of @hathach. After some thinking i came to the conclusion, that the easiest most convenient way would be to have 3 options:
- no SOF and feedback calculation done by user (default as done all the time)
- SOF and feedback value calculation within the audio driver (my approach - nice for rapid prototyping and evaluation if it could work)
- SOF and feedback value calcuation by the user (as desired by @hathach. The user can apply what he tried with option 2 on his own)

So, everybody can do as he wants. Let me know what you think!

hathach:
@PanRe sorry, I am currently way behind my work schedule, I will check this out asap. Sorry for the delay.

hathach:
sorry @PanRe for the huge delay, I have been overwhelmed with work with rp2040 pio usb, and just able to pull it off.

> SOF and feedback value calculation by the user (as desired by @hathach. The user can apply what he tried with option 2 on his own)

I prefer this option more, since this has been awhile and you are probably busy with other work. Hence, I could do the change myself, will make it as an pr to this branch for your review :)

PanRe:
@hathach No worries, glad you are back ;) I already implemented everything, all is now possible, the original solution without SOF, SOF with feedback calculation within the driver (this is a good starting point to start testing if it works with your hardware), and SOF based feedback handling where everything is implemented by the user. You can choose which version you would like by use of defines. So, there is nothing much to implement more, just have a look on what is in now give me feedback please ;)

hathach:
> @hathach No worries, glad you are back ;) I already implemented everything, all is now possible, the original solution without SOF, SOF with feedback calculation within the driver (this is a good starting point to start testing if it works with your hardware), and SOF based feedback handling where everything is implemented by the user. You can choose which version you would like by use of defines. So, there is nothing much to implement more, just have a look on what is in now give me feedback please ;)

thank you, will do will do. Now I completely forgot most of the work, need to revise this a bit :D. Hopefully we will wrap this up asap (in the next few days).

src/class/audio/audio_device.c
```diff
@@ -1970,6 +2038,150 @@ bool audiod_xfer_cb(uint8_t rhport, uint8_t ep_addr, xfer_result_t result, uint3
...
+  if ((f_m % f_s) != 0 || !tu_is_power_of_two(audio->fb_n_frames * f_m / f_s))
+  {
+    TU_LOG2("  FEEDBACK_DETERMINATION_MODE_POWER_OF_TWO_SHIFT not possible!\r\n"); TU_BREAKPOINT(); return false;
+  }
+
+  audio->fb_power_of_two_val = 16 - tu_log2(audio->fb_n_frames * f_m / f_s);
+#endif
+
+#if CFG_TUD_AUDIO_ENABLE_FEEDBACK_DETERMINATION_MODE == CFG_TUD_AUDIO_ENABLE_FEEDBACK_DETERMINATION_MODE_FLOAT
+  audio->fb_float_val = (float)f_s / audio->fb_n_frames / f_m * (1 << 16);
```
hathach, line 2081: still reviewing, but the specs/comment say K=10 for FS, K=13 for HS. Should we do (1 << K) instead of (1 << 16). Sorry if this is noob question.

src/class/audio/audio_device.c
```diff
@@ -1970,6 +2038,150 @@ bool audiod_xfer_cb(uint8_t rhport, uint8_t ep_addr, xfer_result_t result, uint3
...
+#if CFG_TUD_AUDIO_ENABLE_FEEDBACK_DETERMINATION_MODE == CFG_TUD_AUDIO_ENABLE_FEEDBACK_DETERMINATION_MODE_POWER_OF_TWO_SHIFT
+        feedback = (n_cylces - audio->fb_n_cycles_old) << audio->fb_power_of_two_val;
+#endif
+
+#if CFG_TUD_AUDIO_ENABLE_FEEDBACK_DETERMINATION_MODE == CFG_TUD_AUDIO_ENABLE_FEEDBACK_DETERMINATION_MODE_FLOAT
+        feedback = (uint32_t)((float)(n_cylces - audio->fb_n_cycles_old) * audio->fb_float_val);
+#endif
+
+#if CFG_TUD_AUDIO_ENABLE_FEEDBACK_DETERMINATION_MODE == CFG_TUD_AUDIO_ENABLE_FEEDBACK_DETERMINATION_MODE_FIXED_POINT
+        feedback = ((n_cylces - audio->fb_n_cycles_old) << 3) * audio->fb_param_factor_N / audio->fb_param_factor_D;          // feeback_param_factor_N has scaling factor of 13 bits, n_cycles 3 and feeback_param_factor_D 1, hence 16.16 precision
```
hathach, line 2144: could I ask why we does shift `13 on f_s` and `3 on cycles` instead of just `16 on f_s` at once.

src/class/audio/audio_device.c
```diff
@@ -1970,6 +2038,150 @@ bool audiod_xfer_cb(uint8_t rhport, uint8_t ep_addr, xfer_result_t result, uint3
...
+#endif
+
+#if CFG_TUD_AUDIO_ENABLE_FEEDBACK_DETERMINATION_MODE == CFG_TUD_AUDIO_ENABLE_FEEDBACK_DETERMINATION_MODE_FLOAT
+  audio->fb_float_val = (float)f_s / audio->fb_n_frames / f_m * (1 << 16);
+#endif
+
+#if CFG_TUD_AUDIO_ENABLE_FEEDBACK_DETERMINATION_MODE == CFG_TUD_AUDIO_ENABLE_FEEDBACK_DETERMINATION_MODE_FIXED_POINT
+
+  // f_s max is 2^19-1 = 524287 Hz
+  // f_m max is 2^29/(1 ms * n_frames) for full speed and 2^29/(125 us * n_frames) for high speed, this means for f_m fitting in an uint32, n_frames must be < 125 for full speed and < 1000 for high speed (1000 does not fit into uint8 so the maximum possible value cannot even be reached by UAC2 - we don't need to check for it)
```
hathach, line 2087: the bInterval means 2^(interval-1), though I think we almost not going more than 128 (bInterval = 8) in most cases.

image

Though I don't understand why we don't just shift f_s by 16 then multiply with cycle later on instead of shift 13 on f_s and 3 on cycle ?

src/class/audio/audio_device.c
PanRe: Sorry for the late feedback, yes it is the 16.16 format

src/class/audio/audio_device.c
PanRe: That is correct, my original intention was to save N and D as unit32 in order to save space. 13/3 was chosen because i wanted f_s to be able to hold frequencies of bigger than 500 kHz (just to cover a large space also for application other than audio). And i wanted the scale on n_cycles to be a small as possible because n_cycle must cover the same range as f_m, for which we need a huge number. I think 8/8 would be not a good choice therefor.
During rewriting the code, i realized that N * n_cylces must be unit64, so i ended up extending everything. Now, you are correct, we can simple choose to scale N as N << 16 and skip the << 3 later on! I forgot about that, well spotted!

hathach: thanks for confirmation, I have an update proposal for this PR detailed below :)

hathach:
@PanRe I made an PR-to-PR #1463 as update proposal. Thanks to your hard work with the  calculation. I have come up with something hydrid solution without hard-coded into compiler macro. For actual code changes, please check out the other PR, here is my sum up
- First of all `tud_audio_sof_isr()` must be implemented, which will be invoked every polling interval, which is 2^(bInterval-1) in (micro) frames. User can then 
  - compute their own feedback and call tud_audio_fb_set() or
  - use built-in driver computation with tud_audio_feedback_update(cycles), where cycles is the delta from last SOF can be compensated by user own logic. In order for this to work, user must implement the callback `tud_audio_feedback_params_cb()`, which replace the fb_set_params() ( and its need to invoked within the set_itf_cb()).
  - `tud_audio_feedback_params_cb()` will require user to input sample & mclk freq along with fixed/float computation. The extra code to support both fixed and float aren't not that much in both rom and ram and justified for less compiler option I think. The power of 2 method will automatically used if mclk/sample is indeed power of 2.

There is a few pros for this hydrid solution
1. No extra compiler config option
2. User can do their own computation or user built-in one by implementing optional callback
3. User can compensate for isr jitter by adjusting cycle with magic number etc...

There is sure more clean up work, but let me know what do you guys think. Hopefully we could wrap this up

PS: after this is merged, the **__no_inline_not_in_flash_func()** for rp2040 i.e ramfunc/fastfunc will be added as follow-up since that attribute is part of another PR #1412

PanRe:
Hi @hathach,
- if `tud_audio_sof_isr()` is invoked only every `bInterval`, it should be renamed to `tud_audio_interval_isr()`. Otherwise i would expect it to be called every (micro) frame!
- Could you rename `fb_param_factor_N` to `f_s` and `fb_param_factor_D` to `f_m`? (I guess this is what you ment with cleaning up ;))
- We definitely need a minimum example for users such that they get the approach of how to implement everything
I left a few comments in the PR but i like your approach!

hathach:
> * if `tud_audio_sof_isr()` is invoked only every `bInterval`, it should be renamed to `tud_audio_interval_isr()`. Otherwise i would expect it to be called every (micro) frame!
> * Could you rename `fb_param_factor_N` to `f_s` and `fb_param_factor_D` to `f_m`? (I guess this is what you ment with cleaning up ;))
>   I left a few comments in the PR but i like your approach!

Yeah, will rename sof_isr(), and fs fm as you suggested, will also make a union to group all method variable to save footprint as well. I am glad you like my approach, it is still very rough, I will polish it more and try to also include the fifo-count method as suggested by HiFiPhile as well in other PR. 

> * We definitely need a minimum example for users such that they get the approach of how to implement everything

Yeah, please, if you have time, we don't need anything fancy, just enough for demonstration is good enough to get user started :)

hathach (commented):
thank you @PanRe and @HiFiPhile for reviewing and discussing this PR. I think we finally got everything together with frequency method (power of 2 / fixed / float). Regarding the fifo count, due to my inexperience and lack of actual app-on-hardware testing, I comment it out still having place holder to add support later on as follow up PR. Demonstrated example for feedback will also be updated later on as well. It is better to merge this as it is since it has been pending for too long (and we are all busy). At least this give user who need feedback a way to play with. 

detail on fifo count method discussion  https://github.com/hathach/tinyusb/pull/1463#discussion_r885637209

PR #1391, opened 2022-03-09
Renesas ra family support

perigoso:
**Describe the PR**
This PR generalizes the previously named renesas `usba` driver that supported the renesas RX family of MCUs into the `link` driver, named after the name of the usb core according to renesas (LINK core), it also adds support for the cortex m [family RA](https://www.renesas.com/us/en/products/microcontrollers-microprocessors/ra-cortex-m-mcus?).

Along comes the support for one of the evaluation boards [ek-ra4m3](https://www.renesas.com/us/en/products/microcontrollers-microprocessors/ra-cortex-m-mcus/ek-ra4m3-evaluation-kit-ra4m3-mcu-group)

The is also a small change to the way the freertos port is handled in the build system, as it was necessary to use the port provided by renesas

**Additional context**
Most device examples were tested successfully on the ek-ra4m3, no host was tested other than compilation.
No tests were done on the RX family, regression testing is necessary but I do not have access to boards of that family, and have no easy way of installing the toolchain.

Will have access to a ek-ra4m2 and ek-ra6m2 soon, but no tests were done so far.
I expect all RA devices to work, but I have to note that two of the mcus in this family support USB HS on this same core ([RA6M3](https://www.renesas.com/RA6M3), [RA6M5](https://www.renesas.com/RA6M5)), I have not done any work on that though.


perigoso:
I believe I'm done here, the RX family port needs fixes, but I can't deal with that easily so I would appreciate If someone picked that up for me

Wini-Buh:
Why do you made the the missing register definition so complicated ?

The following definition would do it for your case without doing changes umpteen times across the whole usb driver file:

```
#if CFG_TUSB_MCU == OPT_MCU_RAXXX
#include "bsp_api.h"
#include "renesas.h"
#define USB0	R_USB_FS0_BASE
#else
#include "iodefine.h"
#endif
```

perigoso:
> The following definition would do it for your case without doing changes umpteen times across the whole usb driver file

If you looked closer you would see that it's not quite as simple as that, the register definitions are not compatible, that's why I created a local register definition.

I didn't complicate anything, the fundamental driver did not change, only the registers.

perigoso:
the interrups are also handled different in those families, thats why that was abstracted, I tried to a similar approach to the dwc2 driver to keep things consistent, it could have been kept all in the driver files if we wanted

perigoso:
and avoid snarky remarks like that, if it were so easy I wouldn't have gone through all this effort...

src/portable/renesas/link/link_type.h
```diff
@@ -0,0 +1,1624 @@
...
+			volatile uint16_t DMRPU : 1; /* [3..3] D- Line Resistor Control */
+			volatile uint16_t DPRPU : 1; /* [4..4] D+ Line Resistor Control */
+			volatile uint16_t DRPD : 1;  /* [5..5] D+/D- Line Resistor Control */
+			volatile uint16_t DCFM : 1;  /* [6..6] Controller Function Select */
+			uint16_t : 1;
+			volatile uint16_t CNEN : 1; /* [8..8] CNEN Single End Receiver Enable */
+			uint16_t : 1;
+			volatile uint16_t SCKE : 1; /* [10..10] USB Clock Enable */
+			uint16_t : 5;
+		} SYSCFG_bits;
```
kkitayam, line 84: Why not use the symbol name `SYSCFG_b` as defined in `renesas.h` .

The register definitions are very similar to those generated by [SVDConv](https://arm-software.github.io/CMSIS_5/SVD/html/svd_SVDConv_pg.html) with `--fields=struct`.

src/portable/renesas/link/dcd_link.c
```diff
@@ -0,0 +1,826 @@
...
+		uint32_t ep : 8; /* an assigned endpoint address */
+		uint32_t ff : 1; /* `buf` is TU_FUFO or POD */
+		uint32_t : 0;
+	};
+} pipe_state_t;
+
+TU_ATTR_PACKED_END /* End of definition of packed structs (used by the CCRX toolchain) */
+	TU_ATTR_BIT_FIELD_ORDER_END
+
+	typedef struct {
```
kkitayam, line 94: This indentations may be typos.

src/portable/renesas/link/link_type.h
perigoso: was an effort to keep it vaguely similar to what was present before and readable, but in hindsight, the standard is better, will fix

src/portable/renesas/link/dcd_link.c
perigoso: yes, the formatter I'm using (clang-format) doesnt like those

kkitayam:
@perigoso 

Thank you for your great work!
If possible, could you please keep the indentation of `link_dcd.c` and `link_hcd.c`? I would like to see differences other than indentation and symbol names, but it is difficult, I seem to miss them. I don't mind changing the indentation, but would you separate the pull requests.

perigoso:
Hm, that might be hard the way i did the commits, i kind of messed up, if you at the rename commit for instance you'll see what i mean

I'll try to redo this more cleanly when i find some time

kkitayam:
@perigoso 

I see. Could you share the settings file for clang-format? I will temporarily apply it to master branch, and try to generate diff file except indentation changes.

perigoso:
https://gist.github.com/perigoso/79e32321758b1e425212e0984dbe3ce4

perigoso:
rebased everything, now there should be no style changes.

It was quite a bit of work to redo everything properly but it was worth it

perigoso:
hopefully nothing broke, but I tested `cdc_msc` and `cdc_dual_ports` and they behaved as expected so I think I'm good.

perigoso:
@kkitayam did you ever get to test this? not in a rush or trying to pressure, just eager

kkitayam:
Sorry for the late reply. Not yet tested.

src/portable/renesas/link/hcd_link.c
```diff
@@ -27,86 +27,33 @@
 
 #include "tusb_option.h"
 
-#if CFG_TUH_ENABLED && ( CFG_TUSB_MCU == OPT_MCU_RX63X || \
-                               CFG_TUSB_MCU == OPT_MCU_RX65X || \
-                               CFG_TUSB_MCU == OPT_MCU_RX72N )
+#if CFG_TUD_ENABLED && (CFG_TUSB_MCU == OPT_MCU_RX63X || \
```
kkitayam, line 30: It's a typo.
```suggestion
#if CFG_TUH_ENABLED && (CFG_TUSB_MCU == OPT_MCU_RX63X || \
```

perigoso:
ah, that fixed the rx build, wonderful! thank you
hoping that means everything still works in hardware

kkitayam:
I have tested following device examples on RX65N target board with WIndows 10.

|  example |  result   | note |
| ---------- | -------- | ---- |
|  cdc_msc                |  ✔️  | |
|  cdc_dual_ports      | ✔️ | |
| cdc_msc_freertos   | ❌| Enumeration failed |
|hid_composite        |  ❌| Enumeration OK, but the behavior is not correct. |
|msc_dual_lun          | ✔️|  |
|net_lwip_webserver| ❌| Enumeration probably OK, but the behavior is not correct. |
|uac2_headset         | ❌| Enumeration OK, but no audio data recorded |
|video_capture        |❌|  Enumeration OK, but no video data recorded |
|webusb_serial        | ✔️| |

It seems that bulk transfers work fine but Isochroous and interrupt transfers. I'll see if there are any problems.

src/portable/renesas/link/dcd_link.c
```diff
@@ -712,26 +625,26 @@ bool dcd_edpt_open(uint8_t rhport, tusb_desc_endpoint_t const * ep_desc)
...
   unsigned cfg = (dir << 4) | epn;
   if (xfer == TUSB_XFER_BULK) {
-    cfg |= (USB_PIPECFG_BULK | USB_PIPECFG_SHTNAK | USB_PIPECFG_DBLB);
+    cfg |= (LINK_REG_PIPECFG_TYPE_BULK | LINK_REG_PIPECFG_SHTNAK_Msk | LINK_REG_PIPECFG_DBLB_Msk);
   } else if (xfer == TUSB_XFER_INTERRUPT) {
-    cfg |= USB_PIPECFG_INT;
+    cfg |= LINK_REG_PIPECFG_TYPE_ISO;
   } else {
-    cfg |= (USB_PIPECFG_ISO | USB_PIPECFG_DBLB);
+    cfg |= (LINK_REG_PIPECFG_TYPE_INT | LINK_REG_PIPECFG_DBLB_Msk);
```
kkitayam, line 639: Typo
```suggestion
    cfg |= (LINK_REG_PIPECFG_TYPE_ISO | LINK_REG_PIPECFG_DBLB_Msk);
```

src/portable/renesas/link/hcd_link.c
```diff
@@ -726,25 +628,25 @@ bool hcd_edpt_open(uint8_t rhport, uint8_t dev_addr, tusb_desc_endpoint_t const
...
   unsigned cfg = ((1 ^ dir_in) << 4) | epn;
   if (xfer == TUSB_XFER_BULK) {
-    cfg |= USB_PIPECFG_BULK | USB_PIPECFG_SHTNAK | USB_PIPECFG_DBLB;
+    cfg |= LINK_REG_PIPECFG_TYPE_BULK | LINK_REG_PIPECFG_SHTNAK_Msk | LINK_REG_PIPECFG_DBLB_Msk;
   } else if (xfer == TUSB_XFER_INTERRUPT) {
-    cfg |= USB_PIPECFG_INT;
+    cfg |= LINK_REG_PIPECFG_TYPE_ISO;
   } else {
-    cfg |= USB_PIPECFG_ISO | USB_PIPECFG_DBLB;
+    cfg |= LINK_REG_PIPECFG_TYPE_INT | LINK_REG_PIPECFG_DBLB_Msk;
```
perigoso, line 642:
```suggestion
    cfg |= LINK_REG_PIPECFG_TYPE_ISO | LINK_REG_PIPECFG_DBLB_Msk;
```

src/portable/renesas/link/hcd_link.c
```diff
@@ -726,25 +628,25 @@ bool hcd_edpt_open(uint8_t rhport, uint8_t dev_addr, tusb_desc_endpoint_t const
...
-  *ctr = USB_PIPECTR_ACLRM | USB_PIPECTR_SQCLR;
+  *ctr = LINK_REG_PIPE_CTR_ACLRM_Msk | LINK_REG_PIPE_CTR_SQCLR_Msk;
   *ctr = 0;
   unsigned cfg = ((1 ^ dir_in) << 4) | epn;
   if (xfer == TUSB_XFER_BULK) {
-    cfg |= USB_PIPECFG_BULK | USB_PIPECFG_SHTNAK | USB_PIPECFG_DBLB;
+    cfg |= LINK_REG_PIPECFG_TYPE_BULK | LINK_REG_PIPECFG_SHTNAK_Msk | LINK_REG_PIPECFG_DBLB_Msk;
   } else if (xfer == TUSB_XFER_INTERRUPT) {
-    cfg |= USB_PIPECFG_INT;
+    cfg |= LINK_REG_PIPECFG_TYPE_ISO;
```
perigoso, line 640:
```suggestion
    cfg |= LINK_REG_PIPECFG_TYPE_INT;
```

kkitayam:
@perigoso 

Thanks for updating! I have confirmed that these examples work fine except `cdc_msc_freertos`.
Also, I confirmed that `cdc_msc_freertos` didn't work at the HEAD of the master branch either. So, this issue is not caused  by this PR.

kkitayam:
The host driver will test at a later date.

kkitayam (approved):

kkitayam:
@Wini-Buh 

If you don't mind, could you comment in terms of big-endian or CCRX?

Wini-Buh:
@kkitayam 

I had a look into the code and did a quick test with the CCRX toolchain in big and little endian mode. After doing the quick hacks on the code (add the correct `#pragma` directive and add `__evenaccess` keyword, see the description below), the CDC test was successful. 

Unfortunately the code as it is, is not endian independent. For the CCRX we could solve this with a `#pragma` directive. But to be more independent of the toolchain (and invest into the future), we could use solution like it is found in the header file `iodefine.h`, which is provided by Renesas if using the GCC toolchain (see the example code fragment below). This way it should be possible to do get along without the `#pragma` directive.

```
typedef struct st_usb0 {
	union {
		unsigned short WORD;
		struct {
			
#ifdef __RX_LITTLE_ENDIAN__
			unsigned short USBE : 1;
			unsigned short  : 3;
			unsigned short DPRPU : 1;
			unsigned short DRPD : 1;
			unsigned short DCFM : 1;
			unsigned short  : 3;
			unsigned short SCKE : 1;
			unsigned short  : 5;
#else
			unsigned short  : 5;
			unsigned short SCKE : 1;
			unsigned short  : 3;
			unsigned short DCFM : 1;
			unsigned short DRPD : 1;
			unsigned short DPRPU : 1;
			unsigned short  : 3;
			unsigned short USBE : 1;
#endif
		} BIT;
	} SYSCFG;
```

Another small problem is the definition of the `LINK_REG`. To make sure the CCRX accesses the defined registers correctly, the keyword `__evenaccess` is required (this extension guarantees access in the size of the target variable).

```
#if defined(__CCRX__)
#define LINK_REG ((LINK_REG_t __evenaccess*)LINK_REG_BASE)
#else
#define LINK_REG ((LINK_REG_t*)LINK_REG_BASE)
#endif
```

I hope this info is useful for you.

kkitayam:
@Wini-Buh 
Thank you very much for testing and useful comments! I understood we need more modifications to work with gcc for big-endian.

My understanding for the current implementation (`dcd_usba`)  of master branch, it supports big-endian with only CCRX. Because of using [TU_ATTR_BIT_FIELD_ORDER_BEGIN](https://github.com/hathach/tinyusb/blob/9c8c5c1c53214b8954bb28c33a496b5b423d5ecc/src/portable/renesas/usba/dcd_usba.c#L104) for endianness. It is a `#pragma`. Is this correct?

I think it is better to keep the features that the current implementation of master branch has. So, we should keep support big-endian with CCRX. Of course, I think it is more better to support big-endian with gcc. But, I don't mind big-endian with gcc would be supported in another PR and not in this PR.

> After doing the quick hacks on the code (add the correct #pragma directive and add __evenaccess keyword, see the description below), the CDC test was successful.

I understood `__evenaccess` keyword needs to [src/portable/renesas/link/link_rx.h](https://github.com/hathach/tinyusb/pull/1391/files#diff-d20e3ed5d428d2f4866f564cc06db1ea992ccb938dff2503103bad097d91665a). But, It seems that `#pragma` is already added  [src/portable/renesas/link/link_type.h](https://github.com/hathach/tinyusb/pull/1391/files#diff-f11e050731e577e2dfd2f4dd22b03c203ffd95a717e255cc65b1b05cece790dd). It has TU_ATTR_BIT_FIELD_ORDER_BEGIN directive. If you don't mind, would you give me more information regarding correct `#pragma`?

Wini-Buh:
@kkitayam 
> My understanding for the current implementation (`dcd_usba`) of master branch, it supports big-endian with only CCRX. Because of using [TU_ATTR_BIT_FIELD_ORDER_BEGIN](https://github.com/hathach/tinyusb/blob/9c8c5c1c53214b8954bb28c33a496b5b423d5ecc/src/portable/renesas/usba/dcd_usba.c#L104) for endianness. It is a `#pragma`. Is this correct?

Yes

> I think it is better to keep the features that the current implementation of master branch has. So, we should keep support big-endian with CCRX. Of course, I think it is more better to support big-endian with gcc. But, I don't mind big-endian with gcc would be supported in another PR and not in this PR.

OK
> 
> I understood `__evenaccess` keyword needs to [src/portable/renesas/link/link_rx.h](https://github.com/hathach/tinyusb/pull/1391/files#diff-d20e3ed5d428d2f4866f564cc06db1ea992ccb938dff2503103bad097d91665a). But, It seems that `#pragma` is already added [src/portable/renesas/link/link_type.h](https://github.com/hathach/tinyusb/pull/1391/files#diff-f11e050731e577e2dfd2f4dd22b03c203ffd95a717e255cc65b1b05cece790dd). It has TU_ATTR_BIT_FIELD_ORDER_BEGIN directive. If you don't mind, would you give me more information regarding correct `#pragma`?

First of all, I did the test with an old CCRX toolchain version which I normally not use anymore, sorry for that.
So the #pragma is already correctly set. The correct #pragma is `_Pragma("bit_order right")` (that's already correct defined in the file `tusb_compiler.h`). So everything is fine here. Only the `__evenaccess` keyword needs to be added. With that everything is working fine.

Sorry again for the mistake with the old toolchain.

perigoso:
If i understood correctly commit a6139eddc113238b78ff576430682317ee6b925b is all that was needed?

if so this PR is ready, I would just like the good to go from @hathach (I understand you are busy right now, don't worry)

kkitayam:
Sorry for late response. I am busy for family reasons.

@Wini-Buh Thank you for comments. I understood we need `__evenaccess` keyword.

@perigoso Thank you for updating this PR to add the keyword. I think this PR is ready for merge to master branch!

perigoso:
@hathach do you think you could review this? It should be relatively clean, I just don't want to have this become stale and then have to rework it again

hathach:
@perigoso I am so sorry for the massive delay, Unfortunately, this matter slipped off my radar and the notification got lost amidst others. I admit that my lack of familiarity with Renesas MCUs make me not able to follow up on this matter more closely. 

I am gonna try to sync this with master, review and pull out my boards to test, hopefully we could merge this soon as it should be.

Sorry again

perigoso:
Hi @hathach , unfortunately I no longer have access to the renesas development boards (job change), but I'm happy to see this being picked up

hopefully it's not too much work, the more conflicting change might be the RTOS changes

src/portable/renesas/link/hcd_link.c
```diff
@@ -156,28 +106,6 @@ typedef struct
 //--------------------------------------------------------------------+
 static hcd_data_t _hcd;
 
-static uint32_t disable_interrupt(void)
```
hathach, line 159: do you still remember why this disable_interrupt()/enable_interrupt() is dropped ? Or is it not needed in the first place ?

hathach:
> Hi @hathach , unfortunately I no longer have access to the renesas development boards (job change), but I'm happy to see this being picked up
> 
> hopefully it's not too much work, the more conflicting change might be the RTOS changes

Thank you, RTOS change is minor comparing to the whole PR. It is totally my fault, I am trying to fix/merge this as soon as I could :)

src/portable/renesas/link/link_ra.h
```diff
@@ -0,0 +1,64 @@
...
+/* renesas fsp api */
+#include "bsp_api.h"
+
+#define LINK_REG_BASE (0x40090000)
+
+TU_ATTR_ALWAYS_INLINE
+static inline void link_int_enable(uint8_t rhport)
+{
+  (void) rhport;
+  NVIC_EnableIRQ(TU_IRQn);
```
hathach, line 45: everything look great, though I don't understand where is `TU_IRQn` is used, I couldn't find where it got used to setup USB interrupt, maybe I miss something.

hathach:
@perigoso I have made some minor commits and I think we are really close to merge this. I only have an question regarding the IRQn number used by RA chip as mentioned above. That is ok if you don't remember that, we can merge it as it is for now and worry about that later on.

Another thing is the name of USB IP, I did a quick search on renesas page. As far as I can tell, Link is not the name of the controller, it seems to be used in place for connectivity, there is also sata, pcie link. If you don't mind I would suggest to rename the usbip back to simply renesas ? ofc, I will do that myself.

Screenshot from 2023-03-12 16-30-16

src/portable/renesas/link/hcd_link.c
perigoso: This has moved to the link_rx.h file, as it is specific to the RX family

perigoso:
> Another thing is the name of USB IP, I did a quick search on renesas page. As far as I can tell, Link is not the name of the controller, it seems to be used in place for connectivity, there is also sata, pcie link. If you don't mind I would suggest to rename the usbip back to simply renesas ? ofc, I will do that myself.

I don't remember where I got the IP name, but I can't seem to find anything, renaming things back is fine by me

perigoso:
> Another thing is the name of USB IP, I did a quick search on renesas page. As far as I can tell, Link is not the name of the controller, it seems to be used in place for connectivity, there is also sata, pcie link. If you don't mind I would suggest to rename the usbip back to simply renesas ? ofc, I will do that myself.

I don't remember where I got the IP name, I might have misread something, renaming things back is fine by me

src/portable/renesas/link/link_ra.h
hathach: oh, I see thank you, it is a dynamic IRQ binding. I may try to change it to CFG_TUD_*

hathach:
> I don't remember where I got the IP name, I might have misread something, renaming things back is fine by me

Thanks, I will do the rename, maybe calling it rusb2, most of renesas mcu use this IP I think.

hathach (approved):
Superb! Thank you @perigoso very much for generalizing the renesas driver and sorry for the huge delay. I was too busy last time (around last Lunar New Year), and lots of stuffs happening making this fall off my radar. Though that is still hugely my bad. 

I made some changes but should not cause any issue since I tested with both RA and RX boards
- Add ra4m1_ek board, which I have. I also have an ra6m1_ek but that can wait for next PR. Note bsp for RA can be refactor to reduce code a bit but can wait for next PR as well
- rename usb-ip from link to rusb2 (they seem to also have usb 3 supported but only on CPU)
- rename to freertos portable to make it clear that path is to portable source

hathach:
@perigoso thank you very much, I have already rename everything from LINK_REG to RUSB2_REG, let me know if you spot any leftover. I have another ra6m1 board, but running out of time. Will come back to this port later on. 

PS: will also review the other SAM3 pr as soon as I could.

PR #1412, opened 2022-03-30
PIO USB support

hathach:
**Describe the PR**
Add support for [pico-pio-usb](https://github.com/sekigon-gonnoc/Pico-PIO-USB), allow TinyUSB can be used with both native usb (rhport0) and pio-usb (rhport1). Currently pio-usb and native usb must be on different mode. For recommendation, native should be device and pio-usb as host, since device has an stricter timing constraint, while in host mode, rp2040 basically control all the timing.

To test this PR
- check out this branch, and submodule lib/Pico-PIO-USB
- compile example host/hid_to_cdc, which is configured to use native usb as cdc device, and pio usb as host to get hid input. 
- any input event e.g mouse move, button will be printed to cdc on the native usb.

requires https://github.com/sekigon-gonnoc/Pico-PIO-USB/pull/22


kilograham:
@hathach can you please back this out... this touches a bunch of the Pico SDK integrations (and does weird stuff like setting the clock unconditionally).

Perhaps we dropped the ball on reviewing this, but it needs some more work to not just pull in PIO support when the user doesn't want it.

kilograham:
We can start a discussion again on the code. Particularly, the user should explicitly opt into "dual" USB, and the existing TinyUSB base INTERFACE libraries should not include it.

kilograham:
actually, i just added a PR instead #1467

hathach:
> We can start a discussion again on the code. Particularly, the user should explicitly opt into "dual" USB, and the existing TinyUSB base INTERFACE libraries should not include it.

Right, I agree that user need to explicitly specify the dual USBs. I am not too familiar with the way pico-sdk wrapper for tinyusb lib/target definition, but I did try to add a separated target/lib for pio_usb, and user indeed need to initialize that in addition to normal tinyusb_device/host.

> actually, i just added a PR instead https://github.com/hathach/tinyusb/pull/1467

thanks, I am currently off for a couple of days, will review it next week.

kilograham:
> Right, I agree that user need to explicitly specify the dual USBs. I am not too familiar with the way pico-sdk wrapper for tinyusb lib/target definition, but I did try to add a separated target/lib for pio_usb, and user indeed need to initialize that in addition to normal tinyusb_device/host.

Yes, it was actually nice and separate, the weirdness we have with keeping things happy when doing raw TinyUSB examples (which includes pico-sdk within it's build) and SDK (which includes TinyUSB within it's build) and pico-examples which uses TinyUSB examples build as a sub-build just needed a little jiggling!

PR #1422, opened 2022-04-07
Add meson Meson build file

perigoso:
**Describe the PR**
This adds a meson build file that exports a dependecy for use in projects using the meson build buildsystem
This is just for convenience of said projects and won't be used internally, in the future it might be worth considering switching to meson to build the examples too, as it's much easier to maintain than make

As for know projects that will benefit from this, that I know of, [icebreaker](https://github.com/icebreaker-fpga/icebreaker-v2-usb-firmware), that is being worked on by @gregdavill and @ahmedcharless right now (mentioning for possible input), and internal projects I have at work that I can't disclose

This is just a draft, creating the PR just so I can get some input

**Additional context**
[Meson](https://mesonbuild.com/index.html)
> Meson is an open source build system meant to be both extremely fast, and, even more importantly, as user friendly as possible.
> 
> The main design point of Meson is that every moment a developer spends writing or debugging build definitions is a second wasted. So is every second spent waiting for the build system to actually start compiling code.

mention #1423

### DIFF
--- a/.github/workflows/trigger.yml
+++ b/.github/workflows/trigger.yml
@@ -37,13 +37,14 @@ jobs:
         # Copy src and other files
         cp -r src tinyusb_src/
         cp LICENSE tinyusb_src/
+        cp meson.build tinyusb_src/
         cd tinyusb_src
 
         # Commit if there is changes
         if [ -n "$(git status --porcelain)" ]; then
           git add .
           git commit --message "Update from https://github.com/$GITHUB_REPOSITORY/commit/$GITHUB_SHA"
-          git push 
+          git push
         fi
 
     - name: Create tinyusb_src Release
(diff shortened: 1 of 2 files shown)


gregdavill:
This looks handy when making use of tinyusb in a meson based project!

eli-schwartz:
Why are you repeatedly comparing against OPT_MCU_NONE?

This meson.build file doesn't have a `project()` declaration -- is it supposed to be used as a https://mesonbuild.com/Subprojects.html or what?

perigoso:
I'm comparing against OPT_MCU_NONE because it's not done lol

not sure making this a subproject is the best approach, it's not really meant to be compiled as a library, and it needs an external configuration file, which is doable, but i usually just `subdir()` this file where it declares a dependency and use that in the main project

but this is which I made a draft

ernestask:
Honestly, this is kind of useless in its current form.

https://mesonbuild.com/Subprojects.html

Make this a project and expose the dependency, that way it’s easier to include the code.

perigoso:
Closing as this would need a big clean up I won't be doing any time soon, the project is also not planning to switch so for the maintenance might not be worth it

PR #1433, opened 2022-04-17
tusb_types: add edpt_dir and edpt_type strings

Skyler84:
Added descriptive strings for edpt_dir and edpt_type for debug output use rather than ports having to define their own.


hathach (changes requested):
thank you for your PR, all the debug function must be in `common/tusb_debug.h` and wrapped within the CFG_TUSB_DEBUG, check out similar string debug for reference

Skyler84:
ok will do!

PR #1219, opened 2021-12-01
Midihost

rppicomidi:
**Describe the PR**
This pull request adds MIDI host functionality to tinyusb. 

**Additional context**
This code is a work in progress and should be treated as a draft. I am submitting it as a pull request in hopes it will be helpful.

**Working now**
- Compiles and loads on a RP2040 board in the Visual Studio Code IDE
- Both devices I have tried enumerate correctly (Yamaha Reface CS, Korg nanoKONTROL Studio)
- The code will correctly format a MIDI message and send it to the device via the bulk OUT endpoint; the example code will blink the LEDs behind the REC button on the Korg nanoKONTROL Studio configured for use with a DAW that uses Mackie Control messages.

**There are a lot of known defects**
- It is not structured correctly to work with USB hubs. This is laziness on my part and can easily be fixed.
- It has only been tested so far on a Raspberry Pi Pico board
- The bulk IN endpoint code I wrote for the RP2040 does not work reliably. See Blocking Issues below
- It probably will not correctly enumerate USB Audio Class 2 devices with a MIDI interface because of the interface association descriptor. I know how to fix this if it is a problem.
- There are asserts in the code that will fail sometimes on after an unplug or after a re-plug. I fixed one but I have randomly hit others.

**Blocking Issues**
@hathach I am struggling with the IN endpoint. The Korg device I use for most of my testing should be sending a NAK for each transfer request when no controls are touched, but the RP2040 is happily reporting completed transfer for each transfer request. The buffers contain all 0s. I don't have a bus analyzer so I don't know whether the Korg test device or my code is at fault there. Also not clear is what to do with the NAK_POLL register at offset 0x6c; I would think that a bulk IN endpoint that shares hardware with bulk OUT should not automatically retry on NAK, nor should it time out. It should set the NAK flag in the SIE status register, but I don't observe that. Do you know of a support path from the Raspberry Pi foundation where I can get direct answers to questions about the USB on the RP2040?

PR #1434, opened 2022-04-17
Rp2040 hcd bulk

Skyler84:
Add support for BULK endpoint transfers. This is achieved by allocating hw_endpoint for bulk transfers, not just interrupt. These endpoints can perform transfers in a non-blocking manner, whereas using EPx would block all other (non interrupt) transfers.
This change also allows interrupt transfers in both IN and OUT directions now.


bohdan-tymkiv:
I can confirm that this patch works with multiple IN/OUT BULK endpoints simultaneously.
I am using a raw bare API (tuh_edpt_xfer) to communicate with a custom USB device.

rppicomidi:
@Skyler84 I can confirm this patch works with one BULK IN endpoint and and ONE BULK OUT endpoint at the same time with a MIDI Host function.

rppicomidi:
@Skyler84 @bohdan-tymkiv Did you try this patch with a hub? I am finding it works fine when only one device is plugged in but if two devices are plugged in and packets are being sent to both devices, packet corruption occurs. Also, the IN transfers complete on both devices but the second device plugged in has an all-zero packet buffer even though the connected device should be sending data.

Skyler84:
I have not tested this with a hub yet! Definitely something I can take a look at with my logic analyzer. I've also found that it doesn't work with transfers >64B since that's all that's allocated, and the double buffering is not complete in the HCD i dont think?
It's a fairly close solution but I think needs some refinement.

johnkjellberg:
I can confirm that this works for direct connection to a CDC device!

bohdan-tymkiv:
Have not tried it with the HUB. I think implementing HUB support can be a subject of a different patch. This change works and it basically makes RP2040 Host functional.

johnkjellberg:
Any news regarding this? Is the hub support critical for merging, or is there something else that needs to be done before approval?

wrtlprnft:
I did a proof of concept in https://github.com/wrtlprnft/tinyusb/commit/43a245db80b0f293ab9d3fcb55503e6309feb072.

I don't think it's ready to be a pull request of its own because it requires the user to only have one transfer (`tuh_edpt_xfer` or `tuh_control_xfer`) in flight at a time and there are likely many things I didn't think about. Maybe it's useful to someone until there's a better solution.

johnkjellberg:
> I did a proof of concept in [wrtlprnft@43a245d](https://github.com/wrtlprnft/tinyusb/commit/43a245db80b0f293ab9d3fcb55503e6309feb072).
> 
> I don't think it's ready to be a pull request of its own because it requires the user to only have one transfer (`tuh_edpt_xfer` or `tuh_control_xfer`) in flight at a time and there are likely many things I didn't think about. Maybe it's useful to someone until there's a better solution.

@wrtlprnft Thanks! This is also the first attempt I have seen that I got working for both CDC and MSC!
Seems to be some cleanup bugs though(it crashes/don't work after CDC have been unmounted. Neither CDC or MSC). I will investigate that a bit.

rppicomidi:
@wrtlprnft Have you seen the discussion #1261? (TL;DR If a bulk IN transfer does not complete (because the device NAKs the transfer), epx is busy until the device sends some data back) I took an admittedly quick look at the PR and did not see such a mechanism. My attempt a while ago in PR #1219 does address receiving NAK on bulk IN, but it is a bit ugly.

wrtlprnft:
@rppicomidi I did not see that, but I was wondering about that myself. Thanks for the pointer!

My intended application, which I was doing a quick prototype for to see whether it was feasible and fast enough, was doing JTAG over an FTDI dongle. In that application, I have a clear command/response structure where I know more or less exactly how many Bytes of IN transfer I'll get at what point, so the issue never really arose.

As I said before, my code currently doesn't even attempt to support multiple concurrent transactions at the moment, so that's something that will have to be done before I can really consider it in my branch. I'll definitely have a look at your branch because it seems like it solves many of the issues I had.

hathach (approved):
Thank you very much for this brilliant PR and every for helping testing and verifying. Indeed the datasheet seems to hint that these "Interrupt" endpoints are used for Interrupt type. Using them with bulk in asynchronous manner is a great find out. Which simplify the usbh driver.

Basically the interrupt and bulk have the same packet layout (pid, data, crc, ack), the difference is that bulk endpoint will continuously transferring data as long as bus frame (1ms) is allowed. For example, we can have 10 packets of 64 bytes in 1ms with Bulk. For interrupt endpoint, it will only transferring 1 packet per interval (could be 1ms, 2ms, 4ms etc..). To be honest, I am not sure if using these "interrupt" endpoint for bulk this way would hurt the throughput or not (need to confirm with rpi team). If it does then our max throughput would be 64 bytes/1ms --> 64KB/s . IMHO, it will likely be the case here. Though it is still better than nothing now. We can merge this as it is and try to figure out a better implementation later e.g similar to #1219 approach

PR #1444, opened 2022-04-24
Gh action use replacement ruby

Firefishy:
Replace retired github action [actions/setup-ruby ](https://github.com/actions/setup-ruby) with supported replacement [ruby/setup-ruby](https://github.com/ruby/setup-ruby)

Minor actions/checkout version bump.


Firefishy:
The tests pass: https://github.com/Firefishy/tinyusb/runs/6147648724?check_suite_focus=true

hathach (approved):
thank you very much for the bump up

PR #1455, opened 2022-05-03
Fix asserting at 853 in dcd_edpt_close on endpoint re-open

Iktek:
Solves issue discussed in:
https://github.com/hathach/tinyusb/discussions/1439

This is not a complete solution as we only can re-open an endpoint with the same size. 
But it improves the behaviour allowing to close and re-open an endpoint that is not opened at last

We now let fifos allocated if we can't free them.
We check if there is already an allocated fifo for that endpoint.
Assert when that fifo has the wrong size

This could be further improved by collecting the already allocated fifos with it's size in a list and re-assign them if they are fitting and try to allocate a new one if they're not fitting.


HiFiPhile:
Since #2333 there is no more dynamic FIFO size adjustment, should fix the issue.

hathach:
right, we may remove the dcd_edpt_close() API in the future as well. This should be marked as solved. Feel free to re-open or file an new issue/pr should it still be an issue. Thank you

Issue #1457, opened 2022-05-04
some examples don't work for imxrt:  cdc_msc_freertos, dynamic_configuration, hid_composite_freertos,  usbtmc


ccrome:
### Operating System

Linux

### Board

imxrt 1060, imxrt

### Firmware

examples/device/cdc_msc_freertos
examples/device/dynamic_configuration
examples/device/hid_composite_freertos
examples/device/usbtmc

None of these examples work on the teensy 4.0, using etither tag 0.13.0, or the latest fa895ed3dc46d5aa5e11b19ff07beb6fcfc62655.

**all other device examples work**

### What happened ?

I compile all examples using the tools/build_family.py.  They all compile properly with the latest commit (fa895ed3dc46d5aa5e11b19ff07beb6fcfc62655), but some build errors with 0.13.0.  

In any case, when I attempt to load the binaries into the teensy, the following examples simply stay in the teensy bootloader and never actually work properly:
cdc_msc_freertos, dynamic_configuration, hid_composite_freertos,  usbtmc

I get the same behavior on both the teensy and on the 1060 evk

### How to reproduce ?

run the build_family.py imxrt script, and attemp to load and use the aforementioned binaries.

ccrome:
Oh!  Found that on the 1060 EVK, some demos are on USB1 whereas others are on USB2.

So, it appears the 1060 EVK works on all examples, so now just left wondering why it doesn't work on all examples on the tensy 4

| teensy | 1060evk | example |
|--------|---------|---------|
|     OK |  OK USB1  |   audio_4_channel_mic     |
|     OK |  OK USB1  |   audio_test              |
|     OK |  OK       |   board_test              |
|     OK |  OK USB2  |   cdc_dual_ports          |
|     OK |  OK USB2  |   cdc_msc                 |
|     -- |  OK USB2  |   cdc_msc_freertos  device attempts to attach to USB 2, but gets errors on enumeration |
|     OK |  OK USB2  |   dfu                     |
|     OK |  OK USB2  |   dfu_runtime             |
|     -- |  OK USB2  |   dynamic_configuration   |
|     OK |  OK USB2  |   hid_boot_interface      |
|     OK |  OK USB2  |   hid_composite           |
|     -- |  OK USB2  |   hid_composite_freertos  |
|     OK |  OK USB2  |   hid_generic_inout       |
|     OK |  OK USB2  |   hid_multiple_interface  |
|     OK |  OK USB2  |   midi_test               |
[...]

PR #1459, opened 2022-05-04
Fix special pin on teensy

ccrome:
This fixes issue https://github.com/hathach/tinyusb/issues/1457.

The problem is that somehow, someway, the IOMUXC_GPIO_AD_B0_09_GPIO1_IO09 pin gets
configured wrong, which interferes with the way the teensy boots.  I don't
know what's messing it up, or any.  But I do know that this fix corrects
non-functioning usb examples on the teensy:

examples/device/cdc_msc_freertos
examples/device/dynamic_configuration
examples/device/hid_composite_freertos
examples/device/usbtmc

### DIFF
--- a/hw/bsp/imxrt/family.c
+++ b/hw/bsp/imxrt/family.c
@@ -57,6 +57,27 @@ void board_init(void)
 #endif
 #endif
 
+#ifdef TEENSY_SPECIAL_PINMUX
+  // TEENSY jtag pins must be configured in this manner
+  // so that communication works with the bootlaoder chip
+  // (the MKL02Z32VFG4, U2).  by configuring these pins
+  // to JTAG mode, it allows a software reboot to works properly
+  //
+  // With the pins configured thusly, you can reboot to the
+  // application or to the teensy bootloader.
+  //
+  // See the reboot_to_bootloader() and reboot_to_application()
+  // functions below.
+  IOMUXC_SetPinMux(IOMUXC_GPIO_AD_B0_04_SRC_BOOT_MODE00, 0U);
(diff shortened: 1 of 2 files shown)


ccrome:
The one pin I fixed previously allowed the board to boot, but the button on the TEENSY still didn't work properly, presumably since some pins were still being held in an invalid state for the MKL02Z32FG4 (U2) to do its job.  Check out the [schematic](https://www.pjrc.com/teensy/schematic.html).

ccrome:
Okay... sorry for any confusion here, but now I pretty much understand what's going on.  In order for teensy to work as expected, it needs the pins configured in JTAG mode so that the reboot-to-bootloader command works (there's a separate bootloader microcontroller that connects to the jtag pins).  

This patch now fixes it, and allows rebooting to the application or to the bootloader.

hw/bsp/imxrt/family.c
```diff
@@ -169,6 +190,19 @@ void USB_OTG2_IRQHandler(void)
   #endif
 }
 
+void reboot_to_application(void)
```
hathach, line 193: reboot_to_application() / reboot_to_bootloader() aren't used by tinyusb example, could we dropped it ?

hathach (commented):
sorry for late response, I was busy with other works, the reboot_to_bootloader() and reboot_to_application() aren't used by tinyusb example, we should drop it if possible ? I could make the changes if you agree to speed up the merge

hathach (commented):
thank you for your PR, however the root cause and correct fix is described in #1516

ccrome:
Excellent!  I guess I didn't understand what was going on :-)  Thanks again for the amazing library!

hathach:
no problems, and thanks for the issue and your patience.

PR #1460, opened 2022-05-05
set irq priority for freertos

ccrome:
**Describe the PR**
For FreeRTOS, This PR sets the USB interrupt priority to configLIBRARY_MAX_SYSCALL_INTERRUPT_PRIORITY, which is where it should be in order to not cause problems for freeRTOS.  it was initially there, then commented out for some reason.  Not sure exactly why it was removed.  

This sets the interrupt priority for both USB ports, if they exist.


hathach (commented):
Thank you for your PR. Can you put the code directly into the `#elif` instead of having an extra function.

ccrome:
sure, no problem.  Just commited.  Is the right thing here to do a force-push so there's only one commit?  That's what I did.

hathach (approved):
look great, thank you for your PR. This indeed should be set by all ports with freeRTOS.

PR #1463, opened 2022-05-13
changes proposal to audio feedback computation

hathach:
**Describe the PR**
Proposal update to #1381

### DIFF
--- a/src/class/audio/audio_device.c
+++ b/src/class/audio/audio_device.c
@@ -1708,6 +1708,7 @@ static bool audiod_set_interface(uint8_t rhport, tusb_control_request_t const * 
 
       TU_VERIFY(foundEPs == nEps);
 
+#if CFG_TUD_AUDIO_ENABLE_FEEDBACK_EP
       // Invoke one callback for a final set interface
       if (tud_audio_set_itf_cb) TU_VERIFY(tud_audio_set_itf_cb(rhport, p_request));
 
@@ -1728,6 +1729,7 @@ static bool audiod_set_interface(uint8_t rhport, tusb_control_request_t const * 
           set_fb_params(audio, sample_freq, mclk_freq);
         }
       }
+#endif
 
       // We are done - abort loop
       break;
@@ -1737,6 +1739,7 @@ static bool audiod_set_interface(uint8_t rhport, tusb_control_request_t const * 
     p_desc = tu_desc_next(p_desc);
   }
 
+#if CFG_TUD_AUDIO_ENABLE_FEEDBACK_EP
   // Disable SOF interrupt if no driver has any enabled feedback EP
   bool disable = true;
   for(uint8_t i=0; i < CFG_TUD_AUDIO; i++)
@@ -1748,6 +1751,7 @@ static bool audiod_set_interface(uint8_t rhport, tusb_control_request_t const * 
     }
   }
   if (disable) usbd_sof_enable(rhport, false);
+#endif
 
   tud_control_status(rhport, p_request);
 
@@ -2080,7 +2084,6 @@ static bool set_fb_params(audiod_function_t* audio, uint32_t f_s, uint32_t f_m)
 
   return true;
 }
-#endif
 
 uint32_t tud_audio_feedback_update(uint8_t func_id, uint32_t cycles)
 {
@@ -2118,10 +2121,12 @@ uint32_t tud_audio_feedback_update(uint8_t func_id, uint32_t cycles)
 
   return feedback;
 }
+#endif
 
 void audiod_sof_isr (uint8_t rhport, uint32_t frame_count)
 {
   (void) rhport;
+  (void) frame_count;
 
 #if CFG_TUD_AUDIO_ENABLE_EP_OUT && CFG_TUD_AUDIO_ENABLE_FEEDBACK_EP
   // Determine feedback value - The feedback method is described in 5.12.4.2 of the USB 2.0 spec
--- a/src/class/audio/audio_device.c
+++ b/src/class/audio/audio_device.c
@@ -306,21 +306,29 @@ typedef struct
 
 #if CFG_TUD_AUDIO_ENABLE_FEEDBACK_EP
 
-  uint32_t fb_val;                      // Feedback value for asynchronous mode (in 16.16 format).
-  uint8_t fb_n_frames;                  // Number of (micro)frames used to estimate feedback value
-  uint8_t fb_n_frames_shift;
-  uint8_t fb_compute_method;
+  struct {
+    uint32_t value;     // Feedback value for asynchronous mode (in 16.16 format).
+    uint32_t min_value; // min value according to UAC2 FMT-2.0 section 2.3.1.1.
+    uint32_t max_value; // max value according to UAC2 FMT-2.0 section 2.3.1.1.
 
-  uint32_t fb_val_min;                  // Maximum allowed feedback value according to UAC2 FMT-2.0 section 2.3.1.1.
-  uint32_t fb_val_max;                  // Maximum allowed feedback value according to UAC2 FMT-2.0 section 2.3.1.1.
+    uint8_t frame_shift; // bInterval-1 in unit of frame (FS), micro-frame (HS)
+    uint8_t compute_method;
 
-  // should be union
-  uint8_t fb_power_of_two_val;
+    union {
+      uint8_t power_of_2; // pre-computed power of 2 shift
+      float float_const;  // pre-computed float constant
 
-  float fb_float_val;
+      struct {
+        uint32_t sample_freq;
+        uint32_t mclk_freq;
+      }fixed;
 
-  uint32_t fb_sample_freq;
-  uint32_t fb_mclk_freq;
+//      struct {
+//
+//      }fifo_count;
+    }compute;
+
+  } feedback;
 
 #endif // CFG_TUD_AUDIO_ENABLE_FEEDBACK_EP
 #endif // CFG_TUD_AUDIO_ENABLE_EP_OUT
@@ -435,10 +443,6 @@ static inline uint8_t tu_desc_subtype(void const* desc)
 {
   return ((uint8_t const*) desc)[2];
 }
-#endif
-
-#if CFG_TUD_AUDIO_ENABLE_EP_OUT && CFG_TUD_AUDIO_ENABLE_FEEDBACK_EP
-bool tud_audio_n_fb_set(uint8_t func_id, uint32_t feedback);
 #endif
 
 bool tud_audio_n_mounted(uint8_t func_id)
@@ -1059,7 +1063,7 @@ static uint16_t audiod_encode_type_I_pcm(uint8_t rhport, audiod_function_t* audi
 #if CFG_TUD_AUDIO_ENABLE_EP_OUT && CFG_TUD_AUDIO_ENABLE_FEEDBACK_EP
 static inline bool audiod_fb_send(uint8_t rhport, audiod_function_t *audio)
 {
-  return usbd_edpt_xfer(rhport, audio->ep_fb, (uint8_t *) &audio->fb_val, 4);
+  return usbd_edpt_xfer(rhport, audio->ep_fb, (uint8_t *) &audio->feedback.value, 4);
 }
 #endif
 
@@ -1564,8 +1568,7 @@ static bool audiod_set_interface(uint8_t rhport, tusb_control_request_t const * 
 #if CFG_TUD_AUDIO_ENABLE_FEEDBACK_EP
     usbd_edpt_close(rhport, audio->ep_fb);
     audio->ep_fb = 0;
-    audio->fb_n_frames = 0;
-
+    tu_memclr(&audio->feedback, sizeof(audio->feedback));
 #endif
   }
 #endif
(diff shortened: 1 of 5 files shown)


src/class/audio/audio_device.c
```diff
@@ -2038,81 +2041,94 @@ bool audiod_xfer_cb(uint8_t rhport, uint8_t ep_addr, xfer_result_t result, uint3
...
+    TU_LOG1("  UAC2 feedback interval too small\r\n"); TU_BREAKPOINT(); return false;
   }
 
-
-#if CFG_TUD_AUDIO_ENABLE_FEEDBACK_DETERMINATION_MODE == CFG_TUD_AUDIO_ENABLE_FEEDBACK_DETERMINATION_MODE_POWER_OF_TWO_SHIFT
   // Check if parameters really allow for a power of two division
-  if ((f_m % f_s) != 0 || !tu_is_power_of_two(audio->fb_n_frames * f_m / f_s))
+  if ((f_m % f_s) == 0 && tu_is_power_of_two(f_m / f_s))
+  {
+    audio->fb_compute_method = FEEDBACK_COMPUTE_POWER_OF_2;
```
hathach, line 2069: power of 2 is auto detected

src/class/audio/audio_device.c
```diff
@@ -2038,81 +2041,94 @@ bool audiod_xfer_cb(uint8_t rhport, uint8_t ep_addr, xfer_result_t result, uint3
...
-#endif
+    case FEEDBACK_COMPUTE_FLOAT:
+      feedback = (uint32_t) ((float) cylces * audio->fb_float_val);
+    break;
 
-  audio->fb_val_min = ((TUSB_SPEED_FULL == tud_speed_get() ? (f_s/1000) : (f_s/8000)) - 1) << 16;   // Minimal value in 16.16 format for full speed (1ms per frame) or high speed (125 us per frame)
-  audio->fb_val_max = ((TUSB_SPEED_FULL == tud_speed_get() ? (f_s/1000) : (f_s/8000)) + 1) << 16;   // Maximum value in 16.16 format
+    case FEEDBACK_COMPUTE_FIXED:
+    {
+      uint64_t fb64 = (((uint64_t) cycles) * audio->fb_param_factor_N) << (16 - audio->fb_n_frames_shift);
```
hathach, line 2105: since n_frame is power of 2, I just store the sample & mclk freq and do the shift `(16 - (bInterval-1))`. The shift is fast enough, and we don't have to store 2x 64-bit. 

image

src/class/audio/audio_device.c
```diff
@@ -1695,16 +1686,11 @@ static bool audiod_set_interface(uint8_t rhport, tusb_control_request_t const *
           if (tu_edpt_dir(ep_addr) == TUSB_DIR_IN && desc_ep->bmAttributes.usage == 1)   // Check if usage is explicit data feedback
           {
             audio->ep_fb = ep_addr;
-            audio->fb_n_frames = desc_ep->bInterval;
```
PanRe, line 1698: Well spotted!

HiFiPhile:
Since the focus has changed from SOF to UAC feedback, I think we could add something to facilite feedback calculation based on FIFO count.

> It's rather a hardware question, let's take an example of 72MHz STM32F1 & 12.288MHz MCLK.

>Because MCLK frequency is relatively high to system clock, timer capture & compare shouldn't be used to measure MCLK frequency otherwise the obtained result will be horrible (72/6=12MHz ~ 72/5=14MHz).

>Only option is drive the timer with MCLK and count timer_cnt in sof_irq, in this case we have to "press the stop watch" precisely in the irq. DMA can't help as the transfer need to be triggered somehow, I don't aware any MCU which USB SOF can trigger a DMA transfer.

As I said earlier it's impossible to measure MCLK precisely with a MCU, also with @geekbozu's experimentation he also confirmed feedback value need to be tweaked with actual FIFO count.

What do you think ?

PanRe:
I think that would complete all possibilities to determine the feedback value! I think that would be awsome! Do you mean to add some option like `FEEDBACK_COMPUTE_FIFO_COUNT`, which does not need the SOF interrupt?

HiFiPhile:
> Do you mean to add some option like FEEDBACK_COMPUTE_FIFO_COUNT, which does not need the SOF interrupt?

Actually I'm not 100% sure if it should be based on SOF or not. All my tests was based on High-Speed which is easier to regulate the FIFO level, because frame interval is 125us instead of 1ms and less samples are arrived each frame.

It's better to test with FS to see if my method works.

HiFiPhile:
> @HiFiPhile do you have an example code on how to calculate feedback on fifo count work ? I will try to find a way to fit it as well or give some option to do so.

Full code is in https://github.com/HiFiPhile/i2s_bridge, UAC related code is in `audio_callbacks.c`

1. I define a feedback min/max of 0.1%, in UAC spec clock deviation is 0.05% for HS so 0.1% should cover it.
https://github.com/HiFiPhile/i2s_bridge/blob/1d6a24fdd0f70d150a88d85d63ed4b583b3b6b30/audio_callbacks.c#L101
Actual min/max value is updated each time sample frequency is changed: https://github.com/HiFiPhile/i2s_bridge/blob/1d6a24fdd0f70d150a88d85d63ed4b583b3b6b30/audio_callbacks.c#L179

2. Each time a packet is arrived I update a filtered FIFO count value with fixed-point math:
https://github.com/HiFiPhile/i2s_bridge/blob/1d6a24fdd0f70d150a88d85d63ed4b583b3b6b30/audio_callbacks.c#L210
Playback starts after FIFO is half full `buffer_ready == true` to avoid starving on startup.

3. Feedback value is updated in `tud_audio_fb_done_cb`
https://github.com/HiFiPhile/i2s_bridge/blob/1d6a24fdd0f70d150a88d85d63ed4b583b3b6b30/audio_callbacks.c#L237

hathach:
thanks @HiFiPhile for sharing your code, I currently switch (yes) to another works. Give me a few days, I will try to update this PR-to-PR to include your feedback-on-fifo-count method.

src/class/audio/audio_device.c
```diff
@@ -305,38 +305,22 @@ typedef struct
...
-  uint64_t fb_param_factor_D;                                                       // Denominator of feedback parameter coefficient
-#endif
-#endif // CFG_TUD_AUDIO_ENABLE_FEEDBACK_DETERMINATION_OPTION == CFG_TUD_AUDIO_ENABLE_FEEDBACK_DETERMINATION_OPTION_SOF_BY_AUDIO_DRIVER
 
-#if CFG_TUD_AUDIO_ENABLE_FEEDBACK_DETERMINATION_OPTION == CFG_TUD_AUDIO_ENABLE_FEEDBACK_DETERMINATION_OPTION_SOF_BY_USER
-  volatile uint32_t fb_val;                                                               // Feedback value for asynchronous mode (in 16.16 format).
-  uint8_t fb_n_frames;                                                                    // Number of (micro)frames used to estimate feedback value
-#endif
+  uint32_t fb_sample_freq;
+  uint32_t fb_mclk_freq;
```
hathach, line 323: @PanRe renamed

src/class/audio/audio_device.h
```diff
@@ -493,68 +458,83 @@ TU_ATTR_WEAK bool tud_audio_rx_done_post_read_cb(uint8_t rhport, uint16_t n_byte
...
+// TU_ATTR_WEAK void tud_audio_feedback_params_cb(uint8_t func_id, uint8_t alt_itf, audio_feedback_params_t* feedback_param);
+
+// mclk_freq   : Main clock frequency in Hz i.e. master clock to which sample clock is locked
+// sample_freq : sample frequency in Hz
+// fixed_point : 0 float (default), 1 fixed point (for mcu without FPU)
+TU_ATTR_WEAK void tud_audio_feedback_params_cb(uint8_t func_id, uint8_t alt_itf, uint32_t* sample_freq, uint32_t* mclk_freq, uint8_t* fixed_point);
+
+// Callback in ISR context, invoked periodically according to feedback endpoint bInterval.
+// Could be used to compute and update feedback value
+TU_ATTR_WEAK void tud_audio_feedback_interval_isr(uint8_t func_id, uint32_t frame_number, uint8_t interval_log2);
```
hathach, line 536: @PanRe renamed and also add interval_log2 (or shift), just in case user want to use it for their own calculation.

src/class/audio/audio_device.h
```diff
@@ -493,68 +458,83 @@ TU_ATTR_WEAK bool tud_audio_rx_done_post_read_cb(uint8_t rhport, uint16_t n_byte
...
+    struct {
+        uint32_t nominal;
+        uint32_t threshold;
+        // variation etc ..
+    }fifo_count;
+  };
+}audio_feedback_params_t;
+
+
+// TU_ATTR_WEAK void tud_audio_feedback_params_cb(uint8_t func_id, uint8_t alt_itf, audio_feedback_params_t* feedback_param);
```
hathach, line 527: @HiFiPhile I made a draft idea on how to add the fifo count method, the params_cb with struct for input. Fifo count method seems to reply on a few params: std (nominal) value, variation, and buffer threshold. The buffer threshold is a bit tricky, since it requires application to respect this. The rest can be computed within the stack. Let me know if I miss something before adding the actual code.

HiFiPhile: I like the idea of union.

> The buffer threshold is a bit tricky, since it requires application to respect this.
Yes indeed. It can be calculated in the stack because normally it's N*PacketSize, or add a macro for easier threshold calculation in application.

https://github.com/HiFiPhile/i2s_bridge/blob/1d6a24fdd0f70d150a88d85d63ed4b583b3b6b30/audio_callbacks.c#L186
Here 8000 is for HS (125us), FS need 1000.

hathach: @HiFiPhile I have a silly question, `fifo_count_flt` stands for float in 16.16 format right ? 

https://github.com/HiFiPhile/i2s_bridge/blob/1d6a24fdd0f70d150a88d85d63ed4b583b3b6b30/audio_callbacks.c#L113

HiFiPhile: It's "filtered" fifo count. As fifo count is jumpy each time a packet is arrived / consumed, it's hard to regulate the fifo level based on raw count. So I make an average of fifo count to get a stabilized value.
Basically fifo_count_flt = `fifo_count_old * 255 + fifo_count`, I shifted the whole value <<8 to let the result has enough precision so float is not required.

hathach: thanks @HiFiPhile , I kind of getting the idea, filtered ~ average here

```
filtered = 255 * old + raw_count
feedback = nominal - (filtered/256 - threshold)
```

out of curiosity, how would you come up with the filtered/avg formula and how does it depend on sample rate, FS/HS link. Sorry for more question, I am trying to figure out the big picture and normalize the argument for this compute method.

HiFiPhile: > out of curiosity, how would you come up with the filtered/avg formula and how does it depend on sample rate, FS/HS link. 

I used J-Scope to watch FIFO level, it's kind of trial and error :)

image

FIFO is filled by USB packets and drained by I2S.

For the **same** sample rate, since HS packets arrive 8x faster every packet is 8x smaller, FIFO level variation is much smaller so easier to filter.

The higher the sample rate the higher the variation, in theory a higher filter coefficient is needed to achieve a same smoothed result.
But since `feedback_std` is also proportional to sample rate, which cancelled the need of a higher filter coefficient.
So no need to change filter coefficient on sample rate change.

hathach: thanks for the explanation, unfortunately, I still pretty much have no ideas how to generalize this into generic code. Apparently, there is no formula to fit all, I am just hoping to make some skeleton and allow user to simply tweaking a parameter or two with actual hardware. However, this is really out of my knowledge, and don't actually have any real application on hardware to test these out. 

Since this has been pending for quite some time, I would like to merge this as it is and come back later to this later on. The good news is we have an decent API with place holders union for fifo count method (or any new method we could find afterwards). Should you have time feel free to make an PR to enhance this :)

Protoxy22:
> > @HiFiPhile do you have an example code on how to calculate feedback on fifo count work ? I will try to find a way to fit it as well or give some option to do so.
> 
> Full code is in https://github.com/HiFiPhile/i2s_bridge, UAC related code is in `audio_callbacks.c`
> 
>     1. I defined a feedback min/max of 0.1%, in USB HS spec clock deviation is 0.05% so 0.1% should be good if same type of oscillator is used for MCLK.
>        https://github.com/HiFiPhile/i2s_bridge/blob/1d6a24fdd0f70d150a88d85d63ed4b583b3b6b30/audio_callbacks.c#L101
>        Actual min/max value is updated each time sample frequency is changed: https://github.com/HiFiPhile/i2s_bridge/blob/1d6a24fdd0f70d150a88d85d63ed4b583b3b6b30/audio_callbacks.c#L179
> 
>     2. Each time a packet is arrived I update a filtered FIFO count value with fixed-point math:
>        https://github.com/HiFiPhile/i2s_bridge/blob/1d6a24fdd0f70d150a88d85d63ed4b583b3b6b30/audio_callbacks.c#L210
>        Playback starts after FIFO is half full `buffer_ready == true` to avoid starving on startup.
> 
>     3. Feedback value is updated in `tud_audio_fb_done_cb`
>        https://github.com/HiFiPhile/i2s_bridge/blob/1d6a24fdd0f70d150a88d85d63ed4b583b3b6b30/audio_callbacks.c#L237
> 
> 
> My `0.1%` magic number maybe need more proof, only thing close is in UAC spec:
> 
> ```
> In the case of an adaptive isochronous data endpoint that support only a discrete number o
> sampling frequencies, the endpoint must at least tolerate ±1000 PPM inaccuracy on the reported
> Sampling Frequency Control values to accommodate sample clock inaccuracies.  
> ```

Hi @HiFiPhile, I would be really interested to see your i2s_bridge repository for the fifo count feedback method.
Is there a way I can access to it ? 
Thanks

HiFiPhile:
@Protoxy22 just sent you a invite:)

Protoxy22:
> @Protoxy22 just sent you a invite:)

Thank you !

PR #1467, opened 2022-05-20
Minor cleanup of RP2040 code post addition of Pico-PIO-USB

kilograham:
* Removed some compiler warnings, and cleaned out no-longer-necessary warning suppression from CMake `suppress_tinyusb_warnings()`
* Made explicit `family_configure_dual_usb_example()` for DUAL mode examples as `family_configure_target()` may not generally be called multiple times for the same target
* Renamed library `pico_pio_usb` to `tinyusb_pico_pio_usb` to be clearer and avoid conflict if someone already has a `pico_pio_usb` in their project
* Added `family_add_pico_pio_usb()` method for adding Pico-PIO-USB support to an existing example
* Added some explicit guards to check the Pico-PIO-USB submodule is initialized
* Removed explicit setting of a system clock for Pico-PIO-USB in the rp2040 board support unless using Pico-PIO-USB
* Allowed `tinyusb_pico_pio_usb` to be added as a dependency of regular apps using the Pico SDK

### DIFF
--- a/examples/device/net_lwip_webserver/src/lwipopts.h
+++ b/examples/device/net_lwip_webserver/src/lwipopts.h
@@ -49,7 +49,9 @@
 
 #define TCP_MSS                         (1500 /*mtu*/ - 20 /*iphdr*/ - 20 /*tcphhr*/)
 #define TCP_SND_BUF                     (2 * TCP_MSS)
+#ifndef TCP_WND
 #define TCP_WND                         (TCP_MSS)
+#endif
 
 #define ETHARP_SUPPORT_STATIC_ENTRIES   1
 
--- a/src/portable/raspberrypi/rp2040/hcd_rp2040.c
+++ b/src/portable/raspberrypi/rp2040/hcd_rp2040.c
@@ -537,7 +537,11 @@ bool hcd_setup_send(uint8_t rhport, uint8_t dev_addr, uint8_t const setup_packet
     (void) rhport;
 
     // Copy data into setup packet buffer
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Warray-bounds"
+#pragma GCC diagnostic ignored "-Wstringop-overflow"
     memcpy((void*)&usbh_dpram->setup_packet[0], setup_packet, 8);
+#pragma GCC diagnostic pop
 
     // Configure EP0 struct with setup info for the trans complete
     struct hw_endpoint *ep = _hw_endpoint_allocate(0);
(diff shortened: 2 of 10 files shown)


examples/device/net_lwip_webserver/src/lwipopts.h
```diff
@@ -49,7 +49,9 @@
 
 #define TCP_MSS                         (1500 /*mtu*/ - 20 /*iphdr*/ - 20 /*tcphhr*/)
 #define TCP_SND_BUF                     (2 * TCP_MSS)
+#ifndef TCP_WND
```
kilograham, line 52: Note I should probably comment on this... there are pre-existing overrides of this to `2*TCP_MSS` in the CMakeLists.txt for the net_lwip_webserver example. This was causing a compiler warning, as this line just trashes it. So either my change is correct, or the override is unnecessary!

hathach: The defined in cmakelist.txt is added in this commit
https://github.com/hathach/tinyusb/commit/95f2478146ef83d18b52408c0692222bf2a15053#diff-12cfd80973cd9c09e8493c08a1eb623365231a9e2b0db14d109e9b6836985830R71-R74

```
PBUF_POOL_SIZE=2
HTTPD_USE_CUSTOM_FSDATA=0
```

are also redefined with the same values, maybe we should drop it altogether in cmakelist and adjust/use value in the header file only for consistency

src/portable/raspberrypi/rp2040/hcd_rp2040.c
```diff
@@ -537,7 +537,11 @@ bool hcd_setup_send(uint8_t rhport, uint8_t dev_addr, uint8_t const setup_packet
     (void) rhport;
 
     // Copy data into setup packet buffer
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Warray-bounds"
+#pragma GCC diagnostic ignored "-Wstringop-overflow"
```
hathach, line 542: this warnings from gcc11 is bit annoying, hopefully we could find an better way to go around this.

hathach (commented):
everything looks great, except, maybe we should remove re-define of network example in cmakelist

kilograham:
> everything looks great, except, maybe we should remove re-define of network example in cmakelist. If you agree, I could merge this first, and make an changes as separated PR afterwards.

Sure. i wasn't sure why why was overridden in the CMakeLists.txt to start with, so didn't want to change that to remove the error. I 100% defer to your better knowledge of what it does, and if it is safe for all boards to change (or if it is even needed)

hathach:
> > everything looks great, except, maybe we should remove re-define of network example in cmakelist. If you agree, I could merge this first, and make an changes as separated PR afterwards.
> 
> Sure. i wasn't sure why why was overridden in the CMakeLists.txt to start with, so didn't want to change that to remove the error. I 100% defer to your better knowledge of what it does, and if it is safe for all boards to change (or if it is even needed)

ok thanks, I will do that later on with a bit of testing

PR #1473, opened 2022-05-24
fix links in docs

sknebel:
changed markdown formatting to correct rst formatting in rst file


hathach (approved):
thank you for the fix

PR #1480, opened 2022-05-29
Host: Add support for multi-level usb hubs

Ryzee119:
**Describe the PR**
* Add support for multiple levels of USB hubs in the usb host stack. For your consideration Hathach 😄 

To get this working reliably I had to relax the assert on failed port_status control transfer. Currently with multiple hubs it may perform a control xfer if the ctrl buffer is busy. 

**Additional context**
1. Developed on imxrt (Teensy41) with EHCI.
2. Tested with 3 USB hubs levels, combination of low speed and full speed devices. (HID devices) Not tested with Bulk or Iso but this is not expected to be different from upstream.
3. Quite susceptible to 5V rail supply issues. So an externally powered usb hub is recommended.

**Known issues**
1. If you unplug a downstream hub, it will hit this assert which makes sense and can easily be fixed by relaxing this assert to a TU_VERIFY, but not included in this PR. I think this assert is quite strict as is, but will leave for feedback.
https://github.com/hathach/tinyusb/blob/f4efb51fe2b7f4fee38c28ff80d03904d0b5b73e/src/host/hub.c#L324 

May help to review on a per commit basis.

Below is a working log of 2 daisy chained usb hubs, with 3 usb drives connected. One drive in the first hub, 2 in the 2nd hub.
After enumeration of all 5 devices, I unplug the root hub so you can see all downstream devices disconnected successfully.
[multihub-log.txt](https://github.com/hathach/tinyusb/files/8792675/multihub-log.txt)

Thanks for this great USB stack 🚀

### DIFF
--- a/src/host/hcd.h
+++ b/src/host/hcd.h
@@ -204,13 +204,11 @@ void hcd_event_xfer_complete(uint8_t dev_addr, uint8_t ep_addr, uint32_t xferred
     .rhport   = 0, // TODO correct rhport
     .event_id = HCD_EVENT_XFER_COMPLETE,
     .dev_addr = dev_addr,
-    .xfer_complete =
-    {
-      .ep_addr = ep_addr,
-      .result  = result,
-      .len     = xferred_bytes
-    }
   };
+  event.xfer_complete.ep_addr = ep_addr;
+  event.xfer_complete.result = result;
+  event.xfer_complete.len = xferred_bytes;
+
 
   hcd_event_handler(&event, in_isr);
 }
--- a/src/host/usbh.c
+++ b/src/host/usbh.c
@@ -1089,6 +1089,12 @@ uint8_t tuh_descriptor_get_serial_string_sync(uint8_t daddr, uint16_t language_i
 //
 //--------------------------------------------------------------------+
 
+TU_ATTR_ALWAYS_INLINE
+static inline bool is_hub_addr(uint8_t daddr)
+{
+  return (CFG_TUH_HUB > 0) && (daddr > CFG_TUH_DEVICE_MAX);
+}
+
 // a device unplugged from rhport:hub_addr:hub_port
 static void process_device_unplugged(uint8_t rhport, uint8_t hub_addr, uint8_t hub_port)
 {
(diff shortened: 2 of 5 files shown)


src/host/hcd.h
```diff
@@ -204,13 +204,11 @@ void hcd_event_xfer_complete(uint8_t dev_addr, uint8_t ep_addr, uint32_t xferred
...
-    {
-      .ep_addr = ep_addr,
-      .result  = result,
-      .len     = xferred_bytes
-    }
   };
+  event.xfer_complete.ep_addr = ep_addr;
+  event.xfer_complete.result = result;
+  event.xfer_complete.len = xferred_bytes;
+
```
perigoso, line 211: did this not initialize correctly before?

Ryzee119: I was getting `hcd.h:213:3: sorry, unimplemented: non-trivial designated initializers not supported` and failing on error.

perigoso: ah, oh well lgtm

src/host/usbh.c
```diff
@@ -1507,13 +1513,6 @@ static bool _parse_configuration_descriptor(uint8_t dev_addr, tusb_desc_configur
...
     TU_ASSERT(drv_len >= sizeof(tusb_desc_interface_t));
 
-    if (desc_itf->bInterfaceClass == TUSB_CLASS_HUB && dev->hub_addr != 0)
-    {
-      // TODO Attach hub to Hub is not currently supported
-      // skip this interface
-      TU_LOG(USBH_DBG_LVL, "Only 1 level of HUB is supported\r\n");
-    }
-    else
     {
```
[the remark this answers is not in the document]
Ryzee119: Fixed, thanks

Ryzee119:
I can hit issues if we disconnect a device from a hub when we are enumerating another device, the hub status control transfer will begin, then we'll assert in enumeration due to a busy control buffer.

I think there's a few conflicts with hub, enumerating and control buffers to be resolved still for this to be very resilient.

src/host/hcd.h
hathach: just want to chime in, the warning could be suppressed if we initialize all member by their order.

Ryzee119: Thats true. I believe for the most part it is in order, just missing the connection struct in hcd_event_t.
So we could add this the mix
```
.connection =
{
    .hub_addr=0,
    .hub_port=0,
    .speed=0
}
```
I can push this change or whatever you think is best

hathach: no, I think your changes is better, we should only init the one of unioned only.

hathach (approved):
Superb ! thank you very much for your effort and patience. I have tested with mcb1800 (same chipidea controller as imxrt) and it works like a charm. The recursive call to process unplugged hub device is working but not a great approach in long term IMHO. I make an commit to add FIXME to this and also make an issue for reminder as well. Thanks again for your work.

PS: finally I could hide the fact that I am lazy to support this :)
PSS: https://github.com/hathach/tinyusb/issues/1511 is created

PR #1485, opened 2022-06-01
Fix compile error

tfx2001:
**Describe the PR**
Fix compile error for `osal_rtthread.h`.


hathach (approved):
thank you for the fix

PR #1487, opened 2022-06-02
cdc: Fix autoflush for FIFO < MPS

tore-espressif:
**Describe the PR**
If CDC TX buffer size is configured to be less than bulk packet size, the autoflushing condition is never reached.

Changes:
1. TX buff is automatically flushed when TX FIFO is full

**Additional context**
Found out during https://github.com/espressif/esp-idf/issues/9040

### DIFF
--- a/src/class/cdc/cdc_device.c
+++ b/src/class/cdc/cdc_device.c
@@ -171,7 +171,7 @@ uint32_t tud_cdc_n_write(uint8_t itf, void const* buffer, uint32_t bufsize)
   uint16_t ret = tu_fifo_write_n(&p_cdc->tx_ff, buffer, (uint16_t) bufsize);
 
   // flush if queue more than packet size
-  if ( tu_fifo_count(&p_cdc->tx_ff) >= BULK_PACKET_SIZE )
+  if ( (tu_fifo_count(&p_cdc->tx_ff) >= BULK_PACKET_SIZE) || ((CFG_TUD_CDC_TX_BUFSIZE < BULK_PACKET_SIZE) && tu_fifo_full(&p_cdc->tx_ff)) )
   {
     tud_cdc_n_write_flush(itf);
   }


tore-espressif:
This is in great extend reverting this https://github.com/hathach/tinyusb/pull/547

src/class/cdc/cdc_device.c
```diff
@@ -171,7 +166,7 @@ uint32_t tud_cdc_n_write(uint8_t itf, void const* buffer, uint32_t bufsize)
   uint16_t ret = tu_fifo_write_n(&p_cdc->tx_ff, buffer, bufsize);
 
   // flush if queue more than packet size
-  if ( tu_fifo_count(&p_cdc->tx_ff) >= BULK_PACKET_SIZE )
+  if ( (tu_fifo_count(&p_cdc->tx_ff) >= CFG_TUD_CDC_EP_BUFSIZE) || tu_fifo_full(&p_cdc->tx_ff) )
```
perigoso, line 169: shouldn't the packet size check be kept? the buffer may be larger than the packet size, and in that case, we would want to flush when we have enough to fill said packet

tore-espressif: The buffer size is defined here https://github.com/hathach/tinyusb/blob/master/src/class/cdc/cdc_device.h#L42 , so defining `BULK_PACKET_SIZE` seems as a redefinition...

But maybe I'm just missing reasons for change inhttps://github.com/hathach/tinyusb/pull/547 , I can change it back if you will :)

perigoso: sure, it's redefined, but I'm actually talking about the logic that triggers a flush, you just using the bufsize, but I'm arguing both the bufsize and the packet size should be considered

perigoso: this goes for the zero length packet too, we should see if we sent a full packet and there is no more to send, not if we sent a CFG_TUD_CDC_EP_BUFSIZE

perigoso (approved):
LGTM

tore-espressif:
@perigoso @hathach any update about this?

src/class/cdc/cdc_device.c
```diff
@@ -171,7 +171,7 @@ uint32_t tud_cdc_n_write(uint8_t itf, void const* buffer, uint32_t bufsize)
   uint16_t ret = tu_fifo_write_n(&p_cdc->tx_ff, buffer, bufsize);
 
   // flush if queue more than packet size
-  if ( tu_fifo_count(&p_cdc->tx_ff) >= BULK_PACKET_SIZE )
+  if ( (tu_fifo_count(&p_cdc->tx_ff) >= BULK_PACKET_SIZE) || tu_fifo_full(&p_cdc->tx_ff) )
```
hathach, line 174: would you mind changing the code to 

```C
  if ( (tu_fifo_count(&p_cdc->tx_ff) >= BULK_PACKET_SIZE) || ((CFG_TUD_CDC_TX_BUFSIZE < BULK_PACKET_SIZE) && tu_fifo_full(&p_cdc->tx_ff) )
```

hathach (changes requested):
sorry for late response, I was too busy with other works. The reason for change request: although it has more code, the additional condition `CFG_TUD_CDC_TX_BUFSIZE < BULK_PACKET_SIZE` is constant expression and will help the compiler to optimize it out. 

Therefore we don't have to run tu_fifo_full() when bufsize is larger than the packet size (and const expression also not executed at all in other case)

src/class/cdc/cdc_device.c
tore-espressif: Thank you for the review. Updated!

hathach (approved):
thank for the update

Issue #662, opened 2021-02-19
Use LDREX/STREX on ARM cortex


hathach:
**Is your feature request related to a problem? Please describe.**
LDREX/STREX instructions are available on ARM cortex  M3/M4/M7 which can be used for mutex without the need to disable/enable interrupt. 

Reference https://github.com/adafruit/Adafruit_nRF52_Arduino/issues/562#issuecomment-681583256

PR #1489, opened 2022-06-02
nrf5x: Fix DMA access

### LINKED ISSUE #1474: RNF52/NRF53 incorrect size DMA transfers due to hardware undocumented behavior
### Operating System

Linux

### Board

pca10095 or pca10056 or other NRF52/NRF53

### Firmware

Latest master can be 0.13.0.
mynewt/nimble build with BTH and MSC enabled

### What happened ?

When DMA is started for one OUT endpoint values read from  `SIZE.EPOUT[n]` for other endpoint are not valid.
This behavior is not documented (yet) by Nordic but it can be observed on **NRF52/NRF53**.
When two packets arrive on two different endpoints single interrupt will detect and try to handle both.
It calls function `xact_out_dma()` for both endpoints one after the other.
For the first endpoint (one with lower number) DMA will start transferring data from endpoint to RAM.
Second call `xact_out_dma()` for next ready endpoint will not start DMA right away since it's already started, it will call `usbd_defer_func()` with task start register.
This looks like correct sequence however there is problem in following code inside `xact_out_dma()`
```c
    // Following line may compute incorrect value !!!!!!!!!!!!!
    xact_len = tu_min16((uint16_t) NRF_USBD->SIZE.EPOUT[epnum], xfer->total_len - xfer->actual_len);
    // Trigger DMA move data from Endpoint -> SRAM
    NRF_USBD->EPOUT[epnum].PTR = (uint32_t) xfer->buffer;
    NRF_USBD->EPOUT[epnum].MAXCNT = xact_len; // <- Possible wrong value

    edpt_dma_start(&NRF_USBD->TASKS_STARTEPOUT[epnum]);
```
first line can incorrectly computes `xact_len` while DMA is ongoing.
Read from `NRF_USBD->SIZE.EPOUT[epnum]` yields incorrect values.
This incorrect value is then set to `NRF_USBD->EPOUT[epnum].MAXCNT` and later DMA will copy less bytes that it should and resulting in broken protocols.

Following code could be used to trap error
```c
// helper to start DMA
static void start_dma(volatile uint32_t* reg_startep)
{
  for (int i = 0; i < 8; ++i) stored_size_epout1[i] = NRF_USBD->SIZE.EPOUT[i];
  (*reg_startep) = 1;
  __ISB(); __DSB();
  for (int i = 0; i < 8; ++i) stored_size_epout2[i] = NRF_USBD->SIZE.EPOUT[i];

  if (memcmp(stored_size_epout1, stored_size_epout2, 8))
    TU_BREAKPOINT();
```

### How to reproduce ?

I think it maybe it could be reproduce in cdc_msc modified sample (at least two OUT endpoint are needed) and some massive transfer on both endpoints.

### Debug Log as txt file

_No response_

### Screenshots

_No response_

kasjer:
**Describe the PR**
There were two problems:
- dma_running flag could be checked in USB interrupt (not set yet) then higher priority
  interrupt could start transfer, check dma_running (not set yet) set it to true start
  DMA; then when USB interrupt continues it starts another DMA that is not allowed
- when DMA is started some registers can't be safely accessed, read can yield invalid
  values (SIZE.EPOUT, SIZE.EPISO)
  current implementation could start DMA for one OUT endpoint then check that another
  endpoint also has data and while DMA was not started right away, SIZE.EPOUT was copied
  already to MAXCNT register. Later on when DMA was started not all data was read from
  endpoint due to incorrect DMA size previously set.

To prevent both cases dma_running is changed in atomic way.
Only code that actually set this value to true starts DMA, code that tried and
had dma_running flag already set simply defers DMA start to USB task.
This eliminates also need to have mutex that was there to limit access to dma_running flag
to one task only.
transfer also now has started flag that is set only after dcd_edpt_xfer() sets up total_len
and actua_len. Previously USB interrupt was disabled when total_len and actual_len were
setup to prevent race condition when data arrived to ednpoint before transfer was setup
was finished.

**Additional context**
This make obsolete #1407 (same race condition is fixed here in different way)
It also fixes #1474 (discussion about SIZE.EPOUT on Nordic DevZone https://devzone.nordicsemi.com/f/nordic-q-a/88301/nrf5340-usb-size-epout-n-value-not-stable)

### DIFF
--- a/src/portable/nordic/nrf5x/dcd_nrf5x.c
+++ b/src/portable/nordic/nrf5x/dcd_nrf5x.c
@@ -28,6 +28,7 @@
 
 #if CFG_TUD_ENABLED && CFG_TUSB_MCU == OPT_MCU_NRF5X
 
+#include <stdatomic.h>
 #include "nrf.h"
 #include "nrf_clock.h"
 #include "nrf_power.h"
@@ -72,15 +73,13 @@ typedef struct
   // nRF will auto accept OUT packet after DMA is done
   // indicate packet is already ACK
   volatile bool data_received;
+  volatile bool started;
 
   // Set to true when data was transferred from RAM to ISO IN output buffer.
   // New data can be put in ISO IN output buffer after SOF.
   bool iso_in_transfer_ready;
 
 } xfer_td_t;
-
-static osal_mutex_def_t dcd_mutex_def;
-static osal_mutex_t dcd_mutex;
 
 // Data for managing dcd
 static struct
@@ -90,7 +89,7 @@ static struct
   xfer_td_t xfer[EP_CBI_COUNT + 1][2];
 
   // nRF can only carry one DMA at a time, this is used to guard the access to EasyDMA
-  volatile bool dma_running;
+  atomic_bool dma_running;
 }_dcd;
 
 /*------------------------------------------------------------------*/
@@ -121,8 +120,6 @@ TU_ATTR_ALWAYS_INLINE static inline bool is_in_isr(void)
 // helper to start DMA
 static void start_dma(volatile uint32_t* reg_startep)
 {
-  _dcd.dma_running = true;
-
   (*reg_startep) = 1;
   __ISB(); __DSB();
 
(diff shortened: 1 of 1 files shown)


sjanc:
Hi,

Any update on this?  We are using this in product and test it extensively.
In long term we would like to avoid having to use our own fork on tinyusb..

sjanc:
I just wanted to confirm that this patch fixes all the USB issues we've been seeing when stress-testing USB on nRF5340. 
It would be really great if this fix is merged into upstream repo.

tannewt:
Could this bug cause data written out of usb to be read in? That's been a longstanding issue in CircuitPython on the nRF52840.

pigrew:
@hathach , I see this is using the C11 atomics. I was thinking about using them in the STM32 driver, too, but was not sure if TinyUSB can require that new of a compiler?

kasjer:
> Could this bug cause data written out of usb to be read in? That's been a longstanding issue in CircuitPython on the nRF52840.

One of the issue this fixes is that DMA is reported as finished while number of bytes received is less than what was requested (including 0) in that case if same buffer was used for write and read it could manifest like echo I guess. I think that at one point I also had an impression about mismatch between what was send in both directions but it turned out that buffer was not filled even though DMA finished.
So I guess this could be the fix for what you saw in CircuitPython

Nikitarc:
TinyUSB already requires gcc C11 to get rid of warnings about unnamed structs/unions when [-Wpedantic] is used (ISO C99 doesn't support unnamed structs/unions).
Why not use other features of C11?

hathach (approved):
Brilliant !!! Thank you very much for the fix and great explanation. I am sorry for the super late response, I was too off schedule lately, and still catching-up with PRs/Issues.

The fix is spot-on, I have tried the blehci via usb, before this PR it will crash randomly, but got stable now. I remembered trying to use atomic bool previously, but got into a worse racing condition, probably due to missing the 2nd case where SIZE.EPOUT cannot be read after DMA started. Thanks again.

Hopefully this will fix the nrf DMA race condition once and for all (there has been several PRs for this race alone).

hathach:
> Could this bug cause data written out of usb to be read in? That's been a longstanding issue in CircuitPython on the nRF52840.

Yeah, I think this will possibly fix the hiccup issue with nrf cdc fast transferring. 

> @hathach , I see this is using the C11 atomics. I was thinking about using them in the STM32 driver, too, but was not sure if TinyUSB can require that new of a compiler?

I actually really like c11, we already in 22' already. C11 is much better and safer than C99, one of the major C version where it is under lots of improvement for a very long time (since C99). So yeah, please go ahead with your changes. I also made an issue regarding LDREX/STREX on M3/M4 #662. Note since M0 does not support these instruction, atomic operation on M0 will probably is generated to __disable_irq() (I could be wrong, but please check the dissambly to be sure).

PR #1509, opened 2022-06-14
RP2040 HCD Improvements (Hub + Keyboard + Mouse working)

liamfraser:
Improvements to RP2040 HCD driver. The main issue this PR fixes is that interrupt endpoints were sometimes sending packets to the tinyusb host stack that the stack had not requested. This would result in hangs / crashes / corrupted data structures. 

We now disable interrupt endpoint polling unless the host stack has requested a packet. The other improvements were things noticed while debugging this issue. In particular you should check for a stall received before dealing with any other pending buffers.

In practical terms, this means you can use a Raspberry Pi keyboard with hub, plus a Raspberry Pi mouse in the host_cdc_msc_hid demo without problems. Hot plugging also works.

### DIFF
--- a/src/portable/raspberrypi/rp2040/hcd_rp2040.c
+++ b/src/portable/raspberrypi/rp2040/hcd_rp2040.c
@@ -104,6 +104,8 @@ static void __no_inline_not_in_flash_func(hw_xfer_complete)(struct hw_endpoint *
 static void __no_inline_not_in_flash_func(_handle_buff_status_bit)(uint bit, struct hw_endpoint *ep)
 {
     usb_hw_clear->buf_status = bit;
+    // EP may have been stalled?
+    assert(ep->active);
     bool done = hw_endpoint_xfer_continue(ep);
     if (done)
     {
@@ -166,6 +168,8 @@ static void __no_inline_not_in_flash_func(hw_trans_complete)(void)
     pico_trace("Sent setup packet\n");
     struct hw_endpoint *ep = &epx;
     assert(ep->active);
+    // Set transferred length to 8 for a setup packet
+    ep->xferred_len = 8;
     hw_xfer_complete(ep, XFER_RESULT_SUCCESS);
   }
   else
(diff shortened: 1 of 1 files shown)


liamfraser:
cc @kilograham

src/portable/raspberrypi/rp2040/hcd_rp2040.c
```diff
@@ -98,12 +98,21 @@ static void __no_inline_not_in_flash_func(hw_xfer_complete)(struct hw_endpoint *
     uint8_t ep_addr = ep->ep_addr;
     uint xferred_len = ep->xferred_len;
     hw_endpoint_reset_transfer(ep);
+
+    // If this is an interrupt endpoint disable it until host stack requests another
+    // transfer
+    if (ep->transfer_type == TUSB_XFER_INTERRUPT) {
+        usb_hw_clear->int_ep_ctrl = (1 << (ep->interrupt_num + 1));
```
hathach, line 105: I have an question though, according to the rp2040 DS, if the buffer is `not full and available`, there won't be any transfer. This seems to indicate we have an un-scheduled transfer occurred. 

image

src/portable/raspberrypi/rp2040/hcd_rp2040.c
```diff
@@ -166,6 +175,8 @@ static void __no_inline_not_in_flash_func(hw_trans_complete)(void)
     pico_trace("Sent setup packet\n");
     struct hw_endpoint *ep = &epx;
     assert(ep->active);
+    // Set transferred length to 8 for a setup packet
+    ep->xferred_len = 8;
```
hathach, line 179: ah right, nice fix ! we totally skip the xferred_len for setup

hathach (commented):
thank you for the fix, I still have question regarding unscheduled usb transfer on interrupt endpoint. Could you tell more on your setup and how to reproduce the issue.

src/portable/raspberrypi/rp2040/hcd_rp2040.c
liamfraser: Good point. I implemented this a while ago and rebased it, so maybe it is no longer necessary. In a comment here https://github.com/liamfraser/tinyusb/commit/a38b54d1f2d2b3e756eea475f7114f888ad960fc#commitcomment-68710466 you mention there was an issue with queueing multiple transfers that is now fixed. I remember seeing a race condition at the time so that could have fixed it.

hathach: @liamfraser would you mind re-testing it with this specific command along its enabled counterpart when starting new transfer to see if the issue still occurs.

hathach: Ah I remembered that bug, it occurs when plugging hub with already-attached multiple devices to rp2040 host and cause multiple control conflicts. It should be resolved, maybe we should revert this change ?

liamfraser: I will test it now without the interrupt endpoint change

liamfraser:
@hathach you are right the change is unnecessary as it was fixed correctly in the host stack. I would still like the other tweaks to be merged but not 100% necessary. Just helps if you are debugging any future issues.

hathach (approved):
look great thank you for the fix.

hathach:
> @hathach you are right the change is unnecessary as it was fixed correctly in the host stack. I would still like the other tweaks to be merged but not 100% necessary. Just helps if you are debugging any future issues.

sure thing

PR #1512, opened 2022-06-16
rp2040: make moving 1K of code into RAM optional - and off by default

kilograham:
Made the changes in commit dd035b0eb optional because stealing 1K of RAM by default seems undesirable.

It is now controlled by PICO_RP2040_USB_FAST_IRQ which is 0 by default.

It is probably worth documenting what use cases the faster IRQ helps/is-required-for.


kilograham:
cc @liamfraser

PR #1513, opened 2022-06-16
add tud_queue_send_cb

kilograham:
In the Pico SDK for stdio over USB, (by default) we call TinyUSB in a low priority IRQ (so the user does not have to have a polling loop).

Currently this IRQ is called every 1ms which is pretty wasteful. This change adds a callback which can be hooked to know when there is something in the queue that needs processing.

Note, i don't know if:

1) i called this the right thing
2) whether it should be defined somewhere else (if other ports might find it useful)
3) it performs the callback for all queues. This was just to avoid having to put a call after every queue_write (and based on the fact that no other use of osal_queue is made in the code base!

### DIFF
--- a/src/osal/osal_pico.h
+++ b/src/osal/osal_pico.h
@@ -155,19 +155,24 @@ TU_ATTR_ALWAYS_INLINE static inline bool osal_queue_receive(osal_queue_t qhdl, v
   return success;
 }
 
+#define TUD_QUEUE_SEND_CB_SUPPORTED 1
+
+TU_ATTR_WEAK void tud_queue_send_cb(bool in_isr);
+
 TU_ATTR_ALWAYS_INLINE static inline bool osal_queue_send(osal_queue_t qhdl, void const * data, bool in_isr)
 {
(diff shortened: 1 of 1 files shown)


hathach (commented):
I don't think this is the right way to address the issue with pico stdio. There is other place that use queue API as well. I think the better approach would be having USB IRQ as shared IRQ (currently exclusive) and pico-sdk installed its own irq to detect if there is a new usb event and call tud_task() later in main thread

https://github.com/hathach/tinyusb/blob/80121303eb80a1847ad314cb87774e6e390754c6/src/portable/raspberrypi/rp2040/dcd_rp2040.c#L366

I have done something similar to this with circuitpython, though by removing the exclusive isr, then re-install and manually invoke dcd_int_handler()

https://github.com/adafruit/circuitpython/blob/main/ports/raspberrypi/supervisor/usb.c#L41

kilograham:
yeah, that's a good idea

kilograham:
I will open a separate request to change to use irq_add_shared_handler

Issue #1446, opened 2022-04-25
Typo in audio_device.c line 539


maddyaby:
### Operating System

Others

### Board

Feather M4 Express

### Firmware

Custom firmware using the audio functions to make a 2-channel speaker with TYPE_I_PCM decoding

### What happened ?

Code will not build with CFG_TUD_AUDIO_ENABLE_DECODING && CFG_TUD_AUDIO_ENABLE_EP_OUT if the corresponding EP_IN and ENCODING defines are not set.

I believe line 539 of audio_device.c should actually be format_type_I_rx - changing it to that allows code to compile and works as expected.

### How to reproduce ?

(see what happened)

### Debug Log as txt file

\Arduino15\packages\adafruit\hardware\samd\1.7.10\libraries\Adafruit_TinyUSB_Arduino\src\class\audio\audio_device.c:539:22: error: 'audiod_function_t' {aka 'struct <anonymous>'} has no member named 'format_type_I_tx'; did you mean 'format_type_I_rx'? 

  539 |       switch (audio->format_type_I_tx) 

      |                      ^~~~~~~~~~~~~~~~ 

      |                      format_type_I_rx 

exit status 1 

Error compiling for board Adafruit Feather M4 Express (SAMD51) 

### Screenshots

![image](https://user-images.githubusercontent.com/32046405/165146226-f896ea17-24c6-4805-9d49-684c14757838.png)

hathach:
fixed by #1514

PR #1514, opened 2022-06-16
Fix typo in audiod_rx_done_cb in audio_device.c

maddyaby:
To solve issue #1446. Code does not compile when trying to build with  CFG_TUD_AUDIO_ENABLE_DECODING && CFG_TUD_AUDIO_ENABLE_EP_OUT.

After fixing typo it does.


hathach (approved):
Great, thank you very much for the fix, and sorry for the late review

PR #1515, opened 2022-06-16
Stub out dcd_edpt_close for samd

maddyaby:
Not having this prevents the device from finishing the mounting process.

Tested on a SAMD51 and didn't seem to need to actually do anything in the close function. Did repeated close/opens of an audio streaming endpoint.

### DIFF
--- a/src/portable/microchip/samd/dcd_samd.c
+++ b/src/portable/microchip/samd/dcd_samd.c
@@ -250,6 +250,13 @@ bool dcd_edpt_open (uint8_t rhport, tusb_desc_endpoint_t const * desc_edpt)
   return true;
 }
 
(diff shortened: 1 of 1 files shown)


hathach:
`dcd_edpt_close()` is mostly used by audio with ISO endpoint, since SAMD doesn't support/test with ISO, it is left out. Have you manage to get audio example running with samd51 ?

maddyaby:
I got a streaming application working with SAMD51 (audio streaming from computer to device using ISO OUT endpoint). I needed to stub this out though for the calls in audio_device.c / audiod_set_interface to finish correctly.

hathach (approved):
thank you for testing out, samd is not tested with ISO yet. Though indeed, this will allow it to enumerate and do a bit more testing.

PR #1516, opened 2022-06-17
correct bulk size for highspeed endpoint in dynamic_configuration and usbtmc exmaples

### LINKED ISSUE #1457: some examples don't work for imxrt:  cdc_msc_freertos, dynamic_configuration, hid_composite_freertos,  usbtmc

hathach:
**Describe the PR**
fix #1457, close #1459
bulk endpoint size must be 512 for highspeed and 64 for FS, dynamic_configuration and usbtmc examples hard coded to 64 and cause an assert in the stack

```
[ 4949.159993] usb 3-1: USB disconnect, device number 51
[ 4949.484276] usb 3-1: new high-speed USB device number 52 using xhci_hcd
[ 4949.633227] usb 3-1: config 1 interface 0 altsetting 0 bulk endpoint 0x1 has invalid maxpacket 64
[ 4949.633230] usb 3-1: config 1 interface 0 altsetting 0 bulk endpoint 0x81 has invalid maxpacket 64
```

### DIFF
--- a/examples/device/dynamic_configuration/src/usb_descriptors.c
+++ b/examples/device/dynamic_configuration/src/usb_descriptors.c
@@ -163,7 +163,7 @@ uint8_t const desc_configuration_0[] =
   TUD_CONFIG_DESCRIPTOR(1, ITF_0_NUM_TOTAL, 0, CONFIG_0_TOTAL_LEN, 0x00, 100),
 
   // Interface number, string index, EP notification address and size, EP data address (out, in) and size.
-  TUD_CDC_DESCRIPTOR(ITF_0_NUM_CDC, 0, EPNUM_0_CDC_NOTIF, 8, EPNUM_0_CDC_OUT, EPNUM_0_CDC_IN, 64),
+  TUD_CDC_DESCRIPTOR(ITF_0_NUM_CDC, 0, EPNUM_0_CDC_NOTIF, 8, EPNUM_0_CDC_OUT, EPNUM_0_CDC_IN, TUD_OPT_HIGH_SPEED ? 512 : 64),
 
   // Interface number, string index, EP Out & EP In address, EP size
   TUD_MIDI_DESCRIPTOR(ITF_0_NUM_MIDI, 0, EPNUM_0_MIDI_OUT, EPNUM_0_MIDI_IN, TUD_OPT_HIGH_SPEED ? 512 : 64),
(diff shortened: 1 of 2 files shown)


pigrew:
@hathach, perhaps you can explain this to me?

Will this cause any issues when a high-speed device is talking to a full-speed host (or using a full-speed hub)?

Since it would be a full-speed connection, the packets shouldn't be more than 64 bytes, but the descriptor will statically report a maxPacketLen of 512?

(I'm not so familiar with high-speed USB...)

hathach:
For complete compatible with FS hub/host, highspeed-capable device must prepare 2 configuration descriptors: fs and hs and return correct one according to the actual link speed. When HS-capable device working in FS link, it may need to answer to the `get_other_speed_configuration()` from host. I did implement all of this for cdc_msc example sometime ago

https://github.com/hathach/tinyusb/blob/ae8b8f0c93885e3078f12c9ef00d65faa835d507/examples/device/cdc_msc/src/usb_descriptors.c#L143-L233

Note: The driver reserve max 512 bytes for buffering, but should only use the actual packet size from actual descriptor if needed. For cdc driver it simply schedule the max possible (512) since typically, cdc transfer end with short packet and have an stream buffered.

hathach:
PS: I will update the usbtmc example later to also support this scenario :)

pigrew:
> PS: I will update the usbtmc example later to also support this scenario :)

Thanks! I didn't want to create more work for you (but on the other hand, I don't have any HS devices to test...)

I unfortunately used `USBTMCD_MAX_PACKET_SIZE` a bunch of times in the driver, too. These have to be changed to be based on the value in the endpoint descriptor table....

hathach:
No problems at all, you added usbtmc driver before I have sorted this hs/fs thing out. I will double check the example and driver when having time.

PR #1521, opened 2022-06-17
Re-add some warning suppression for rp2040

kilograham:
I previously removed all the warning suppression in the rp2040 family.cmake, because I thought all the warnings had been fixed, but I must have been testing the wrong thing, as all the `-Wconversion` are still there.

Re-adding to get clean SDK build with master of TinyUSB


hathach:
I didn't encounter these warnings when building with `make BOARD=raspberry_pi_pico`, could you add some comment `# -Wyour-warning-option` in the cmake, maybe I could find a way to fix the source rather than suppressing it.

kilograham:
Yeah; they can certainly be fixed... this file shows what we do a test SDK compile with (which seems to about match the sensible limit of warning options that our users use)

https://github.com/raspberrypi/pico-sdk/blob/426e46126b5a1efaea4544cdb71ab81b61983034/test/kitchen_sink/CMakeLists.txt#L52

You can reproduce by

* cloning the SDK and `cd pico-sdk`
* doing a `git submodule update`
* `mkdir build`
* `cd build`
* `cmake -DPICO_SDK_TESTS_ENABLED ..`
* `make` or `make -j<procs>`

Obviously having commented out the suppression in lib/tinyusb

hathach:
thank @kilograham for the detail instructions, I have reproduced the all the warnings with rp2040. The most annoying one is -Wconversion, I am trying to "fix" these altogether. Will make an PR for this soon enough.

hathach (commented):
tinyusb should have no warnings with `-Wconversion` after #1534 . Please double check to see if this PR is still needed.

kilograham:
Sadly there are still a large number of `-Wconversion` errors (at least with gcc 9.2.1) some of them are a bit silly.

Can we please get this workaround PR merged in the interim, as we would like to pick up from master by tomorrow ! :-)

kilograham:
sadly there still seem to be a huge bunch under gcc 9.2.1  (perhaps they were there before, and it just wasn't in my test matrix).

Anyway, there also turns out to have been a backwards incompatible change in cmake 3.18

Therefore i think it is best to merge this in the short term

hathach:
that is weird, I tested with gcc 10 and 11 and it seems to be fine. Could you attach the txt file of compile output here for analysis

hathach:
I have been doing more tests with various gcc version. In sum up gcc 9 and prior will throw incorrect warnings with Wconversion. gcc 10/11 or later fixed this issue. Therefore I just push some more update to the PR with the version check

if (CMAKE_C_COMPILER_VERSION VERSION_LESS 10.0)

In general, all example will skip the Wconversion option with gcc9, also the suppress_tinyusb_warnings() function will do nothing if the gcc is 10 or later. And the "-Wno-stringop-overflow -Wno-array-bounds" is no longer needed. Let me know what do you think, and please do the test, if everything is ok on your side, we can merge it

PS: for `-Wstringop-overread/write`, this is false warnings, and has been fixed in gcc 11.3.
https://gcc.gnu.org/git/gitweb.cgi?p=gcc.git;h=91f7d7e1bb6827bf8e0b7ba7eb949953a5b1bd18

hathach:
merge now, feel free to push more update with additional PR if needed.

PR #1523, opened 2022-06-17
Allow more endpoint packet sizes for SAMD

maddyaby:
Tested on SAMD51 in an audio application with ISO OUT endpoint - we can allow more packet sizes by checking that the set size value is greater than the requested packet size instead of exactly the same.

### DIFF
--- a/src/portable/microchip/samd/dcd_samd.c
+++ b/src/portable/microchip/samd/dcd_samd.c
@@ -222,14 +222,14 @@ bool dcd_edpt_open (uint8_t rhport, tusb_desc_endpoint_t const * desc_edpt)
   UsbDeviceDescBank* bank = &sram_registers[epnum][dir];
   uint32_t size_value = 0;
   while (size_value < 7) {
-    if (1 << (size_value + 3) == tu_edpt_packet_size(desc_edpt)) {
+    if (1 << (size_value + 3) >= tu_edpt_packet_size(desc_edpt)) {
       break;
     }
     size_value++;
   }
 
   // unsupported endpoint size
-  if ( size_value == 7 && tu_edpt_packet_size(desc_edpt) != 1023 ) return false;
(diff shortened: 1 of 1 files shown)


pigrew:
I see that a `==` was changed to `>=`. I'm not so familiar with SAMD, but I figure this could cause issues with INT IN endpoints:

1. Class defines INT IN endpoint with `wMaxPacketSize=10`
2. Driver sets hardware to use `wMaxPacketSize=16`
3. Class starts transfer of 25 bytes.
4. Peripheral transmits packet of size 10 which would confuse the host since it expects a max packet size of 10 bytes.

hathach (approved):
thank you for the PR and sorry for late response. I have been too busy with other works.

hathach:
> I see that a `==` was changed to `>=`. I'm not so familiar with SAMD, but I figure this could cause issues with INT IN endpoints:
> 
> 1. Class defines INT IN endpoint with `wMaxPacketSize=10`
> 2. Driver sets hardware to use `wMaxPacketSize=16`
> 3. Class starts transfer of 25 bytes.
> 4. Peripheral transmits packet of size 10 which would confuse the host since it expects a max packet size of 10 bytes.

Thank you @pigrew for reviewing, this is SAMD port specific. SAMD endppoint can only have endpoint with power of two size i.e 8/16/.../512/1023. Therefore normally application on SAMD must take this into account and must define endpoint size accordingly. The use case here is for ISO endpoint which will have arbitrary depending on constant data rate e.g audio with 44.1 Mhz may have endpoint of 131 bytes while 48Mh would be 150 bytes. Normally ISO endpoint will only send 1 packet per constant interval, which is handle by audio driver, which make sure that it send data at constant rate. Of course for other use-case such as HID, SAMD user should only stick with power of two size. Hopefully this explanation make sense to you.

image

PR #1526, opened 2022-06-21
Added The Teensy 4.1 board

ccrome:
The Teensy 4.1 board has an 8MB W25Q64JV as opposed to the 2MB flash
on the Teensy 4.0.  This largely doesn't matter for TinyUsb, except
when you attempt to access any memory in the flash above the 2MB
boundary, when it fails with a hard fault.

This patch adds Teensy 4.1 support with the 8MB flash chip.


hathach (approved):
thank you for the addition.

PR #1531, opened 2022-06-25
USBTMC: Fixes and support for high-speed USB

### LINKED ISSUE #1529: USBTMC triggers assertion soon after configuration, due to extra CLEAR_FEATURE(ENDPT_HALT)
### Operating System

Windows 10

### Board

stm32f070rb

### Firmware

`master` branch (ae8b8f0c93885e3078f12c9ef00d65faa835d507)
Driver is NI VISA for windows v21.5

### What happened ?

_Summary_: Soon after device configuration, the host sends a CLEAR_FEATURE(ENDPOINT_HALT) on the USBTMC bulk-out endpoint, triggering an assertion that a transfer is being started on a busy endpoint (usbd_edpt_xfer:1243). 

At configuration, the class driver initiates a bulk-in transfer, so the endpoint is correctly being marked as busy. Then, the host sends CLEAR_FEATURE(ENDPOINT_HALT), and the class driver tries to initiate a second transfer on the already busy EP.

I'm not sure what the device _should_ be doing:
1. Should the device initiate a bulk-out transfer after configuration, or leave the EP in a STARLL state?
2. What should happen when CLEAR_FEATURE(ENDPOINT_HALT) when the EP is already listening?

I'm proposing the following, though additional input is appreciated:
1. Do start a bulk-outtransfer upon configuration. (as is currently implemented)
2. During a redundant clear-feature(halt), stall then unstall the endpoint, so the data toggle is reset to DATA0, followed by the start of a new transfer.

I'm going to continue trying to figure out what the correct behavior should be, and propose a patch.... Triggering an assert is bad.

### How to reproduce ?

Flash with usbtmc example, and connect to host...

### Debug Log as txt file

[log_clearFeature.txt](https://github.com/hathach/tinyusb/files/8959412/log_clearFeature.txt)

### Screenshots

_No response_

pigrew:
This updates USBTMC so that it should work with USB high-speed (untested, I don't have a HS device handy...).

It also fixes a few existing bugs:
1. Updates python test code to use newer versions of pyvisa
2. Manually stalls and unstalls an endpoint when receiving clear_feature(ENDPOINT_HALT), fixes #1529 
3. Reads the max packet size from the endpoint descriptors, instead of using a compile-time define.

The fix for the endpoint halt clearing likely should be put in the USBD core itself, instead of the class driver, but that's up for discussion. The USB488 spec mentions that the class may want to prevent the HALT from being cleared... I'd be happy for this patch to be merged, and then I can propose a larger fix.

### DIFF
--- a/examples/device/usbtmc/src/usb_descriptors.c
+++ b/examples/device/usbtmc/src/usb_descriptors.c
@@ -37,6 +37,9 @@
 #define USB_PID           (0x4000 | _PID_MAP(CDC, 0) | _PID_MAP(MSC, 1) | _PID_MAP(HID, 2) | \
                            _PID_MAP(MIDI, 3) | _PID_MAP(VENDOR, 4) )
 
+#define USB_VID   0xCafe
+#define USB_BCD   0x0200
+
 //--------------------------------------------------------------------+
 // Device Descriptors
 //--------------------------------------------------------------------+
@@ -51,9 +54,9 @@ tusb_desc_device_t const desc_device =
 
     .bMaxPacketSize0    = CFG_TUD_ENDPOINT0_SIZE,
 
-    .idVendor           = 0xCafe,
+    .idVendor           = USB_VID,
     .idProduct          = USB_PID,
-    .bcdDevice          = 0x0100,
+    .bcdDevice          = USB_BCD,
 
     .iManufacturer      = 0x01,
     .iProduct           = 0x02,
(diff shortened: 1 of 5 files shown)


pigrew:
Thinking more, it would be good to have a way to clear the event queue of transfers on the particular endpoint during the clear_feature call, but we can leave that for another day. This patch is a step in the proper direction.

hathach:
sorry for late response, I will pull out one of my board with HS to test with as soon as I could :) .

pigrew:
Great. Let me know if you have any questions.

Do you use Windows? If not, I believe that Linux comes with a kernel driver for USBTMC, which will operate with python's pyvisa library. On Windows, the driver I mentioned comes with the "NI MAX" software that allows manual interaction with the device. In the tinyusb USBTMC example, I do have a python script that runs a number of testcases using pyvisa (which is how I found this bug, though it is intermittent).

I'm also contemplating creating testcases for the CLEAR_FEATURE bug. I don't think that's possible with the standard USBTMC drivers, so it would need to use something like libusb, which makes it much harder to run on Windows.

pigrew:
I updated this branch to fix #1532.

pigrew:
Were you able to do testing with HS USB to confirm it works?

hathach:
> Were you able to do testing with HS USB to confirm it works?

I would love to, but not yet. The changes look great, but you are right, it may not be enough. I will try to pull out my old windows laptop and get it set up to test with. Will post the testing result here as soon as I could.

hathach:
Thanks for the reminder, to be honest, I haven't tried usbtmc before, I will give it a try this week.

PR #1536, opened 2022-06-28
samd: Fix a lock-up situation at high traffic.

robert-hh:
This is the companion PR #1535 as a branch to version 0.12.0. I'm not sure if that is the right way, but at least it changes the right file.
-------------------
This PR fixes a transmit lock-up, which happens, when data is received
and sent at the same time at moderate to high speeds, like code
which just echoes incoming data.

In my case, an issue was reported here:
https://github.com/micropython/micropython/issues/8521

**Describe the PR**

**Additional context**

### DIFF
--- a/src/portable/microchip/samd/dcd_samd.c
+++ b/src/portable/microchip/samd/dcd_samd.c
@@ -271,14 +271,14 @@ bool dcd_edpt_xfer (uint8_t rhport, uint8_t ep_addr, uint8_t * buffer, uint16_t 
   {
     bank->PCKSIZE.bit.MULTI_PACKET_SIZE = total_bytes;
     bank->PCKSIZE.bit.BYTE_COUNT = 0;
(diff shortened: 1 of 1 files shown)


hathach:
this PR is still actually toward the master branch, and have no difference with #1535 . At current stage, tinyusb does not have maintaining branch per release. It will only move forward in single release.

robert-hh:
Sorry. My fault. Is it possible to create a branch for 0.12.0 and change that one?

robert-hh:
The MicroyPython project uses tinyusb for five different MCU families. All work fine with v0.12.0. I'm taking some care of two of them, mimxrt and samd. The maintainer of MicroPython does not want to have different commits of the same library for different MCU ports. When switching a library version, all ports have to be tested against that and eventually changed. Someone has to do it, and it will take time, in which the fix for SAMD is not available. I can of course use a local copy of the changed file for this transition period.

PR #1538, opened 2022-06-29
Fix actual compiler warning on gcc 10.3.1

kilograham:


hathach (approved):
ah this is only compiled when debug, totally forgot this

PR #1541, opened 2022-06-30
comment out example.cmake warnings which break builds of rp2040 sdk or older compilers

kilograham:


kilograham:
Note also, that lots of code like

```
# Example common such as compiler warnings
include(${CMAKE_CURRENT_SOURCE_DIR}/../../example.cmake)
```

was added to each example CMakeLists.txt, however there was a function call in each, e.g.

```
# Example common such as compiler warnings
# Configure compilation flags and libraries for the example... see the corresponding function
# in hw/bsp/FAMILY/family.cmake for details.
family_configure_device_example(${PROJECT})
```
which would seem like a better place to have allowed each port to specify warnings if it wanted to

kilograham:
converting this to a draft as I'm not going to apply this quick fix at the moment

hathach:
> which would seem like a better place to have allowed each port to specify warnings if it wanted to

Yeah, I agree, it is a better place to add warnings. Though since `family.cmake` is also used by pico-sdk for and I am not sure if the  family_configure_device_example/family_configure_host_example() function are used by other sdk examples and/or user projects. I play a bit safe by only introducing these warnings to tinyusb's examples only, we can refactor this later on.

hathach:
Update: when build with LOG eanbled e.g make `BOARD=raspberry_pi_pico LOG=2` I got 

```
 error: "CFG_TUSB_DEBUG" redefined [-Werror]
```

due to the macro are defined in 2 places

- https://github.com/hathach/tinyusb/blob/19956f070d4dbf75b7fe4e8108e20837a70007f4/hw/bsp/rp2040/family.cmake#L58
- https://github.com/hathach/tinyusb/blob/19956f070d4dbf75b7fe4e8108e20837a70007f4/hw/bsp/rp2040/family.cmake#L122

hathach:
should we close this now ?

hathach:
Close since this is obsolete. Feel free to open another one should the issue still exists

PR #1543, opened 2022-06-30
mimxrt: Change the definitions for MIMXRT boards.

robert-hh:
The MIMXRT11XX definitions are mostly identical to the MIMXRT10XX, and the PR just adds OPT_MCU_MIMXRT11XX where OPT_MCU_MIMXRT10XX was mentioned. Only in ci_hs_imxrt.h to additional #defines are used to cater for different symbols being used. I tested the dcd part of change with a MIMXRT1176_EVK board.


hathach:
I just look at 1100 overview, I think we probably don't need an extra define for OPT_MCU_MIMXRT11XX, since NXP/Freescale did a great job with their mcusdk. We could rename this one to simply `OPT_MCU_MIMXRT` (to also account for rt600/500 in the future as well), the existing one is defined to new one to also allow backward-compatible. And we will update all example to use the new one. Over-time user will migrate (or not) 

```C
#define OPT_MCU_MIMXRT          700
#define OPT_MCU_MIMXRT10XX      OPT_MCU_MIMXRT11XX
```

src/portable/chipidea/ci_hs/ci_hs_imxrt.h
```diff
@@ -29,6 +29,11 @@
 
 #include "fsl_device_registers.h"
 
+#if CFG_TUSB_MCU == OPT_MCU_MIMXRT11XX
```
hathach, line 32: for this we can replace it with 

```C
#if !defined(USB1_BASE) && defined(USB_OTG1_BASE)
#endif
```

NXP should really need to add USB1_BASE to USB_OTG1_BASE for backward compatible like stm does.

robert-hh:
Thank you for your suggestion. I implemented and tested these with MIMXRT10XX and a MIMXRT1176 board, and it works. What I did not include was the line:
```
#define OPT_MCU_MIMXRT10XX      OPT_MCU_MIMXRT11XX
```
I did not understand it. In the new scheme, OPT_MCU_MIMXRT11XX is not defined any more.

hathach:
sorry, it is a typo, I mean

``C
#define OPT_MCU_MIMXRT          700
#define OPT_MCU_MIMXRT10XX      OPT_MCU_MIMXRT
```

**Comment by robert-hh** [comment]
That's a good addition for backward compatibility. I also extended that to 
```
#define OPT_MCU_MIMXRT            700 ///< NXP iMX RT Series
#define OPT_MCU_MIMXRT10XX        OPT_MCU_MIMXRT
#define OPT_MCU_MIMXRT11XX        OPT_MCU_MIMXRT
```
Maybe this is not needed, but does not hurt. I can also make a PR at NXP to add USB1_BASE to USB_OTG1_BASE for backward compatibility.

**Comment by hathach** [comment]
just removed the `OPT_MCU_MIMXRT11XX` since it is not used by anyone yet. OPT_MCU_MIMXRT should be the one to use for all imxrt mcu

**Comment by robert-hh** [comment]
Thanks. Did that. I pushed my actual chanegd version to the PR.

**Comment by robert-hh** [comment]
On a second thought: We could also add the redefinition of 
```
#if !defined(USB1_BASE) && defined(USB_OTG1_BASE)
.....
#endif
```
into our build scripts. I looked at the NXP header files defining USB_OTG1_BASE. These are >100000 lines monsters. I do not expect that NXP will change these.

**Comment by robert-hh** [comment]
On a third thought: Maybe we leave the substitution  in TinyUSB. Otherwise a note would be required, and it has to be considered in every potential use of MIMXRT11XX. Even if there is little hope for success, I'll make that PR at NXP.

**Comment by robert-hh** [comment]
OK. Then let's go with that state.

File: src/portable/chipidea/ci_hs/ci_hs_imxrt.h (hunk 1 above)
Comment (robert-hh) [reply src/portable/chipidea/ci_hs/ci_hs_imxrt.h]: Assuming that NXP will not do that, I'll declare this conversation as resolved.

**Review by hathach — APPROVED** [review]
thank you for you PR. I change my mind a bit, and re-add the `OPT_MCU_MIMXRT11XX`. Application should continue to use either `OPT_MCU_MIMXRT10XX` or `OPT_MCU_MIMXRT11XX` , the OPT_MCU_MIMXRT is used internally. Although they have no differences now, but in the future, NXP may just introduce some breaking changes e.g to rt1200, so we are better to be safe. 

It is a bit late, since it has a compile issue with rt1050 when bumping up mcu-sdk and I have to switch to troubleshoot other issue. Turn out it is only a minor macro rename from nxp. All is good now.

**Comment by robert-hh** [comment]
Thank you for merging. I have tested it with both a MIMXRT1170EVK and a MIMXRT1050_EVKB board. Both work. In the Makefile I use OPT_MCU_MIMXRT, being happy that there is one difference less ni the Makefile for the boards.

File: hw/bsp/imxrt/boards/mimxrt1050_evkb/board.h [hunk 2]
```diff
@@ -45,7 +45,7 @@
 
 // UART
 #define UART_PORT             LPUART1
-#define UART_RX_PINMUX        IOMUXC_GPIO_AD_B0_13_LPUART1_RX
-#define UART_TX_PINMUX        IOMUXC_GPIO_AD_B0_12_LPUART1_TX
+#define UART_RX_PINMUX        IOMUXC_GPIO_AD_B0_13_LPUART1_RXD
+#define UART_TX_PINMUX        IOMUXC_GPIO_AD_B0_12_LPUART1_TXD
```
hathach, line 49: @robert-hh rt1050 only has issue with the symbol naming in later mcu-sdk, fixed already.

PR #1554, opened 2022-07-12
Rework CMake example warnings some more

kilograham:
* Remove example.cmake and it's calls from each example, as we already have `family_` CMake functions that configure examples
* Added new CMake function to add the warnings to a target, and (conditiaonlly) called it from the rp2040 family.cmake allowing turning warnings off if it breaks some future SDK release.
* Note: I did not add this to any other family.cmake since there don't seem to be any that define the family_configure_xxx_example` functions


src/host/usbh.c
```diff
@@ -655,7 +655,9 @@ static bool usbh_control_xfer_cb (uint8_t dev_addr, uint8_t ep_addr, xfer_result
           TU_ASSERT( hcd_edpt_xfer(rhport, dev_addr, tu_edpt_addr(0, request->bmRequestType_bit.direction), _ctrl_xfer.buffer, request->wLength) );
           return true;
         }
+#if __GNUC__ >= 7
```
hathach, line 658: we still need to tell gcc 6 this is intended fallthough, also since usbh.c is compiled with other gcc we should check if GNUC is defined first at well

hathach (commented):
thank you very much, I think we need a bit extra works with fallthrough. I push an update asap

src/common/tusb_compiler.h
```diff
@@ -131,6 +131,12 @@
   #define TU_ATTR_BIT_FIELD_ORDER_BEGIN
   #define TU_ATTR_BIT_FIELD_ORDER_END
 
+  #if __has_attribute(__fallthrough__)
+    #define TU_ATTR_FALLTHROUGH         __attribute__((fallthrough))
+  #else
+    #define TU_ATTR_FALLTHROUGH         do {} while (0)  /* fallthrough */
```
hathach, line 137: the dowhile didn't do anything, but fallthough in comment is marker to suppress warnings for compiler that does not support the attribute.

hathach (approved):
everything is perfect now, ci in my pico-examples forks is happy as well from gcc6 to 11 https://github.com/hathach/pico-examples/runs/7339158529?check_suite_focus=true

PR #1580, opened 2022-07-28
Fixing Renesas register write protection and some compiler warnings in static code analysis

MatiMcFly:
**Describe the PR**
The Renesas RX controllers possess system-wide write protection for certain registers. Setup functions within the tinyUSB stack have to disable this write protection. This happens in order to disable the USB hardware's module stop function (e.g. `MSTP(USB0)` in `dcd_init`).
I changed the the write protection register's value handling in a way that the previous state is recovered. This leads to a more intuitive behavior of tinyUSB's initialization code

**Additional context**
Normally, peripheral modules are set up sequentially e.g. in a startup task. Bigger projects require lots of different modules whereas every single peripheral needs to have the module stop function disabled.
Threrefor, it is wise to disable the write protection at the very beginning of the initialization routine. At the end it would be reenabled again.

With this new behavior, developers don't have to think about register write protection when calling tinyUSB init-functions.

Here are two examples that show how the simplified behavior works:

Old behavior:
```
void StartTask(void *pv) {
  (void) pv;
  /* disable write protection of module stop functions */
  SYSTEM.PRCR.WORD = SYSTEM_PRCR_PRKEY | SYSTEM_PRCR_PRC1;
  
  /* call peripheral init functions */
  SetupADC();
  SetupXY();

  /* call tinyUSB init */
  UsbDeviceInit();   /* calls e.g. dcd_init() => This function in turn enables register write protection... */
  
  /* disable register protection again */
  SYSTEM.PRCR.WORD = SYSTEM_PRCR_PRKEY | SYSTEM_PRCR_PRC1;

  /* call further init functions
  lwIPInit();

  /*
  :
  :
  */

  /* reenable register write protection */
  SYSTEM.PRCR.WORD = SYSTEM_PRCR_PRKEY;
}
```

New behavior:
```
void StartTask(void *pv) {
  (void) pv;
  /* disable write protection of module stop functions */
  SYSTEM.PRCR.WORD = SYSTEM_PRCR_PRKEY | SYSTEM_PRCR_PRC1;
  
  /* call peripheral init functions */
  SetupADC();
  SetupXY();

  /* call tinyUSB init */
  UsbDeviceInit();   /* calls e.g. dcd_init() => Write protection will be disabled (again), but the previous state is recovered. */

  /* call further init functions, without having to disable the write protection a second time */
  lwIPInit();

  /*
  :
  :
  */

  /* reenable register write protection */
  SYSTEM.PRCR.WORD = SYSTEM_PRCR_PRKEY;
}
```


MatiMcFly:
I noticed an additional change that needs to be made. It has to do with the software configurable peripheral interrupt of the USB hardware.

The interrupt vector (`PERIB_INTB185`) needs  to know which peripheral triggers it, therefore the following line needs to be added:

  `ICU.SLIBR185.BYTE = 62;    /* set software configurable interrupt vector 185 to USBI0 */`

Please add this line to the files "dcd_usba.c" (insert at 632) and "hcd_usba.c" (insert at 554).

Not adding the line means that users of the stack have to add it manually in the stack or outside of `tusb_init`. Let me know if you prefer a new pull request or if you'd like to get further clarification.

perigoso:
consider using `uint32_t` and declaring the variable at the time of attribution

MatiMcFly:
@perigoso I totally agree with you. I prefer the stdint types over the underlying types, too. But in this specific file, the underlying types were used. Hence I tried to stay consistent with the existing code style.
Declaring a variable anywhere else than at the start of a scope can lead to problems when using older C standards such as C89 (I think). I just wanted to avoid that.

Thanks for your proposals though!
Since the build seems to have failed it looks like I'm going to have to make an additional PR anyways...

MatiMcFly:
Commit [c1c1238](https://github.com/hathach/tinyusb/pull/1580/commits/c1c123875d0edc8b32631e830c2dda39d47468f2):

Static code checks generate unnecessary warnings when using the CCRX toolchain. 
This happens if the macro `TU_VERIFY_STATIC` is used on the same line numbers in different files. The macro only uses the current line number as "unique ID".

The new solution includes the __COUNTER__ macro in the static code check which gets rid of that issue.

src/common/tusb_compiler.h
```diff
@@ -56,7 +56,7 @@
 #elif defined (__STDC_VERSION__) && __STDC_VERSION__ >= 201112L
   #define TU_VERIFY_STATIC   _Static_assert
 #elif defined(__CCRX__)
-  #define TU_VERIFY_STATIC(const_expr, _mess) typedef char TU_XSTRCAT(Line, __LINE__)[(const_expr) ? 1 : 0];
+  #define TU_VERIFY_STATIC(const_expr, _mess) typedef char TU_XSTRCAT3(Line, __LINE__, _TU_COUNTER_)[(const_expr) ? 1 : 0];
```
MatiMcFly, line 59: This is the macro that expands to something like:

`typedef char Line<XY>[(const_expr) ? 1 : 0];`

This commit changes the macro to:

`typedef char Line<XY><Z>[(const_expr) ? 1 : 0];`
With Z being the value of `_TU_COUNTER_`.

Yes, the number does not contain the exact line number (because `_TU_COUNTER_` was appended). But this is not an issue since the compiler error generated in case of a failed verification is the following:
`E0520094: The size of an array must be greater than zero`
 
Meaning the line index was unused anyways.

hathach: thanks, actually we don't need the line for indicator, gcc would be able to emit the exact location. We can just replace __LINE__ with __TU_COUNTER__ to make it consistent  with the followed ``#else`.

hathach (approved):
sorry for late response, thank you for the PR. Everything makes sense. Though the usba driver is rename to rusb2 and has been generalized to work with ra family. Though I noticed there is some hipcup espcially msc example does not work anymore with rx. Haven't spent much time to testing more. But it has nothing to do with this PR. Maybe I will revisit and/or file an seperated issues for that later. Will merge when ci passes.

MatiMcFly:
Thank you very much for merging the PR.
Yes, I also noticed that there is an issue with the msc example code. If I remember correctly, the issues start as soon as USB version 2.1 is specified and certain msc descriptors are read. 
I think there are even some open discussions regarding this topic.

I'll be back to work next week. If you want, I could probably provide further information.

hathach:
> Thank you very much for merging the PR. Yes, I also noticed that there is an issue with the msc example code. If I remember correctly, the issues start as soon as USB version 2.1 is specified and certain msc descriptors are read. I think there are even some open discussions regarding this topic.
> 
> I'll be back to work next week. If you want, I could probably provide further information.

that would be awesome, since I am not really familliar with rx much.

hathach:
maybe worth retested after (will try myself as well) https://github.com/hathach/tinyusb/pull/2587

PR #1588, opened 2022-08-03
add 'stm32l052dap52' (STM32L052K8 MCU) board target

majbthrd:
This adds the ST STM32L052 since I encountered a target with said MCU and didn't find an existing board type in TinyUSB.

I tried to follow the newer TinyUSB convention of "boards" sub-directory for a given "family" (stm32l0).  There is an existing "stm32l0538disco" board using the older convention that I left untouched since I don't know the reasons behind it still being in the main bsp directory.

I set the board name to "stm32l052dap52".  The reason for the naming suffix is that the target that I encountered is the [Olimex LoRa-STM32WL-DevKIT](https://github.com/OLIMEX/LoRa-STM32WL-DevKIT), which uses a STM32L052K8 as a CMSIS-DAP adapter for the board's main target, which is a STM32WLE5-series (no USB capability).  '[dap52](https://github.com/OLIMEX/LoRa-STM32WL-DevKIT/tree/main/SOFTWARE/cmsis-dap)' using the VID:PID 15ba:0044 appears to be Olimex's name of their (slightly buggy) fork of [dap42](https://github.com/devanlai/dap42)

Olimex did not do the user any favors.  Although there is talk in their github repository about a bootloader, that appears to just be leftover verbiage from the dap42 project.  There is a SB1 in the schematic labelled "Bootloader", but it doesn't invoke anything.  The only practical way to reprogram the STM32L052K8 is via the SWD pins on CON1 (1x10 IDC header).

### DIFF
--- a/examples/device/uac2_headset/skip.txt
+++ b/examples/device/uac2_headset/skip.txt
@@ -4,3 +4,4 @@ mcu:NUC121
 mcu:SAMD11
 mcu:SAME5X
 mcu:SAMG
+board:stm32l052dap52
--- a/hw/bsp/stm32l0/boards/stm32l052dap52/STM32L052K8Ux_FLASH.ld
+++ b/hw/bsp/stm32l0/boards/stm32l052dap52/STM32L052K8Ux_FLASH.ld
@@ -1,0 +1,169 @@
+/*
+*****************************************************************************
+**
+
+**  File        : LinkerScript.ld
+**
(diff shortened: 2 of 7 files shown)


hathach (approved):
thank you for PR, this is indeed an odd target with dap mcu instead of main mcu. Though I am sure other will find this useful as well, we did have stlink v3 as target as well. 

Don't worry about `stm32l0538disco`, it is added before the family thing, and still haven't got migrated yet. I will do it later on.

PR #1603, opened 2022-08-15
Host hub clear port and device interrupts

Ryzee119:
**Describe the PR**
I have a USB Hub that generates a `HUB_FEATURE_HUB_LOCAL_POWER_CHANGE` interrupt. This is not currently handed in TinyUSB so this hub just didnt work.

This PR handles hub device changes and clears the respective feature.

I also have another hub that generates `HUB_FEATURE_PORT_ENABLE_CHANGE`, which was never cleared so I was just spammed with hub interrupts. This PR also clears other HUB port interrupts to prevent this happening.

This PR should improve USB hub compatibility.

### DIFF
--- a/src/host/hub.c
+++ b/src/host/hub.c
@@ -408,7 +408,24 @@ static void hub_port_get_status_complete (tuh_xfer_t* xfer)
     hub_port_clear_feature(daddr, port_num, HUB_FEATURE_PORT_CONNECTION_CHANGE, connection_clear_conn_change_complete, 0);
   }else
   {
-    // Other changes are: Enable, Suspend, Over Current, Reset, L1 state
+    // Clear other port status change interrupts. TODO Not currently handled - just cleared.
+    if (p_hub->port_status.change.port_enable)
+    {
+      hub_port_clear_feature(daddr, port_num, HUB_FEATURE_PORT_ENABLE_CHANGE, hub_clear_feature_complete_stub, 0);
+    }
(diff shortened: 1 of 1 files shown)


src/host/hub.c
```diff
@@ -365,7 +407,16 @@ static void connection_get_status_complete (tuh_xfer_t* xfer)
     hub_port_clear_feature(daddr, port_num, HUB_FEATURE_PORT_CONNECTION_CHANGE, connection_clear_conn_change_complete, 0);
   }else
   {
-    // Other changes are: Enable, Suspend, Over Current, Reset, L1 state
+    // Clear other port status change interrupts. TODO Not currently handled - just cleared.
+    if (p_hub->port_status.change.port_enable)
```
hathach, line 411: can you add bracket `{ }` to if-else, it is minor, but I try to avoid if-else on newline without brackets.

hathach (commented):
superb ! thank you for the PR. Indeed, my hub does not generate these events. Though would you mind adding the brackets for the if-else.

Note: CI failed is due to an recent changes of pico-sdk (unrelated to this PR). I will try to fix that soon enough.

src/host/hub.c
Ryzee119: right ofcourse, Ill fix this soon

Ryzee119: Fixed

hathach (approved):
perfect thank you very much for the PR

PR #1614, opened 2022-08-26
nrf5x: Fix endpoint internal state when closed

kasjer:
**Describe the PR**
Field started (regarding transfer) was only cleared when transfer
was finished.
For audio devices set interface is called many times.
When there is no audio (silence) set interface requests zero
length bandwidth that in turn calls dcd_edpt_close().

When endpoint is closed due to set interface request transfer
should not longer be started since it will block next start transfer
with assert.

This just sets 'started' to false when endpoint is closed.

**Additional context**
This mostly affects audio where dcd_edpt_close() is called often but
maybe is could be observed on other classes.


hathach (approved):
great, thank for the fix.

kasjer:
> great, thank for the fix.

It was me who introduced this bug in the first place.
Thanks for quick review.

hathach:
> > great, thank for the fix.
> 
> It was me who introduced this bug in the first place. Thanks for quick review.

haha, not really, the bug is just an side-effect introduced along with a much more complicated race condition :D

PR #1621, opened 2022-08-31
Introduced declaration that supports STM32L4xx

lukebayes:
There are a number of compile errors for missing definitions when building against the library for an STM32L4 series IC.

It's possible there are other issues that need to be resolved to fully support this family of chips, but this did get us closer to being up and running.

While there are many compilation errors, here is one example:

```bash
tinyusb/src/portable/st/stm32_fsdev/dcd_stm32_fsdev_pvt_st.h:119:15: error: unknown type name '__IO'
  119 | static inline __IO uint16_t* pcd_ep_rx_cnt_ptr(USB_TypeDef * USBx, uint32_t bEpNum);
```

### DIFF
--- a/src/portable/st/stm32_fsdev/dcd_stm32_fsdev_pvt_st.h
+++ b/src/portable/st/stm32_fsdev/dcd_stm32_fsdev_pvt_st.h
@@ -98,6 +98,11 @@
   #undef USB_PMAADDR
   #define USB_PMAADDR USB1_PMAADDR
 
+// NOTE(lbayes): Added the following block for L4 support
+#elif CFG_TUSB_MCU == 309 // OPT_MCU_STM32L4
+  #include "stm32l4xx.h"
+  #define PMA_LENGTH (1024u)
+
 #else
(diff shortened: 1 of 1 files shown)


src/portable/st/stm32_fsdev/dcd_stm32_fsdev_pvt_st.h
```diff
@@ -98,6 +98,11 @@
   #undef USB_PMAADDR
   #define USB_PMAADDR USB1_PMAADDR
 
+// NOTE(lbayes): Added the following block for L4 support
```
hathach, line 101: please remove the Note line, also use the `OPT_MCU_STM32L4` macro instead of number 309. Since L4 has two series with different usb controller. You may want to make sure it is L4x2 or L4x3. (L4x5,x6 use dwc2).

hathach (changes requested):
thanks for the PR, please follow the review

hathach:
closed by #1727

PR #1623, opened 2022-09-02
Initial PIC32MM/MX & PIC24 support

ReimuNotMoe:
**Describe the PR**
Initial Microchip PIC32MM/MX & PIC24 support. Based on the KHCI implementation.

Only device mode is supported at this moment.

**Additional context**
Tested on the Microchip PIC32MM0064GPM028.


hathach (commented):
superb! thank you very much for the ports. Currently, I don't have any pic32mm board to test this out, but can you add the bsp for pic32mm board so that it can be easily tested for other users. If that is too much work, we can merge this as it is and do that later on.

Note: since pic32mm and nxp khci use the same chipidea fs usb, we can merge them into general drivers later on in the future. (just a head up).

ReimuNotMoe:
The test code is a bit complicated for PIC devices. Basically there are 2 factions when it comes to PIC development - some people would just use everything supplied by Microchip (including their huge IDE & build system), and other people (like me) use as many free software components as possible. Currently the test firmware project of the both PIC32MM and PIC24 uses CMake and contains some out-of-tree BSP files, and it would be unwise to include all of them in the TinyUSB code tree (or you think otherwise?). So I suggest you to merge them directly. I can provide a dedicated repo for the testing firmwares if you want.

Thanks.

hathach (approved):
thank you very much for the explanation. I understand that providing out-of-box example is a bit complicated, especially as you said, most people would stick with microchip IDE. We will merge this as working piece of code and will look at the running example later on when having time. Thanks again

PR #1605, opened 2022-08-16
RP2040 MIDI host

atoktoto:
**Describe the PR**
This PR adds MIDI host support for RP2040 boards based on work done by @rppicomidi in https://github.com/hathach/tinyusb/pull/1219 and @Skyler84 in https://github.com/hathach/tinyusb/pull/1434. I rebased all changed on top of master and fixed some compilation errors. 

MIDI controller (behringer x-touch mini) enumerates and the data is received. 

Creating this pull request as a draft for comments. I'm very new at this and I'm kind of surprised that it works :D

PR #1627, opened 2022-09-07
RP2040 MIDI Host

atoktoto:
This pull request replaces the previous draft: #1605 

This PR adds MIDI host support for RP2040 boards based on work done by @rppicomidi here: https://github.com/rppicomidi/tinyusb/tree/pio-midihost and @Skyler84 in #1434.

It seems to work pretty OK and has no issue with hot-plugging (re-plugging) the device that #1605 had.


src/portable/raspberrypi/rp2040/hcd_rp2040.c
```diff
@@ -142,16 +142,20 @@ static void __tusb_irq_path_func(hw_handle_buff_status)(void)
...
+        // IEP3 OUT is bit 7
         // etc
-        bit = 1 << (i*2);
-
-        if (remaining_buffers & bit)
-        {
-            remaining_buffers &= ~bit;
-            _handle_buff_status_bit(bit, &ep_pool[i]);
+        for(int j = 0; j < 2; j++){
+            bit = 1 << (i*2+j);
```
todbot, line 153: I'm getting error on this line:
``` 
error: conversion to 'uint' {aka 'unsigned int'} from 'int' may change the sign of the result [-Werror=sign-conversion]
  153 |             bit = 1 << (i*2+j);

File: src/class/midi/midi_host.c [hunk 2]
```diff
@@ -0,0 +1,944 @@
...
+// Enumeration
+//--------------------------------------------------------------------+
+bool midih_open(uint8_t rhport, uint8_t dev_addr, tusb_desc_interface_t const *desc_itf, uint16_t max_len)
+{
+  (void) rhport;
+
+  midih_interface_t *p_midi_host = get_midi_host(dev_addr);
+
+  TU_VERIFY(p_midi_host != NULL);
+  p_midi_host->num_string_indices = 0;
```
todbot, line 245: Getting error on this line: 
```
error: 'midih_interface_t' has no member named 'num_string_indices'
  245 |   p_midi_host->num_string_indices = 0;
```

todbot: Looks like adding `#define CFG_MIDI_HOST_DEVSTRINGS 1`  to `tusb_config.h` fixes this error.

src/class/midi/midi_host.c
```diff
@@ -0,0 +1,944 @@
...
+      }
+      else if (p_mdij->bDescriptorSubType == MIDI_CS_INTERFACE_IN_JACK)
+      {
+        // Then it is an in jack. 
+        TU_LOG2("Found in jack\r\n");
+#if CFG_MIDI_HOST_DEVSTRINGS
+        if (p_midi_host->next_in_jack < MAX_IN_JACKS)
+        {
+          p_midi_host->in_jack_info[p_midi_host->next_in_jack].jack_id = p_mdij->bJackID;
+          p_midi_host->in_jack_info[p_midi_host->next_in_jack].jack_type = p_mdij->bJackType;
```
todbot, line 315:
```
error: conversion to 'int8_t' {aka 'signed char'} from 'uint8_t' {aka 'const unsigned char'} may change the sign of the result [-Werror=sign-conversion]
  315 |           p_midi_host->in_jack_info[p_midi_host->next_in_jack].jack_type = p_mdij->bJackType;
      |                                                                            ^~~~~~

File: src/portable/raspberrypi/rp2040/hcd_rp2040.c (hunk 1 above)
Comment (atoktoto) [reply src/portable/raspberrypi/rp2040/hcd_rp2040.c:153]: changed j to uint

File: src/class/midi/midi_host.c (hunk 3 above)
Comment (atoktoto) [reply src/class/midi/midi_host.c:315]: No idea why jack_type was signed... changed it to uint8_t.

**Comment by atoktoto** [comment]
I will add a simple example in the next few days if time permits.

**Comment by atoktoto** [comment]
Thanks for being persistent on this one. I'm new at this and was not even aware of the standard way to build TinyUSB. 
I'm building this starting from pico-examples CMakeLists.txt as the project includes both pico-sdk and TinyUSB examples.

**Comment by atoktoto** [comment]
Generally I'm following this guide: https://datasheets.raspberrypi.com/pico/getting-started-with-pico.pdf and building from CLion IDE, but the commands it's running are:

`cmake.exe -G Ninja -S E:\pico\pico-examples -B E:\pico\pico-examples\build`
`cmake.exe --build E:\pico\pico-examples\build --target tinyusb_host_midi_rx -j 9`

(I ommited the env variables specifying the paths to compilers for the second command)

For any other example it would be as simple as:
`cd pico-examples`
`mkdir build`
`cd build`
`cmake ..`
`cd blink` (in the build directory)
`make -j4`

But the TinyUSB examples are pulled in from outside of project root and I don't know where to look for CMake outputs of those.

**Comment by atoktoto** [comment]
In the meantime (while waiting for #1434 to be merged) I started on adding a nice-looking MIDI API. Turns out the simplest way is to reuse https://github.com/FortySevenEffects/arduino_midi_library. Fortunately it does not depend on Arduino itself and has a transport plug-in architecture. So I wrote a TinyUSB based plugin for it: https://github.com/atoktoto/pico-midi-usb-transport

**Comment by atoktoto** [comment]
> Seems odd to me you're not following standard `tinyusb` build process for this.

I will try to fix this

**Comment by atoktoto** [comment]
@todbot On my machine (Windows, under WSL2) the example now builds with `make BOARD=raspberry_pi_pico all` in the example directory and works as intended when uf2 is dropped onto RP2040. I did not get the `no macro name given in #define directive` error at any point so not sure where to look for the issue.

**Comment by atoktoto** [comment]
That's a naming inconsistency on my side. 

The key fact here is the name of the project is not important. AFAIK `add_library(pico-midi-usb INTERFACE)` adds a target that can be linked to with `target_link_libraries(example-print-notes pico-midi-usb)`.

File: src/class/midi/midi_host.h [hunk 4]
```diff
@@ -0,0 +1,137 @@
...
+#ifndef CFG_TUH_HID_EPOUT_BUFSIZE
+#define CFG_TUH_HID_EPOUT_BUFSIZE 64
+#endif
+
+
+//--------------------------------------------------------------------+
+// Application API (Single Interface)
+//--------------------------------------------------------------------+
+bool     tuh_midi_configured      (uint8_t dev_addr);
+uint32_t tuh_midi_available    (uint8_t dev_addr);
```
dequis, line 55: The implementation of this doesn't seem to exist?

On the device side it's just a static inline with no parameters that calls `tud_midi_n_available(0, 0)`, the parameters being device and cable, and returns number of bytes available to read.

Not really an equivalent in this code. Just remove it, I guess?

rppicomidi:
@atoktoto I found a fix for hubs not working with this. I made the error in my original pull request. Would you consider adding this change to your pull request?
```
diff --git a/src/class/midi/midi_host.c b/src/class/midi/midi_host.c
index d242d33c5..bc87898cc 100644
--- a/src/class/midi/midi_host.c
+++ b/src/class/midi/midi_host.c
@@ -490,7 +490,7 @@ bool midih_set_config(uint8_t dev_addr, uint8_t itf_num)
   p_midi_host->configured = true;
 
   // TODO I don't think there are any special config things to do for MIDI
-
+  usbh_driver_set_config_complete(dev_addr, p_midi_host->itf_num);
   return true;
 }
```

AndrewCapon:
Hi @atoktoto,

The last update from @rppicomidi fixed the enumeration problems but there are still data transfer problems, this fixes it:

```
diff --git a/src/class/midi/midi_host.c b/src/class/midi/midi_host.c
index bc87898c..8df1b200 100644
--- a/src/class/midi/midi_host.c
+++ b/src/class/midi/midi_host.c
@@ -529,7 +529,7 @@ bool tuh_midi_read_poll( uint8_t dev_addr )
   if (in_edpt_not_busy)
   {
     TU_LOG2("Requesting poll IN endpoint %d\r\n", p_midi_host->ep_in);
-    TU_ASSERT(usbh_edpt_xfer(p_midi_host->dev_addr, p_midi_host->ep_in, _midi_host->epin_buf, _midi_host->ep_in_max), 0);
+    TU_ASSERT(usbh_edpt_xfer(p_midi_host->dev_addr, p_midi_host->ep_in, p_midi_host->epin_buf, p_midi_host->ep_in_max), 0);
     result = true;
   }
   else
```

raveslave:
checking in on this one

rppicomidi:
@raveslave I wrote a USB MIDI host application driver for TinyUSB in the [usb_midi_host](https://github.com/rppicomidi/usb_midi_host) project. You might find that helpful. I have tested it on the RP2040 on a Raspberry Pi Pico board and a Pico W board.

raveslave:
> @raveslave I wrote a USB MIDI host application driver for TinyUSB in the [usb_midi_host](https://github.com/rppicomidi/usb_midi_host) project. You might find that helpful. I have tested it on the RP2040 on a Raspberry Pi Pico board and a Pico W board.

thanks!!

brendena:
What are the steps holding this back from being merged?

Ownasaurus:
+1 person interested in seeing this merged before it becomes any more stale! :)

nevvkid:
+1 for this feature.

ctag-fh-kiel:
+1 for this feature.

jfedor2:
You can just use the external application driver linked above (usb_midi_host), it works great. There's no need for this to be part of tinyusb.

nevvkid:
@jfedor2 thats what we are currently doing. But there are huge performance problems with that if you do more advanced stuff via USB Host. So i would it say it depends. Generally there is room for improvement and better integration of the MIDI Host feature with TinyUSB.

rppicomidi:
@nevvkid Can you please describe where the performance problems from application drivers come from, and what advanced stuff you are doing that suffers when you use application drivers?

As far as I understand it, the USB host code accesses external driver functions using function pointers; same as built-in drivers. External application driver function pointers are stored in an array; built-in driver function pointers are stored in a different array. To get a pointer to the function pointers of the driver, the host code calls the [get_driver()](https://github.com/hathach/tinyusb/blob/35e9f033248e2a816d8f391c3ebf55cb7f3e5019/src/host/usbh.c#L225) function. Is there something I am missing?

hathach (commented):
Thank you very much of your PR. It has been a quite and I really appreciate your effort and patient. I have been wanting to review/merge this but keep getting pulled by other works and kind of forgot sometimes. Adding an new host driver is significant works, and reviewing it may take a bit longer. Though I have sync it up with master, also fix some out of sync naming/signature/make changes. I run the example and it works well (great work). 

We will need to migrate the endpoint to use the new endpoint stream, maybe rename/move thing around if neccessary. Before that since I am not familliar with MIDI, can you explain what `CFG_MIDI_HOST_DEVSTRINGS` means ? Is it a terminology since it enable mananging series of midi jack ? 


```
[0:] USBH Device Attach                                                                                                                                                                         
MIDI opening Interface 1 (addr = 1)                                                                                                                                                             
MIDI descriptor parsed successfully                                                                                                                                                             
MIDI device address = 1, IN endpoint 1 has 1 cables, OUT endpoint 1 has 1 cables                                                                                                                
Read bytes 3 cable 0  0000:  90 51 7F                                         |.Q.|                                                                                                             
Read bytes 3 cable 0  0000:  80 4E 00                                         |.N.|              
```

hathach:
I made more changes to update the driver to use the new endpoint stream API (similar to cdc host). I have only tested with midi_rx example, didn't test with tx at all. I will try to look up what the string doing with midi and continue to make more update, I think we should also need to make some driver API changes as well. Hopefully we could wrap this up and merge this soon enough.

Ownasaurus:
Thank you so much for allocating some of your limited time to this. Us MIDI people really appreciate it!!

ctag-fh-kiel:
Is #2814 relevant (and maybe #2776) could they be accounted for in this merge?

hathach:
> Is #2814 relevant (and maybe #2776) could they be accounted for in this merge?

That PR solve another issue, and isn't part of this.

rppicomidi:
@hathach Set `CFG_MIDI_HOST_DEVSTRINGS` to 1 to process the virtual cable names of the attached device for MIDI IN and MIDI OUT from the MIDI configuration descriptors and the string descriptors during enumeration and to enable the API for applications to read the strings. For example, on my Arturia KeyLab Essential 88 keyboard, virtual cable 0 is labeled "MIDI" and virtual cable 1 is labeled "DAW" for both MIDI IN and MIDI OUT.

rppicomidi:
@hathach I recommend that you reference the [usb_midi_host](https://github.com/rppicomidi/usb_midi_host) project when merging this pull request. I have been maintaining it for quite a while and have fixed a number of issues since this pull request was submitted.

hathach:
> @hathach Set `CFG_MIDI_HOST_DEVSTRINGS` to 1 to process the virtual cable names of the attached device for MIDI IN and MIDI OUT from the MIDI configuration descriptors and the string descriptors during enumeration and to enable the API for applications to read the strings. For example, on my Arturia KeyLab Essential 88 keyboard, virtual cable 0 is labeled "MIDI" and virtual cable 1 is labeled "DAW" for both MIDI IN and MIDI OUT.

Thank you for explanation

> @hathach I recommend that you reference the [usb_midi_host](https://github.com/rppicomidi/usb_midi_host) project when merging this pull request. I have been maintaining it for quite a while and have fixed a number of issues since this pull request was submitted.

Ah thank you for the head up, I will try to merge this first, then do compare and update later with follow-up PR afterwards. I don't own/use any midi comercial at all, therefore it is all up to you and others to fix its issues. I mostly make sure it is up-to-date with stack API, compiling with other ports and working with simple examples (like midi_rx).

hathach:
PS: I made more changes to the API and its signature and feeling good for now. I will check the string thing later, need to switch to other works for now. So the string is mostly for displaying the label ? Besides that does it has any behavior or setting impact ?

rppicomidi:
@hathach Yes, `CFG_MIDI_HOST_DEVSTRINGS` is for displaying labels for virtual cables. Enabling the setting requires slightly more RAM and code space. Enabling it slows enumeration slightly because the driver needs to parse the Jack Descriptors. The setting does not change run-time behavior after enumeration. Applications with no UI do not need it.

Many MIDI devices do not implement Jack Descriptor strings. Some implement them, but do it poorly. An example RP2040 device that uses Jack Descriptor strings is [midi-multistream2usbdev](https://github.com/rppicomidi/midi-multistream2usbdev). You may find that project useful to test the Jack Strings when you get to that.

hathach:
@rppicomidi thank you for your explanation, if string is only used for displaying label. I think we can just add an optional callback that have all these information when parsing descriptor instead of storing them in the interface. If application need these, they should store it.

hathach (approved):
Thank you very much everyone for putting time and effort for this PR. Especially @atoktoto and @rppicomidi for getting most of the work done. I am sorry for being super late, I have to constant switching works and haven't worked much with rp2040 until now. 

Though since there is lots of changes since this PR is created, I have update and also make lots of changes, notably:
- lost of API rename and signature changes to match other class driver consistency e.g num_rx_cables --> rx_cable_count
- remove the devstring and added tuh_midi_descriptor_cb() along with callback data to leave extracting midi jack info to application if needed
- use the new endpoint stream API for streamging data
- allow to turn off stream API `CFG_TUH_MIDI_STREAM_API=0` (which consume lots of logic/footprint). Since more recent high level MIDI library can support multiple cables and can parse 4-byte usb packet natively
- change the API input from dev_addr --> interface index, consistency with cdc, but also allow to support more than 1 midi interface per device (rare case).

It is quite a bit of changes, and may break compatible with existing host driver from application space, @rppicomidi let me know what you think and how you want to chagne/resolve the confict. I am open to all feedback, thank you.

rppicomidi:
@hathach I am just back from travel. I may need a few days to get you feedback. Sorry for the delay.

hathach:
> @hathach I am just back from travel. I may need a few days to get you feedback. Sorry for the delay.

no problems, please take your time.

rppicomidi:
The `midi_rx` example does not build.  I tried the build with VS Code using the Raspberry Pi `pico-examples` project and pico-sdk 2.1.1. The reasons the build failed are likely out of scope for this pull request. I am writing them here because I do not know where else to file them.

It does not build for two reasons:
1. The pico-sdk file `src/rp2_common/tinyusb/CMakeLists.txt` sets `BOARD` to `pico-sdk`, which does not work with the latest TinyUSB code. Even the develop branch of the `pico-sdk` does not set `BOARD` and `FAMILY` based on the `PICO_BOARD` setting for this `CMakeLists.txt` file; it also does not test if `BOARD` or `FAMILY` are already defined on the cmake command line. I believe this is something that needs to be worked out between the Raspberry Pi pico-sdk project and this one.
2. There are 3 examples in the TinyUSB project whose `CMakeLists.txt` files call `family_configure_device_example(${EXE_NAME} ${RTOS})` when they should call `family_configure_device_example(${EXE_NAME} noos)`. The files are examples in the `examples/device` directory: `board_test`, `msc_dual_lun` and `cdc_msc`. Because these files do not build, the whole `pico-examples` project appears broken.

Are these known issues? I did not see bugs filed, but it may be my search. People who follow the Raspberry Pi Pico build workflow from current Raspberry Pi Pico C/C++ documentation will have trouble here and may give up.

I will try to get to a test on hardware and a review of the actual driver when I am able.

hathach:
@rppicomidi how to do build the example, I normally just go to tinyusb/exmaples/ and cmake there. I surely can try to make it backward compatible.

rppicomidi:
@hathach Refer to issue #3004 for reproduce steps. Yes, putting an empty `board.h` file in the `pico-sdk` directory seems to allow it to build.

hathach:
> @hathach Refer to issue #3004 for reproduce steps. Yes, putting an empty `board.h` file in the `pico-sdk` directory seems to allow it to build.

I don't use vscode, but if this fix the compile issue, let just do it. I guess I miss this board config

rppicomidi:
Also need to fix 3 CMakeList.txt files. I added command line steps to reproduce to #3004.

rppicomidi:
In `midi_host.c`, lines 173-176
```
    if (xferred_bytes) {
      tu_edpt_stream_read_xfer_complete(&p_midi_host->ep_stream.rx, xferred_bytes);
      tuh_midi_rx_cb(idx, xferred_bytes);
    }
```
Many MIDI devices, especially from Korg, will always respond to an IN transfer request from the host with 64 bytes of data. If they have no data to send, they will send back all 64-bytes of zero instead of NAK. If they have only a single message to send, they will send back the first 4 bytes encoded as a MIDI message packet and the remaining 60 bytes all zero. And so on.

My `usb_midi_host` application driver handles this case by parsing the received data in the transfer callback and filtering out all 4-byte packets that are all zero. It only calls the `tuh_midi_rx_cb()` if the received data contains valid MIDI data. If you choose not to do this filtering in the midi_host driver, then the test midi_rx application's `tuh_midi_rx_cb()` function should test if `bytes_read == 0` before printing. For example:

```
  uint32_t bytes_read = tuh_midi_stream_read(idx, &cable_num, buffer, sizeof(buffer));
  if (bytes_read == 0) return; // <---rppicomidi recommended change
  printf("Cable %u rx %lu bytes: ", cable_num, bytes_read);
  for (uint32_t i = 0; i < bytes_read; i++) {
    printf("%02X ", buffer[i]);
  }
  printf("\r\n");
```
This change to driver behavior may change behavior of existing applications that use the `usb_midi_host` application host driver.

rppicomidi:
A question about the example program: Why does it not also demonstrate MIDI TX? For example, see https://github.com/rppicomidi/usb_midi_host/blob/main/examples/C-code/usb_midi_host_example/usb_midi_host_example.c

The example will either light up the transport LEDs on MIDI controller devices that support play and record transport buttons, or it will play musical notes on sound generating MIDI devices. If the issue is you have no commercial MIDI hardware to test this, perhaps just hooking it to a Pico or other board with the TinyUSB MIDI Device example will suffice.

rppicomidi:
There are some MIDI devices, especially guitar pedals, that are designed for high speed USB 2.0 instead of full speed. They also do not provide an alternate descriptor when the host is full speed. As a result, their bulk endpoint descriptor max transfer sizes are always 512 bytes. That is too large for chips like the RP2040 that only support full speed USB. The code in the `usb_midi_host` application driver limits the endpoint sizes to the maximum that the hardware supports before opening the endpoint. I would recommend that this driver limit the endpoint buffer sizes before calling `tuh_edpt_open()`. Alternatively, the processor specific hcd code should limit the buffer sizes

rppicomidi:
`tuh_midi_stream_write()` should handle running status to make converting serial streams with running status encoded easier. This feature was added to the `usb_midi_host` application driver by pull request. See https://github.com/rppicomidi/usb_midi_host/blob/82a5c8e04bb16d638beafbb594c8f241279c6764/usb_midi_host.c#L616

hathach:
> Many MIDI devices, especially from Korg, will always respond to an IN transfer request from the host with 64 bytes of data. If they have no data to send, they will send back all 64-bytes of zero instead of NAK. If they have only a single message to send, they will send back the first 4 bytes encoded as a MIDI message packet and the remaining 60 bytes all zero. And so on.

Ah sorry, I didn't pay attention to this, I just did a quick check to see if all zeroes is a valid packet to make sure we don't drop a meaningful payload. The CIN = 0 is lisetd as `Miscellaneous function codes. Reserved for future extensions.` seem like we can just drop it, I will make the changes 

> A question about the example program: Why does it not also demonstrate MIDI TX? For example, see https://github.com/rppicomidi/usb_midi_host/blob/main/examples/C-code/usb_midi_host_example/usb_midi_host_example.c
> 
> The example will either light up the transport LEDs on MIDI controller devices that support play and record transport buttons, or it will play musical notes on sound generating MIDI devices. If the issue is you have no commercial MIDI hardware to test this, perhaps just hooking it to a Pico or other board with the TinyUSB MIDI Device example will suffice.

Example is part of the PR, I don't know the MIDI enough or have any consumer device to make a meaningful tx example. I mostly test with tinyusb midi device to print out console. If possible, please make an follow-up PR to update the example to both rx & tx.

> There are some MIDI devices, especially guitar pedals, that are designed for high speed USB 2.0 instead of full speed. They also do not provide an alternate descriptor when the host is full speed. As a result, their bulk endpoint descriptor max transfer sizes are always 512 bytes. That is too large for chips like the RP2040 that only support full speed USB. The code in the usb_midi_host application driver limits the endpoint sizes to the maximum that the hardware supports before opening the endpoint. I would recommend that this driver limit the endpoint buffer sizes before calling tuh_edpt_open(). Alternatively, the processor specific hcd code should limit the buffer sizes

Hmm, this is out of USB specs, what if the device send 512bytes or simply more than 64 bytes on the bus ? that would corrupt buffer/buffer overflow on the rp2040 side. Maybe giving a LOG warnin, I will check, some of the sanity check may freakout.

> `tuh_midi_stream_write()` should handle running status to make converting serial streams with running status encoded easier. This feature was added to the `usb_midi_host` application driver by pull request. See https://github.com/rppicomidi/usb_midi_host/blob/82a5c8e04bb16d638beafbb594c8f241279c6764/usb_midi_host.c#L616

thanks for the info, maybe we can do that in follow-up PR.

hathach:
@rppicomidi I made more changes to addres 2 of your reviewed issues
- all zero response: it is special case, I think we can just simply check if all rx bytes are zeroes instead of doing it in 4 bytes  which is quicker. I guess there is no device that has zero packet between meanningful packets in the same response ?
- force bulk 512 -> 64 bytes is done in tu_edpt_validate() which will help other drivers using bulk/stream ep API as well.

Other issues can be resolved in follow-up PR, please make one if you have time, I will try to review it faster :). Also let me know that the callback with descriptors are good enough for your application to display, since I found bundling it into driver make thing a bit more complicated and may not extract/inteprete all infomration that user want. So it is better to leave that to application.

rppicomidi:
I will attempt to integrate this driver with some of my applications after you merge it. Any issues I encounter will result in smaller pull requests that should be simpler to review and manage.

hathach:
> I will attempt to integrate this driver with some of my applications after you merge it. Any issues I encounter will result in smaller pull requests that should be simpler to review and manage.

Thank you very much for the effort and patient so far. Feel free to submit more follow-up PRs, I will try to review them as soon as I could.

PR #1639, opened 2022-09-12
Fixes for building with IAR toolchain

bavison:
I'm aiming to get TinyUSB to build with the IAR Embedded Workbench toolchain (I'm testing against Arm version 9.20.4) and specifically, I'm trying to build it as part of the Raspberry Pi Pico SDK. In doing so, I've identified a number of issues. Some of these are in rp2040-specific source files, but many others are in platform-neutral ones.

I've grouped the changes into a series of commits based upon the diagnostic generated by IAR. The first four are errors, and prevent the build from proceeding. The remainder are fixes for warnings (I'm more generally aiming to get the entire Pico SDK to build without warnings).


src/class/video/video_device.c
```diff
@@ -251,39 +250,39 @@ static void const* _find_desc_ep(void const *beg, void const *end)
...
+      return cur;
     }
     cur = tu_desc_next(cur);
   }
   return end;
 }
 
 /** Return the end of the video streaming descriptor. */
-static inline void const* _end_of_streaming_descriptor(void const *desc)
+static inline uint8_t const* _end_of_streaming_descriptor(uint8_t const *desc)
```
kkitayam, line 271: The modification makes caller require cast operations. Does it need for IAR?

I think it is better to avoid cast operations at the caller. Cast operations sometimes occur unexpected bugs, it is better to gather cast operations in a low-level function than scatter at caller functions.

bavison: It was just the case that there was a lot of pointer arithmetic on void pointers, and it was easiest to simply replace all of them in that source file with `uint8_t const *`. I have gone through and reverted them to `void const *` where that reduces the overall number of casts involved. Overall, the commit now results in a net reduction in the number of casts, which is nice.

HiFiPhile:
Thanks for your PR, most changes are good for me, especially explicit enum type conversion.

I don't use RP2040 so I can't test these changes. ~~I feel like most IAR users use traditional MCUs~~

@pigrew Could you take a look at changes in USBTMC class ?

src/class/video/video_device.c
kkitayam: Thank you for your comments and updating!

hathach (approved):
Superb! thank you very much for the PR, since then I have made some changes to also fix IAR build for some stm32 but haven't try to compile with rp2040. The changes look great, I have synced with master and updated, will merge when ci passed.

kilograham:
actually this code

https://github.com/hathach/tinyusb/blob/f1a859d90701564dc6e152197454fae1f49ee846/src/portable/raspberrypi/rp2040/rp2040_usb.h#L39

is wrong (and likely won't actually compile on IAR)

It should read

```
#define usb_hw_set    ((usb_hw_t *) hw_set_alias_untyped(usb_hw))
#define usb_hw_clear  ((usb_hw_t *) hw_clear_alias_untyped(usb_hw))
```
(Note the `hw_xxx_alias_untyped` vs `hw_xxx_alias`)

hathach:
@kilograham ah thanks, I haven't tried IAR with rp2040, will make PR per your suggestion. I saw rpi team is also doing IAR support within pico-sdk as well, nice timing.

kilograham:
indeed - that is no coincidence ;-) and yeah the change to this PR is a result of me making changes to what was merged there ;-)

PR #1642, opened 2022-09-13
nrf5x: Fix reception of large ISO packets

kasjer:
**Describe the PR**
ISO packet size is up to 1023 for full speed device. Upon completion of ISO reception, reported length of incoming packet was truncated to one byte only.
This results in incorrect data stream for higher bit rates 48 samples * 4 bytes per sample * 2 channels = 384 bytes of data and 128 was reported.

There is no change in logic extending xact_len to uint16_t fixes the issue.


hathach (approved):
superb!! thank you for the fix. Sorry for these type of issue, it should be avoidable. I could imagine you having hard time to troubleshoot this.

kasjer:
> superb!! thank you for the fix. Sorry for these type of issue, it should be avoidable. I could imagine you having hard time to troubleshoot this.

Thank you for quick response. No need to be sorry, ISO support was added late.
16 bit PCM was good enough so far but I tried to use AUDIO_DATA_FORMAT_TYPE_I_IEEE_FLOAT and that change subslot size to 4 hence error.

hathach:
thanks for the fix and the extra info :)

PR #1662, opened 2022-10-04
Check for event queue overflow in usbd / dcd_event_handler

Erlkoenig90:
When using `osal_queue_send` to send events to the queue `_usbd_q` in `udbd.c`'s `dcd_event_handler`, the queue may overflow (particularly if the application calls `usbd_defer_func` too often). In that case, events will be missed, leading to spurious and hard-to-debug errors (for example, IN data is sent, ISR is called, but nothing else happens).

Therefore I'd like to suggest adding `TU_ASSERT` around the calls to `osal_queue_send` in order to make finding these bugs easier.

### DIFF
--- a/src/device/usbd.c
+++ b/src/device/usbd.c
@@ -296,9 +296,9 @@ tu_static osal_queue_t _usbd_q;
 #endif
 
 TU_ATTR_ALWAYS_INLINE static inline bool queue_event(dcd_event_t const * event, bool in_isr) {
-  bool ret = osal_queue_send(_usbd_q, event, in_isr);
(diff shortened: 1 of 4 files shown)


hathach:
which RTOS you are using, the osal_none does have an TU_ASSERT() when adding item to queue.

Erlkoenig90:
I'm using FreeRTOS, via ESP-IDF. What's the point of the return value in `osal_queue_send` if there is a `TU_ASSERT`, i.e. the return value is supped to be always `true`?

hathach (approved):
thank you for the Pr, and sorry for late response. I have rebased and add check for queuing new usb event

PR #1676, opened 2022-10-11
dcd_lpc_ip3511: isochronous support and endpoint accidental write fix

tswan-quasi:
**Describe the PR**

Added support for isochronous endpoints as well as stopped the USB from accidentally writing the incorrect address upon endpoint resets for dcd_lpc_ip3511

**Additional context**

Tested on an [LPC55S69JBD100](https://www.nxp.com/products/processors-and-microcontrollers/arm-microcontrollers/general-purpose-mcus/lpc5500-cortex-m33/high-efficiency-arm-cortex-m33-based-microcontroller-family:LPC55S6x) using the [LPC55S69-EVK](https://www.nxp.com/design/development-boards/lpcxpresso-boards/lpcxpresso55s69-development-board:LPC55S69-EVK). Only the FS port was tested, not HS.

Isochronous Examples Tested: headset, video

Other Examples Tested: cdc_msc, hid

### DIFF
--- a/src/portable/nxp/lpc_ip3511/dcd_lpc_ip3511.c
+++ b/src/portable/nxp/lpc_ip3511/dcd_lpc_ip3511.c
@@ -421,7 +421,7 @@ static void bus_reset(uint8_t rhport)
   tu_memclr(&_dcd, sizeof(dcd_data_t));
   edpt_reset_all(rhport); 
 
-  // disable all endpoints as epecified by LPC55S69 UM Table 778
+  // disable all endpoints as specified by LPC55S69 UM Table 778
   for(uint8_t ep_id = 0; ep_id < 2*MAX_EP_PAIRS; ep_id++)
   {
     _dcd.ep[ep_id][0].disable = _dcd.ep[ep_id][1].disable = 1;
--- a/src/portable/nxp/lpc_ip3511/dcd_lpc_ip3511.c
+++ b/src/portable/nxp/lpc_ip3511/dcd_lpc_ip3511.c
@@ -375,6 +375,8 @@ void dcd_edpt_close_all (uint8_t rhport)
 
 void dcd_edpt_close(uint8_t rhport, uint8_t ep_addr)
 {
+  (void) rhport;
+  
   uint8_t ep_id = ep_addr2id(ep_addr);
   _dcd.ep[ep_id][0].active = _dcd.ep[ep_id][0].active = 0; // TODO proper way is to EPSKIP then wait ep[][].active then write ep[][].disable (see table 778 in LPC55S69 Use Manual)
   _dcd.ep[ep_id][0].disable = _dcd.ep[ep_id][1].disable = 1;
@@ -403,8 +405,6 @@ static void prepare_ep_xfer(uint8_t rhport, uint8_t ep_id, uint16_t buf_offset, 
 
 bool dcd_edpt_xfer(uint8_t rhport, uint8_t ep_addr, uint8_t* buffer, uint16_t total_bytes)
 {
-  (void) rhport;
-
   uint8_t const ep_id = ep_addr2id(ep_addr);
 
   tu_memclr(&_dcd.dma[ep_id], sizeof(xfer_dma_t));
--- a/src/portable/nxp/lpc_ip3511/dcd_lpc_ip3511.c
+++ b/src/portable/nxp/lpc_ip3511/dcd_lpc_ip3511.c
@@ -178,7 +178,7 @@ typedef struct
 //    For example: LPC55s69 port1 Highspeed must be USB_RAM (0x40100000)
 //    Use CFG_TUSB_MEM_SECTION to place it accordingly.
 CFG_TUSB_MEM_SECTION TU_ATTR_ALIGNED(256) static dcd_data_t _dcd;
-CFG_TUSB_MEM_SECTION TU_ATTR_ALIGNED(256) static volatile uint8_t dummy[64]; // TODO temp fix to prevent accidental overwrite due to ep[][].buffer_xx.offset being 0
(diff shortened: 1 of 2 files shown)


tswan-quasi:
It seems that the checks are failing based on some MCUs that I did not modify files for? Did I do something wrong or how should I go about resolving this?

tswan-quasi:
> > It seems that the checks are failing based on some MCUs that I did not modify files for? Did I do something wrong or how should I go about resolving this?
> 
> no worries, it is ci issue, sometime when running multiple jobs too often. It runs out of bandwidth or got into network issue. I have re-run ci. Please give me a bit of time for reviewing and pulling out an lpc55 board for testing.

Ok no problem, thanks for letting me know

src/portable/nxp/lpc_ip3511/dcd_lpc_ip3511.c
```diff
@@ -174,6 +178,7 @@ typedef struct
 //    For example: LPC55s69 port1 Highspeed must be USB_RAM (0x40100000)
 //    Use CFG_TUSB_MEM_SECTION to place it accordingly.
 CFG_TUSB_MEM_SECTION TU_ATTR_ALIGNED(256) static dcd_data_t _dcd;
+CFG_TUSB_MEM_SECTION TU_ATTR_ALIGNED(256) static volatile uint8_t dummy[64]; // TODO temp fix to prevent accidental overwrite due to ep[][].buffer_xx.offset being 0
```
hathach, line 181: do we really need to have this. Have you experience issue with dcd data is overwritten. USB data shouldn't be accepted when active bit is not set

hathach (commented):
Thank you very much for the PR. I have tested both audio on HS and video on FS. Both works great, video HS does not seem to work (though it is probably issue with UVC or something else). The PR is almost ready for merging. Though I still have an question regarding dummy bytes, do we really need it. Have you experience issues without it ?

PS: I made some push to rename the enum name (also add an TODO for myself, since I found some mistake with fs/hs buffer layout when checking with LPC55 UM)

src/portable/nxp/lpc_ip3511/dcd_lpc_ip3511.c
tswan-quasi: I think there is likely a better solution, but it was causing me some issues not assigning the offsets to another location.

In functions such as ```dcd_init()``` and others, the USB would write to the offset in ```_dcd.ep[...][...].buffer_xx.offset``` and I think because the value of the ```DATABUFSTART``` register is 0x2000000, the USB was overwriting some other values I had in SRAM. The reason I noticed it was because the LED ```blink_interval_ms``` in ```main()``` was not changing correctly and it happened to be placed at 0x20000000 in my build. First place it happened was  after ```dcd_reg->DEVCMDSTAT  |= ...``` in ```dcd_init()```

tswan-quasi:
> video HS does not seem to work (though it is probably issue with UVC or something else).
I think the UVC was only setup with FS support unless something has changed from #1118 original description

src/portable/nxp/lpc_ip3511/dcd_lpc_ip3511.c
hathach: it is not clear enough that USB overwrite the data, even if that is the case, I think we should properly fix it e.g explicitly set SKIP/DISABLE bit for endpoint first before zero it out. Otherwise we could run into issue that host may think we already received data (overwritten to zero), while it is ended up in dummy. Maybe we should do it in a separated PR.

hathach:
> > video HS does not seem to work (though it is probably issue with UVC or something else).
> 
> I think the UVC was only setup with FS support unless something has changed from #1118 original description

agreed, it is probably the issue with UVC than the driver.

src/portable/nxp/lpc_ip3511/dcd_lpc_ip3511.c
tswan-quasi: I think I noticed it on the video example when I was testing because ```blink_interval_ms``` was getting overwritten at address 0x20000000. I will take a look again as well as seeing if I can find proper solution.

tswan-quasi: A repeatable setup for me:
- example: video_capture
- modifications: revert all lines using ```dummy``` to before
- OS: Windows
- when USB is asleep it blinks find. open the camera app, once the device is awake, the value of ```blink_interval_ms``` is 0 (address 0x20000000)

tswan-quasi: I think I have narrowed it down to ```dcd_edpt_xfer()``` when ```buffer``` is ```NULL```, occurring for EP0 OUT ZLPs. The code below makes it modify ```dummy```, and the other previous locations of ```dummy``` can be removed. From what I have seen, the USB writes 4 bytes of zeros to its offset location in this scenario. I am not sure if there is a proper solution to this aside from a dummy buffer?
```
bool dcd_edpt_xfer(uint8_t rhport, uint8_t ep_addr, uint8_t* buffer, uint16_t total_bytes)
{
  uint8_t const ep_id = ep_addr2id(ep_addr);

  tu_memclr(&_dcd.dma[ep_id], sizeof(xfer_dma_t));
  _dcd.dma[ep_id].total_bytes = total_bytes;

  if (!buffer && !ep_id)
  {
    buffer = (uint8_t*)(uint32_t)dummy;
  }
  prepare_ep_xfer(rhport, ep_id, get_buf_offset(buffer), total_bytes);

  return true;
}
```

hathach (approved):
Thank you very much for the PR. I am able to verified the issue with ZLP. It probably caused by USB/DMA controller, even though we schedule zero byte. The controller still clear/reset buffer probably due to its bug. I made it a bit generic to use dummy whenever we use ZLP for all endpoint (not only EP0) and for both direction. Will merge when ci passed.

PS: on my makefile setup, the overwritten memory is `SystemCoreClock` variable, and has no visual impact once setup and running.

PR #1683, opened 2022-10-14
Solution for usb memory dynamic allocated

mingpepe:
It should be asserted just like other weak function.


mingpepe:
Since the previoused pull request not merged, I change the title.  As disscussed in #540, memory fragmentation force drivers to manage memory. Some driver implement as release all memory unitl all endpoints(except ep0) closed. 

It may solve the problem for certain usb class devie, however for composite device, like me, report a HID & UVC device, since HID host driver never clear the interface, the endpoints for HID never closed, this cause memory for UVC endpoints not released.

Of course we can handle it in driver, but I think possibly many of drivers may encounter similar issue, we need a common memory allocator to manage memory for drvier.

Currently I have only rp2040, hence only verify on it.

mingpepe:
As discussed in #540, we can do it in a simple way.

PR #1690, opened 2022-10-18
esp: Fix Isochronous transfers

tore-espressif:
On ESP32-S2/S3 ISO IN packets must be configured for even or odd frame. If they're not, they are transmitted only in even frames, so every second packet is lost.

Closes https://github.com/hathach/tinyusb/issues/1382

**Describe the PR**
Fix Isochronous IN transfers on Espressif's SoCs.

cc @kaspernyhus @leeebo

### DIFF
--- a/src/portable/espressif/esp32sx/dcd_esp32sx.c
+++ b/src/portable/espressif/esp32sx/dcd_esp32sx.c
@@ -59,6 +59,7 @@ typedef struct {
     uint16_t queued_len;
     uint16_t max_size;
     bool short_packet;
+    uint8_t interval;
 } xfer_ctl_t;
 
 static const char *TAG = "TUSB:DCD";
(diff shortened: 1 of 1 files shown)


PanRe (approved):

tore-espressif:
cc @me-no-dev

PanRe:
@hathach I would vote to merge the PR, you are good with that?

leeebo:
The bugfix works for me, no frame is missed now.

iso_bug_fixed

By the way, the common dwc2 driver may have the same issue, which misses such configurations also https://github.com/hathach/tinyusb/blob/master/src/portable/synopsys/dwc2/dcd_dwc2.c#L662, as

PanRe:
@leebo Could you test if [there](https://github.com/hathach/tinyusb/blob/55db123a85d464462562f4c64efa8d6ff5734dce/src/portable/synopsys/dwc2/dcd_dwc2.c#L662) is a potential bug? If so, could you please open a new issue therefore? Many thanks in advance!

hathach (approved):
thank you very much for the fix. This is indeed out of my knowledge, we may also need to update the dwc2 as well to adopt this fix. Thank you @PanRe for reviewing and @leeebo for testing this out.

leeebo:
@PanRe The dwc2 port has handled this condition [here](https://github.com/hathach/tinyusb/blob/55db123a85d464462562f4c64efa8d6ff5734dce/src/portable/synopsys/dwc2/dcd_dwc2.c#L252), I'll test again to make sure.

leeebo:
@PanRe It works fine when using dwc2 driver for esp32s2/s3, no packet missed.

截图 2022-10-25 19-36-06

PR #1691, opened 2022-10-18
Fix typo in getting_started guide

hadess:
Fixes: 1234111 ("add get-deps target")


hathach (approved):
thank you for the fix

PR #1698, opened 2022-10-20
do not apply storage attributes to member of struct

battlesnake:
**Describe the PR**
Compilation fails when audio interrupt-endpoint is enabled, due to storage attributes on a struct member.
I removed the attributes.

**Additional context**

### DIFF
--- a/src/class/audio/audio_device.c
+++ b/src/class/audio/audio_device.c
@@ -341,7 +341,7 @@ typedef struct
 
   // Audio control interrupt buffer - no FIFO - 6 Bytes according to UAC 2 specification (p. 74)
 #if CFG_TUD_AUDIO_INT_CTR_EPSIZE_IN
-  CFG_TUSB_MEM_SECTION CFG_TUSB_MEM_ALIGN uint8_t ep_int_ctr_buf[CFG_TUD_AUDIO_INT_CTR_EP_IN_SW_BUFFER_SIZE];
+  uint8_t ep_int_ctr_buf[CFG_TUD_AUDIO_INT_CTR_EP_IN_SW_BUFFER_SIZE];
 #endif
 
   // Decoding parameters - parameters are set when alternate AS interface is set by host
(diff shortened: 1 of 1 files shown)


hathach (changes requested):
it does not fail with my compiling. This is required for some mcu. Could you provide info on your setup

battlesnake:
This is using GCC 12.2.0, st/synopsys port, STM32H7.

The attributes (`CFG_TUSB_MEM_SECTION CFG_TUSB_MEM_ALIGN`) are being applied to a _member_ of a struct.
I don't think it makes sense (regarding memory layout) for some members of a struct to be in a different linker section to the rest of the struct?

The only instance of the struct is (around line 413):
```
CFG_TUSB_MEM_SECTION audiod_function_t _audiod_fct[CFG_TUD_AUDIO];
```

So the struct (and therefore its members) are already in the section denoted by `CFG_TUSB_MEM_SECTION`.
The alignment is a separate issue, perhaps that attribute should stay.

Setting `CFG_TUD_AUDIO_INT_CTR_EPSIZE_IN` to non-zero (e.g. 6) enables that struct member.
If you have `CFG_TUSB_MEM_SECTION` defined with a specific section, then this combination should be sufficient to trigger a compiler error.

For configuration, I have:
```
#define _d1ram_bss __attribute__((__section__(".ram_d1.bss")))
...
#define dma_align_bytes (32)
#define _dma_align __attribute__((__aligned__(dma_align_bytes)))
...
#define CFG_TUSB_MEM_SECTION _d1ram_bss
#define CFG_TUSB_MEM_ALIGN _dma_align
```

Which, expanded, is:
```
#define CFG_TUSB_MEM_SECTION __attribute__((__section__(".ram_d1.bss")))
#define CFG_TUSB_MEM_ALIGN __attribute__((__aligned__(32)))
```

battlesnake:
Currently, the `ep_int_ctr` member is used, but never assigned.

So I'm guessing that I'm the first person to try building a UAC2 configuration with interrupt/status endpoint enabled.

hathach:
> I don't think it makes sense (regarding memory layout) for some members of a struct to be in a different linker section to the rest of the struct? GCC certainly doesn't like this.

Oh, this is right, please remove the CFG_TUSB_MEM_SECTION but keep the `CFG_TUSB_MEM_ALIGN`

battlesnake:
Done

hathach (approved):
perfect, thank you

Issue #1697, opened 2022-10-20
Interrupt-endpoint for audio class


battlesnake:
### Related area

Audio class UAC2

### Hardware specification

STM32H7 (probably not relevant)

### Is your feature request related to a problem?

UAC2 supports an interrupt endpoint to provide notifications to the host about control changes (e.g. mute button pressed on the device).

I see that there is some code for this, but it seems incomplete.
I see `audio_device.c` never assigns a value to `audio->ep_int_ctr`.
Compare to feedback endpoint (which I have working), which assigns and clears `audio->ep_fb` as needed.

### Describe the solution you'd like

Support for the interrupt endpoint in Audio Class devices.

I can implement this myself, but any advice/guidance beforehand would be appreciated!

### I have checked existing issues, dicussion and documentation

- [X] I confirm I have checked existing issues, dicussion and documentation.

battlesnake:
Working, but possibly ugly implementation: https://github.com/hathach/tinyusb/pull/1702

Gadgetoid:
I came across the linked PR while searching for a means to synchronise device control changes (volume and mute in particular) to the host. I'd like to register my interest in this feature.

battlesnake:
Sure.  My fork of tinyusb has a few extra features implemented for audio class, but didn't get merged as I don't have the boards used by the mainline project for testing.
If you don't need any newer features than my fork (or official support from the project), you could potentially use my fork.

Gadgetoid:
> If you don't need any newer features than my fork (or official support from the project), you could potentially use my fork.

Unfortunately I think I need both, I'd hoped I could get away with cherry-picking some bits from your fork but so far it's mostly lead to eclectic bugs and crashes.

Thanks for blazing this trail, anyway, there are surely others out there like me who think using media keys to have the host control the volume is barbaric 😆

Issue #1684, opened 2022-10-15
ST Synopsys port lacks SOF support


battlesnake:
### Related area

ST Synopsys port

### Hardware specification

Cortex M7, STM32H750, USB3300, custom board

### Is your feature request related to a problem?

The port currently lacks support for USB SOF frames.

### Describe the solution you'd like

I will provide a PR with my proposed solution

### I have checked existing issues, dicussion and documentation

- [X] I confirm I have checked existing issues, dicussion and documentation.

battlesnake:
PR: https://github.com/hathach/tinyusb/pull/1685

battlesnake:
I didn't use try it.  I didn't find info in the docs about which port(s) to use, so I tried `st/synopsys` (since using ST part), and it worked.

battlesnake:
Besides (presumably) supporting SOF already, what are other differences of that synopsys/dw2 port vs. st/synopsys?

PanRe:
It fuses the device driver for multiple MCUs as far as i know. I don't know why the `st/synopsis` driver is still there, maybe due to backward compatibilities.

battlesnake:
Thanks, I'll give it a try in an hour or so.

battlesnake:
Using synopsis/dwc2, I get into a crash-loop.
Almost certainly due to configuration on my side though.

I'll go back to using st/synopsys for now.
Once I've got my UAC2 interrupt status endpoint working, then I'll try synopsys/dwc2 again and see if I can either fix the issue in my code, or file a PR if it's due to a bug in tinyusb.

PR #1702, opened 2022-10-23
UAC2 supports interrupt-endpoint for providing control-change notifications to the host

battlesnake:
Relates to issue https://github.com/hathach/tinyusb/issues/1697

**Describe the PR**

UAC2 supports an interrupt endpoint to provide notifications to the host about control changes (e.g. mute button pressed on the device).

The interrupt endpoint lives in the control interface, and transmits 6-byte messages to notify the host about control values changing.
The host may then decide to issue a get-entity request via the control endpoint in order to read the new control value.

These interrupts can notify about changes to endpoints, interfaces, entities/controls, and also provide vendor-specific notifications.

The implementation of this functionality in TinyUSB was incomplete.

I have tested this on a custom board using STM32H7 and no RTOS.
On Linux, I see ALSAMIXER's UI controls on the host PC change, when I interact with the physical controls on the device.

I'm not too familiar with the coding style or architecture for TinyUSB, so please give suggestions on making this PR conform with the existing style.

This PR also includes the fix from https://github.com/hathach/tinyusb/pull/1698, where compilation was failing (when interrupt-endpoint enabled) due to a "section" attribute being applied to a _member_ within a struct.

**Additional context**


src/class/audio/audio_device.h
```diff
@@ -531,7 +541,33 @@ TU_ATTR_WEAK TU_ATTR_FAST_FUNC void tud_audio_feedback_interval_isr(uint8_t func
 
 #endif // CFG_TUD_AUDIO_ENABLE_EP_OUT && CFG_TUD_AUDIO_ENABLE_FEEDBACK_EP
 
-#if CFG_TUD_AUDIO_INT_CTR_EPSIZE_IN
+#if CFG_TUD_AUDIO_ENABLE_INTERRUPT_EP
+// UAC2 §9.6
```
HiFiPhile, line 545: Thanks for your PR !

Which UAC2 document are you using ? In my side `Interrupt Data Message` is section 6.1 p129.

battlesnake: Reference fixed

battlesnake:
Also, to get this (and feedback) working on my STM32H7 board, I also had to use this patch (for enabling SOF interrupt):
https://github.com/hathach/tinyusb/issues/1684

This is with the st/synopsys port.
I recently and briefly tried the synopsis/dwc2 port, but it didn't work out-of-the-box.
So I switch back to st/synopsys for now, and will look into the dwc2 issue sometime in the future.

src/class/audio/audio_device.c
```diff
@@ -1079,6 +1081,52 @@ static inline bool audiod_fb_send(uint8_t rhport, audiod_function_t *audio)
...
+  uint8_t const *p_desc = audio->p_desc;
+  // Get pointer at end
+  uint8_t const *p_desc_end = audio->p_desc + audio->desc_length - TUD_AUDIO_DESC_IAD_LEN;
+
+
+  // p_desc starts at required interface with alternate setting zero
+  while (p_desc < p_desc_end)
+  {
+    // Find correct interface
+    if (tu_desc_type(p_desc) == TUSB_DESC_INTERFACE && ((tusb_desc_interface_t const * )p_desc)->bInterfaceNumber == 0 && ((tusb_desc_interface_t const * )p_desc)->bAlternateSetting == 0)
```
HiFiPhile, line 1097: `bInterfaceNumber` could be other than 0, it's defined by `_intfnum` https://github.com/hathach/tinyusb/blob/28f49c088bb0c498d730d80943017172061cfd05/src/device/usbd.h#L442.

@PanRe could you take a look ? I havn't work on UAC for a moment 🤣

battlesnake: Ah sorry, this is leftover from when I initially hacked it together for my specific project.  I'll generalise that soon.

Also, since there are several places where we sequentially search the descriptors for some specific descriptor, should I create a separate function for doing that, which takes a predicate as an argument?
It should reduce code-size a bit while not burning a significant amount of extra cycles as a result.
And the CPU overhead shouldn't matter much anyway since this is set-up stuff rather than "in the loop" stuff.

HiFiPhile: > Also, since there are several places where we sequentially search the descriptors for some specific descriptor, should I create a separate function for doing that, which takes a predicate as an argument?

I think there is no need to do it now, later we can generalize them with a function like `tu_desc_search()` if needed.

PanRe: True, this number can be any interface number!

PanRe:
Thanks a lot! Indeed the interrupt EP was not implemented correctly! Since you are the first one who needs it, i am glad you fix the issues!
One thing, the naming convention `CFG_TUD_AUDIO_INT_CTR_EPSIZE_IN` was wrongly introduced and is misleading - `CTR` has nothing to do with the interrupt EP, so it should be skipped. Could you fix that if you have the time?

src/class/audio/audio_device.c
PanRe: You may also skip that check since you already work on the descriptor corresponding to the right interrupt EP. Do we need to remember which INT EP corresponds to which interface? For sending messages via INT EP?

battlesnake:
INT_CTR to INT refactor done

src/class/audio/audio_device.c
battlesnake: Done

PanRe:
Thanks a lot and sorry for the late reply! I have but a few questions on this:
- do we need a separate buffer for the interrupt messages to be sent?
- where is the interrupt EP communication handled? I guess its outside the UAC2 driver currently. Does it make sense to incorporate the INT EP comunications somehow into the UAC2 driver?
- i know this could be a lot of work but i wanted to ask: could you come up with some minimum working example such that other users could start on INT EPs within UAC2 right away? That would be awsome! :)

battlesnake:
The buffer within the tinyusb audio device structure was there already when I started this work - I have no idea if it's required or if we can bypass it.

My board + build system are quite custom so I can't easily extract an example out or test an example project.

Here's the relevant code though for using this feature.  I have a buffer containing three pre-determined messages, which I send incrementally when the audio controls have changed (which can happen by user using physical controls, or me testing via serial interface, or by the USB host setting parameters via existing tinyusb functionality).

## usb.c (implementations of tinyusb event handlers and tinyusb basics)
```
...

/* Called as part of main loop */
void usb_update()
{
#if CFG_TUD_AUDIO_ENABLE_INTERRUPT_EP
	usb_control_status_update();
#endif
#if CFG_TUD_AUDIO_ENABLE_FEEDBACK_EP
	usb_feedback_config_update();
#endif
	tud_task_ext(0, false);
}

...

#if CFG_TUD_AUDIO_ENABLE_INTERRUPT_EP
bool tud_audio_int_done_cb(_unused uint8_t rhport, _unused uint16_t n_bytes_copied)
{
	return usb_control_status_done();
}
#endif
```

## usb_control_status.c (manages notifying host about control changes)
```
#include ...

#include <tusb.h>

#if CFG_TUD_AUDIO_ENABLE_INTERRUPT_EP

#define MAX_STATUS_MESSAGES_QUEUED 3
#define AUDIO_ATTR_CUR AUDIO_CS_REQ_CUR
#define STATUS_INTERRUPT_INFO_VENDOR_SPECIFIC 0x01
#define STATUS_INTERRUPT_INFO_ORIGIN_ENDPOINT 0x02
#define STATUS_INTERRUPT_INFO_ORIGIN_INTERFACE 0x00
#define STATUS_INTERRUPT_CONTROL_MASTER 0x00

static const audio_status_update_t status_message_queue[MAX_STATUS_MESSAGES_QUEUED] = {
	{
		.bInfo = STATUS_INTERRUPT_INFO_ORIGIN_INTERFACE,
		.bAttribute = AUDIO_CS_REQ_CUR,
		.wValue_cn_or_mcn = STATUS_INTERRUPT_CONTROL_MASTER,
		.wValue_cs = AUDIO_FU_CTRL_VOLUME,
		.wIndex_ep_or_int = ITF_NUM_AUDIO_CONTROL,
		.wIndex_entity_id = TERM_FEATURE_ID,
	},
	{
		.bInfo = STATUS_INTERRUPT_INFO_ORIGIN_INTERFACE,
		.bAttribute = AUDIO_CS_REQ_CUR,
		.wValue_cn_or_mcn = STATUS_INTERRUPT_CONTROL_MASTER,
		.wValue_cs = AUDIO_FU_CTRL_MUTE,
		.wIndex_ep_or_int = ITF_NUM_AUDIO_CONTROL,
		.wIndex_entity_id = TERM_FEATURE_ID,
	},
	{
		.bInfo = STATUS_INTERRUPT_INFO_ORIGIN_INTERFACE,
		.bAttribute = AUDIO_CS_REQ_CUR,
		.wValue_cn_or_mcn = STATUS_INTERRUPT_CONTROL_MASTER,
		.wValue_cs = AUDIO_FU_CTRL_GRAPHIC_EQUALIZER,
		.wIndex_ep_or_int = ITF_NUM_AUDIO_CONTROL,
		.wIndex_entity_id = TERM_FEATURE_ID,
	},
};

static uint32_t status_message_next = MAX_STATUS_MESSAGES_QUEUED;

/* Sends the next message and increments the message counter */
static void status_message_send_next()
{
	if (status_message_next >= MAX_STATUS_MESSAGES_QUEUED) {
		return;
	}
	usb_control_status_log("msg#=%d", status_message_next);
	const void *msg = &status_message_queue[status_message_next];
	status_message_next++;
	tud_audio_int_write(msg, 6);
}

void usb_control_status_update()
{
	/* If any control values have changed since our observer last polled, notify host */
	if (control_update(CONTROL_OBSERVER_USB)) {
		status_message_next = 0;
		status_message_send_next();
	}
}

bool usb_control_status_done()
{
	status_message_send_next();
	return true;
}

#endif
```

battlesnake:
Almost forgot, relevant part of descriptor.  Goes in control-interface part of descriptor, i.e. before any of the streaming interfaces/alternates.

## descriptor
```
#if USB_ENABLE_STATUS_EP
#define STATUS_INT_EP_DESC_LEN TUD_DESC_EP_LEN
#define STATUS_INT_EP_DESC() \
		TUD_DESC_EP( \
			ENDPOINT_AUDIO_INTERRUPT, \
			HOSTIN_AUDIO_INTERRUPT_TYPE, \
			6 /* Buffer size as per standard */, \
			CONTROL_UPDATE_INTERVAL_LOG2P1 \
		),
#else
#define STATUS_INT_EP_DESC_LEN 0
#define STATUS_INT_EP_DESC()
#endif
```

battlesnake:
Project-specific implementation details in my case:

Basically - each thing that cares about the audio controls has a value in a `CONTROL_OBSERVER_*` enum to represent it.
These are "observers", example:
 - The audio processor unit is an observer of the controls.
 - The USB control/status unit is an observer of the controls.
 - The serial user-interface is also an observer of the controls.

There's an array of observer flags (`bool`), one for each observer.

Whenever a control value changes (i.e. `new value != old value`, not just a "*_set method called") then _all_ observer flags are set to true.
When an observer is polled (e.g. `if (control_update(CONTROL_OBSERVER_USB))` in the above code), then we reset the flag to false while returning its previous value.

This way, whenever anything changes audio parameters, all observers get to know about it whenever they next poll during their update, and they can handle it each their own way.

Probably over-complicated for a basic demo, where you could probably just have a single boolean flag.

silvergasp:
Looks like your CI failure for the RP2040 build could be fixed in this PR #1716

battlesnake:
@maxsimmonds1337 Info is here if you fancy building an example/demo.

maxsimmonds1337:
> @maxsimmonds1337 Info is here if you fancy building an example/demo.

Perfect, thanks @battlesnake

Gadgetoid:
Has anyone got this working?

I've attempted to reconcile these changes with the version of tinyusb used by Pico SDK, but at best nothing happens and at worst I can reliably crash my board. This happens when I attempt to adjust the volume from the host after calling `tud_audio_int_ctr_n_write` .

There's an awful work in progress mess here:

https://github.com/Gadgetoid/pico-tinyusb-i2s-speaker/commit/84c127f5296ab82b9262720d78d13360cd917d4f

This doesn't show my futile modifications to tinyusb, which mostly involve trying to set `ep_int_ctr` to the correct value, since afaict everything else is covered.

HiFiPhile:
I can take a look later, but there are too much pending changes in UAC class not merged yet...

Gadgetoid:
> I can take a look later

Thank you!

> I can take a look later, but there are too much pending changes in UAC class not merged yet...

😬 I guess I'm going to have some fixes to do in future!

HiFiPhile (approved):

Gadgetoid:
> Sorry for the extra long delay

Looks like three weeks since I bumped it, that's a fast turnaround IMO and it's hugely appreciated! (We'll pretend the PR isn't years old because something something glass houses and stones 😬)

I have updated my code and replaced the tinyusb submodule in my local clone of pico-sdk. Everything works as I'd expect with our RP2040-based design. Including mute/unmute. Full changeset on my not-actually-a-headset-but-two-speakers codebase here: https://github.com/Gadgetoid/pico-tinyusb-i2s-speaker/commit/6032314ca90401bccc654a3788b161abaca304ee

Tested on RP2040, macOS (Ventura) and Linux (Raspberry Pi OS bookworm, 6.6.20+rpt-rpi-2712). 🥳 

Thanks, both, for making audio on the Pico just that little bit slicker.

battlesnake:
Hooray!

I can switch back to mainline for my audio projects, instead of using my fork :D

This PR should hopefully bring in 2-way sync between device and host for control values.

Sometime I'll try to add support to the Linux UAC2 driver for multi-band equaliser controls too, which it currently seems to totally lack.

PR #1709, opened 2022-11-01
Add audio_test_freertos & audio_4_channel_mic_freertos

kaspernyhus:
Add freertos versions of the two audio examples for use on the ESP32-s2/3


PanRe (approved):

HiFiPhile:
Sorry for huge delay, since then both examples received many updates, if you still work on it would be appreciated to port them.

PR #1718, opened 2022-11-03
Fixed wrong FreeRTOS port for EFM32GG12

hfegran:
Changed Freertos port for EFM32GG12 to ARM_CM4F as EFM32GG12 is not an M3, but rather an M4F based MCU
Fixes hid_composite_freertos example for SLTB009a


hathach (approved):
yeah, right. Thank you for the fix.

PR #1738, opened 2022-11-11
MIDI: support port name strings

kaysievers:
This allows to set custom descriptive names for virtual MIDI ports/cables. A MIDI port offers different functionality on the same physical USB device. The names make it easier to distinguish one from the other.

This is the tinyusb part needed for https://github.com/adafruit/Adafruit_TinyUSB_Arduino/pull/215


hathach (approved):
look great, thank you

PR #1767, opened 2022-11-30
Allow the use of non-static allocation for FreeRTOS mutexes & queues

pete-pjb:
## ***Allow the use of non-static allocation for FreeRTOS mutexes & queues***

Added some conditional code to the file ```src/osal/osal_freertos.h``` to check the FreeRTOS configuration option ```configSUPPORT_STATIC_ALLOCATION``` and if it is set to **'0'** We use the non-static variants of the queues, mutexes and associated creation functions.

### DIFF
--- a/src/host/usbh.c
+++ b/src/host/usbh.c
@@ -509,7 +509,7 @@ void tuh_task_ext(uint32_t timeout_ms, bool in_isr)
       default: break;
     }
 
-#if CFG_TUSB_OS != OPT_OS_NONE && CFG_TUSB_OS != OPT_OS_PICO
+#if CFG_TUSB_OS != OPT_OS_NONE && CFG_TUSB_OS != OPT_OS_PICO && CFG_TUSB_OS != OPT_OS_FREERTOS
     // return if there is no more events, for application to run other background
     if (osal_queue_empty(_usbh_q)) return;
 #endif
--- a/src/host/usbh.c
+++ b/src/host/usbh.c
@@ -246,8 +246,8 @@ static uint8_t _usbh_ctrl_buf[CFG_TUH_ENUMERATION_BUFSIZE];
 
 // Control transfer: since most controller does not support multiple control transfer
 // on multiple devices concurrently. And control transfer is not used much except enumeration
-// We will only execute control transfer one at a time.
-struct
+// We will only execute control transfer one at a time.CFG_TUSB_MEM_SECTION TU_ATTR_ALIGNED(4096)
+CFG_TUSB_MEM_SECTION struct
 {
   tusb_control_request_t request TU_ATTR_ALIGNED(4);
   uint8_t* buffer;
--- a/src/host/usbh.c
+++ b/src/host/usbh.c
@@ -246,8 +246,8 @@ static uint8_t _usbh_ctrl_buf[CFG_TUH_ENUMERATION_BUFSIZE];
 
 // Control transfer: since most controller does not support multiple control transfer
 // on multiple devices concurrently. And control transfer is not used much except enumeration
-// We will only execute control transfer one at a time.CFG_TUSB_MEM_SECTION TU_ATTR_ALIGNED(4096)
-CFG_TUSB_MEM_SECTION struct
+// We will only execute control transfer one at a time.
+struct
 {
   tusb_control_request_t request TU_ATTR_ALIGNED(4);
   uint8_t* buffer;
@@ -509,7 +509,7 @@ void tuh_task_ext(uint32_t timeout_ms, bool in_isr)
       default: break;
     }
 
-#if CFG_TUSB_OS != OPT_OS_NONE && CFG_TUSB_OS != OPT_OS_PICO && CFG_TUSB_OS != OPT_OS_FREERTOS
+#if CFG_TUSB_OS != OPT_OS_NONE && CFG_TUSB_OS != OPT_OS_PICO
     // return if there is no more events, for application to run other background
     if (osal_queue_empty(_usbh_q)) return;
 #endif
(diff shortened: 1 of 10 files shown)


pete-pjb:
Hi @hathach ,

It seems one of the CI tests is failing (***Build ARM / build-arm (rp2040) (pull_request)***) ,  I have looked at the 'Details', there seems to be no clue as to why?  If I need to change something, please let me know.

Kind Regards,

Pete

hathach:
it is network/bandwidth issue, happen quite often now and then. Just re-run it.

pete-pjb:
Hi @hatach ,

I have also added another 'FreeRTOS' related commit to this pull request to eliminate the need to repeatedly call the tuh_task() when using FreeRTOS.  You can just start tuh_task() as a normal task from FreeRTOS like this:

````c
xTaskCreate( (TaskFunction_t)tuh_task, "USB", THREAD_STACKSIZE, NULL, USB_PRIORITY, NULL );
````

This eliminates the very small overhead incurred when calling the function continuously in a loop from another FreeRTOS task which is not necessary whilst the tuh_task_ext() is being executed in it's own loop and blocking using the OS specific calls...  I hope that makes sense.  

Let me know your thoughts...

Kind Regards,

Pete

**Comment by pete-pjb** [comment]
Hi @hathach ,

I have also added another commit here, I added the ```CFG_TUSB_MEM_SECTION``` macro to the declaration of ```struct _ctrl_xfer``` in ```usbh.c```, this is required for my port also.

I hope that is okay.

Kind Regards,

Pete

File: src/host/usbh.c [hunk 1]
```diff
@@ -246,8 +246,8 @@ static uint8_t _usbh_ctrl_buf[CFG_TUH_ENUMERATION_BUFSIZE];
 
 // Control transfer: since most controller does not support multiple control transfer
 // on multiple devices concurrently. And control transfer is not used much except enumeration
-// We will only execute control transfer one at a time.
-struct
+// We will only execute control transfer one at a time.CFG_TUSB_MEM_SECTION TU_ATTR_ALIGNED(4096)
+CFG_TUSB_MEM_SECTION struct
```
hathach, line 250: please revert this change, only the buffer needed to be in the usb section, the xfer info is not

src/host/usbh.c
```diff
@@ -509,7 +509,7 @@ void tuh_task_ext(uint32_t timeout_ms, bool in_isr)
       default: break;
     }
 
-#if CFG_TUSB_OS != OPT_OS_NONE && CFG_TUSB_OS != OPT_OS_PICO
+#if CFG_TUSB_OS != OPT_OS_NONE && CFG_TUSB_OS != OPT_OS_PICO && CFG_TUSB_OS != OPT_OS_FREERTOS
```
hathach, line 512: please revert this change, since we want application to be able to share usb with other background if needed

src/osal/osal_freertos.h
```diff
@@ -62,12 +62,20 @@ TU_ATTR_ALWAYS_INLINE static inline void osal_task_delay(uint32_t msec)
...
+typedef SemaphoreHandle_t osal_semaphore_def_t;
+#endif
 typedef SemaphoreHandle_t osal_semaphore_t;
 
 TU_ATTR_ALWAYS_INLINE static inline osal_semaphore_t osal_semaphore_create(osal_semaphore_def_t* semdef)
 {
+#if configSUPPORT_STATIC_ALLOCATION == 1
   return xSemaphoreCreateBinaryStatic(semdef);
+#else
+  return xSemaphoreCreateBinary();
```
hathach, line 77: this will throw unused parameter warnings with semdef

hathach (changes requested):
thank you for the PR, though it is better to revert some of your recent changes.

src/osal/osal_freertos.h
```diff
@@ -145,7 +162,11 @@ typedef QueueHandle_t osal_queue_t;
 
 TU_ATTR_ALWAYS_INLINE static inline osal_queue_t osal_queue_create(osal_queue_def_t* qdef)
 {
-  return xQueueCreateStatic(qdef->depth, qdef->item_sz, (uint8_t*) qdef->buf, &qdef->sq);
+#if configSUPPORT_STATIC_ALLOCATION == 1
+	return xQueueCreateStatic(qdef->depth, qdef->item_sz, (uint8_t*) qdef->buf, &qdef->sq);
```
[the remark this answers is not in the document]
pete-pjb: Hi @hathach ,

I did check this and it seems it should always be available the way FreeRTOS is written, but for good measure I have added the ```#if configSUPPORT_STATIC_ALLOCATION``` clause to the ```sq``` variable, just in case anything changes in the future. 🙂 

Cheers,

Pete

src/host/usbh.c
pete-pjb: As you have requested I have reverted this commit but, I still have an issue with this as my port will not function without it,  I will look into this further and let you know what I find...  I suspect it is some sort of caching problem.

New commit will follow shortly...

Cheers,
Pete

src/osal/osal_freertos.h
pete-pjb: My bad sorry...  I have added (void) parameter...

New commit to follow...

Cheers,

Pete

src/host/usbh.c
pete-pjb: I have also reverted this commit but I would like to discuss this further with you, I will create a discussion topic for it if you don't mind?

Cheers,

Pete

hathach (approved):
thank you for your PR. I have pulled and tested with existing examples on actual hardware. It worked great, I also clean up and put more update to existing examples to allow it to run with `configSUPPORT_STATIC_ALLOCATION=0`

PR #1779, opened 2022-12-06
rp2040: avoid device-mode state machine hang

P33M:
Please see the commit text for a description of the change.

Note that I now export hw_endpoint_start_next_buffer() and hw_endpoint_lock_update() in the header file, and because both the IRQ and driver code are doing read-modify-write operations the endpoint lock really should be implemented.


hathach:
can you provide context and reason for this PR ?

P33M:
This avoids a bug in rp2040. The device controller stops responding to host packets if it hits the condition described in the commit.

The state is effectively unrecoverable without a complete peripheral reset (not a USB bus reset) so it is far better to avoid it in the first place - host drivers do not respond well to a device that suddenly disappears from the bus.

hathach:
thanks for explaining, do we have an document for this bug in rp2040 i.e is it an errata or a note somewhere in the rp2040 document. If possible, could you detail how is your hw setup and step to reproduce the issue if possible. I would like to try out the before and after pr to see if that would fix the issue and/or making change update later on without breaking this fix.

P33M:
An errata note will be published in due course.

I have pushed a testcase to https://github.com/P33M/rp2040-e15-testcase - this is an SDK-compatible widget that defines a vendor-specific interface using TinyUSB. Bulk OUT packets are discarded, bulk IN packets are pseudorandomly sized and submitted to the hardware at arbitrary times.
The expected behaviour with the included python script is that the transfer will continue forever, but with a pico plugged into a Pi 4 directly, will crash after a few seconds.

P33M:
I don't know why the CI build is failing. It builds fine locally under CMake.

P33M:
Please merge - we consider it important enough that we are doing a point release of the Pico C SDK to pick up this change.

hathach:
> Please merge - we consider it important enough that we are doing a point release of the Pico C SDK to pick up this change.

In this case, I will try to review the code changes without hw testing as soon as I could in the next couple of days. And will try to do hand-on testing later on.

src/portable/raspberrypi/rp2040/rp2040_usb.c
```diff
@@ -57,6 +55,36 @@ TU_ATTR_ALWAYS_INLINE static inline bool is_host_mode(void)
...
+      ep->transfer_type == TUSB_XFER_INTERRUPT ||
+      ep->transfer_type == TUSB_XFER_ISOCHRONOUS)
+    return 0;
+
+  /* Avoid the last 200us (uframe 6.5-7) of a frame, up to the EOF2 point.
+   * The device state machine cannot recover from receiving an incorrect PID
+   * when it is expecting an ACK.
+   */
+  now = time_us_32();
+  if (now < (last_sof + 800) || now > (last_sof + 998)) {
```
kilograham, line 77: i would do
```
delta =  time_us_32() - last_sof
```
to avoid rollover issues (not a biggy but still)

src/portable/raspberrypi/rp2040/dcd_rp2040.c
```diff
@@ -247,16 +249,35 @@ static void __tusb_irq_path_func(reset_non_control_endpoints)(void)
...
     if (status & USB_INTF_DEV_SOF_BITS)
     {
       handled |= USB_INTF_DEV_SOF_BITS;
 
+#if TUD_OPT_RP2040_USB_DEVICE_UFRAME_FIX
+      parked_eps = (uint32_t)pending_nonperiodic_xfer;
+      last_sof = time_us_32();
+
+      // Kick any deferred nonperiodic IN endpoints
+      for (i = 0; i < USB_MAX_ENDPOINTS; i++) {
```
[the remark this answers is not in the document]
P33M: True, most applications will use an average of 2 low-numbered EPs for bulk.

src/portable/raspberrypi/rp2040/rp2040_usb.c
P33M: Rollover would mean the time check would not apply for one frame in every 4.29 million, which is arguably a bug.

src/portable/raspberrypi/rp2040/dcd_rp2040.c
```diff
@@ -390,6 +412,9 @@ void dcd_init (uint8_t rhport)
   usb_hw->sie_ctrl = USB_SIE_CTRL_EP0_INT_1BUF_BITS;
   usb_hw->inte     = USB_INTS_BUFF_STATUS_BITS | USB_INTS_BUS_RESET_BITS | USB_INTS_SETUP_REQ_BITS |
                      USB_INTS_DEV_SUSPEND_BITS | USB_INTS_DEV_RESUME_FROM_HOST_BITS |
+#if TUD_OPT_RP2040_USB_DEVICE_UFRAME_FIX
+                     USB_INTS_DEV_SOF_BITS |
```
hathach, line 416: SOF should only be enabled when needed.

src/portable/raspberrypi/rp2040/dcd_rp2040.c
```diff
@@ -247,16 +249,36 @@ static void __tusb_irq_path_func(reset_non_control_endpoints)(void)
...
+      for (i = 0; parked_eps && i < USB_MAX_ENDPOINTS; i++) {
+        if (parked_eps & (1u << i)) {
+          parked_eps &= ~(1u << i);
+          ep = hw_endpoint_get_by_num(i, TUSB_DIR_IN);
+          hw_endpoint_lock_update(ep, 1);
+          hw_endpoint_start_next_buffer(ep);
+          hw_endpoint_lock_update(ep, -1);
+        }
+      }
+      pending_nonperiodic_xfer = 0;
```
hathach, line 276: we should only clear what we handled i.e `parked_eps`

```C
pending_nonperiodic_xfer &= ~parked_eps
```

However, since pending_nonperiodic_xfer is accessed by both dcd_edpt_xfer() and sof isr. There is still chance for race condition since both try to read-then-set the variable.

src/portable/raspberrypi/rp2040/rp2040_usb.c
```diff
@@ -218,8 +246,14 @@ void hw_endpoint_xfer_start(struct hw_endpoint *ep, uint8_t *buffer, uint16_t to
   ep->active        = true;
   ep->user_buf      = buffer;
 
-  _hw_endpoint_start_next_buffer(ep);
-  _hw_endpoint_lock_update(ep, -1);
+  if(!rp2040_critical_frame_period(ep)) {
+    hw_endpoint_start_next_buffer(ep);
+  } else {
+#if TUD_OPT_RP2040_USB_DEVICE_UFRAME_FIX
+    pending_nonperiodic_xfer |= (uint16_t)(1 << tu_edpt_number(ep->ep_addr));
```
hathach, line 253: `pending_nonperiodic_xfer` are both read-then-set in both here via dcd_edpt_xfer() and sof isr. There is a race condition e.g 
- we need to set with 0x02 mask, after the read here when pending_nonperiodic_xfer = 0x01
- sof isr occurred, process the 0x01 and clear the bit, sof return
- pending_nonperiodic_xfer is set with 0x03 therefore incorrectly schedule an redundant transfer.

src/portable/raspberrypi/rp2040/rp2040_usb.c
```diff
@@ -329,15 +363,15 @@ bool __tusb_irq_path_func(hw_endpoint_xfer_continue)(struct hw_endpoint *ep)
...
                ep->xferred_len, tu_edpt_number(ep->ep_addr), ep_dir_string[tu_edpt_dir(ep->ep_addr)]);
     // Notify caller we are done so it can notify the tinyusb stack
-    _hw_endpoint_lock_update(ep, -1);
+    hw_endpoint_lock_update(ep, -1);
     return true;
   }
   else
   {
-    _hw_endpoint_start_next_buffer(ep);
+    hw_endpoint_start_next_buffer(ep);
```
hathach, line 371: rp2040_critical_frame_period() should also be checked here as well, since this will also trigger an usb transaction. I think this place will likely to cause the issue than one in dcd_edpt_xfer() since it schedule packet back-to-back here.

hathach (commented):
late review since I was busy with other works. After going through the change, although the pr fix a severe issue, however due to the impact of this PR in performance i.e 20% off bandwidth and force to have sof (which will hurt power consumption). I would like to do actual hand-on testing with the PR before merging. Though I couldn't do that until another week. 

Meanwhile there is a few issues I think we should fix first
1. Since this only occurs with bulk in transfer, we should only enable SOF when needed i.e when there is active bulk-in within 20% last frame. This will minimize the impact on current consumption. Since not all device make use of Bulk in endpoint e.g HID device. And even with MSC/cdc device, it is more often than not, device sitting idle than transferring data. and when it does, it is more often within first 80% frame.
2. hw_endpoint_xfer_continue() should also check of critical frame i.e all places where hw_endpoint_start_next_buffer() is called to trigger usb transaction
3. There is a race condition with `pending_nonperiodic_xfer` since both dcd_edpt_xfer() and sof isr try to read-then-set it. We should use an bit in hw_endpoint to address this so that each endpoint is set and clear invidually.

Let me know if above request makes sense and feel free to provide feedback on them. I am open to all discussion.

hathach:
Update: SOF is only used to sync time, maybe we don’t need to sync it every frame, and could only need to sync every 1000ms or so. Since sof is very consistent from host.

P33M:
> Meanwhile there is a few issues I think we should fix first
> 
> 1. Since this only occurs with bulk in transfer, we should only enable SOF when needed i.e when there is active bulk-in within 20% last frame. This will minimize the impact on current consumption. Since not all device make use of Bulk in endpoint e.g HID device. And even with MSC/cdc device, it is more often than not, device sitting idle than transferring data. and when it does, it is more often within first 80% frame.

I can address this by making SOF enabled iff pending_nonperiodic_xfer != 0, i.e. we need to pay attention to SOF. The buffer enqueue check will filter out endpoint types that don't need the fix.

> 2. hw_endpoint_xfer_continue() should also check of critical frame i.e all places where hw_endpoint_start_next_buffer() is called to trigger usb transaction
> 3. There is a race condition with `pending_nonperiodic_xfer` since both dcd_edpt_xfer() and sof isr try to read-then-set it. We should use an bool/flag in hw_endpoint to address this so that each endpoint is set and clear individually.

These two points mean that a proper critical section implementation needs to be added to hw_endpoint_lock_update(). Which of the pico-sdk synchronisation primitives would be best to use in tinyUSB's case?

> PS: I have one extra question does the 20% frame take in account of the double buffered i.e
> 
> * if we are about to schedule a double buffered transfer (buf0 and buf1) which result in 2 consecutive packet, should the critical frame now is increase to 40% ?
> * or if we are about to schedule a single buffer transfer, should the critical frame is reduced to 10% ?

I believe the testcase exercises the double-buffered behaviour.  The 20% reservation should also be enough to cover the cases where you get:
- Two or more IN endpoints activated at the same time before the reservation period, whether or not they are double-buffered
- An OUT scheduled by the HC before an IN transfer, and the IN was delayed by the hub's TT due to the large OUT packet.

src/portable/raspberrypi/rp2040/dcd_rp2040.c
P33M: Updated to guard the entire section with the endpoint lock. Note that the loop exits if parked_eps = 0, and pending_nonperiodic_xfer is volatile. So taking a local copy avoids emitting tons of loads/stores inside the for loop.

src/portable/raspberrypi/rp2040/rp2040_usb.c
P33M: Accesses to this variable are now guarded by the endpoint lock (which implements a pico-sdk critical_section). Would you prefer folding the variable into the hw_endpoint array?

kilograham (commented):
one thing i had been vaguely wondering about before this change is why the existing methods have +1, -1; this implies to me that re-entrancy is required, however this implementation does not support it.

If re-entrancy is not required, I think we should make that explicit.

src/portable/raspberrypi/rp2040/rp2040_usb.c
hathach: thank you for adding the guard, I think we should move the variable to hw_endpoint array, since the tinyusb stack only submit 1 transfer per endpoint at a time. This will remove the need to run critical section which is an rather expensive function.

hathach (commented):
sorry for late response again, the Lunar New Year aka TET in my country forcing me to clean my house/desk and leave not much time for the work. I don't think I will have time to test this out on actual RPI4 in another couple of weeks, and this has been on pending for too long (my bad). If you could update the PR 
- to move the parked_ep into hw_endpoint array without the use of critical_section
- comment out/revert the hw_endpoint_lock_update() implementation since there is other places that use spinlock as well, Some is not neccessary since usbd only schedule 1 transfer per endpoint until it is complete. We could re-add spinlock later on with more closer look.

Afterwards, I think we could merge this as it is. I will do the tests after merged and make changes when & where needed with PR mentioned you for review.

P33M:
I've addressed the review comments and updated the PR but unfortunately it looks like the SOF interrupt has to be constantly enabled. I cache the time of the previous SOF event and compute delta modulo 1000, but this will fail after several seconds as it appears there is enough clock drift to push the software blanking period away from the critical period.

We have no other way of keeping a stable count of where we are in a frame without logging every interrupt so whenever a bulk IN endpoint is active SOF must also be active.

src/portable/raspberrypi/rp2040/rp2040_usb.h
```diff
@@ -11,11 +11,20 @@
...
 #if defined(PICO_RP2040_USB_DEVICE_ENUMERATION_FIX) && !defined(TUD_OPT_RP2040_USB_DEVICE_ENUMERATION_FIX)
 #define TUD_OPT_RP2040_USB_DEVICE_ENUMERATION_FIX PICO_RP2040_USB_DEVICE_ENUMERATION_FIX
 #endif
 
+#if defined(PICO_RP2040_USB_DEVICE_UFRAME_FIX) && !defined(TUD_OPT_RP2040_USB_DEVICE_UFRAME_FIX)
+#define TUD_OPT_RP2040_USB_DEVICE_UFRAME_FIX PICO_RP2040_USB_DEVICE_UFRAME_FIX
+#endif
+
+#if TUD_OPT_RP2040_USB_DEVICE_UFRAME_FIX
+#define PICO_RP2040_USB_FAST_IRQ 1
```
kilograham, line 25: if we want to unconditionally set it to 1, we should probably #undef it first

src/portable/raspberrypi/rp2040/rp2040_usb.h
```diff
@@ -74,11 +90,24 @@ typedef struct hw_endpoint
 #endif
 } hw_endpoint_t;
 
+#if TUD_OPT_RP2040_USB_DEVICE_UFRAME_FIX
+extern volatile uint32_t last_sof;
+#endif
+
+uint32_t rp2040_critical_frame_period(struct hw_endpoint *ep);
```
kilograham, line 97: maybe consider making this a macro that returns 0 when !TUD_OPT_RP2040_USB_DEVICE_UFRAME_FIX rather than having an external function that always returns false.

src/portable/raspberrypi/rp2040/dcd_rp2040.c
```diff
@@ -249,11 +246,28 @@ static void __tusb_irq_path_func(dcd_rp2040_irq)(void)
...
+#if TUD_OPT_RP2040_USB_DEVICE_UFRAME_FIX
+    struct hw_endpoint *ep;
+#endif
 
     if (status & USB_INTF_DEV_SOF_BITS)
     {
       handled |= USB_INTF_DEV_SOF_BITS;
 
+#if TUD_OPT_RP2040_USB_DEVICE_UFRAME_FIX
+    hw_endpoint_lock_update(ep, 1);
```
kilograham, line 258: looks like indent here is wrong

src/portable/raspberrypi/rp2040/dcd_rp2040.c
```diff
@@ -447,6 +461,15 @@ void dcd_sof_enable(uint8_t rhport, bool en)
 
   _sof_enable = en;
 
+#if TUD_OPT_RP2040_USB_DEVICE_UFRAME_FIX
```
kilograham, line 464: should this be in the `if (en)` branch

kilograham: oh, i guess not

src/portable/raspberrypi/rp2040/rp2040_usb.h
```diff
@@ -64,7 +76,11 @@ typedef struct hw_endpoint
 
     // Interrupt, bulk, etc
     uint8_t transfer_type;
-    
+
```
kilograham, line 79: since your here, how about moving `active` from above with suitable comment into this group of 1 byte fields... we have 32 of these structures and we were wasting 4 bytes per.

src/portable/raspberrypi/rp2040/dcd_rp2040.c
P33M: The whole file uses 2 and 4-space indents interchangeably throughout, but yes it looks off

src/portable/raspberrypi/rp2040/rp2040_usb.h
P33M: Yes, compilation will fail if someone's set this already

src/portable/raspberrypi/rp2040/rp2040_usb.h
P33M:
```
(gdb) ptype /o struct hw_endpoint
/* offset    |  size */  type = struct hw_endpoint {
/*    0      |     1 */    _Bool configured;
/*    1      |     1 */    _Bool rx;
/*    2      |     1 */    uint8_t ep_addr;
/*    3      |     1 */    uint8_t next_pid;
/*    4      |     4 */    io_rw_32 *endpoint_control;
/*    8      |     4 */    io_rw_32 *buffer_control;
/*   12      |     4 */    uint8_t *hw_data_buf;
/*   16      |     1 */    _Bool active;
/* XXX  1-byte hole  */
/*   18      |     2 */    uint16_t remaining_len;
/*   20      |     2 */    uint16_t xferred_len;
/* XXX  2-byte hole  */
/*   24      |     4 */    uint8_t *user_buf;
/*   28      |     2 */    uint16_t wMaxPacketSize;
/*   30      |     1 */    uint8_t transfer_type;
/*   31      |     1 */    uint8_t pending;

                           /* total size (bytes):   32 */
                         }
```
three according to gdb, but yes holes are bad

kilograham: ah, ok, i thought it would make it smaller, but i was thinking of the 2 uint16_t s as being 4 byte aligned for some reason :-(

src/portable/raspberrypi/rp2040/rp2040_usb.h
kilograham: which i think they may have

src/portable/raspberrypi/rp2040/rp2040_usb.h
P33M: In host mode the two extra fields end up growing the structure to 36B, so it is worth rearranging so that pad bytes at the end of the structure get used instead.

src/portable/raspberrypi/rp2040/dcd_rp2040.c
P33M: Reimplemented in v3

hathach (approved):
Thank you very much for the PR and your patient. I still haven't got time to tried this out on actual hardware described with Errata 15 (with Pi4 USB Hub). But it looks great now. 

I have made some minor changes mostly whitespace, rename, and minor refactor. Let me know these changes made sense to you. If all is good, I think we could merge this as it is.

lurch:
@hathach I think your comments mentioning "walkaround" / "walk-around" should actually say "workaround" ?

hathach:
> @hathach I think your comments mentioning "walkaround" / "walk-around" should actually say "workaround" ?

Indeed, that is my bad

hathach:
> I have no objections to the edits.

perfect !! Thank you very much, merged now. Feel free to make more follow up PR should this needs more changes to fix.

PR #1783, opened 2022-12-08
usbtmc: correct packet size bug

scoudreau:
Code was only reading the first 64 bytes of a 512 bytes packet.

For some reason, code was working if receiving a single packet strictly below 512 bytes. But when sending a USBTMC message which lead to a data stage longer than a single packet, only the first 64 bytes where read.
This seems to be only due to the "arming" of transfer which where obviously only asking for 64 bytes (constant).


hathach (approved):
thank you, the main developer for usbtmc @pigrew only has FS device to test with. There is a recent update to get it running with HS #1531 , too bad, I keep "forgetting" to test it out. I will do a full revise later on.

PR #1784, opened 2022-12-08
Correct interrupt polling interval in descriptor

scoudreau:
Use macro argument instead of 0x16 constant value.
At least Linux kernel was complaining about this value, giving the following warning:
"config 1 interface 0 altsetting 0 endpoint 0x82 has an invalid bInterval 22, changing to 8"
(warning obtained using usbtmc device config)


hathach (approved):
thank you for the fix

Issue #1729, opened 2022-11-07
Vulnerability reporting procedure


silvergasp:
### Related area

Security/vulnerability reporting

### Hardware specification

N/A

### Is your feature request related to a problem?

I think I've found a security vulnerability, but I don't want to publicly describe the problem on a GitHub issue until the problem is fixed. Having a security problem published publicly would open people up to attack.

### Describe the solution you'd like

I'd like some procedures to be in place to;
- Contact the maintainers of this repository with security issues without publicly posting the problem to a GitHub issue. e.g. https://github.com/hathach/tinyusb/security/policy
- Post security problems publicly once they have been fixed, and in the latest release. e.g. https://github.com/hathach/tinyusb/security/advisories

NOTE: There are some templates on the links above which offer a good starting point.

### I have checked existing issues, dicussion and documentation

- [X] I confirm I have checked existing issues, dicussion and documentation.

cr1901:
I agree something should be done about this, apparently Github now has a feature to contact project maintainers securely/directly in case of a vuln? Idk much about it tho...

silvergasp:
I wasn't aware of that. But it looks like a great option!

Link to the docs for posterity;

https://docs.github.com/en/code-security/security-advisories/repository-security-advisories/configuring-private-vulnerability-reporting-for-a-repository

silvergasp:
It appears pretty easy to setup. But it would need to be done by someone with admin privileges.

cr1901:
Well, @hathach seems to be taking a break for now, but I imagine he'll be back soon. Thanks for the link, since I completely forgot other than hearing about the new feature thirdhand :D!

silvergasp:
Yeah, I've emailed @hathach directly with directly with details of the vulnerability that I found. Got a reply mentioning that he was keen to take a look, but is currently away from his computer. 

I don't think there is any need to rush into getting it fixed :). Though it would be nice to get a fix together in the next couple of weeks.

hathach:
just be back to my PC, will check this out in this week or so. Thank you for the issue/pr.

hathach:
this should be fixed by #1789, thank you @silvergasp very much for bringing this up and also making effort to add fuzzer as well and pr to integrate tinyusb to google fuzzing.

PR #1716, opened 2022-11-02
fuzz: Add support for fuzzing

silvergasp:
Adds support for fuzzing with basic a cdc fuzzer.

**Describe the PR**
This fuzz harness adds the necessary build infrastructure for fuzzing tinyusb. As a proof of concept there is a cdc fuzz harness written in c++. The intention of this PR is to improve security and robustness of the tinyusb stack by tracking down bugs and performance problems using fuzz testing.

PR #1789, opened 2022-12-12
Fix fifo memory overflow

### LINKED ISSUE #1039: CDC device receives garbage if it transmits first
### Operating System

Linux

### Board

Custom board with ULPI PHY and STM32H730

### Firmware

cdc_dual example, with these changes:
- with only one CDC enabled
- loopback code disabled
- Instead of looping back data from host back to host, device transmits data periodically since boot
- CDC buffers raised to 4 kiB

```cpp
#include <inttypes.h>

const uint32_t coreFrequency = 16000000; // has to be configured according to core frequency

volatile uint32_t ticks = 0; // ticks are volatile variable incremented every 1 ms (systick).
void SysTick_Handler() {
	ticks = ticks + 1;
}

/*------------- MAIN -------------*/
int main(void)
{

	SysTick_Config(coreFrequency/1000); // 1 kHz systick ticking
	NVIC_EnableIRQ(UsageFault_IRQn);

	uint32_t prev = ticks; // save last TX timestamp (at first, we don't do tx, but nobody cares)
	uint8_t data = 0; // data for TX
	uint8_t buffer[64]; // buffer to catch spuriously received data
	unsigned int offset = 0; // buffer writing offset

	board_init();
	tusb_init();

	for(;;) {
		tud_task();

		const auto now = ticks;
		if (now - prev > 200) {
			prev = now;

			tud_cdc_write(&data, 1); // TX one byte
			tud_cdc_write_flush();
			data++; // increment data value for each transmission
		}

		if (tud_cdc_available()) {
			// capture all received data
			// assert(offset < sizeof(buffer) - 1);
			offset += tud_cdc_read(&buffer[offset], sizeof(buffer) - offset);
			// __BKPT(0);
		}
	}
	return 0;
}
```

### What happened ?

1) power on device and let i live (1s or so, just to have some data in TX buffer)
2) open UART (`ACM0`) in serial terminal (iserterm in ma case)
3) data in TX buffer are received by host, but also some new data ends in RX buffer (ERR led lit up and some garbage is in `buffer`, but I no (explicit) data transmission from host occurred

### How to reproduce ?

Basically replicate "What happened" scenario. I am not sure if this is issue of Linux or tinyUSB stack.

### Debug Log
[tusb_log.txt](https://github.com/hathach/tinyusb/files/7027747/tusb_log.txt)

### Screenshots

_No response_

hathach:
**Describe the PR**

While trying to fix an issue reported in https://github.com/hathach/tinyusb/issues/1729 (detected using fuzzing #1716). I am able to fix a couple of issues with current tusb_fifo when using with overwritable mode
- Prevent double overflowed to happen which make fifo count invalid. Though due to the cost of moving memory, the data at read index is not replaced by newer one.
- fix _tu_fifo_remaining() return incorrect value which cause memory overflowed, detected by fuzzing
- minimize tu_fifo size to 16
- simplify and remove _tu_fifo_empty, _tu_fifo_full. Also correct full condition check
- simplify _tu_fifo_count() and _tu_fifo_remaining(), also rename to _ff_count() and _ff_remaining()
- update advance_pointer/backward_pointer to use depth instead of fifo, also rename to advance/backward_index
- simplify _ff_correct_read_index()
- remove _ff_overflowed() due to lack of use
- using clang with ceedling unit-test with -fsanitize=address

Both of above cases can be reproduced/tested with 2 new unit tests. I also take the chance to add fifo implementation note (since I kind of forgot about it) to help future troubleshooting and rename absolute/relative naming to index/pointer and other minor changes. There is still room to improve fifo, thus submitted this as draft first for feedback.

ref reading: https://www.snellman.net/blog/archive/2016-12-13-ring-buffers/9 

probably fix #1039 , and also related to https://github.com/adafruit/circuitpython/pull/7041 since this could be the bug of cdc fifo instead of what we thought is Linux with echo mode.

### DIFF
--- a/src/common/tusb_fifo.c
+++ b/src/common/tusb_fifo.c
@@ -358,8 +358,9 @@ static inline uint16_t idx2ptr(uint16_t idx, uint16_t depth)
   return idx;
 }
 
-// Works on local copies of w and r - return only the difference and as such can be used to determine an overflow
-static inline uint16_t _tu_fifo_count(tu_fifo_t* f, uint16_t wr_idx, uint16_t rd_idx)
+// return only the index difference and as such can be used to determine an overflow i.e overflowable count
+TU_ATTR_ALWAYS_INLINE static inline
+uint16_t _ff_count(uint16_t depth, uint16_t wr_idx, uint16_t rd_idx)
 {
   // In case we have non-power of two depth we need a further modification
   if (wr_idx >= rd_idx)
@@ -367,24 +368,33 @@ static inline uint16_t _tu_fifo_count(tu_fifo_t* f, uint16_t wr_idx, uint16_t rd
     return (uint16_t) (wr_idx - rd_idx);
   } else
   {
-    return (uint16_t) (2*f->depth - (rd_idx - wr_idx));
+    return (uint16_t) (2*depth - (rd_idx - wr_idx));
   }
 }
 
-// Works on local copies of w and r
 // BE AWARE - THIS FUNCTION MIGHT NOT GIVE A CORRECT ANSWERE IN CASE WRITE POINTER "OVERFLOWS"
 // Only one overflow is allowed for this function to work e.g. if depth = 100, you must not
 // write more than 2*depth-1 items in one rush without updating write pointer. Otherwise
 // write pointer wraps and you pointer states are messed up. This can only happen if you
 // use DMAs, write functions do not allow such an error.
-static inline bool _tu_fifo_overflowed(tu_fifo_t* f, uint16_t wr_idx, uint16_t rd_idx)
+TU_ATTR_ALWAYS_INLINE static inline
+bool _ff_overflowed(uint16_t depth, uint16_t wr_idx, uint16_t rd_idx)
 {
-  return _tu_fifo_count(f, wr_idx, rd_idx) > f->depth;
+  return _ff_count(depth, wr_idx, rd_idx) > depth;
+}
+
+// return remaining slot in fifo
+TU_ATTR_ALWAYS_INLINE static inline
+uint16_t _ff_remaining(uint16_t depth, uint16_t wr_idx, uint16_t rd_idx)
+{
+  uint16_t const count = _ff_count(depth, wr_idx, rd_idx);
+  return (depth > count) ? (depth - count) : 0;
 }
 
 // Works on local copies of w
 // For more details see _tu_fifo_overflow()!
-static inline void _tu_fifo_correct_read_pointer(tu_fifo_t* f, uint16_t wAbs)
+TU_ATTR_ALWAYS_INLINE static inline
+void _tu_fifo_correct_read_pointer(tu_fifo_t* f, uint16_t wAbs)
 {
   f->rd_idx = backward_pointer(f, wAbs, f->depth);
 }
@@ -393,7 +403,7 @@ static inline void _tu_fifo_correct_read_pointer(tu_fifo_t* f, uint16_t wAbs)
 // Must be protected by mutexes since in case of an overflow read pointer gets modified
 static bool _tu_fifo_peek(tu_fifo_t* f, void * p_buffer, uint16_t wr_idx, uint16_t rd_idx)
 {
-  uint16_t cnt = _tu_fifo_count(f, wr_idx, rd_idx);
+  uint16_t cnt = _ff_count(f->depth, wr_idx, rd_idx);
 
   // Check overflow and correct if required
   if (cnt > f->depth)
@@ -417,7 +427,7 @@ static bool _tu_fifo_peek(tu_fifo_t* f, void * p_buffer, uint16_t wr_idx, uint16
 // Must be protected by mutexes since in case of an overflow read pointer gets modified
 static uint16_t _tu_fifo_peek_n(tu_fifo_t* f, void * p_buffer, uint16_t n, uint16_t wr_idx, uint16_t rd_idx, tu_fifo_copy_mode_t copy_mode)
 {
-  uint16_t cnt = _tu_fifo_count(f, wr_idx, rd_idx);
+  uint16_t cnt = _ff_count(f->depth, wr_idx, rd_idx);
 
   // Check overflow and correct if required
   if (cnt > f->depth)
@@ -441,13 +451,6 @@ static uint16_t _tu_fifo_peek_n(tu_fifo_t* f, void * p_buffer, uint16_t n, uint1
   return n;
 }
 
-// Works on local copies of w and r
-static inline uint16_t _tu_fifo_remaining(tu_fifo_t* f, uint16_t wr_idx, uint16_t rd_idx)
-{
-  uint16_t const count = _tu_fifo_count(f, wr_idx, rd_idx);
-  return (f->depth > count) ? (f->depth - count) : 0;
-}
-
 static uint16_t _tu_fifo_write_n(tu_fifo_t* f, const void * data, uint16_t n, tu_fifo_copy_mode_t copy_mode)
 {
   if ( n == 0 ) return 0;
@@ -460,12 +463,12 @@ static uint16_t _tu_fifo_write_n(tu_fifo_t* f, const void * data, uint16_t n, tu
   uint8_t const* buf8 = (uint8_t const*) data;
 
   TU_LOG(TU_FIFO_DBG, "rd = %3u, wr = %3u, count = %3u, remain = %3u, n = %3u:  ",
-                       rd_idx, wr_idx, _tu_fifo_count(f, wr_idx, rd_idx), _tu_fifo_remaining(f, wr_idx, rd_idx), n);
+                       rd_idx, wr_idx, _ff_count(f->depth, wr_idx, rd_idx), _ff_remaining(f->depth, wr_idx, rd_idx), n);
 
   if ( !f->overwritable )
   {
     // limit up to full
-    uint16_t const remain = _tu_fifo_remaining(f, wr_idx, rd_idx);
+    uint16_t const remain = _ff_remaining(f->depth, wr_idx, rd_idx);
     n = tu_min16(n, remain);
   }
   else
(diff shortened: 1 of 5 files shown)


test/unit-test/test/test_fifo.c
```diff
@@ -61,85 +67,134 @@ void test_normal(void)
...
-  tu_fifo_write_n(ff, data+8, 6); // wr = 3, count = 9
+  tu_fifo_write_n(ff, test_data+32, 40); // wr = 72 -> 8, count = 56
+
+  tu_fifo_read_n(ff, rd_buf, 32); // count = 24
+  TEST_ASSERT_EQUAL_MEMORY( test_data+16, rd_buf, rd_count);
+
+  TEST_ASSERT_EQUAL(24, tu_fifo_count(ff));
+}
+
+void test_write_double_overflowed(void)
```
hathach, line 141: new test to test fifo double overflowed handling.

src/common/tusb_fifo.h
```diff
@@ -52,18 +52,74 @@ extern "C" {
 #define tu_fifo_mutex_t  osal_mutex_t
 #endif
 
+/* Write/Read index is always in the range of:
```
hathach, line 55: @PanRe I added implementation note for fifo. If you have time, please review this to see if I miss anything.

src/common/tusb_fifo.c
```diff
@@ -180,6 +179,7 @@ static void _ff_push_n(tu_fifo_t* f, void const * app_buf, uint16_t n, uint16_t
         memcpy(ff_buf, app_buf, nLin_bytes);
 
         // Write data wrapped around
+        // TU_ASSERT(nWrap_bytes <= f->depth, );
```
hathach, line 182: running test_write_overwritable2() with the pre-pr fifo will cause nWrap_bytes > f->depth (out of boundary access)

src/common/tusb_fifo.c
```diff
@@ -462,33 +469,74 @@ static uint16_t _tu_fifo_write_n(tu_fifo_t* f, const void * data, uint16_t n, tu
...
+        // TODO should read from hw fifo to discard data, however reading an odd number could
+        // accidentally discard data.
+      }
+
+      n = f->depth;
+
+      // We start writing at the read pointer's position since we fill the whole buffer
+      wr_idx = rd_idx;
+    }
+    else if (overflowable_count + n >= 2*f->depth)
```
hathach, line 509: double overflowed detection and handling with overwritable fifo

src/common/tusb_fifo.c
```diff
@@ -340,44 +343,47 @@ static uint16_t backward_pointer(tu_fifo_t* f, uint16_t p, uint16_t offset)
...
+  return new_p;
 }
 
-// get relative from absolute pointer
-static uint16_t get_relative_pointer(tu_fifo_t* f, uint16_t p)
+// index to pointer, simply an modulo with minus
+static inline uint16_t idx2ptr(uint16_t idx, uint16_t depth)
 {
-  return _ff_mod(p, f->depth);
+  while ( idx >= depth ) idx -= depth;
```
PanRe, line 358: You may comment here, that this loop is executed at most 3 times since the usable index space is limited to 2*depth and the index will be reset to the wrapped value if idx >= depth - for all who think this is a computational bottleneck

src/common/tusb_fifo.c
```diff
@@ -442,18 +448,19 @@ static uint16_t _tu_fifo_peek_n(tu_fifo_t* f, void * p_buffer, uint16_t n, uint1
...
   return n;
 }
 
 // Works on local copies of w and r
-static inline uint16_t _tu_fifo_remaining(tu_fifo_t* f, uint16_t wAbs, uint16_t rAbs)
+static inline uint16_t _tu_fifo_remaining(tu_fifo_t* f, uint16_t wr_idx, uint16_t rd_idx)
 {
-  return f->depth - _tu_fifo_count(f, wAbs, rAbs);
+  uint16_t const count = _tu_fifo_count(f, wr_idx, rd_idx);
+  return (f->depth > count) ? (f->depth - count) : 0;
```
PanRe, line 463: Does this prevent the out of bound error to happen? Would it not be better to take care `_tu_fifo_count()` does not return a value bigger than `f->depth`? Then you could avoid the overflow check here and at the same time sanitize `_tu_fifo_count()`! Also `f->depth == count` should be ok, this means the buffer is exactly full!

src/common/tusb_fifo.c
```diff
@@ -340,44 +343,47 @@ static uint16_t backward_pointer(tu_fifo_t* f, uint16_t p, uint16_t offset)
...
-  uint16_t cnt = wAbs-rAbs;
+  uint16_t cnt = (uint16_t) (wr_idx-rd_idx);
 
   // In case we have non-power of two depth we need a further modification
-  if (rAbs > wAbs) cnt -= f->non_used_index_space;
+  if (rd_idx > wr_idx)
+  {
+    // 2*f->depth - (rd_idx - wr_idx);
+    cnt = (uint16_t) (cnt - f->non_used_index_space);
+  }
```
PanRe, line 373: How about checking here if the `cnt` value is ok?
`return (f->depth >= cnt) ? cnt : 0;`

src/common/tusb_fifo.h
```diff
@@ -52,18 +52,74 @@ extern "C" {
...
+ *      0 .. 2*depth-1
+ * The extra window allow us to determine the fifo state of empty or full with only 2 indexes
```
PanRe, line 57: indexes -> indices ;)

src/common/tusb_fifo.c
hathach: right, I will add a note here

src/common/tusb_fifo.c
hathach: yes, this prevent out of bound to occur. I think we can use the public version of tu_fifo_count() which does limit the count to depth. The internal version _tu_fifo_count() seems to be used to determine an overflowed condition so that we could correct/adjust read index in a fifo_read() function.

src/common/tusb_fifo.c
PanRe: No you are right, i just saw it... sorry it is some time i had this code in my head by heart :D - forget this comment here please :D

src/common/tusb_fifo.c
hathach: They are overlapped a bit but not the same. I did put a note on the header file (did mentioned you there). Overflowed (single/double) only occurred with over-writable fifo where we repeatedly write() without doing any read(). Which is a case of cdc when first enumerated but dtr/rts is not set. Or with DMA but the writer is running much faster than reader. 

There is also a unite test case for this https://github.com/hathach/tinyusb/pull/1789/files/04a5c03ea8ea70510c332f0fafb44daf0ec7c7fb#diff-f49c2fc269132b486ec507c3e772fb981b9fddf2a1961ae5b9ad6d566d8d3c23R141

src/common/tusb_fifo.h
hathach: I hate English :D

src/common/tusb_fifo.c
hathach: yeah, memory fades as well, and take a bit of time to revise and re-understood the fifo implementation. That is why I put a lots of effort in the header the implementation note.

src/common/tusb_fifo.c
```diff
@@ -462,33 +469,74 @@ static uint16_t _tu_fifo_write_n(tu_fifo_t* f, const void * data, uint16_t n, tu
...
+      // re-position write index to have a full fifo after pushed
+      wr_idx = advance_pointer(f, rd_idx, f->depth - n);
+
+      // TODO we should also shift out n bytes from read index since we avoid changing rd index !!
+      // However memmove() is expensive due to actual copying + wrapping consideration.
+      // Also race condition could happen anyway if read() is invoke while moving result in corrupted memory
+      // currently deliberately not implemented --> result in incorrect data read back
+    }else
+    {
+      // normal + single overflowed: just increase write index
```
PanRe, line 521: Maybe a comment: Overflow is in the range of [0,2*depth) and thus detect and recoverable. Recovering is handled in `fifo_read()`.

src/common/tusb_fifo.c
PanRe: I got it, maybe a comment for later: Overflow is bigger than the allowed range [0,2*depth) and thus not detect and as such not recoverable in the read function - we need to handle the overflow condition here

src/common/tusb_fifo.h
```diff
@@ -52,18 +52,74 @@ extern "C" {
...
+ *
+ *  - Double Overflowed: write(3), write(1), write(2)
+ *    Continue to write after overflowed to 2nd overflowed.
+ *    We must prevent 2nd overflowed since it will cause incorrect computed of count, in above example
+ *    if not handled the fifo will be empty instead of continue-to-be full. Since we must not modify
+ *    read index in write() function, which cause race condition. We will re-position write index so that
+ *    after data is written it is a full fifo i.e W = depth - R
+ *
+ *      re-position W = 1 before write(2)
+ *      Note: we should also move data from mem[4] to read index as well, but deliberately skipped here
```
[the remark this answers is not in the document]
hathach: ah right, I mean move data from mem[3] to read index, will update the comment. an typical off-by-one mistake :D

PanRe:
Looks good! Nice that the buffer would be fuzzing-proof :)

Is there anything else to improve since you mentioned there would be room therefore!

hathach:
> Looks good! Nice that the buffer would be fuzzing-proof :)
> 
> Is there anything else to improve since you mentioned there would be room therefore!

just update the PR with your feedback, thank you for reviewing. For room to improve, I feel that we could do better to simplify the current code, but it is not urgent. Like merging get_realitve_poiinter() and _ffmod() to idx2ptr(), mostly for readability and maintenance only. I will try to wrap this up soon enough.

silvergasp:
Just a suggestion. It might be worth enabling `-sanitizer=address` which will detect any overflows in your unit test.

hathach:
> Just a suggestion. It might be worth enabling `-sanitizer=address` which will detect any overflows in your unit test.
> 
> However that particular flag does require clang.

yeah, this is great idea. I will check to see we could switch to clang. Unit test like fuzzing run on host, so that wouldn't be an issue at all.

silvergasp:
Apologies the flag is `-fsanitize=address` docs here https://clang.llvm.org/docs/AddressSanitizer.html

hathach:
> Apologies the flag is `-fsanitize=address` docs here https://clang.llvm.org/docs/AddressSanitizer.html

unit-test is now configured to run with clang and `-fsanitize=address` https://github.com/hathach/tinyusb/pull/1789/commits/e885ced0fed0a59d093db09c3e7325457d863fc1 , thanks for your suggestion.

Issue #1782, opened 2022-12-08
Missing fields in audio_desc_input_terminal_t


s13n:
### Operating System

Linux

### Board

Raspberry Pi Pico

### Firmware

Raspberry Pi SDK - master branch, using TinyUSB 0.12.0

### What happened ?

A struct definition in header file src/class/audio/audio.h is missing two fields, compared to the definition in "USB Device Class Definition for Audio Devices Release 2.0".

See the definition of `audio_desc_input_terminal_t` starting in line 719 of file audio.h
Compare with table 4-9 in the USB document.

Missing fields are: `bTerminalID` and `iChannelNames`.

### How to reproduce ?

Read the mentioned source file and the USB document.

### I have checked existing issues, dicussion and documentation

- [X] I confirm I have checked existing issues, dicussion and documentation.

hathach:
can you paste the link to the code in master branch, the 0.12.0 version is too old and line number may not be correct. Also please try to update tinyusb as well to see if it fixed recently.

s13n:
The bug is still there in master, and the line in the file is still the same.

pete-pjb:
Hi @hathach ,

I can add this to my PR if that is helpful - let me know...

Cheers,

Pete

hathach:
@pete-pjb sure thing, if you already have the fix ready.

hathach:
fixed by #1837

PR #1795, opened 2022-12-13
Minor changes to support Zynq host port. Fix a typo in audio.h

pete-pjb:
Add USB memory macro to USB host control structure, to enable full un-cached implementation.  See [this](https://github.com/hathach/tinyusb/discussions/1769#discussioncomment-4354189) for details.

Add NULL pointer check to ***EHCI host*** module as discussed [here](https://github.com/hathach/tinyusb/discussions/1769#discussioncomment-4352979) 

Add weak function prototype to ```hcd.h``` for cache control, see discussion [here](https://github.com/hathach/tinyusb/discussions/1769#discussioncomment-4377690)

Add support for caching to ***EHCI host*** module.

This will also address issue raised here by @s13n :
https://github.com/hathach/tinyusb/issues/1782

### DIFF
--- a/src/host/hcd.h
+++ b/src/host/hcd.h
@@ -124,6 +124,9 @@ void hcd_int_disable(uint8_t rhport);
 // Get frame number (1ms)
 uint32_t hcd_frame_number(uint8_t rhport);
 
+// Optional MCU Cache control if you require this please implement the function in your hcd_platform.c file
+TU_ATTR_WEAK extern void hcd_cache_flush(uint32_t addr, uint32_t data_size);
+
 //--------------------------------------------------------------------+
 // Port API
 //--------------------------------------------------------------------+
--- a/src/host/hcd.h
+++ b/src/host/hcd.h
@@ -124,8 +124,9 @@ void hcd_int_disable(uint8_t rhport);
 // Get frame number (1ms)
 uint32_t hcd_frame_number(uint8_t rhport);
 
-// Optional MCU Cache control if you require this please implement the function in your hcd_platform.c file
-TU_ATTR_WEAK extern void hcd_cache_flush(uint32_t addr, uint32_t data_size);
+// Optional MCU Data Cache control if you require this please implement the functions in your hcd_platform.c file
+TU_ATTR_WEAK extern void hcd_dcache_flush(uint32_t addr, uint32_t data_size);
+TU_ATTR_WEAK extern void hcd_dcache_invalidate( uint32_t addr, uint32_t data_size);
 
 //--------------------------------------------------------------------+
 // Port API
--- a/src/host/usbh.c
+++ b/src/host/usbh.c
@@ -245,7 +245,7 @@ CFG_TUSB_MEM_SECTION CFG_TUSB_MEM_ALIGN
 static uint8_t _usbh_ctrl_buf[CFG_TUH_ENUMERATION_BUFSIZE];
 
 // Control transfers: since most controllers do not support multiple control transfers
-// on multiple devices concurrently, and control transfers are not used much except for
+// on multiple devices concurrently and control transfers are not used much except for
 // enumeration, we will only execute control transfers one at a time.
 CFG_TUSB_MEM_SECTION struct
 {
@@ -413,7 +413,6 @@ void tuh_task_ext(uint32_t timeout_ms, bool in_isr)
   {
     hcd_event_t event;
     if ( !osal_queue_receive(_usbh_q, &event, timeout_ms) ) return;
-
     switch (event.event_id)
     {
       case HCD_EVENT_DEVICE_ATTACH:
@@ -1262,6 +1261,7 @@ static void process_enumeration(tuh_xfer_t* xfer)
       // Get first 8 bytes of device descriptor for Control Endpoint size
       TU_LOG2("Get 8 byte of Device Descriptor\r\n");
       TU_ASSERT(tuh_descriptor_get_device(addr0, _usbh_ctrl_buf, 8, process_enumeration, ENUM_SET_ADDR), );
+
     }
     break;
 
@@ -1381,6 +1381,9 @@ static void process_enumeration(tuh_xfer_t* xfer)
       enum_full_complete();
     break;
   }
+  // If required invalidate data cache after DMA transfer complete - see hcd.h for details.
+  hcd_dcache_invalidate((uint32_t)_usbh_ctrl_buf, sizeof(_usbh_ctrl_buf));
+
 }
 
 static bool enum_new_device(hcd_event_t* event)
--- a/src/host/usbh.c
+++ b/src/host/usbh.c
@@ -244,10 +244,10 @@ static osal_queue_t _usbh_q;
 CFG_TUSB_MEM_SECTION CFG_TUSB_MEM_ALIGN
 static uint8_t _usbh_ctrl_buf[CFG_TUH_ENUMERATION_BUFSIZE];
 
-// Control transfer: since most controller does not support multiple control transfer
-// on multiple devices concurrently. And control transfer is not used much except enumeration
-// We will only execute control transfer one at a time.
-struct
+// Control transfers: since most controllers do not support multiple control transfers
+// on multiple devices concurrently, and control transfers are not used much except for
+// enumeration, we will only execute control transfers one at a time.
+CFG_TUSB_MEM_SECTION struct
 {
   tusb_control_request_t request TU_ATTR_ALIGNED(4);
   uint8_t* buffer;
--- a/src/host/usbh.c
+++ b/src/host/usbh.c
@@ -413,6 +413,7 @@ void tuh_task_ext(uint32_t timeout_ms, bool in_isr)
   {
     hcd_event_t event;
     if ( !osal_queue_receive(_usbh_q, &event, timeout_ms) ) return;
+
     switch (event.event_id)
     {
       case HCD_EVENT_DEVICE_ATTACH:
@@ -1261,7 +1262,6 @@ static void process_enumeration(tuh_xfer_t* xfer)
       // Get first 8 bytes of device descriptor for Control Endpoint size
       TU_LOG2("Get 8 byte of Device Descriptor\r\n");
       TU_ASSERT(tuh_descriptor_get_device(addr0, _usbh_ctrl_buf, 8, process_enumeration, ENUM_SET_ADDR), );
-
     }
     break;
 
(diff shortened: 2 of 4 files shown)


pete-pjb:
Further commits to come to support caching...

hathach (approved):
all look good, thank you very much for the PR

PS: just realize, you seem to have more commit to add cache flush, let me know if you could do it quickly within this PR or prefer to submit another PR for cache flush.

hathach:
take your time, there is no need to rush

pete-pjb:
Hi @hathach ,

A quick update, I am still not quite ready, I have also not been well today so not on the ball.... 

I will continue with it tomorrow now and hopefully request your review then...

Kind Regards,

Pete

hathach:
> Hi @hathach ,
> 
> A quick update, I am still not quite ready, I have also not been well today so not on the ball....
> 
> I will continue with it tomorrow now and hopefully request your review then...
> 
> Kind Regards,
> 
> Pete

no problem at all, no rush as mentioned above. Take care of yourself, we can always do this later on.

pete-pjb:
Hi @hathach ,

Sorry for the delay but I am back on the ball now!

I have now added two new external weak function prototypes to ***hcd.h*** as follows:

```c
// Optional MCU Data Cache control if you require this please implement the functions in your hcd_platform.c file
TU_ATTR_WEAK extern void hcd_dcache_flush(uint32_t addr, uint32_t data_size);
TU_ATTR_WEAK extern void hcd_dcache_invalidate( uint32_t addr, uint32_t data_size);
```
It also turned out with my platform I needed to invalidate the cache after the reception of a DMA transfer, hence the inclusion of the ```hcd_dcache_invalidate()``` function also.
I have also added the function calls as required to the ***usbh.c*** and ***ehci.c*** modules.  
I have tested compilation with GCC and the code works un-cached without the functions coded in the hcd_platform file and cached with the two functions coded for my platform as expected, there are no warnings or errors being generated so all looks good for me...

If you want me to change anything or something is not as expected please let me know and I will sort it out.

Kind Regards,

Pete

src/host/hcd.h
```diff
@@ -124,6 +124,10 @@ void hcd_int_disable(uint8_t rhport);
 // Get frame number (1ms)
 uint32_t hcd_frame_number(uint8_t rhport);
 
+// Optional MCU Data Cache control if you require this please implement the functions in your hcd_platform.c file
+TU_ATTR_WEAK extern void hcd_dcache_flush(uint32_t addr, uint32_t data_size);
```
hathach, line 128: minor: could you remove the `extern`, since this is supposed to be implemented by hcd.

src/portable/ehci/ehci.c
```diff
@@ -388,6 +390,8 @@ bool hcd_edpt_open(uint8_t rhport, uint8_t dev_addr, tusb_desc_endpoint_t const
 
   // TODO might need to disable async/period list
   list_insert(list_head, (ehci_link_t*) p_qhd, EHCI_QTYPE_QHD);
+  // Flush data cache if required, see hcd.h for details
+  hcd_dcache_flush((uint32_t)p_qhd, sizeof(ehci_qhd_t));
```
hathach, line 394: for all calls of weak functions, it must be check with `if` first. Since not all compiler could skip un-implemented weak function (which is typically resolved to NULL in symbol table).

```C
if (hcd_dcache_flush) 
{
  hcd_dcache_flush((uint32_t)ep_desc, sizeof(tusb_desc_endpoint_t));
}
```

src/portable/ehci/ehci.c
```diff
@@ -343,6 +343,8 @@ bool hcd_edpt_open(uint8_t rhport, uint8_t dev_addr, tusb_desc_endpoint_t const
 {
   (void) rhport;
 
+  // Flush cache if required, see hcd.h for details
+  hcd_dcache_flush((uint32_t)ep_desc, sizeof(tusb_desc_endpoint_t));
```
hathach, line 347: do we really need to flush this ? We don't make any transfer yet, would you mind trying without this flush.

src/portable/ehci/ehci.c
```diff
@@ -418,32 +427,36 @@ bool hcd_edpt_xfer(uint8_t rhport, uint8_t dev_addr, uint8_t ep_addr, uint8_t *
 {
   (void) rhport;
 
+  ehci_qhd_t* p_qhd;
+  ehci_qtd_t* p_qtd;
   uint8_t const epnum = tu_edpt_number(ep_addr);
   uint8_t const dir   = tu_edpt_dir(ep_addr);
 
+  // Flush data cache if required, see hcd.h for details
+  hcd_dcache_flush((uint32_t)buffer, buflen);
```
hathach, line 436: we should both invalidate (for rx) and flush (tx) this buffer. In general, I think we should move this dcache flush + invalidate into qtd_init() function.

src/portable/ehci/ehci.c
```diff
@@ -399,6 +403,8 @@ bool hcd_setup_send(uint8_t rhport, uint8_t dev_addr, uint8_t const setup_packet
   ehci_qhd_t* qhd = &ehci_data.control[dev_addr].qhd;
   ehci_qtd_t* td  = &ehci_data.control[dev_addr].qtd;
 
+  // Flush data cache if required, see hcd.h for details
+  hcd_dcache_flush((uint32_t)setup_packet, 8);
```
hathach, line 407: can be removed once moved into qtd_init() check below feedback.

src/portable/ehci/ehci.c
```diff
@@ -457,6 +470,9 @@ bool hcd_edpt_xfer(uint8_t rhport, uint8_t dev_addr, uint8_t ep_addr, uint8_t *
     // attach head QTD to QHD start transferring
     p_qhd->qtd_overlay.next.address = (uint32_t) p_qhd->p_qtd_list_head;
   }
+  // Flush data cache if required, see hcd.h for details
+  hcd_dcache_flush((uint32_t)p_qhd, sizeof(ehci_qhd_t));
```
hathach, line 474: we probably need to also do invalidate + flush since EHCI does write back to update qhd/qtd as well ? or we could do the invalidate in the hcd_int_handler() I guess.

src/host/usbh.c
```diff
@@ -1362,6 +1362,8 @@ static void process_enumeration(tuh_xfer_t* xfer)
       enum_full_complete();
     break;
   }
+  // If required invalidate data cache after DMA transfer complete - see hcd.h for details.
+  hcd_dcache_invalidate((uint32_t)_usbh_ctrl_buf, sizeof(_usbh_ctrl_buf));
```
hathach, line 1366: should be removed since I think invalidate should be done within the HCD (EHCI in this case) see below feedback in hcd_edpt_xfer().

hathach (commented):
Thank you for your update, I have a couple of questions/suggestions regarding the usage of dcache API. Feel free to correct me if necessary.

src/portable/ehci/ehci.c
pete-pjb: I wasn't aware of this, so no problem I will make the changes. 🙂

src/portable/ehci/ehci.c
pete-pjb: I will check this, I am still getting to grips with the USB protocol in general and with the code base, I haven't had time to extensively study the code yet, so please bare with me.  😊

pete-pjb: As above...
> I will check this, I am still getting to grips with the USB protocol in general and with the code base, I haven't had time to extensively study the code yet, so please bare with me. 😊

src/portable/ehci/ehci.c
pete-pjb: As above...
> I will check this, I am still getting to grips with the USB protocol in general and with the code base, I haven't had time to extensively study the code yet, so please bare with me. 😊

src/portable/ehci/ehci.c
pete-pjb: Personally, I wasn't keen on using cache functions inside the interrupt handler as it adds latency to the interrupt, on my platform that might be around 6uS worst case which isn't hideous but it will vary on other platforms.  Do we want to try it this way or try and keep the cache code away from the ISR?  

What do you think? 

Thanks for your feedback @hathach.

Kind Regards,

Pete

src/host/usbh.c
pete-pjb: I'll check this too... 🙂

src/portable/ehci/ehci.c
pete-pjb: Hi @hathach, 

Yes this is my bad don't know what I was thinking 😊

src/host/hcd.h
hathach: thanks, It is only minor things, I sometimes mix thing up a bit.

src/portable/ehci/ehci.c
hathach: sure we can skip dcache or any function within ISR if possible. I think you could try to both invalidate and flush following memory location in hcd_edpt_xfer() an hopefully this would be enough:
- data buffer
- qtd (since ehci controller update this)
- qhd (ehci also update it via qtd overlay).

src/host/usbh.c
hathach: it should be removed, we will do it in hcd layer.

hathach:
> Hi @hathach ,
> 
> Just a quick update, this has turned out to be much more work than I expected. I am now spending some time reading up on the USB/EHCI specifications so I can do this correctly. I have also had a request from my boss to add MSC support also which means I really should get a good cached version of the EHCI driver working as the performance will be much better. It wasn't so important previously for my application as I was just handling HID devices. This also means I need to include a hub in my hardware configuration so I have also enabled that in my configuration too. So am now working on a fully cached version with hub support, although I don't think the hub really matters as it appears to use the same underlying EHCI functions so once it is all correct it should just fall in to place I hope blush
> 
> I will add a commit and come back to you when I have something new for you to take a look at. I also realise in my previous post I mentioned I wasn't keen on using d-cache functions within interrupt handlers but I don't think in this case it can be avoided so you can disregard that question/thread.
> 
> Kind Regards,
> 
> Pete

no problems at all, it surely take some time. In the meanwhile, could you revert following changes so that we could wrap up this PR even though it is only a minor update
- all changes in ehci.c
- hcd_dcache_flush() call in usbh.c

That will make your next PR cleaner with complete focus on dcache in ehci. Also this avoid conflict since master branch move often enough.

pete-pjb:
Hi @hathach ,

Happy New Year!

I am going to close this PR and submit a new one which just makes the small changes as my development system is having issues right now.  I will then create a new PR for the caching code later.

Kind Regards,

Pete

PR #1796, opened 2022-12-13
Fix ifdefs for disabled uart stdio in rp2040 family

Sinitax:
Fix ifdefs such that uart is not used for stdio by tinyusb when it is disabled via pico_enable_stdio_uart in CMakeLists.txt.

Without the extra guards, the compiler infers that a null pointer deref will take place when the pico stdio uart object is accessed in tinyusb and aborts compilation.


hathach (approved):
thank you for your PR. I made some changes to move the macro check to board.h to make it easier to follow

PR #1810, opened 2022-12-22
fix incorrect rx buf in cdc

hathach:
**Describe the PR**
This typo exists a long time ago, and it is a total mystery why tinyusb does not crash and works flawlessly so far !!

### DIFF
--- a/src/class/cdc/cdc_device.c
+++ b/src/class/cdc/cdc_device.c
@@ -432,7 +432,7 @@ bool cdcd_xfer_cb(uint8_t rhport, uint8_t ep_addr, xfer_result_t result, uint32_
   // Received new data
   if ( ep_addr == p_cdc->ep_out )
   {
-    tu_fifo_write_n(&p_cdc->rx_ff, &p_cdc->epout_buf, (uint16_t) xferred_bytes);
+    tu_fifo_write_n(&p_cdc->rx_ff, p_cdc->epout_buf, (uint16_t) xferred_bytes);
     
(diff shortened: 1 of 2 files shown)


dhalbert:
I compiled CircuitPython with `p_cdc->epout_buf` and with `&p_cdc->epout_buf`, and the resulting .elf and .asm's are identical.

This is a peculiarity of references to C arrays: https://stackoverflow.com/questions/2528318/how-come-an-arrays-address-is-equal-to-its-value-in-c.

So though it's a little "wrong", adding the `&` doesn't do anything.

There's a demo program in the link above, and here's one I wrote before finding that:
```c
#include <stdio.h>

typedef struct {
    char buf[2];
} s_t;

int main() {
    s_t s;
    s_t *sp = &s;

    void *p1 = sp->buf;
    void *p2 = &sp->buf;

    printf("p1: %p, p2: %p\n", p1, p2);
}
```
```
$ ./a.out
p1: 0x7ffcc06db176, p2: 0x7ffcc06db176
```

hathach:
thank you @dhalbert for looking deeper and detail explanation. Luckily `array and `&array` have the same value, C is really a mystery language :D

hathach:
I don't understand your point. This is bug/typo and as mentioned, it should crash previously in my opinion but it doesn't (pure luck). The code speaks for itself, what else I should put the comment for ?

> I use tinyusb and most of the changes that I see do not have consequences for usability.

Then don't upgrade if you have no need for those changes.

PR #1820, opened 2022-12-27
Fix UVC probe and commit on MacOS

Staacks:
**Describe the PR**
This is a minor change to the USB Video class implementation to accept a too small wLength during probe and commit requests in order to fix MacOS compatibility.

**Additional context**
My (UVC project)[https://github.com/Staacks/gbinterceptor] as well as a minimal example on a Raspberry Pi Pico based on the examples here fails on MacOS during SET CUR Probe because wLength is set to 0x22 instead of the expected 0x30. This behavior can actually be observed with WireShark on MacOS for any commercial webcam that I plugged in and it is marked by Wireshark as a malformed package.

However, if I understand section 4.1 of the UVC 1.5 definition correctly, this behavior is allowed:

> If the parameter block is longer than is indicated in the wLength field, only the initial bytes of the
parameter block are returned. If the parameter block is shorter than is indicated in the wLength
field, the device indicates the end of the control transfer by sending a short packet when further
data is requested.

But even if I misinterpret this (this is the first time I really look into these specs) and it is a bug by Apple, I still recommend to accept too small wLength values here as I do not see any relevant downside to doing so.

### DIFF
--- a/src/class/video/video_device.c
+++ b/src/class/video/video_device.c
@@ -917,7 +917,7 @@ static int handle_video_stm_cs_req(uint8_t rhport, uint8_t stage,
       switch (request->bRequest) {
         case VIDEO_REQUEST_SET_CUR:
           if (stage == CONTROL_STAGE_SETUP) {
-            TU_VERIFY(sizeof(video_probe_and_commit_control_t) == request->wLength, VIDEO_ERROR_UNKNOWN);
+            TU_VERIFY(sizeof(video_probe_and_commit_control_t) >= request->wLength, VIDEO_ERROR_UNKNOWN);
             TU_VERIFY(tud_control_xfer(rhport, request, self->ep_buf, sizeof(video_probe_and_commit_control_t)),
                       VIDEO_ERROR_UNKNOWN);
           } else if (stage == CONTROL_STAGE_DATA) {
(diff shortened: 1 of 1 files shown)


kkitayam:
Thank you for the PR.

> My (UVC project)[https://github.com/Staacks/gbinterceptor] as well as a minimal example on a Raspberry Pi Pico based on the examples here fails on MacOS during SET CUR Probe because wLength is set to 0x22 instead of the expected 0x30. 

I guess MacOS runs as a UVC 1.1 host. UVC 1.1 spec. shown below
image
And UVC 1.5 spec. shown below.
image

> However, if I understand section 4.1 of the UVC 1.5 definition correctly, this behavior is allowed:

It seems to be the description for `Get Request`, not for `Set Request`. SET CUR is a `Set Request`. So, I guess the behavior does not apply to SET CUR.

> But even if I misinterpret this (this is the first time I really look into these specs) and it is a bug by Apple, I still recommend to accept too small wLength values here as I do not see any relevant downside to doing so.

Yes, that's right. I think it is better to be able to run on a actual machine than to strictly comply with the specifications.

I will check the PR.

kkitayam:
I have checked that there are no side-effects. No problem.
On macOS,  6 bytes following UVC 1.1 data member are not initialized by the host. However, the current implementation does not use these data.

kkitayam (approved):

hathach:
thank you @Staacks for the fix and @kkitayam for reviewing

Staacks:
Great, thanks! Now I have to figure out why M1/M2 platforms discard half the transmitted data while it is working perfectly on MacBooks with Intel CPUs...

PR #1823, opened 2022-12-28
fix(fuzz): Make sanitizer flags optional

silvergasp:
**Describe the PR**
Currently OSS fuzz expects to have complete control over the sanitizer flags. As we currently have these set it's causing problems with the OSS fuzz build. Instead we should use the provided variables from the OSS fuzz build environment. For local testing we'll create a set a well defined defaults.

**Additional context**
https://github.com/google/oss-fuzz/actions/runs/3761908047

### DIFF
--- a/.github/workflows/pre-commit.yml
+++ b/.github/workflows/pre-commit.yml
@@ -38,6 +38,8 @@ jobs:
 
     - name: Build Fuzzer
       run: |
+        export CC=clang
+        export CXX=clang++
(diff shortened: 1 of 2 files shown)


silvergasp:
This is been locally tested for integration with OSS fuzz, and everything is working. I had locally tested before, though I missed a few things. Fingers crossed this fixes those issues.

hathach (approved):
Super! Thank you very much, I am on travel and haven’t tested it locally but it look great

PR #1825, opened 2022-12-29
Minor audio_test example fix.

HiFiPhile:
**Describe the PR**
Fix CFG_TUD_AUDIO_EP_SZ_IN size.

**Additional context**
Tested with STM32L072 on WIP stm32_fsdev driver.

### DIFF
--- a/examples/device/audio_test/src/main.c
+++ b/examples/device/audio_test/src/main.c
@@ -71,7 +71,7 @@ audio_control_range_2_n_t(1) volumeRng[CFG_TUD_AUDIO_FUNC_1_N_CHANNELS_TX+1]; 		
 audio_control_range_4_n_t(1) sampleFreqRng; 						// Sample frequency range state
 
 // Audio test data
-uint16_t test_buffer_audio[CFG_TUD_AUDIO_EP_SZ_IN/2];
+uint16_t test_buffer_audio[(CFG_TUD_AUDIO_EP_SZ_IN - 2) / 2];
 uint16_t startVal = 0;
 
 void led_blinking_task(void);
@@ -400,7 +400,7 @@ bool tud_audio_tx_done_pre_load_cb(uint8_t rhport, uint8_t itf, uint8_t ep_in, u
   (void) ep_in;
   (void) cur_alt_setting;
 
-  tud_audio_write ((uint8_t *)test_buffer_audio, CFG_TUD_AUDIO_EP_SZ_IN);
+  tud_audio_write ((uint8_t *)test_buffer_audio, CFG_TUD_AUDIO_EP_SZ_IN - 2);
 
   return true;
 }
@@ -413,7 +413,7 @@ bool tud_audio_tx_done_post_load_cb(uint8_t rhport, uint16_t n_bytes_copied, uin
   (void) ep_in;
   (void) cur_alt_setting;
 
-  for (size_t cnt = 0; cnt < CFG_TUD_AUDIO_EP_SZ_IN/2; cnt++)
+  for (size_t cnt = 0; cnt < (CFG_TUD_AUDIO_EP_SZ_IN - 2) / 2; cnt++)
   {
     test_buffer_audio[cnt] = startVal++;
   }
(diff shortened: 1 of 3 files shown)


HiFiPhile:
With Bus Hound capture I've something funky....
In every 10ms interval there are 9 concussive sample lost (0xcf66 to 0xcf70) and 9 zero samples... I can ensure what is written into the PMA buffer is right.

image

HiFiPhile:
With my logic analyzer decoder dump (limited USB analyze capacity) it seems streaming data is correct, which makes me wondering if Windows has some UAC2 implementation issue...

[2022-12-29_16-15-24.csv](https://github.com/hathach/tinyusb/files/10320184/2022-12-29_16-15-24.csv)

PanRe:
Have you noticed this behaviour with the other MCUs as well? Do you have the chance to test it on a linux system?

PanRe:
I did not notice such missing samples on windows when i tested it with the saw tooth signal, there were no missing samples

HiFiPhile:
> I did not notice such missing samples on windows when i tested it with the saw tooth signal, there were no missing samples

How did you make Windows to enumerate without adding one sample to CFG_TUD_AUDIO_EP_SZ_IN ?

PanRe:
I mean the audio_test example available in the example folder.

HiFiPhile:
> I mean the audio_test example available in the example folder.

Yep, without changing CFG_TUD_AUDIO_EP_SZ_IN to 49*n Windows will report error 10, that's why I'm fixing it :)

You did the same thing in `audio_4_channel_mic` example
```
#define CFG_TUD_AUDIO_EP_SZ_IN                        (48 + 1) * CFG_TUD_AUDIO_FUNC_1_N_BYTES_PER_SAMPLE_TX * CFG_TUD_AUDIO_FUNC_1_N_CHANNELS_TX      // 48 Samples (48 kHz) x 2 Bytes/Sample x CFG_TUD_AUDIO_N_CHANNELS_TX Channels - the Windows driver always needs an extra sample per channel of space more, otherwise it complains... found by trial and error
```

If I export the recording to csv I don't have empty sample but sometimes the value is jumpy, seems not bit-perfect:
```
31008
31009
31009
31011
31012
31013
31014
31014
31016
31017
31018
31019
31020
31021
31022
31023
31024
31025
31026
31027
31028
31029
31029
31031
31032
31033
31034
31035
31036
31037
31038
31039
31041
31041
```

PanRe:
What with debian it works? At first glance i thought the MCU might have been too slow in updating the saw tooth values between the frames. 

I once also found this 49*n value by trial and error and thought that might have been a precausion from the windows UAC2 driver to have a sufficiently big buffer.

PanRe:
And what i also found was "noisy" recordings i.e. a saw tooth signal with noisy values added. You can test for that by sending only zeros.

This was an impact of the UAC driver chosen in Windows. One (i dont recall which) had this problem, another (the one for the 4 channel if i remember correctly) did not show this noisy samples. My guess was that some drivers conduct some digital filtering in some weired manner...

HiFiPhile:
>  At first glance i thought the MCU might have been too slow in updating the saw tooth values between the frames.

My logic analyzer attached to D+/D- line dumped correct 48 samples so it can't be the cause.

> And what i also found was "noisy" recordings i.e. a sawtooth signal with noisy values added.

For a moment I had noisy sample on Linux, but it disappeared adjusting volume and I can't reproduce it.

My finding is Linux handle volume in host side, but volume set by SET_REQUEST is very strange:
vol=-90 if volume slider < 0dB
vol=0 if volume slider = 0dB
vol=90 if volume slider > 0 dB

While Windows handle volume in device side, in `audio_test` since samples are not multiplied by volume, no matter how the volume slider is changed recorded samples are not changed.

HiFiPhile:
I think it's safe now, captured with Wireshark looks good.
[audio_test_capture.zip](https://github.com/hathach/tinyusb/files/10324049/audio_test.zip)

I've also redid a BusHound capture on a Realtek sound card which gives also zero samples, so it should be a BusHound / Windows issue.

PanRe:
Great to hear! I will do the review today once i get some spare time! As always, thanks for your effort in enhacing our audio driver! :)

HiFiPhile:
@PanRe Hello, did you get some time lately :)

PanRe (approved):
Nice work and VERY sorry this delay!!

PR #1828, opened 2022-12-29
stm32_fsdev & ISO EP buffer allocation improvements 

HiFiPhile:
**Describe the PR**
For `stm32_fsdev`:
- Minor fixes
- Rework FIFO based transfer
- Add an endpoint allocator to overcome ISO ep limitation

For ISO transfer:
- Add dcd_edpt_iso_alloc() / dcd_edpt_iso_activate() for ISO EP buffer allocation
- Handle new allocation method in UAC2.

**Additional context**


HiFiPhile:
I think it's better to add a flag somewhere to choose new buffer allocation method or continue use old one.
For example in `tusb_mcu.h` add TUP_DCD_ISO_ALLOCATION, or add it inside `audio_device.c` like how USE_LINEAR_BUFFER is handled.

HiFiPhile:
@kkitayam Hi, I'm working on ISO EP buffer allocation based on #540, since video class is also ISO, does it benefit from this ?
Is this new dcd_edpt_iso_alloc() / dcd_edpt_iso_activate() suitable for video ?

kkitayam:
@HiFiPhile 

Hi, thanks for your comments.

Maybe yes, the APIs would be suitable for UVC too.
My understanding is that UVC controls ISO EPs in basically the same manner as UAC.

By the way, UVC can also use bulk EPs instead of ISO EPs for non-linear editing. Current UVC implementation has not supported this.
If we try to add support bulk EPs for UVC, we might be need some APIs like as dcd_edpt_iso_alloc() / dcd_edpt_iso_activate() for bulk EPs. It's just a guess. I am not familiar for bulk video data.

HiFiPhile:
A new audio test example with multiple formats / data rates is added, which benefit from the new buffer allocation method :)

image

PanRe:
Yes i think this is a good idea! For us it might be a bit more work but the user does not need to be aware of this new allocation feature and we thus will get less issues reported ;)

Great work btw!

hathach:
I am on travel, will check this out next week.

src/portable/st/stm32_fsdev/dcd_stm32_fsdev_pvt_st.h
```diff
@@ -119,26 +119,50 @@
...
+// By default the tinyusb stack chooses the endpoint register using the endpoint number
+// However when using ISOCHRONOUS endpoints, the entire endpoint register (RX _and_ TX)
+// are used for a double buffer in a single direction. If you now want to use the same endpoint number with
+// different directions (e.g. 0x02 and 0x82), they would be mapped to the same register
+// In this case, this serves as a crude helper to overcome this limitation by implementing
+// this callback and defining a custom mapping in your application
+TU_ATTR_WEAK uint8_t tu_stm32_edpt_number_cb(uint8_t addr);
+
+// This callback is called on SOF and can be used to e.g. capture a timer value for timing purposes
+TU_ATTR_WEAK void tu_stm32_sof_cb(void);
```
HiFiPhile, line 123: @skuep 

For the moment I kept your SOF callback but for the stack it's not quite the right place to do it.

Did you test the `tud_audio_feedback_interval_isr` and found out it has too much jitter ?

If a prioritized SOF callback is needed we need do it in a generalized way for other DCD drivers.

skuep: Yeah I am also still on the fence with this callback. 
From the looks of it, the generated feedback values in the speaker feedback endpoint had way less jitter for the UAC2 application. However I did not check if this actually makes a difference for the (stability of the) feedback control loop running on the Host. Since I can only test Linux hosts, I left it in there for a better gut feel in terms of compatibility.

I do however think it would be a nice addition, if this would be a more general solution across all DCD drivers.

skuep:
Sorry for the delay in testing your branch. However I have found a serious bug in the code that you based your commits on.
I updated the original branch that you used. See the latest commit here:
https://github.com/hathach/tinyusb/compare/master...skuep:tinyusb:0.14.0-sk

I first noticed corrupted memory on only some of the devices, which was really hard to debug. I narrowed it down to the TU_ASSERT that I added to the code in this commit. It turns out, that you need to set the receive buffer-size with ISOCHRONOUS endpoints for both double buffer count registers in the buffer descriptor table. I only set it for the ``USB_COUNTn_RX`` register, because the fields are not even there in the manual for the corresponding ``USB_COUNTn_TX`` register. Turns out, they are there after all and need to be set. Otherwise wildly varying (unitialized?) number of received bytes where reported, leading the ``dcd_read_packet_memory``call to corrupt more buffer than it was allowed to.

HiFiPhile:
@skuep Thanks for your update ! Indeed I have only tested IN EP.

skuep:
Alrighty!
I have switched my tinyusb submodule reference to your branch. I adjusted the SOF timing measurement code (due to removal of the ``tu_stm32_sof_cb`` callback. What can I say, it works very well so far!
The PMA and hardware endpoint allocation works quite nice, verified everything looks good in the ``xfer_status``map.

From my side of view, testing was successful! Thanks a lot

HiFiPhile:
@skuep 
That's a great news !

PS: One thing very useful is J-Scope, I used it to visualize FIFO level while adjusting feedback loop.

skuep (approved):
Tested successfully with the AIOC project
https://github.com/skuep/AIOC

src/portable/st/stm32_fsdev/dcd_stm32_fsdev.c
```diff
@@ -1027,64 +1224,74 @@ static bool dcd_write_packet_memory(uint16_t dst, const void *__restrict src, si
...
-    temp2 = temp1 | ((uint16_t)((uint16_t) ((*srcVal) << 8U))) ;
+    temp2 = temp1 | ((uint16_t)(((uint16_t)(*srcVal)) << 8U)) ;
     *pdwVal = temp2;
     pdwVal += PMA_STRIDE;
     srcVal++;
   }
+
+  if (wNBytes & 0x01)
+  {
+    temp1 = *srcVal;
```
hathach, line 1256: this would cause an read to out-of-bound for the memory. We should cast it to `uin8_t*` first before doing a read ?

hathach (commented):
Superb!! Sorry for huge delay, I am in the middle for reviewing this PR. Everything looks great, I am doing mostly clean up to make it more readable. Number of #ifdef in audio driver is starting to look scary now. We should do some refactoring in the future. Overall this PR is awesome and ready to merge, however there is a note in the review, that I think we should change to prevent a read out-of-bound of an array.

src/portable/st/stm32_fsdev/dcd_stm32_fsdev.c
HiFiPhile: No problem I know you are busy:) I'm on vacation I'll look at it later (I think I thought about out of bounds access)

hathach: Then dont worry about this, I can make changes myself, just want to confirm with you first

hathach: oh right, sorry, I got confused and mixed-up, all is good then

hathach (approved):
everything looks good. Thank you very much for putting your effort to this brilliant PR.

PR #1836, opened 2023-01-02
Fix IAR warning: Warning[Pe381]: extra ";" ignored

HiFiPhile:
**Describe the PR**
Seems IAR is not happy with empty `;` in struct
```c
typedef struct {
  bool is_host; // host or device most
  union {
      uint8_t daddr;
      uint8_t rhport;
      uint8_t hwid;
  };
  uint8_t ep_addr;
  uint8_t ep_speed;

  uint16_t ep_packetsize;
  uint16_t ep_bufsize;

  // TODO xfer_fifo can skip this buffer
  uint8_t* ep_buf;

  tu_fifo_t ff;

  // mutex: read if ep rx, write if e tx
  OSAL_MUTEX_DEF(ff_mutex);

}tu_edpt_stream_t;

```
**Additional context**
https://stackoverflow.com/questions/18611303/warning-extra-ignored-by-using-c-macros-with-armcc


HiFiPhile:
Hummm it's either GCC or IAR not happy :(

PR #1837, opened 2023-01-03
Fix typo in audio.h. Specifiy _ctrl_xfer struct in CFG_TUSB_MEM_SECTION in usbh.c

pete-pjb:
Add NULL check to loop in list_remove_qhd_by_addr() function in ehci.c

See discussion [here](https://github.com/hathach/tinyusb/discussions/1769#discussioncomment-4377690) for some background about this PR


hathach (approved):
thank you for your PR. This should fix #1782

PR #1844, opened 2023-01-09
feat(ci): Adds oss-fuzz integration on PRs

silvergasp:
**Describe the PR**
This PR will trigger the fuzzer to run for 10min on every PR with the goal of catching bugs **before** they are merged.

This is intended as a more "surface level" fuzzing to catch obvious bugs before they are merged, the complete OSS-fuzz system will find deeper bugs that you can't find in 10min minutes or so of fuzzing. We can of course extend this time from the default of 10min if you want a "deeper" fuzzing analysis for the PR tests.

**Additional context**
Documentation: https://google.github.io/oss-fuzz/getting-started/continuous-integration/
Testing trigger of ci (on my local fork): https://github.com/silvergasp/tinyusb/actions/runs/3877844318/jobs/6613332627

### DIFF
--- a/.github/workflows/cifuzz.yml
+++ b/.github/workflows/cifuzz.yml
@@ -1,0 +1,33 @@
+name: CIFuzz
+on:
+  pull_request:
+    branches:
+      - master
+    paths:
+      - '**.c'
+      - '**.cc'
+      - '**.cpp'
+      - '**.cxx'
+      - '**.h'
+jobs:
+ Fuzzing:
+   runs-on: ubuntu-latest
+   steps:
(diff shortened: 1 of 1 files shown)


silvergasp:
This is intended as a more "surface level" fuzzing to catch obvious bugs before they are merged, the complete OSS-fuzz system will find deeper bugs that you can't find in 10min minutes or so of fuzzing. We can of course extend this time from the default of 10min it's you want a "deeper" fuzzing analysis for the PR tests.

hathach (approved):
superb !! Thank you very much for your effort to make tinyusb safer

PR #1847, opened 2023-01-11
dwc2: fix iar warnings

HiFiPhile:
**Describe the PR**
Fix:
Warning[Pe188]: enumerated type mixed with another type	dwc2_stm32.h	125	
Warning[Pa082]: undefined behavior: the order of volatile accesses is undefined in this statement	dcd_dwc2.c	532	
Warning[Pa082]: undefined behavior: the order of volatile accesses is undefined in this statement	dcd_dwc2.c	1221


hathach (approved):
thank you. IAR seems to be even more picky than gcc

PR #1853, opened 2023-01-13
examples/uac2: Fix mute and volume array lengths

NullMember:
**Describe the PR**
mute and volume arrays in main.c have wrong number of elements (2 instead of 3) because CFG_TUD_AUDIO_FUNC_1_N_CHANNELS_TX was used to calculate array lengths. CFG_TUD_AUDIO_FUNC_1_N_CHANNELS_TX changed to CFG_TUD_AUDIO_FUNC_1_N_CHANNELS_RX.

**Additional context**
Changing the volume of second channel ends up soft brick of device in my case because OS will remember last volume control.


HiFiPhile (approved):
Thanks for the fix.

PR #1854, opened 2023-01-13
Update to building the NXP frdm_k32l2b examples

bpaddock:
The NXP frdm_k32l2b BSP, as currently shipped with TinyUSB no longer build due to changes in the upstream NXP IDE and SDK.
The start up file and linker files currently referenced in board.mk no longer exist, making it impossible to build the BSP. 

This PR modifies board.mk to reference the current NXP IDE and SDK files.
It also adds a README file to explain which files are needed from the SDK and how to set up the build using them.

### DIFF
--- a/hw/bsp/frdm_k32l2b/README.rst
+++ b/hw/bsp/frdm_k32l2b/README.rst
@@ -1,0 +1,51 @@
+Jan/13/2023 13:04
+
+The FRDM-K32L2B3 Freedom development board provides a platform for
+evaluation and development of the K32 L2B MCU Family. -
+https://www.nxp.com/part/FRDM-K32L2B3#/
+
+TinyUSB does not include the board specific drivers. Those drivers
+need to be extracted from the MCUXpresso IDE and SDK.
+
+Install MCUXPresso version 11.6 or later and SDK 2.12 or later.  Then
+build the example project "frdmk32l2b_hellow_worldvirual_com".
+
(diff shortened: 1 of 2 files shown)


hw/bsp/frdm_k32l2b/README.rst
```diff
@@ -0,0 +1,51 @@
...
+https://www.nxp.com/part/FRDM-K32L2B3#/
+
+TinyUSB does not include the board specific drivers. Those drivers
+need to be extracted from the MCUXpresso IDE and SDK.
+
+Install MCUXPresso version 11.6 or later and SDK 2.12 or later.  Then
+build the example project "frdmk32l2b_hellow_worldvirual_com".
+
+From the frdmk32l2b_hellow_worldvirual_com project copy these files to
+this directory structure, in this directory:
```
hathach, line 14: we should only need to init the submodule for nxp hw/mcu/nxp/mcux-sdk

hathach (changes requested):
have you tried init nxp submodule either with git command or `make BOARD=yourboard get-deps` ? if not please try it first

Most if not all of the board.mk modification should be reverted.

hw/bsp/frdm_k32l2b/README.rst
bpaddock: I will close this pull request and continue in #1858 where we are discussing merging the 2A and 2B families.

PR #1858, opened 2023-01-16
Add support for the NXP FRDM-K32L2A4S eval board.

bpaddock:
Add support NXP FRDM-K32L2A4S evaluation board.

The K32L2A4S uses a different core, and in some cases different I/O configuration registers, than the existing FRDM-K32L2B3 evaluation board port.


hathach (commented):
thank you for your PR. The K32L2AXX is similar to K32L2BXX, that we could merge them together as K32L2 family. If you are not familiar, just simply use the K32L2BXX for this new board. I will make a rename later on.

PS: BSP need to use the git submodule nxp mcux as mentioned in the review of #1854

bpaddock:
I will take a look at the submodules.  I was unaware of their use here.

While the two FRDM boards are similar, their cores, memory map and power management are different.
The B series is aimed at communication apps with dual core M4/M0+ while the A series is for security and has only single M0+ core. The B series is similar to most Kineits parts with SIM_OPTS etc.  The A series is its own thing with Peripheral Clock Control (PCC) etc.


To unify the code how do I tell which board is being built at the C level?  Is there some #define of $(BOARD) passed to C?  May just be BOARD, I've not looked yet.

hw/bsp/frdm_k32l2a4s/board.mk
```diff
@@ -0,0 +1,80 @@
...
+# "VERBOSE=1" to the make command line.
+ifndef VERBOSE
+.SILENT:
+$(info Non-Verbose Output)
+else
+$(info Verbose Output)
+endif
+
+SDK_DIR = hw/mcu/nxp/mcux-sdk
+MCU_DIR = $(SDK_DIR)/devices/K32L2A4S
```
hathach, line 12: the mcu on this board should be K32L2A41A though

hathach (commented):
thank you for your PR, I spotted a couple of issue. Though I have pulled and make more changes, so no worries

hathach (approved):
perfect now, thank you

PR #1860, opened 2023-01-18
Allow config OS for RP2040

howard0su:
**Describe the PR**
The RP2040 cmake script hardcoded TUSB_OS, make it configurable when using other OS like FreeRTOS

**Additional context**
The change will default to OPT_OS_PICO to keep back compatiblity


hathach (approved):
Look good, thank you for your PR

PR #1864, opened 2023-01-20
Add _in and _out variants of tud_control_xfer for type cleanliness

mikeditto:
This allows transferring data IN from a const buffer without compiler warnings.

Also changed several uses of `tud_control_xfer` on const data to use `tud_control_xfer_in` instead, eliminating ugly casts.  Did not change all uses - only the ones that were clearly acting on const data and had ugly casts that could be eliminated.

A more aggressive approach would be to modify all calls in the IN direction to use the new function.
An even more aggressive approach would be to modify all `tud_control_xfer` calls in either direction to use the new functions, and perhaps deprecate the old function as well.

### DIFF
--- a/examples/device/webusb_serial/src/main.c
+++ b/examples/device/webusb_serial/src/main.c
@@ -179,7 +179,7 @@ bool tud_vendor_control_xfer_cb(uint8_t rhport, uint8_t stage, tusb_control_requ
         case VENDOR_REQUEST_WEBUSB:
           // match vendor request in BOS descriptor
           // Get landing page url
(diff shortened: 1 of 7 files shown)


mikeditto:
Unfortunately, I can't say I've tested this on the current master branch.
I have tested it in my project based on 0.12.0.

HiFiPhile:
Thanks for your PR but it's not something we want.

Normally `tud_control_xfer()` is an internal API won't be accessed outside, in the case of custom class one can case the const pointer in user land.

PR #1866, opened 2023-01-20
Add support for NXP's mimxrt1024_evk evaluation board.

DRNadler:
Tested AOK with device\cdc_msc_freertos example.


hathach (approved):
thank you for the PR, I have pushed some update to suppress warnings from system_.c and fix the typo. Everything looks great now.

Issue #1715, opened 2022-11-02
Add fuzz testing harnesses


silvergasp:
### Related area

Fuzz testing

### Hardware specification

Host only

### Is your feature request related to a problem?

No

### Describe the solution you'd like

I'd like to improve the security and robustness of the tinyusb stack by adding a set of fuzz tests.

### I have checked existing issues, dicussion and documentation

- [X] I confirm I have checked existing issues, dicussion and documentation.

silvergasp:
I'm happy to spend some time contributing to this one. I've got some general questions first though;
1. Is it acceptable to you to have the fuzz harnesses written in a combo of c/c++. These fuzz harnesses would be run from the host only, so I think the overhead of c++ is acceptable given the ease of use of the [fuzzing libs available in c++](https://github.com/llvm/llvm-project/blob/main/compiler-rt/include/fuzzer/FuzzedDataProvider.h).
2. Do you have the bandwidth to give some guidance on WIP PRs?

Here is a list of the things that I would like to add;
- [ ] Add build infrastructure for fuzz tests. (WIP) #1716
- [ ] Add a fuzz harness for each class.
    - [ ] audio
    - [ ] bth
    - [ ] cdc (WIP) #1716
    - [ ] dfu
    - [ ] hid
    - [ ] midi
    - [ ] msc
    - [ ] net
    - [ ] usbtmc
    - [ ] vendor
    - [ ] video
- [ ] Add CI configuration to continuously fuzz.
- [ ] Add dynamic configs for composite class (e.g. cdc+msc, or cdc+cdc+audio etc.).

silvergasp:
I'd like to add an additional question. Would you be open to using google's OSS-fuzz service for continuous fuzzing? It's a free service and will fuzz and send you updates when bugs are found.

https://github.com/google/oss-fuzz

hathach:
Awesome, thank you very much for spending time and effort to adding fuzzing support. 

> 1. Is it acceptable to you to have the fuzz harnesses written in a combo of c/c++. These fuzz harnesses would be run from the host only, so I think the overhead of c++ is acceptable given the ease of use of the [fuzzing libs available in c++](https://github.com/llvm/llvm-project/blob/main/compiler-rt/include/fuzzer/FuzzedDataProvider.h).

Sure thing, I have no problems with C/C++ combo at all as long as it is easy to use/maintain. 

> 2. Do you have the bandwidth to give some guidance on WIP PRs?

[...]

PR #1867, opened 2023-01-20
fix: Change all static variables to thread when fuzzing

silvergasp:
**Describe the PR**
Currently the AFL fuzzing engine makes the assumption that it can run a library in multiple threads without any shared state. To make this assumption consistent with tinyusb we need to make static globals thread local when fuzzing.

**Additional context**
_Discussed here:_
https://github.com/hathach/tinyusb/issues/1715#issuecomment-1396436517
_Actual issue here (see stability)_
https://oss-fuzz.com/fuzzer-stats?project=tinyusb&fuzzer=afl&job=afl_asan_tinyusb&group_by=by-fuzzer

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,3 +1,0 @@
-{
-    "editor.formatOnSave": false
-}
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,3 @@
+{
+    "editor.formatOnSave": false
+}
--- a/src/class/cdc/cdc_device.c
+++ b/src/class/cdc/cdc_device.c
@@ -76,7 +76,7 @@ typedef struct
 //--------------------------------------------------------------------+
 // INTERNAL OBJECT & FUNCTION DECLARATION
 //--------------------------------------------------------------------+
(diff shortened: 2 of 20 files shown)


hathach (approved):
thank you very much for your PR and sorry for the delay, I was too busy with other paid works. I back off a bit on using TU_STATIC, instead I would prefer to add _fuzz_thread as additional keyword. Even though it is a bit more verbose, the `static` keyword is more apparent to other user and easier to read. 

I also rename macro FUZZ to _FUZZ to avoid if user application define it in the future . 

Note: we could change `_fuzz_thread` to less verbose word in the future if you have any other suggestion. For now, I think this is good for merge

HiFiPhile:
Personally I prefer to use `TU_STATIC` instead of adding `_fuzz_thread` keyword.

For most people who doesn't use fuzzing it's confusing what this keyword does, and I feel also the code is less neat putting a test keyword inside the stack.

For anyone who contribute they also need to learn what fuzzer does instead of copy `TU_STATIC` from existing code.

hathach:
_fuzz_thread is `__thread` which is an actual keyword for `thread_local` (https://gcc.gnu.org/onlinedocs/gcc/Thread-Local.html#Thread-Local) . I was thinking to actual make use of it later on. Though you made the point, I think we should stick with `tu_static` instead.

PR #1870, opened 2023-01-22
Windows build fix for some mingw gnu make situations

DRNadler:
When tools/top.mk finds it is running on Windows, it expects SHELL set to cmd.exe.
Some Windows development environments invoke a mingw version of gnu make with SHELL set to /bin/sh or /bin/bash (either explicitly or by default), in which case tinyUSB make fails when a $(shell...) command feeds DOS syntax to sh. We can't just use sh, as many Windows environments won't have unix utilities like realpath (used by this makefile for sh shell commands).

This fix forces SHELL=cmd.exe on Windows and documents the issue. Also adds additional documentation to help the next person...

With this fix, tinyUSB can be easily built and debugged in environments like NXP's MCUxpresso (import as a makefile project).

### DIFF
--- a/tools/top.mk
+++ b/tools/top.mk
@@ -6,22 +6,26 @@ endif
 # https://stackoverflow.com/questions/714100/os-detecting-makefile/52062069#52062069
 ifeq '$(findstring ;,$(PATH))' ';'
 # PATH contains semicolon - so we're definitely on Windows.
-CMDEXE := 1# makefile shell commands should use syntax for DOS CMD, not unix sh
+CMDEXE := 1
+
+# makefile shell commands should use syntax for DOS CMD, not unix sh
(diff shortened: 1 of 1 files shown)


hathach (approved):
superb! thank you for your PR, I made a minor clean up on the comment. Haven't tested it on windows machine, but look good

DRNadler:
Thanks! Dumb question though: We all hate comment syntax in make and shell that essentially prohibits space before comment # mark. But, I put the following on one line so that it would be easy to see when someone greps for CMDEXE to find out it's exact meaning. So, sorry for the question, why did you break it into multiple lines?
Thanks again for your great work!
Best Regards, Dave
CMDEXE := 1# makefile shell commands should use syntax for DOS CMD, not unix sh

hathach:
it is just my personal preference, I found having the inline comment for makefile not as readable as one line above the actual command.

PR #1875, opened 2023-01-29
feat(fuzz): Adds seed corpus for cdc and msc classes

silvergasp:
**Describe the PR**
Adds a [seed corpus](https://google.github.io/oss-fuzz/advanced-topics/ideal-integration/#seed-corpus) for the fuzzer based on real USB packets captured using wireshark.

**Additional context**
By default, a fuzzer will try to randomly search for inputs that increase code coverage. In some cases this can take a really long time for the fuzzer to "learn" the structure of a protocol. One approach to mitigate this problem is to give the fuzzer a good starting point, with valid inputs. To do this I've captured a bunch of USB packets using wireshark. To make adding to this corpus easier I've added a tool to parse wireshark captures and add them to the zip corpus.


hathach (approved):
Superb !! Thank you very much again.

PR #1876, opened 2023-01-30
Support iar build

hathach:
**Describe the PR**
This PR add support for IAR build with self-hosted ci hosted by @HiFiPhile. Initial port includes stm32 f0, f1, f4, f7, g4, h7, l4. It is almost complete except for freeRTOS example where portasm.c (in the freeRTOS kernel) include FreeRTOSCOnfig.h which in turn include the tusb_option.h. Initially I thought the IAR Assembler cannot handle complicated macros, however, It look more like the issue is including C header file inside an .s file.

https://github.com/FreeRTOS/FreeRTOS-Kernel/blob/main/portable/IAR/ARM_CM4F/portasm.s#L29

UPDATE: portasm.s only needs `configPRIO_BITS`, since this is defined within mcu header which cannot be included directly by .s file (by IAR Asssembler). We will hardcorded it to 2 to pass CI with warning message for IAR user to update this manually according to target MCU.

Note: self-hosted seems to be under-powered with 7 jobs  running around 20 minutes. We may need to only enable selected ports (e.g with different dcd etc .. in the future).


hathach:
> I have only used the assembler for basic things such as copy vector table to ram.
> 
> It seems only thing needed by port.asm is `configPRIO_BITS`, is it possible to wrap other things by `#ifndef _ _IAR_SYSTEMS_ASM_ _` ?

@HiFiPhile let me move the discussion here to make it easier to follow. Indeed, I just figured it out, that including C header in the .s file is the root cause (after trying to conform with IAR Assembler). configPRIO_BITS is unfortunately defined within mcu header such as
https://raw.githubusercontent.com/STMicroelectronics/cmsis_device_f7/master/Include/stm32f722xx.h

which cannot be included in .s file . I think I will just hard coded it to minimum value of 2 to get passed CI and put an note there for IAR user to correct it per MCU by hand. Otherwise there will be a mismatch between Compiler & Assembler for the configPRIO_BITS

PR #1880, opened 2023-02-01
Bundle mulitple familes for self-hosted iar to reduce build time

hathach:
**Describe the PR**
IAR Build on self-hosted taking too long (~20mins), reduce the iar build to only family with different enough architecture/usb controller. Also update hardware test on self-hosted to download artifacts to reduce compile time for L4


HiFiPhile:
Didn't know it took such long time. Is it compile time or clone sdk ?

hathach:
> Didn't know it took such long time. Is it compile time or clone sdk ?

Checkout sumodule take about 30s or so, each build takes 2 minutes for each matrix. Actually, I think the issue is due to the matrix. Matrix scale well with a farm of PC. In our case it hurt the self-hosted since it run one after another on 1 instance. Let me try to bundle several family in 1 run, it would possible reduce the time to run

HiFiPhile:
> Let me try to bundle several family in 1 run, it would possible reduce the time to run
Yes I saw CPU usage is quite low.

hathach:
bundled everything, the build time only takes 11 minutes now.
 https://github.com/hathach/tinyusb/actions/runs/4063998654/jobs/6996895275
If the cpu usage is still low, it is probably due to my build_family.py script that doesn't do parallel enough. Though it is reasonable for now

hathach:
> Just doubled assigned RAM & Core, after reboot it should be quicker :)

that is great, I trigger rebuild to see if we got it faster :)

PR #1885, opened 2023-02-04
dhserver: Fix a potential DoS vulnerability accidentially introduced by #1712

kripton:
**Describe the PR**
My fix in dhcpserver.c for some DHCP clients (#1712) accidentally introduced a potential DoS vulnerability (I'm sorry!): In case a DHCP-packet WITHOUT `DHCP_MESSAGETYPE` was received, the code jumped out of the function without freeing the pbuf as it is required. When enough such messages are received, not pbufs are available anymore to handle traffic and the system might not be able to receive any more packets.
This PR fixes the problem by properly free()ing the pbuf in case no `DHCP_MESSAGETYPE` is detected in the packet.

Plus a trivial whitespace fix.


hathach (approved):
look great thank you for the fix :)

PR #1887, opened 2023-02-06
RP2040 build fixes

kilograham:
* don't force dependency on pico-pio-usb submodule
* suppress compiler warning/errors especially in pico-pio-usb
* disable use of pico-pio-usb on gcc 11.3+ for now as it doesn't compile

### DIFF
--- a/hw/bsp/rp2040/family.cmake
+++ b/hw/bsp/rp2040/family.cmake
@@ -161,14 +161,22 @@ if (NOT TARGET _rp2040_family_inclusion_marker)
 		rp2040_family_configure_example_warnings(${TARGET})
 	endfunction()
 
+	function(family_add_pico_pio_usb TARGET)
+		target_link_libraries(${TARGET} PUBLIC tinyusb_pico_pio_usb)
+	endfunction()
+
 	function(family_configure_host_example TARGET)
 		family_configure_target(${TARGET})
 		target_link_libraries(${TARGET} PUBLIC pico_stdlib tinyusb_host)
 		rp2040_family_configure_example_warnings(${TARGET})
-	endfunction()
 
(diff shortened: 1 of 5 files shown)


hw/bsp/rp2040/family.cmake
```diff
@@ -161,14 +161,22 @@ if (NOT TARGET _rp2040_family_inclusion_marker)
...
 		family_configure_target(${TARGET})
 		target_link_libraries(${TARGET} PUBLIC pico_stdlib tinyusb_host)
 		rp2040_family_configure_example_warnings(${TARGET})
-	endfunction()
 
-	function(family_add_pico_pio_usb TARGET)
-		target_link_libraries(${TARGET} PUBLIC tinyusb_pico_pio_usb)
+		# For rp2040 enable pico-pio-usb
+		if (TARGET tinyusb_pico_pio_usb)
+			# code does not compile with GCC 12+
```
kripton, line 175: The comment doesn't match the line below.
Does it fail with gcc 12+ (as the comment suggests) or with 11.3+ as the line below and your explanations in this PR suggest it?

liamfraser (approved):

hw/bsp/rp2040/family.cmake
kilograham: ah, it fails with 11.3 too (I updated the line but not the comment)

kilograham: I plan to push a fix to pico-pio-usb anyway

lurch: > I plan to push a fix to pico-pio-usb anyway

Just for cross-linking purposes: this happened in https://github.com/sekigon-gonnoc/Pico-PIO-USB/pull/70

PR #1893, opened 2023-02-11
Adjust reserved bitfields to make SDCC happy

### LINKED ISSUE #1892: bit field arrangement is not guaranteed by the C standard (cdc.h)
### Operating System

Linux

### Board

n/a

### Firmware

examples/device/cdc_msc

### What happened ?

Compilation under sdcc (4.0 and 4.2) fails on [the assert at line 410 of cdc.h](https://github.com/hathach/tinyusb/blob/86c416d4c0fb38432460b3e11b08b9de76941bf5/src/class/cdc/cdc.h#L410) which checks the size of `cdc_line_control_state_t`. The exact layout of bitfields is apparently implementation-defined and sdcc makes a different choice than gcc in this case. See [the SDCC discussion](https://sourceforge.net/p/sdcc/bugs/3530/) for more info. So this seems to be a kind of C standard conformance issue.

There is a straightforward workaround which I will supply in a PR.

### How to reproduce ?

Build tinyusb under sdcc for 8051, for example by using [my branch](https://github.com/jefftrull/tinyusb/tree/mcs51). Change src/class/cdc/cdc.h to match master and observe this error message:

`src/class/cdc/cdc.h:410: warning 215: static assertion failed: "size is not correct"`

### Debug Log as txt file

_No response_

### Screenshots

_No response_

jefftrull:
The open-source SDCC compiler has many limitations, but I have managed, after considerable hacking, to compile TinyUSB with it. This is one issue that I think is legitimately universal: compiler-specific assumptions about bitfield alignment which are implementation-defined per the Standard. Please see [the discussion with the SDCC maintainers](https://sourceforge.net/p/sdcc/bugs/3530/).

**Describe the PR**
This PR tweaks some reserved bitfields so the assumptions made elsewhere in the code (particularly in an assert) are more portable, and so they hew to implementation-defined limits on size. In particular:

- keep each field at or under 16b
- For optimal packing, segment bitfields to 8b boundaries

**Additional context**
N/A

If merged this will resolve #1892

### DIFF
--- a/src/class/cdc/cdc.h
+++ b/src/class/cdc/cdc.h
@@ -377,7 +377,9 @@ typedef struct TU_ATTR_PACKED
     uint32_t incoming_distinctive   : 1; ///< 0 : Reports only incoming ringing. 1 : Reports incoming distinctive ringing patterns.
     uint32_t dual_tone_multi_freq   : 1; ///< 0 : Cannot report dual tone multi-frequency (DTMF) digits input remotely over the telephone line. 1 : Can report DTMF digits input remotely over the telephone line.
     uint32_t line_state_change      : 1; ///< 0 : Does not support line state change notification. 1 : Does support line state change notification
-    uint32_t TU_RESERVED            : 26;
+    uint32_t TU_RESERVED0           : 2;
+    uint32_t TU_RESERVED1           : 16;
+    uint32_t TU_RESERVED2           : 4;
   } bmCapabilities;
 }cdc_desc_func_telephone_call_state_reporting_capabilities_t;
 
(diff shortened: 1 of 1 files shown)


src/class/cdc/cdc.h
```diff
@@ -377,7 +377,9 @@ typedef struct TU_ATTR_PACKED
     uint32_t incoming_distinctive   : 1; ///< 0 : Reports only incoming ringing. 1 : Reports incoming distinctive ringing patterns.
     uint32_t dual_tone_multi_freq   : 1; ///< 0 : Cannot report dual tone multi-frequency (DTMF) digits input remotely over the telephone line. 1 : Can report DTMF digits input remotely over the telephone line.
     uint32_t line_state_change      : 1; ///< 0 : Does not support line state change notification. 1 : Does support line state change notification
-    uint32_t TU_RESERVED            : 26;
+    uint32_t TU_RESERVED0           : 2;
+    uint32_t TU_RESERVED1           : 16;
+    uint32_t TU_RESERVED2           : 4;
```
hathach, line 382: 2+16+4 = 22, last one should be 8

hathach (commented):
only a typo in arithmetic

src/class/cdc/cdc.h
jefftrull: How embarrassing! I changed it at the last minute and got it wrong :(

hathach (approved):
look good now thank you

Issue #1894, opened 2023-02-11
commit 197689 breaks USB on mimxrt1010


ccrome:
### Operating System

Linux

### Board

mimxrt1010_evk

I have checked on the mimxrt1060_evk and it is *NOT* a problem on the 1060. 

### Firmware

As far as I can see, everything is broken. 

examples/device/msc_dual_lun, as an example is definitely broken.

### What happened ?

commit 19768936ae6379b4c86947f70ab5cf2c0f8ed8f4 completely breaks the USB on the mimxrt1010.  It appears that it is a problem in the SDK that was updated.

### How to reproduce ?

```
# git checkout 19768936ae6379b4c86947f70ab5cf2c0f8ed8f4
# git submodule update hw/mcu/nxp/mcux-sdk
# cd examples/device/msc_dual_lun
# rm -rf _build && make BOARD=mimxrt1010_evk flash && pyocd reset -t mimxrt1010
```

Expecting it to work.  It doesn't.  Sometimes I get "Cannot enable, Maybe the USB cable is bad".  Other times I just get nothing in syslog.

Now at this same tinyusb commit, do the following:
```
# cd hw/mcu/nxp/mcux-sdk
# git checkout cb21c660 # (This is the previous rev that worked for the 1010)
# cd examples/device/msc_dual_lun
# rm -rf _build && make BOARD=mimxrt1010_evk flash && pyocd reset -t mimxrt1010
```
And.... Voila! the 1010 works again.

In fact, you can even check out the master branch of tinyusb, and as long as you don't update the mcux-sdk submodule from cb21c660, it still works fine:
```
# git checkout master
# # ensure that mcux-sdk is still at cb21c660
# rm -rf _build && make BOARD=mimxrt1010_evk flash && pyocd reset -t mimxrt1010
# This works!
```

### I have checked existing issues, dicussion and documentation

- [X] I confirm I have checked existing issues, dicussion and documentation.

ccrome:
I have further "narrowed" the problem to commit c84411fd of mcux-sdk.

Narrowed is in quotes because it's a wholesale update of the SDK, so it's not trivial (for me anyway) to see where the issue is.

ccrome:
Upon further investigation, it appears that the SDK broke enabling the DCACHE on the 1011.  Not sure if that was intentional, but if you apply this diff, everything works again:

```
[...]

PR #1898, opened 2023-02-12
Turn data cache on

ccrome:
This fixes https://github.com/hathach/tinyusb/issues/1894.  I'm not really sure if this is the correct way to fix it, and I have not tested on all the rest of the family members, however, this lets the i.MX1010 work again.

The problem:  the latest SDK update does not enable the data cache by default This causes an assert in board_init() when attemping to control clock gating.  I haven't investigated further as to *why* it's a problem, but it is a problem.

**Describe the PR**
This PR solves an issue with the latest SDK, as reported in 1984.

**Additional context**
On the i.MX RT 1011, the board immediately crashes here:
```
Thread #1 57005 (Suspended : Signal : SIGTRAP:Trace/breakpoint trap)	
	HardFault_Handler() at startup_MIMXRT1011.S:457 0x600024ea	
	<signal handler called>() at 0xfffffff9	
	_vfprintf_r() at nano-vfprintf.c:531 0x60010660	
	fprintf() at fprintf.c:54 0x60010302	
	__assert_func() at assert.c:58 0x600102a8	
	CLOCK_ControlGate() at fsl_clock.h:1,085 0x60004640	
	CLOCK_EnableClock() at fsl_clock.h:1,098 0x60004692	
	GPIO_PinInit() at fsl_gpio.c:84 0x6000474a	
	board_init() at family.c:81 0x6000541e	
	main() at main.c:99 0x60005786	
```

Once the data cache is enabled, everything works swimmingly again.

### DIFF
--- a/hw/bsp/imxrt/family.c
+++ b/hw/bsp/imxrt/family.c
@@ -55,6 +55,12 @@ const uint8_t dcd_data[] = { 0x00 };
 
 void board_init(void)
 {
+  // make sure the dcache is on.
+#if defined(__DCACHE_PRESENT) && __DCACHE_PRESENT
+    if (SCB_CCR_DC_Msk != (SCB_CCR_DC_Msk & SCB->CCR))
+        SCB_EnableDCache();
+#endif
+
   // Init clock
   BOARD_BootClockRUN();
   SystemCoreClockUpdate();


hathach (approved):
great, thank you for the fix

PR #1931, opened 2023-02-28
For STM32H7 BSP boards, initialize RCC_...TypeDef structs as empty

branalba:
Following how ST's CubeMX/CubeIDE tools generate RCC init structs and general best practice for usage of a struct like this, instances of `RCC_ClkInitTypeDef`, `RCC_OscInitTypeDef`, and `RCC_PeriphClkInitTypeDef` should all be initialized as empty with `= {0};`. I've confirmed that failing to do this in the stm32h723nucleo example causes some undefined behavior when implementing a TUSB stack on my own STM32H723 board (I didn't use the example directly but I lifted the contents of the
`board_stm32h7_clock_init()` function and tested both with/without initializing the struct). 

From looking at the examples, the initialization may not be needed in all cases since the subsequent population of struct members is sufficient to eliminate any issues, but just to be consistent I applied the change to all occurrences of the aforementioned structs in the STM32H7 BSP examples.

This is my first PR on Github so apologies in advance if I've made any errors; I know this is a fairly minor set of changes but thought it would be good to include.

### DIFF
--- a/hw/bsp/stm32h7/boards/daisyseed/board.h
+++ b/hw/bsp/stm32h7/boards/daisyseed/board.h
@@ -57,9 +57,9 @@
 //--------------------------------------------------------------------+
 static inline void board_stm32h7_clock_init(void)
 {
-  RCC_ClkInitTypeDef RCC_ClkInitStruct;
-  RCC_OscInitTypeDef RCC_OscInitStruct;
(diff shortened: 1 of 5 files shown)


hathach (approved):
Look good, thank you

PR #1933, opened 2023-03-02
make rp2040 code compile cleanly with LLVM embedded toochain for ARM.

kilograham:
A couple of CMake changes related to warnings, and some tiny code changes


hathach (approved):
look great, thank you for your PR

PR #1946, opened 2023-03-10
dcd_nrf5x: ISO OUT handling

kasjer:
For incoming ISO OUT packets it was possible to start DMA from endpoint to RAM before transfer was started resulting in unrelated memory corruption.
This is scenario that causes memory corruption:
- ISO OUT packet is received
- Packet is transferred by DMA to transfer buffer
- xfer->started is cleared and xfer->buffer is updated as it is in every case
- Application takes to long to handle it (it happens when debugger is connected breakpoint is hit slowing down software).
- Next ISO OUT packet arrives At this point there was no check if transfer was started and packet was copied by DMA to location beyond previous data, possibly overwriting unrelated memory.

This solves the issue by checking that transfer was started and there is buffer ready for incoming packet.

**Describe the PR**

**Additional context**

### DIFF
--- a/src/portable/nordic/nrf5x/dcd_nrf5x.c
+++ b/src/portable/nordic/nrf5x/dcd_nrf5x.c
@@ -187,11 +187,16 @@ static void xact_out_dma(uint8_t epnum)
     }
     else
     {
-      // Trigger DMA move data from Endpoint -> SRAM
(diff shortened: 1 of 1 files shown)


hathach (approved):
superb! thank you for yet another race condition with nrf. I guess this never ends :)

PR #1953, opened 2023-03-13
Flush the dcd data on reset

tannewt:
On iMX RT TinyUSB didn't work with the USB buffers behind dcache. (This happens if the RAM region is marked shareable in the MPU.)

This adds a cache flush on reset so that the reset of the shared data structures is seen by the peripheral. Otherwise, it gets confused.


hathach (approved):
perfect, thank you. Note: doc build has some issues, I don't know why, but we could ignore it for now.

PR #1958, opened 2023-03-16
fix HAL_Delay no effect bug on st platform

tyustli:
**Describe the PR**
fix `HAL_Delay` no effect bug on st platform

**Additional context**

due overwrite `SysTick_Handler`, so when my hardware driver call `HAL_Delay` function ,it's can work!!!.  add `HAL_IncTick` in `SysTick_Handler`


hathach (approved):
look good, thank you

PR #1965, opened 2023-03-21
UAC2: fix feedback EP buffer alignment.

HiFiPhile:
**Describe the PR**
Which cause error in `get_buf_offset()` of lpc_ip3511.

### DIFF
--- a/src/class/audio/audio_device.c
+++ b/src/class/audio/audio_device.c
@@ -312,7 +312,7 @@ typedef struct
 
 #if CFG_TUD_AUDIO_ENABLE_FEEDBACK_EP
   struct {
-    uint32_t value;       // Feedback value for asynchronous mode (in 16.16 format).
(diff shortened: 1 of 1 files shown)


src/class/audio/audio_device.c
```diff
@@ -337,6 +336,7 @@ typedef struct
     }compute;
 
   } feedback;
+  CFG_TUSB_MEM_ALIGN uint32_t feedback_value; // Feedback value for asynchronous mode (in 16.16 format).
```
hathach, line 339: we can put the CFG_TUSB_MEM_ALIGN to the struct feedback, which will make the value as 1st member to be aligned correctly ?

HiFiPhile: Yes that will work. I was thinking to avoid aligment on struct member so I moved `value` outside, but it should be ok.

Issue #718, opened 2021-03-14
tud_network_mac_address should be initialized by hardware ID but is const


maxnet:
**Is your feature request related to a problem? Please describe.**

https://github.com/hathach/tinyusb/blob/master/examples/device/net_lwip_webserver/src/main.c#L61 mentions that ideally the MAC should be generated by hardware ID.
However the variable is declared const in https://github.com/hathach/tinyusb/blob/master/src/class/net/net_device.h#L62 and in  https://github.com/hathach/tinyusb/blob/master/lib/networking/rndis_reports.c#L39

**Describe the solution you'd like**

Remove "const" so that the MAC can be changed by program in an initialization routine at the start of program that queries the device's serial number and derives MAC from that.

hathach:
yeah, it even has an TODO remove later as well, apparently I plan to remove it. I will check it out later on to see what is best to change.

MattMills:
@hathach Any progress on this? Having an address that needs to be different on every board be const seems like a bad plan...

If there is currently no plans to move the TODO to TODONE, is there any chance the const can be removed?

hathach:
I haven't got time to reviewt net driver, please submit PR if you have suggestion.

hathach:
fixed now

Slion:
I'm pretty sure this should be reopened, see: https://github.com/hathach/tinyusb/pull/1971#issuecomment-2335141214

HiFiPhile:
> Proper fix should be to use a callback.

If you are up to something a PR is welcomed.

Slion:
> If you are up to something a PR is welcomed.

Currently working with an older version of TinyUSB but I'll try to come up with a PR at some point.

Slion:
Ok the fix works if you init the variable early enough. I updated my comment there: https://github.com/hathach/tinyusb/pull/1971#issuecomment-2335141214

PR #1971, opened 2023-03-22
Update net_device.h to make MAC addresses not const.

MattMills:
Per https://github.com/hathach/tinyusb/issues/718, MAC addresses are intended to be device unique, and thus should NOT be defined as consts.

I'm not sure what the longer term fix in the TODO should be, but this is a simple fix.


hathach (approved):
look good, thank you. extern with variable is not prefered, we should use an callback or something in the future.

Slion:
I need that too. Could have used a callback instead of just removing the `const` to make sure the variable as been initialized before it is being used by RNDIS. As it is it could be tricky to know by when you need to have it initialized.

PR #1977, opened 2023-03-23
Add missing dependency tusb.h for tusb_private.h

JayToltTech:
**Describe the PR**
When working with TinyUSB, my IDE (VSCode) sorted the #includes for tusb_private.h as part of automated code formatting which revealed an implicit dependency on tusb.h via build breaks when tusb_private.h is #included without/before tusb.h.

This PR simply adds the implicit dependency as an explicit dependency, fixing the build break if the #include statements are ordered differently (for example in tusb.c).

**Additional context**
VSCode auto-formatting changed tusb.c from:

```
#include "tusb.h"
#include "common/tusb_private.h"
```

to 

```
#include "common/tusb_private.h"
#include "tusb.h"
```

which revealed the implicit dependency via a build break.

### DIFF
--- a/src/common/tusb_private.h
+++ b/src/common/tusb_private.h
@@ -34,6 +34,8 @@
  extern "C" {
 #endif
 
+#include "tusb.h"
+
 typedef struct TU_ATTR_PACKED
 {
   volatile uint8_t busy    : 1;


hathach:
tusb_private.h is not meant to be included by application. Also include order is also relevant. even if it is public header, fixing code format by changing the code is not accepted.

JayToltTech:
Please take a look at the code change.  There's an implicit dependency in tusb_private that I simply made explicit.

Having build order issues in your .h files isn't preferable @hathach.  It will likely cause an issue for someone else in the future.

JayToltTech:
FYI, saw you opened a (very nice) checklist for static analysis.  FYI, your cppcheck run is going to fail with this issue unfixed.

Issue #1470, opened 2022-05-23
Vendor Device race condition


kasjer:
### Operating System

Windows 10

### Board

pca10095

### Firmware

custom firmware running mynewt OS

### What happened ?

Problem was detected on IN transfers over vendor interface on NRF5340 (but it's not MCU specific).
It may be that other driver has similar problem.
Here is sequence that fails.
- Vendor interface IN packet is sent to host.
- `vendord_xfer_cb()` is called
- it calls `maybe_transmit()`
- endpoint is not busy (yet) `TU_VERIFY( !usbd_edpt_busy(TUD_OPT_RHPORT, p_itf->ep_in) );` goes through
- there is more data in FIFO already; endpoint data is prepared
- next transfer is started: `TU_ASSERT( usbd_edpt_xfer(TUD_OPT_RHPORT, p_itf->ep_in, p_itf->epin_buf, count) );`
- assert fails `TU_ASSERT(_usbd_dev.ep_status[epnum][dir].busy == 0);`

In time between checking if endpoint is busy and setting busy state additional transfer was started from
other thread or interrupt.
From time to time  `TU_ASSERT(_usbd_dev.ep_status[epnum][dir].busy == 0);` goes through in both threads and both
threads try to start next transfer.

To workaround this problem I call `usbd_defer_func()` to run function that would start next transfer in USBD task context to eliminate race condition.
This way it works as it would in non-OS scenario where race is eliminated by single thread.
However **usbd_defer_func** is declared in header that indicates that it's privet and I'm not sure that I should use it application.

Any advice how to solve this properly?
Is starting of transfer from interrupt allowed?
Is scheduling function to run in USBD task from other thread/isr supported?

I did not check but it may be that problem is not limited to IN transfer of vendor interface only.

### How to reproduce ?

There is no easy way to reproduce as there is no public example of vendor interface usage.

hathach:
The issue seems to be the **stack's vendor driver**, currently it doesn't guard against race condition as cdc. I haven't updated the vendor driver once introducing the usbd_edpt_claim() for race protection. I actually have an plan to abstract these to stream/buffered endpoints API for driver and haven't got time to do so.

https://github.com/hathach/tinyusb/blob/master/src/class/vendor/vendor_device.c#L109
https://github.com/hathach/tinyusb/blob/master/src/class/cdc/cdc_device.c#L83

_Originally posted by @hathach in https://github.com/hathach/tinyusb/issues/900#issuecomment-870261839_

[...]

PR #1989, opened 2023-03-28
Vendor device race condition fix

epatstarkey:
**Use usbd_edpt_claim, usbd_edpt_release in vendor device class**

An issue similar to https://github.com/hathach/tinyusb/issues/1470 was discovered in my testing on a PCA10059 running MyNewt/Nimble. This PR ports over the use of the endpoint claim and release mechanisms used in the cdc class, which seems similar. This resolved the race condition I encountered that would result in IN transfers getting stuck until the next OUT transfer was started.

### DIFF
--- a/examples/device/webusb_serial/src/main.c
+++ b/examples/device/webusb_serial/src/main.c
@@ -112,7 +112,7 @@ void echo_all(uint8_t buf[], uint32_t count)
   if ( web_serial_connected )
   {
     tud_vendor_write(buf, count);
-    tud_vendor_flush();
+    tud_vendor_write_flush();
   }
 
   // echo to cdc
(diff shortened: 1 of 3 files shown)


epatstarkey:
Sample updated that I missed earlier. It should pass the gated build now.

epatstarkey:
Trailing whitespace removed that cause the pre-commit to fail. The rest passed.

HiFiPhile (approved):
Thanks for your PR, tested ok on my LPC55S66 custom board.

hathach (approved):
look good, thank you. There is quite of duplication code for driver with stream/buffered endpoint. I have added tu_edpt_stream (read/write) API. In the future we will migrate to those instead 
 https://github.com/hathach/tinyusb/blob/master/src/common/tusb_private.h#L124

currently only cdc host used this stream edpt https://github.com/hathach/tinyusb/blob/master/src/class/cdc/cdc_host.c#L61

PR #1993, opened 2023-03-29
Rp2040 freertos

rppicomidi:
**Describe the PR**
This draft pull request shows a possible way to implement FreeRTOS support for RP2040.
To use FreeRTOS with RP2040, you must install the FreeRTOS-Kernel source in a way that family.cmake can find it. Easiest place to put it is
```
    ${PICO_SDK_PATH}/../FreeRTOS-Kernel
```
Make sure to use the smp branch. That is:
```
    cd ${PICO_SDK_PATH}/../FreeRTOS-Kernel
    git checkout smp
```
In your project's CMakeFiles.txt, use libraries
```
    tinyusb_board_freertos, tinyusb_device_freertos,
    tinyusb_host_freertos
```
instead of the libraries  without the '_freertos' suffix

At some point, the FreeRTOS-Kernel smp branch will merge with the main branch. Hopefully, the large number of compiler warnings will be cleared by then.

**Additional context**
See discussion #1951.

Edit: I meant to make this a draft pull request. Anyone know how to change it?


hw/bsp/rp2040/family.cmake
```diff
@@ -23,133 +23,218 @@ if (NOT TARGET _rp2040_family_inclusion_marker)
...
+	unset(SUPPORTED_OS_LIST)
+	if (DEFINED TINYUSB_OPT_OS)
+		# for backward compatibility, let the user force the OS option
+		list(APPEND SUPPORTED_OS_LIST ${TINYUSB_OPT_OS})
+	else()
+		# otherwise, build both the pico-sdk and FreeRTOS versions of the library
+		list(APPEND SUPPORTED_OS_LIST OPT_OS_PICO OPT_OS_FREERTOS)
 	endif()
+	set(PROJ_SUFFIX "")
+	foreach(SUPPORTED_OS IN LISTS SUPPORTED_OS_LIST)
```
kilograham, line 35: why did you copy this here when you have FreeRTOS_Kernel_import.cmake too?

rppicomidi: Because FreeRTOS_Kernel_import.cmake has a fatal error message lines 47-49.
```
if (NOT FREERTOS_KERNEL_PATH)
    message(FATAL_ERROR "FreeRTOS location was not specified. Please set FREERTOS_KERNEL_PATH.")
endif()
```
As I understand it, we do not want pico-sdk to fail to build if FreeRTOS-Kernel is not installed.

rppicomidi:
@hathach The build pre-comit is failing. I do not understand why file `FreeRTOS_Kernel_import.cmake` requires a blank line at the end of the file. The file `pico_sdk_import.cmake` does not have one.

hathach:
> @hathach The build pre-comit is failing. I do not understand why file `FreeRTOS_Kernel_import.cmake` requires a blank line at the end of the file. The file `pico_sdk_import.cmake` does not have one.

You need to leave an empty line at the end of the file to make ci check happy. On my editor the `pico_sdk_import.cmake` does have one.

image

hathach:
I am making some chagnes to the cmake as well to better support both esp32 and rp2040 as well as others later on. https://github.com/hathach/tinyusb/tree/cmake . I am currently busy, but once I get that merged first, I will hep to sync up this branch

hathach:
still haven't decided which is best way to do this, but I may follow the approach using by FreeRTOS-kernel https://github.com/FreeRTOS/FreeRTOS-Kernel/blob/main/CMakeLists.txt which also requires an external file config I.e adding CMakelists.txt that define tinyusb as a lib but requires an target that expose tusb_config.h in the include path. Pico-sdk can define both no os and freertos config, and one that does not speciy the freertos config in the library link will use the no os one. Let me know what you think about this approach.

rppicomidi:
@hathach That works for a single, stand-alone project but I think it does not  work if you build the project from the pico-examples project? The [top level CMakeLists.txt](https://github.com/raspberrypi/pico-examples/blob/master/CMakeLists.txt) contains `include(pico_sdk_import.cmake)` which includes tinyusb build which prevents using the proper CMake variables later.

kilograham:
fyi, i have been crazy busy, and will be for a bit longer, so apologies for not getting to this yet.

hathach:
> fyi, i have been crazy busy, and will be for a bit longer, so apologies for not getting to this yet.

No problems, we are all busy, I am actually working to add cmake build to tinyusb repo as well (for src/ and all other mcu families). Also latest freertos-kernel is adding support for CMake as well https://github.com/FreeRTOS/FreeRTOS-Kernel/blob/main/CMakeLists.txt

hathach:
Thank you for making PR, after doing quite a bit of works to support cmake, I am getting better with it. I agree that adding tinyusb_device_freertos and tinyusb_host_freertos is best approach. Though I think we should do it in pico-sdk, I also think the freertos-kernel lib should be also added there rather than within tinyusb. That will help pico user by just simply link with freertos-kernel and or tinyusb_device_freertos. Check out my PR to pico-sdk here https://github.com/raspberrypi/pico-sdk/pull/1438 . @kilograham @rppicomidi would you mind review/comment and move our discussion to that.

hathach:
implemented by https://github.com/hathach/tinyusb/pull/3598

PR #1995, opened 2023-03-30
Request for comment : first go at wch fs/otg driver

mean00:
**Describe the PR**
This is a very early driver for the fs/otg usb IP found in wch chips (device mode only)
Tested on CH32V307 and partially on CH32V303.
That IP may be present in other WCH chips such as CH32V2xx etc...
This PR is a request for comment mostly (see below)

**Additional context**
The driver is derived from the HS one but targets the FS/OTG usb controller, with the same restrictions (resume/stall not supported, no cleanup...).
These IPs are similar but not compatible. I'm not sure it would be a good idea to try to factorize both drivers.
Please use the other USB connector on the ch32v307 eval board (the one far away from the debug port), it can power the board also.

I could  use some pointers on the following points:
- i did not put the code in ch32v307 as it is a generic ip from wch, instead in the wch/wch_fs_otg folder
- the code may look a bit complicated, but it's to make it easier to debug for now
- the PR branch is hijacking the ch32v307 board and redirects it to fs instead of hs (the hs init code is still there, clock etc..). I have no idea how to make that in a clean way. Help ?
- i'm mainly doing CDC with the freertos osal and i'm not using the WCH SDK, it seems to work fine for me.
I used tinyUsb demo code (cdc+msc) with  osal_none,  it sort of works but it's not very robust. 
That's why some of the platform dependant code is in dcd_fs_platform.c, it makes it easier to move code.

Any pointers would be appreciated
(and thanks for tinyusb, it's a great project)

### DIFF
--- a/hw/bsp/ch32v307/family.c
+++ b/hw/bsp/ch32v307/family.c
@@ -31,11 +31,16 @@
 #include "bsp/board.h"
 #include "board.h"
 
+#define SYSCLK_FREQ_48MHz   48000000
+#define SYSCLK_FREQ_96MHz   96000000
+#define SYSCLK_FREQ_144MHz 144000000
+
 //--------------------------------------------------------------------+
 // Forward USB interrupt events to TinyUSB IRQ Handler
 //--------------------------------------------------------------------+
-
-void USBHS_IRQHandler (void) __attribute__((naked));
+// using interrupt is maybe inefficient but it does not depend
+// on the compiler used
+void USBHS_IRQHandler (void) __attribute__((interrupt));
 void USBHS_IRQHandler (void)
 {
   __asm volatile ("call USBHS_IRQHandler_impl; mret");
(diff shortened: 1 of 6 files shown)


hathach:
superb !! thank you, give me a bit of time, I will try to review this as soon as I could.

nikitos1550:
Hello! I am also trying bring up usbfs (and usbd) IP on ch32.
Why stall is not supported on USBFS? In the datasheet I can see:
```
R8_UEPn_TX_CTRL
MASK_UEP_T_RES -> 11: Response STALL or error
``` 
Could you explain me a bit?

mean00:
@nikitos1550  : I started from the HS driver and it was not present there at that time.
I've been using the FS port for a while now and it works ok if not perfect.
(sorry for the late reply)

hathach:
superseded by #2362

PR #1999, opened 2023-04-03
openh743i: initialize USB3300 reset pin

raiden00pl:
The USB3300 reset pin for Waveshare openh743i board was never initialized, so resetting the chip doesn't work correctly. This PR initialize the pin as OUTPUT before any use.


hathach (approved):
look good, thank you

PR #2000, opened 2023-04-03
Add STM32L5 support - no OTG similar to some L4s

jncronin:
**Describe the PR**
Adds support for STM32L5 MCUs

**Additional context**
Tested on STM32L552RCT on custom board in crystalless mode using clock recovery system (CRS).  Only additional setup required is APB clock to USB, CRS setup and enabling VDDUSB via PWR->CR2.

USB_PMAADDR is not defined in stm cmsis headers but rather USB_PMAADDR_S and USB_PMAADDR_NS (secure and non-secure modes respectively).  The definition here is relative to USB_BASE which defaults to USB_BASE_NS but the current implementation allows the user to redefine this to USB_BASE_S and the USB_PMAADDR define will be adjusted appropriately.

### DIFF
--- a/src/common/tusb_mcu.h
+++ b/src/common/tusb_mcu.h
@@ -229,6 +229,12 @@
   #define TUP_USBIP_DWC2_STM32
   #define TUP_DCD_ENDPOINT_MAX    6
 
+#elif TU_CHECK_MCU(OPT_MCU_STM32L5)
+  #define TUP_USBIP_FSDEV
+  #define TUP_USBIP_FSDEV_STM32
+  #define TUP_DCD_ENDPOINT_MAX    8
+
+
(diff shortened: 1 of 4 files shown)


hathach:
which board you are testing with, would you mind adding an bsp as well if possible ?

jncronin:
I have tested on a custom PCB and a generic HID in/out test works as expected.  Unfortunately I do not have a development board (Nucleo/Disco etc) to test on.

The STML412Nucleo bsp seems to be the most similar with only the GPIO assignments for LEDs/button/UART differing.  I can edit this and attach if requested however it would be untested.

hathach (approved):
thank you for your PR. Look good, I could add an bsp for devboard later on.

Issue #1997, opened 2023-04-03
Data Seq Error when connecting HID device through USB hub


Xanvarok:
### Operating System

Linux

### Board

Raspberry Pi Pico

### Firmware

/examples/host/cdc_msc_hid

### What happened ?

When connecting a USB keyboard (or mouse) to the Pico via a USB hub, the process fails when trying to setup the keyboard (or mouse) throwing a Data Seq Error.

### How to reproduce ?

1. Apply power to Pico
2. Connect USB hub to Pico
3. Connect a USB keyboard (or mouse) to the USB hub
Or
1. Connect USB hub to Pico
2. Connect a USB keyboard (or mouse) to the USB hub
3. Apply power to Pico

### Debug Log as txt file

[TinyUSB_Data_Seq_Error.txt](https://github.com/hathach/tinyusb/files/11133843/TinyUSB_Data_Seq_Error.txt)

### I have checked existing issues, dicussion and documentation

- [X] I confirm I have checked existing issues, dicussion and documentation.

hathach:
do you have other USB hub to test with, just to make sure it is not as generic issue ? I tested previously fine with my hub.

Xanvarok:
Unfortunately I do not currently. 

The hub does work with both a raspberry pi 4 and with my desktop; however, it does look to be based around a HS8836 which other folks online seem to have issues with in other situations.

I'll see if I can get my hands on another hub to test with.

hathach:
when you are able to test with other hub and can narrow the issue to the specific chipset, please detail your hub make and model as well as link, picture etc.. If it is the case, I will try to see if I could find it locally here for testing.

Xanvarok:
So I managed to get my hands on a Sabrent HB-MCRM and everything seems to be working fine with that  ([log of working one here](https://github.com/hathach/tinyusb/files/11205540/TinyUSB_No_Error_With_Sabrent.txt)).

The Sabrent one isn't mine (so I didn't tear it apart) but it seems to claim it's using a Genesys Logic controller... probably a GL850G.

The one that doesn't work is a no-name brand... looks like the controller is actually a Huasheng Electronics HS8836A which seems to be found in a variety of cheap Chinese no-name USB hubs. There are no identifying marks on the case and the PCB is very basic... only silk-screened with "HS8836A-2022-12-02". I'll see if I can track down where it came from if you're interested.

morio:
[...]

PR #2016, opened 2023-04-09
Even when we get an empty "status change" interrupt from the hub, schedule another interrupt poll

ipopov:
**Describe the PR**
Even when we get an empty "status change" interrupt from a hub, schedule another interrupt poll.

During enumeration, when there are multiple devices attached to the hub as it's plugged into the Pi Pico, enumeration hangs, because we get a "status change" callback with value zero. With this patch, we retry several times on "zero" status change callbacks, until eventually we succeed.

**Additional context**
This is the cheapo hub that exhibits this behavior, but I assume it's not the only one: https://www.amazon.com/gp/product/B083RQMC7S.

While debugging this, I consulted the implementation in the Linux kernel. There, hub setup explicitly checks each port individually, before starting to depend on "status change" interrupts: https://elixir.bootlin.com/linux/latest/source/drivers/usb/core/hub.c#L1133. We probably should do something like that here, but it's a much bigger change.

### DIFF
--- a/src/host/hub.c
+++ b/src/host/hub.c
@@ -361,6 +361,13 @@ bool hub_xfer_cb(uint8_t dev_addr, uint8_t ep_addr, xfer_result_t result, uint32
         break;
       }
     }
+
+    // The status change event was neither for the hub, nor for any of
+    // its ports. (For example `p_hub->status_change == 0`.) This
+    // shouldn't happen, but it does with some devices. Initiate the
+    // next interrupt poll here, because we've scheduled no other work
+    // whose completion can initiate it.
+    hub_edpt_status_xfer(dev_addr);
   }
 
   // NOTE: next status transfer is queued by usbh.c after handling this request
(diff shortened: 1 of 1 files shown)


morio:
With `#define CFG_TUSB_DEBUG 3` this breaks connecting to an USB keyboard, it is an old Apple M7803. I attached the log with ipopov's pull request code and the previous commit in his repo, hash: `28817a715024c82b78f9b9d19f37a72771870759` 

The error with ipopov's code comes up as `Queue EP 81 with 1 bytes ... usbh_edpt_xfer_with_callback 781: ASSERT FAILED`

ipopov's code works find with `#define CFG_TUSB_DEBUG 0`

I was hoping this pull request would fix an issue with an unbranded two port hub that gets a `Data seq error` whenever I plug a Logitech B100 mouse into it. If I plug the mouse into another hub downstream, like the one on the Apple M7803 keyboard,  the mouse works.

I think this issue has been reproduced here: https://github.com/hathach/tinyusb/issues/1997#issuecomment-1507818958

[prev-commit.log](https://github.com/hathach/tinyusb/files/11237082/prev-commit.log)
[ipopov-commit.log](https://github.com/hathach/tinyusb/files/11237083/ipopov-commit.log)

aperezbios:
@hathach ping, is this by chance on your radar as something in need of improvement? There has been no further discussion since @morio wrote about this a couple of weeks ago.

hathach:
will review this soon, just come back from vacation and there is quite a bit of work I have to do first.

hathach (approved):
thank you for your PR, I slightly move the check for status_change not zero to make it clearer. Will merge when ci passed.

Issue #1830, opened 2022-12-31
tuh_hid_set_report fails after multple calls to change a keyboards LEDs. Number of calls is non determinsitic


morio:
### Operating System

Others

### Board

Custom RP2040

### Firmware

https://github.com/rabbitholecomputing/RP2040-TinyUsb-keyboard-leds

### What happened ?

After multiple calls toggling a USB keyboard LEDs `tuh_hid_set_report` starts returning false.

Tracing my original code (https://github.com/rabbitholecomputing/QuokkADB-firmware) the issue is [_ctrl_xfer.stage == CONTROL_STAGE_IDLE);](https://github.com/hathach/tinyusb/blob/7004914d8c6c7d88908776eb153661b6b987e513/src/host/usbh.c#L518)
fails because _ctrl_xfer.stage is equal to CONTROL_STAGE_ACK.

 The test code doesn't actually exercise that line because it uses `tuh_hid_set_report_complete_cb` to initiate another   `tuh_hid_set_report`.  And since tuh_hid_set_report never finishes it can't be called again in the test code.

### How to reproduce ?

1. Upload firmware to RP2040 TinyUSB host connected to a USB keyboard
2. Hook up board to a pico probe
3. Open pico probe COM port in a terminal like Tera Term
4. Power board
5. Press "Caps Lock"
6. Watch the output fly by in a terminal window
7. When the LEDs stop blinking tuh_hid_set_report has failed.
8. The terminal should output "TEST: Set report completed, Toggled leds X times so far." Where X is the number of successful LED toggles.

### Debug Log as txt file

The last three toggles of the LEDs using `tuh_hid_set_report`. The very last one hangs and never completes.
Full in the .log file linked at the bottom

```
HID Set Report: id = 0, type = 2, len = 1

[0:1] Class Request: 21 09 00 02 00 00 01 00

hw_endpoint_init dev 1 ep 0 out xfer 0

dev 1 ep 0 out setup buffer @ 0x0x50100180

endpoint control (0x0x50100100) <- 0xa0000180

Transfer complete

Sent setup packet

on EP 00 with 8 bytes

hcd_edpt_xfer dev_addr 1, ep_addr 0x0, len 1

  Prepare BufCtrl: [0] = 0xf401  [1] = 0x0000

Buffer complete

buf_status 0x00000001

Single Buffered: ep_ctrl = A0000180

  Sync BufCtrl: [0] = 0x6001  [1] = 0x0000

[...]

PR #2024, opened 2023-04-13
[rp2040] Make writes to SIE_CTRL aware of concurrent access

jfedor2:
This fixes https://github.com/hathach/tinyusb/issues/1830

Some context: https://github.com/raspberrypi/pico-feedback/issues/324

This commit makes it so that when setting the START_TRANS bit in the SIE_CTRL register, along with some other bits, we first set all the other bits, then wait some cycles, and then set the START_TRANS bit.

Doing so protects against a situation where the USB controller is reading the register at the same time and gets an incorrect value.

The assembler parts I just copied from _hw_endpoint_buffer_control_update32 where a similar procedure is already followed for the buffer control registers.


liamfraser:
I approve this change so please merge if you are happy @hathach

src/portable/raspberrypi/rp2040/hcd_rp2040.c
```diff
@@ -562,6 +562,9 @@ bool hcd_edpt_xfer(uint8_t rhport, uint8_t dev_addr, uint8_t ep_addr, uint8_t *
     uint32_t flags = USB_SIE_CTRL_START_TRANS_BITS | SIE_CTRL_BASE |
                      (ep_dir ? USB_SIE_CTRL_RECEIVE_DATA_BITS : USB_SIE_CTRL_SEND_DATA_BITS) |
                      (need_pre(dev_addr) ? USB_SIE_CTRL_PREAMBLE_EN_BITS : 0);
+    // write everything except the START_TRANS bit first, then wait
+    usb_hw->sie_ctrl = flags & ~USB_SIE_CTRL_START_TRANS_BITS;
+    __asm volatile ("b 1f\n1: b 1f\n1: b 1f\n1: b 1f\n1: b 1f\n1: b 1f\n1:\n" : : : "memory");
```
hathach, line 567: is this 6 nop instruction, could we just write it as folows. 

```C
__asm volatile("nop; nop; nop; nop; nop; nop;" : : : "memory")
```

Also It would be nice to add comment why we need this e.g `START_TRANS "seems" to have the same behavior as AVAILABLE bit described in section 4.1.2.5.1 Concurrent access rp2040 datasheet v2.1" etc...

hathach (commented):
Thank you very much for the PR, I have an question whether we could rewrite the asm to do nop instead since that what I would write when wanting to kill a few cycles.

src/portable/raspberrypi/rp2040/hcd_rp2040.c
jfedor2: I don't have an opinion on this, I just followed what's already in _hw_endpoint_buffer_control_update32 in rp2040_usb.c. That was written by @kilograham I believe. I don't know if it makes any difference. If you want NOPs tell me if you want 6 or 12 of them as existing code delays for 12 cycles and a single NOP is 1 cycle (I think). This should only make a difference on massively overclocked chips as 3 cycles should be enough at 133MHz.

kilograham: `nop` is 1 cycle, `b 1f` is 2 cyles so they aren't the same (and the `b 1f` is smaller`)

as of SDK1.3.1 there is a `busy_wait_at_least_cycles(cycles)` method which could be used instead (i.e. `busy_wait_at_least_cycles(12)`

hathach: thanks @kilograham for explanation, @jfedor2 what do you think, would you mind chaing it to `busy_wait_at_least_cycles(12)` as suggested.

jfedor2: Of course!

hathach (approved):
perfect, thank you

PR #2030, opened 2023-04-14
Update porting.rst

slc-tl:
Adds more in depth documentation of how tud_task does not need to be polled in implementations that use an RTOS / scheduler.

**Describe the PR**
This adds documentaiton showing how tud_task doesn't need to hold the CPU through polling when tusb_config.h is set to use an RTOS. 

**Additional context**


hathach (commented):
look good, thank you. Though can you move the note for RTOS to OSAL porting guide per review

hathach (approved):
thank you for your PR, I moved the note to OSAL

PR #2033, opened 2023-04-17
Fix compatibility with the latest Microchip XC16 compiler

ReimuNotMoe:
**Describe the PR**

The Microchip XC16 compiler is just a rebranded GCC 4.5.1 with some modifications, so, it defines the `__GNUC__` macro as well. However, GCC 4.5.1 is so ancient that it doesn't support the `__has_attribute` function. This resulted in a compilation error. Fortunately there's way to detect it via the `__XC16` macro. This pull request solves the problem.

### DIFF
--- a/src/common/tusb_compiler.h
+++ b/src/common/tusb_compiler.h
@@ -138,10 +138,14 @@
   #define TU_ATTR_BIT_FIELD_ORDER_BEGIN
   #define TU_ATTR_BIT_FIELD_ORDER_END
 
-  #if __has_attribute(__fallthrough__)
-    #define TU_ATTR_FALLTHROUGH         __attribute__((fallthrough))
-  #else
+  #if defined(__XC16)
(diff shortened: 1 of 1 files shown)


src/common/tusb_compiler.h
```diff
@@ -138,10 +138,14 @@
   #define TU_ATTR_BIT_FIELD_ORDER_BEGIN
   #define TU_ATTR_BIT_FIELD_ORDER_END
 
-  #if __has_attribute(__fallthrough__)
-    #define TU_ATTR_FALLTHROUGH         __attribute__((fallthrough))
-  #else
+  #if defined(__XC16)
```
hathach, line 141: __has_attribute() is introduced since gcc version 5. we could replace XC16 with more generic `__GNUC__ < 5`

hathach (commented):
look great, thank you. Though could you update the PR to use `__GNUC__ < 5` with has_attribute()

hathach (approved):
perfect thank you

PR #2042, opened 2023-04-25
optimization for HID class

jiadingxu:
Hello hathach:
     Here're some optimizatoins for the HID class and a sample configuraiton.

### DIFF
--- a/examples/device/hid_generic_inout/src/tusb_config.h
+++ b/examples/device/hid_generic_inout/src/tusb_config.h
@@ -97,6 +97,26 @@
 #define CFG_TUD_MIDI              0
 #define CFG_TUD_VENDOR            0
 
+//------------- REPORT ID  -------------//
+#define REPORT_ID_COUNT           2   
+	
+#if REPORT_ID_COUNT > 2
+	#error This value must not be greater than 2 in this demo!
+#endif 
+
+#if REPORT_ID_COUNT == 2
+    #define REPORT_ID_OUT           1
+    #define REPORT_ID_IN            2
+#elif REPORT_ID_COUNT == 1
+  #define REPORT_ID               1
+  #define REPORT_ID_OUT           REPORT_ID
+  #define REPORT_ID_IN            REPORT_ID
+#endif
+
+#ifndef CFG_TUSB_REPORT_ID_COUNT
+    #define CFG_TUSB_REPORT_ID_COUNT  REPORT_ID_COUNT
+#endif
+
 // HID buffer size Should be sufficient to hold ID (if any) + Data
 #define CFG_TUD_HID_EP_BUFSIZE    64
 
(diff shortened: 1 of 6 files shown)


src/class/hid/hid_device.h
```diff
@@ -380,7 +380,30 @@ static inline bool  tud_hid_gamepad_report(uint8_t report_id, int8_t x, int8_t y
 // HID Generic Input & Output
 // - 1st parameter is report size (mandatory)
 // - 2nd parameter is report id HID_REPORT_ID(n) (optional)
-#define TUD_HID_REPORT_DESC_GENERIC_INOUT(report_size, ...) \
+#if (CFG_TUSB_REPORT_ID_COUNT == 2)
+  #define TUD_HID_REPORT_DESC_GENERIC_INOUT(report_size, ...) \
```
hathach, line 384: please explain what is this template used for ?

src/device/dcd.h
```diff
@@ -82,6 +83,11 @@ typedef struct TU_ATTR_ALIGNED(4)
       uint32_t frame_count;
     }sof;
 
+    // Plugged
+    struct {
```
hathach, line 87: please explain why adding this change

examples/device/hid_generic_inout/src/tusb_config.h
```diff
@@ -97,6 +97,26 @@
...
+#if REPORT_ID_COUNT == 2
+    #define REPORT_ID_OUT           1
+    #define REPORT_ID_IN            2
+#elif REPORT_ID_COUNT == 1
+  #define REPORT_ID               1
+  #define REPORT_ID_OUT           REPORT_ID
+  #define REPORT_ID_IN            REPORT_ID
+#endif
+
+#ifndef CFG_TUSB_REPORT_ID_COUNT
```
hathach, line 116: report ID count is application-specific and should not be part of configuration parameter.

hathach (commented):
thank you for your PR, please check out review for my questions on the changes.

src/class/hid/hid_device.h
jiadingxu: I found that the macro didn't support two different Report IDs(in / out), so I tried to separate them.

jiadingxu: For example: If I need a report id(1) for out and another one(2) for in,  I won't get the data throuth this original macro.

jiadingxu: During designing a usb generic HID communication, I found that the macro didn't support two different Report IDs(in / out), so I tried to separate them.
For example: 
If I need a report id(Report ID: 1) for out reports and another one(Rreport ID: 2) for in reports, I won't get the data through this original macro.

hathach: if template descriptor does not fit your need, you don't need to use it, just make your own descritpor. These template only cover the most common usage by user. Therefore this should be reverted !

HiFiPhile:
Close deperated PRs.

PR #2047, opened 2023-04-28
cdc: fix line_coding aligment

jbtheou:
While calling tud_cdc_n_get_line_coding, the structure is copied into the destination.

Dump of assembler code for function tud_cdc_n_get_line_coding:
   0x000193f4 <+0>:	mov.w	r2, #2112	@ 0x840
0x000193f8 <+4>:	ldr	r3, [pc, #20]	@ (0x19410
<tud_cdc_n_get_line_coding+28>)
   0x000193fa <+6>:	mla	r0, r2, r0, r3
=> 0x000193fe <+10>:	ldr.w	r3, [r0, #6]
   0x00019402 <+14>:	str	r3, [r1, #0]

On some platform (tested on LPC55S28), the address needs to be 4-bytes aligned. Without this, the address is

(gdb) p &_cdcd_itf.line_coding
$3 = (cdc_line_coding_t *) 0x40100006 <_cdcd_itf+6>

which leads to a HardFault. With this fix

(gdb) p &_cdcd_itf.line_coding
$5 = (cdc_line_coding_t *) 0x40100008 <_cdcd_itf+8>

and the function can be called properly

### DIFF
--- a/src/class/cdc/cdc_device.c
+++ b/src/class/cdc/cdc_device.c
@@ -53,7 +53,7 @@ typedef struct
 
   /*------------- From this point, data is not cleared by bus reset -------------*/
   char    wanted_char;
-  cdc_line_coding_t line_coding;
+  CFG_TUSB_MEM_ALIGN cdc_line_coding_t line_coding;
 
   // FIFO
   tu_fifo_t rx_ff;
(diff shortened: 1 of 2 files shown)


hathach (approved):
Thank you, I push an update to add aligned 4 for host driver as well. Actually, this isn't issue since M4+ can access non-aligned data. However iirc lpc55 eventhough is m33 but its highspeed usb can only access 4-bytes aligned memory region which cause the issue. (fs usb probably work just fine).

jbtheou:
Thanks a lot for extending the fix.
Indeed, used it on SAMD21/51 in the past without issue, so I was a bit confused initially about the issue.

Thank you for the quick review and approval!

PR #2050, opened 2023-04-28
Dwc2 interrupts

### LINKED ISSUE #1737: DWC2: Invalid interrupt status is checked
### Operating System

MacOS

### Board

STM32 with OTG device (Synopsys DWC)

### Firmware

--

### What happened ?

The dwc2 code checks the wrong status register when processing the Rx FIFO:

```
    // Loop until all available packets were handled
    do
    {
      handle_rxflvl_irq(rhport);
    } while(dwc2->gotgint & GINTSTS_RXFLVL);
```

See https://github.com/hathach/tinyusb/blob/master/src/portable/synopsys/dwc2/dcd_dwc2.c#L1316

The correct line should be:

```
    } while(dwc2->gintsts & GINTSTS_RXFLVL);
```

The code still works as the loop is executed at least once and as the interrupt will be immediately retriggered if the Rx FIFO is not empty. But the intention of efficient processing is not achieved.

### How to reproduce ?

--

### Debug Log as txt file

--

### Screenshots

_No response_

### LINKED ISSUE #2049: DWC2: TX fifo's can hold only a single packets while interrupts fire when the fifo is half-empty
### Operating System

Windows 11

### Board

Custom board with STM32H753

### Firmware

Custom firmware with a combination of uac-headset and 2 cdc serial ports

### What happened ?

The symptom is that the interrupt handler sometimes ends up in a neverending loop. This occurs because the TX-fifo-empty interrupts fires. This happens by default when the fifo is half-empty. The interrupt-handler only wants to write a full packet. If it doesn't fir, nothing happens and the interrupt fires again.
But the fifo size is such that only one packet fits.  So when the fifo is half empty, it can't send the packet and the interrupt loops. Until the fifo is fully empty.

Solution is to either make the fifo's twice as large, or configure to fire the interrupt when the fifo is empty.
I'm not 100% sure what the best solution is.

### How to reproduce ?

use UAC headset
Repeatedly play and record audio for a second or so, and then stop again. Loop forever. (can provide python script for that)
Wait until the interrupt hangs forever

No log available

### Screenshots

_No response_

electretmike:
Double the size of TX fifo so that a new packet can be pushed when the fifo is half-empty

Fixes #2049
Fixes #1737

### DIFF
--- a/src/portable/synopsys/dwc2/dcd_dwc2.c
+++ b/src/portable/synopsys/dwc2/dcd_dwc2.c
@@ -534,6 +534,9 @@ void dcd_init (uint8_t rhport)
   int_mask = dwc2->gotgint;
   dwc2->gotgint |= int_mask;
 
+  // Configure TX FIFO to set the TX FIFO empty interrupt when half-empty
+  dwc2->gahbcfg &= ~GAHBCFG_TXFELVL;
+
   // Required as part of core initialization.
   // TODO: How should mode mismatch be handled? It will cause
   // the core to stop working/require reset.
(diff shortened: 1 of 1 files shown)


electretmike:
@hathach  How can we get some traction on this?

hathach (approved):
Superb! the while loop is my terrible typo. Thank you for the fix and sorry for late response. I was buried in pile of works. Though since your PR does not allow maintainer to push to resovle the conflict. To reduce further delay, I have create https://github.com/hathach/tinyusb/pull/2584 based on this commit log, with history/author reserved, once that is merged, this will be marked as merged as well.

PR #2061, opened 2023-05-12
iMX.RT EHCI add dcache support and other fixes + refactor

### LINKED ISSUE #1511: un-roll recursive call when unplug an hub device
**Is your feature request related to a problem? Please describe.**
PR #1480 add support for multiple-level hub using recursive call of process_device_unplugged()

However in limited MCU, recursive call chain can cause stack overflow. We should find an way to un-roll the recursive chain. After all, the device loop is rather small (4-10) only.

**Describe the solution you'd like**
It is always possible to un-roll an recursive chain to a loop with/without an queue. We could use a status (disconnected/connected) and/or usbh event queue to achieve this.

https://github.com/hathach/tinyusb/blob/d7b579a9785a7be62174a912ddb4e34aea8a1156/src/host/usbh.c#L1110-L1152

hathach:
**Describe the PR**
- EHCI: fix xfer failed with disconnected device as stalled
- change CFG_TUH_ENDPOINT_MAX to 16 (max endpoint pair per device) if
not defined
- change QHD_MAX for EHCI, should be user configurable and more
optimized in the future
- fix #1511 un-roll recursive hub removal with usbh queue. (not perfect but good enough)
- fix error on EHCI causes xfer error in non-queued qhd which cause memory fault
- Add support for dcache clean/invalidate when memory is in cachable memory. Rule of thumb is:
  - clean buffer **before** queuing OUT transfer
  - invalidate buffer **before and after** queuing IN transfer 
  - clean and invalidate queue head and TD before transfer and invalidate after transfer
- fix issue with removing queue head (similar to one in OCHI)
- Lots of EHCI refactor: simplify driver to only support 1 TD per QHD (since that is what usbh will schedule). 

** other minor usbd tweak **
- usbd: fix control transfer issue for chipidea hs when previous status and new setup complete in the same isr frame
- change usbd edpt busy/stalled/claimed value to 0/1 instead of
(true/false) since they are 1-bit field.


src/portable/chipidea/ci_hs/hcd_ci_hs.c
```diff
@@ -50,6 +50,10 @@
     SCB_InvalidateDCache_by_Addr(addr, (int32_t) data_size);
   }
 
+void hcd_dcache_clean_invalidate(void* addr, uint32_t data_size) {
```
HiFiPhile, line 53: How about make the cache related function optional (default ON) ?

For IMX.RT in order to have best performance, RAM will be configured to ITCM & DTCM, so they are accessed at CPU clock instead of bus clock (600MHz vs 133MHz)

Since TCM bypass the cache there is no need to flush or invalidate.

image

hathach: ah you are right, I think we can check the address to see if it is within the DTCM (non-cachable) and skip the clean/invalidate will update in the next push or so. Circuitpython is currently placing ehci_data and/or user memory randomly and may not be in DTCM.

hathach: DTCM is 1MB reserved in 2000000-2010000,this should be good enough.

image

PR #2077, opened 2023-05-24
fix(RA Host Portable): git rid of the IAR warnings for packed

abakosh:
This is a daughter PR from PR#2063

I am using the -e flag, for IAR workbench users it is under :
Options/ C_C++ compiler/ Language 1 / Language conformance / Standard with IAR extensions

TU_ATTR_PACKED is not solving the compiler warnings, I thought it should be enough. However, I am using IAR compiler 9.1, which seems insufficient for these structs. When I use the __packed I am getting rid of the warnings

Here is an example given by IAR, which somehow makes it seems that only __packed is supported.
https://www.iar.com/contentassets/e20cfa542d1a4734aa6b234bea42f11a/example-5.pdf
Another link that led me to use __packed on these structs:
https://www.iar.com/knowledge/support/technical-notes/compiler/accessing-unaligned-data/

### DIFF
--- a/src/portable/renesas/rusb2/rusb2_type.h
+++ b/src/portable/renesas/rusb2/rusb2_type.h
@@ -1599,6 +1599,7 @@ TU_ATTR_BIT_FIELD_ORDER_END
 //--------------------------------------------------------------------+
 
 TU_VERIFY_STATIC(sizeof(RUSB2_PIPE_TR_t) == 4, "incorrect size");
+TU_VERIFY_STATIC(sizeof(RUSB2_REG_t) == 1032, "incorrect size");
 
 TU_VERIFY_STATIC(offsetof(RUSB2_REG_t, SYSCFG     ) == 0x00000000, "incorrect offset");
 TU_VERIFY_STATIC(offsetof(RUSB2_REG_t, BUSWAIT    ) == 0x00000002, "incorrect offset");
--- a/src/portable/renesas/rusb2/rusb2_type.h
+++ b/src/portable/renesas/rusb2/rusb2_type.h
@@ -41,11 +41,11 @@ extern "C" {
 TU_ATTR_PACKED_BEGIN
 TU_ATTR_BIT_FIELD_ORDER_BEGIN
 
-typedef struct TU_ATTR_PACKED {
+typedef struct {
   union {
     volatile uint16_t E; /* (@ 0x00000000) Pipe Transaction Counter Enable Register */
 
-    struct {
+    struct TU_ATTR_PACKED {
       uint16_t : 8;
       volatile uint16_t TRCLR : 1; /* [8..8] Transaction Counter Clear */
       volatile uint16_t TRENB : 1; /* [9..9] Transaction Counter Enable */
@@ -56,18 +56,18 @@ typedef struct TU_ATTR_PACKED {
   union {
     volatile uint16_t N; /* (@ 0x00000002) Pipe Transaction Counter Register */
 
-    struct {
+    struct TU_ATTR_PACKED {
       volatile uint16_t TRNCNT : 16; /* [15..0] Transaction Counter */
     } N_b;
   };
 } RUSB2_PIPE_TR_t; /* Size = 4 (0x4) */
 
 /* LINK_REG Structure */
-typedef struct TU_ATTR_PACKED {
+typedef struct {
   union {
     volatile uint16_t SYSCFG; /* (@ 0x00000000) System Configuration Control Register */
 
-    struct {
+    struct TU_ATTR_PACKED {
       volatile uint16_t USBE  : 1; /* [0..0] USB Operation Enable */
       uint16_t                : 2;
       volatile uint16_t DMRPU : 1; /* [3..3] D- Line Resistor Control */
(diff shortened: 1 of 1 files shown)


HiFiPhile:
The root problem is not packed support, you can read more in #431 
image

In this case since USB registers are 16bit aligned, there is no need to put `TU_ATTR_PACKED` at top level (for example stm32_fsdev also has 16bit registers), put `TU_ATTR_PACKED` in the bitfield struct level will make the purpose more clear.

HiFiPhile:
I'm agreed IAR has some problem handling packed struct, behavior of `__attribute__ ((packed))` and `__packed` should be the same, either:

- Emit warning for all packed struct
- Smart enough to know that address is in fact aligned

Could you try to change packed keyword into bitfields level, such as:
```
typedef struct {
  union {
    volatile uint16_t E; /* (@ 0x00000000) Pipe Transaction Counter Enable Register */

    struct TU_ATTR_PACKED {
      uint16_t : 8;
      volatile uint16_t TRCLR : 1; /* [8..8] Transaction Counter Clear */
      volatile uint16_t TRENB : 1; /* [9..9] Transaction Counter Enable */
      uint16_t : 6;
    } E_b;
  };

  union {
    volatile uint16_t N; /* (@ 0x00000002) Pipe Transaction Counter Register */

    struct TU_ATTR_PACKED {
      volatile uint16_t TRNCNT : 16; /* [15..0] Transaction Counter */
    } N_b;
  };
} RUSB2_PIPE_TR_t; /* Size = 4 (0x4) */
```

abakosh:
@HiFiPhile Could you try to change packed keyword into bitfields level?

I did, the warning is still there, but when I use __packed it is suppressing the warnings for sure. I have tested it. The commit in this PR is getting rid of the warnings for sure. The only I am thinking of, if there is a neater way to put it in

abakosh:
behavior of __attribute__ ((packed)) and __packed should be the same. 

I agree with you on this, but the compiler doesn't see it this way from my side. The compiler is not doing anything with this __attribute__ ((packed)), maybe in the newer IAR compiler it is not supported or ignored, I don't know

HiFiPhile:
> I did, the warning is still there

In my test there is no more warning moving TU_ATTR_PACKED to the bitfield struct level.
image

But since `RUSB2_REG_t` heavily use bitfield, I feel like adding TU_ATTR_PACKED everywhere is not elegant.

@hathach what's your opinion?

hathach:
> In my test there is no more warning moving TU_ATTR_PACKED to the bitfield struct level. image

Yeah, we should move the packed to bitfield if that fixes the issue. And also having TU_VERIFY_STATIC() to make sure the size of the whole struct is correct as well.

> But since `RUSB2_REG_t` heavily use bitfield, I feel like adding TU_ATTR_PACKED everywhere is not elegant.

I don't really look closed enough at the driver, but declaring it with bit field is not a problem at all. Though if we access different field in consecutive statements, maybe it is better to use bitwise to save |=, &=, other than that a single bitfield access would generate the same bitwise op as we could do manually.

abakosh:
Okay, I think I figured it out.  It is a little bit strange but __packed on the higher struct level suppresses the warnings, but __attribute__ ((packed)) only works on the bit field level. To maintain the TU_ATTR_PACKED, I agree with Hathach that maybe it is better to put it on the bit field level and solve the problem. I will rework this PR.

hathach:
look good, though can we still keep the TU_ATTR_PACKED of the overal srtuct ? this is just to prevent any padding between sub-struct member ?

abakosh:
> look good, though can we still keep the TU_ATTR_PACKED of the overall struct? is this just to prevent any padding between sub-struct member?

For some reason, the compiler will keep the warnings, if I do that. This is why I have removed them

hathach:
> > look good, though can we still keep the TU_ATTR_PACKED of the overall struct? is this just to prevent any padding between sub-struct member?
> 
> For some reason, the compiler will keep the warnings, if I do that. This is why I have removed them

I see, what I interperte is the bit-field struct is packed -> size = 2, however, since the global struct is not packed, it is techically valid if compiler inserted padding between 2 struct/union e.g to have each 4-byte aligned. Since the global packed throw warnings, would you mind adding the TU_VERIFY_STATIC() on the total struct size just in case future compiler cause issue with this. It is nice to have anyway, I could add those veriy myself if it is not convenient for you atm since the PR is good to merge already.

abakosh:
> > > look good, though can we still keep the TU_ATTR_PACKED of the overall struct? is this just to prevent any padding between sub-struct member?
> > 
> > 
> > For some reason, the compiler will keep the warnings, if I do that. This is why I have removed them
> 
> I see, what I interperte is the bit-field struct is packed -> size = 2, however, since the global struct is not packed, it is techically valid if compiler inserted padding between 2 struct/union e.g to have each 4-byte aligned. Since the global packed throw warnings, would you mind adding the TU_VERIFY_STATIC() on the total struct size just in case future compiler cause issue with this. It is nice to have anyway, I could add those veriy myself if it is not convenient for you atm since the PR is good to merge already.

Done

hathach (approved):
perfect, thank you

PR #2085, opened 2023-05-28
Fixes bug in tuh_msc_ready

georgeboc:
**Description**
The documentation in [msc_host.h](https://github.com/hathach/tinyusb/blob/069e1ef84f495d2cdd4e5b2fb056a882d2f07787/src/class/msc/msc_host.h#L61-L62) specifies that:
```c
// Check if the interface is currently ready or busy transferring data
bool tuh_msc_ready(uint8_t dev_addr);
```
An interface can be busy either receiving or sending data. The current code contemplates only the data reception, but not the data transmission.

See the following example for reference:
```c
tuh_msc_read10(dev_addr, 0, &buffer, 0, 1, read_complete_cb, 0);

while (!tuh_msc_ready(dev_addr)) {
    tuh_task();
}

tuh_msc_inquiry(dev_addr, 0, &inquiry_resp, inquiry_complete_cb, 0);

while (!tuh_msc_ready(dev_addr)) {
    tuh_task();
}
```
The above example currently fails because the USB BULK transfers have no reserved bandwidth on the bus, so it is possible to have a race condition when the following steps happen:
1. send read command when calling `tuh_msc_read10`,
2. there is no bandwith available to send the command yet, but the processor continues,
3. we run the inquiry command when calling `tuh_msc_inquiry`,
4. we reach [this line of code](https://github.com/hathach/tinyusb/blob/069e1ef84f495d2cdd4e5b2fb056a882d2f07787/src/host/usbh.c#L790) and we fail the assertion, as the interface is still busy trying to send the previous command.

This PR solves the problem by verifying in the `tuh_msc_ready` function that the output endpoint is not busy.

### DIFF
--- a/src/class/msc/msc_host.c
+++ b/src/class/msc/msc_host.c
@@ -118,7 +118,7 @@ bool tuh_msc_mounted(uint8_t dev_addr)
 bool tuh_msc_ready(uint8_t dev_addr)
 {
   msch_interface_t* p_msc = get_itf(dev_addr);
-  return p_msc->mounted && !usbh_edpt_busy(dev_addr, p_msc->ep_in);
+  return p_msc->mounted && !usbh_edpt_busy(dev_addr, p_msc->ep_in) && !usbh_edpt_busy(dev_addr, p_msc->ep_out);
 }
 
 //--------------------------------------------------------------------+


hathach (approved):
spot on, thank you for the fix

PR #2093, opened 2023-06-02
osal_none: make it possible to override the task delay function

abakosh:
I am having an issue with the current implementation of the osal_task_delay function. It was also not stable if you are trying to connect and disconnect the USB device very fast. This will make it possible to override the function to your implemented delay function. Otherwise, it will keep the current implementation.


hathach (approved):
look great, thank you for your PR. I think making this as weak make sense.

PR #2100, opened 2023-06-07
movement of TOP in c0ecf8b5 broke RP2040 SDK build

kilograham:
not sure how this would have worked for any of the others either

### DIFF
--- a/hw/bsp/family_support.cmake
+++ b/hw/bsp/family_support.cmake
@@ -4,6 +4,7 @@ include(CMakePrintHelpers)
 
 # TOP is path to root directory
 set(TOP "${CMAKE_CURRENT_LIST_DIR}/../..")
+get_filename_component(TOP ${TOP} ABSOLUTE)
 
 # Default to gcc
 if (NOT DEFINED TOOLCHAIN)


hathach:
should be no issue, though I found that for cmake, we don't need TOP to be absolute path as with makefile. Do pi example requires TOP to be absolute to build ?

kilograham:
> should be no issue, though I found that for cmake, we don't need TOP to be absolute path as with makefile. Do pi example requires TOP to be absolute to build ?

Well you changed TOP from `${CMAKE_CURRENT_LIST_DIR}/../../..` to `${CMAKE_CURRENT_LIST_DIR}/../..` when you moved it to the parent `family_support.cmake`.

All the uses are a directory deeper `${TOP}/some_file` whic no longer matches up. I figured if you wanted `TOP` to be usable anywhere in that sub-tree I'd make it absolute. Otherwise you could have left it as `${CMAKE_CURRENT_LIST_DIR}/../../..` but that would seem a bit weird

hathach (approved):
thanks for the pr, sorry, I didn't notice this would break pi examples.

PR #2104, opened 2023-06-12
Initial support for USB PD stack

hathach:
**Describe the PR**
Initial support for USB PD stack tested with STM32 G4. Able to communicate via CC channel to:
- retrieve Source Capabilities from a charger 

  PD Source Capabilities
  [Fixed] 5000 mV 3000 mA
  [Fixed] 9000 mV 3000 mA
  [Fixed] 15000 mV 2000 mA
  [Fixed] 20000 mV 1500 mA

- Able to request power to supply 5v, 9V and 15V (which is maximum B-G474E-DPOW1 can withstand).
- Although possible, this PR didn't tested USB communication along with PD.
- API is all very early stage, should only be used for testing purpose only. And is subject to change in the future.


HiFiPhile:
wow you are super fast 😄

PR #2107, opened 2023-06-13
fix usbh issue when device genreate multiple attach/detach/attach when plugging in

hathach:
Some device can generate multiple event attach/detach/attach (bounching) when plugging into usb host. These are literally generated by physical contact which could cause issue with usbh.

Symptom would be seeing **USBH Defer Attach until current enumeration complete** all over.


hathach:
I encounter this issue when working on imx host, though each port will be different. This pr only fixes the issue in usbh, if you are still seeing the issue, it can come from the hcd layer. Feel free to open an discussion or bug issue to provide more detail as well as further analysis.

PR #2132, opened 2023-06-30
tusb_fifo: split constant address functions

mndza:
**Describe the PR**
Due to a missed optimization in the compiler, code for constant address handling is being included in all builds. This change splits the code in different functions to avoid that.

**Additional context**
In some `samd11` builds there's a difference of ~480 bytes in the final binary size.


HiFiPhile:
Hi, sorry for the delay. It's hard to say, I believe you got reduced size as constant address functions are not used in your case, however on stm32 it increases code size of ~250 bytes.

HiFiPhile:
@mndza Please allow maintainer write access to branch. I've splitted const addr functions with macro so no more size increase when they are actually used.

mndza:
I can't do that, but I can recreate the PR within a personal repository. Is that OK?

HiFiPhile:
> I can't do that, but I can recreate the PR within a personal repository. Is that OK?

I'll do a PR then since the approach is different. You can test it first: 

https://github.com/HiFiPhile/tinyusb/tree/fifo_const_split

Issue #1987, opened 2023-03-28
Custom implementation of osal_task_delay for OS_NONE


SelvinPL:
### Related area

Host, osal_task_delay

### Hardware specification

Any

### Is your feature request related to a problem?

I'm using non supported here MCU/Board and I wan use OS_NONE but provide custom `osal_task_delay`

### Describe the solution you'd like

in `osal_none.h`
```
#if CFG_TUH_ENABLED
// currently only needed/available in host mode
#ifdef SOME_CONFIG_FLAG
#include "tusb_os_none.h" //provided by platform implemetation of osal_task_delay
#else
void osal_task_delay(uint32_t msec);
#endif
#endif
```
in usbh.c

```

#if CFG_TUSB_OS == OPT_OS_NONE && ! defined(SOME_CONFIG_FLAG)
// TODO rework time-related function later
void osal_task_delay(uint32_t msec)
{
  const uint32_t start = hcd_frame_number(_usbh_controller);
  while ( ( hcd_frame_number(_usbh_controller) - start ) < msec ) {}
}
#endif
```

### I have checked existing issues, dicussion and documentation

- [X] I confirm I have checked existing issues, dicussion and documentation.

hathach:
I see some bsp/sdk provide an better delay than using hcd frame number, which only exist if there is an device attached to bus. I think we can have an optinal callback to get timestamp instead of having an custom OS. There is other place that need this callback as well. Let me think which name we should use

PR #2134, opened 2023-07-01
Non-bus-powered MSP430 support.

clhenry:
**Describe the PR**
Adding supported for hot plugging while using external power.

**Additional context**


cr1901:
It looks fine to me (I think at the time, I only focused on bus-powered applications b/c I found it difficult to reliably put the [MSP-EXP430F5529LP](https://www.ti.com/tool/MSP-EXP430F5529LP) into self-powered mode.

I do have a few questions:

1. Is `USBCTL &= ~FRSTE;` and `USBCTL |= FRSTE;` intended to prevent a race due to power transients causing a second reset while in the process of handling the first reset? What happens if the reset vector is entered while processing the current reset?
2. I think I can't even find a definition for `osal_task_delay` when [`osal_none` is being used](https://github.com/hathach/tinyusb/blob/f5d0510064bd2a0ffd040697de1fac707fa70069/src/osal/osal_none.h#L40). Presumably it's a noop and then is optimized out in release mode. I think the whole build should fail due to missing definition when compiled in debug mode.

   It looks like other platforms use an [include guard](https://github.com/hathach/tinyusb/blob/f5d0510064bd2a0ffd040697de1fac707fa70069/src/portable/microchip/pic32mz/dcd_pic32mz.c#L164) or a [`board_millis`](https://github.com/hathach/tinyusb/blob/f5d0510064bd2a0ffd040697de1fac707fa70069/src/portable/sunxi/dcd_sunxi_musb.c#L102C1-L102C1) function provided by the application (which is not how I'd do it).

   Personally, I'd use the msp430 `__delay_cycles` if I absolutely _have_ to have a hardcoded delay, but that's tied to the current CPU frequency. I'm assuming you're testing without an OS, and so the code works fine without working delays. I would put an include guard for now that doesn't call `osal_task_delay` when `CFG_TUSB_OS == OPT_OS_NONE`.

clhenry:
1. As this was 6 months ago I had to wrack my brain to try and remember the justification. I believe this was a partial implementation of TI's software fix for the USB4 [errata](https://www.ti.com/lit/er/slaz314ad/slaz314ad.pdf?ts=1688268237081) affecting the MSP430F5529. According to the linked document `FRSTE` should be cleared on receipt of the reset interrupt and set on receipt of the suspend, setup and endpoint 0 interrupts.
I'll be pushing a change later that correctly implements their recommend fix. In the mean time, if you have any suggestions to address your original concern, I'm all ears.

2. I'm using FreeRTOS for my project. The initial delay in `handle_bus_power_event` was the shortest I could do in order to get the functionality working. The delay in the loop can be as short as 500 ns per TI's recommendation. As you pointed out, the intrinsic delay function is frequency dependent. I don't mind modify the PR to support a non-OS environment but I'm at a loss as to what would work for everyone if not `board_millis`. Again, I'm open to recommendations.

cr1901:
Ack re: the errata; when I did this port originally, I don't recall looking at the errata. That's fine then (incl. your upcoming changes).

> I'm using FreeRTOS for my project. 
Ahhh I didn't know FreeRTOS ran on MSP430, tbh :D!

I want to wait and hear what @hathach has to say re: "using `board_millis` in `portable/`. #1987 is also relevant.

clhenry:
@hathach I see that the non-os builds are failing on `osal_task_delay`. Can you chime in with your thoughts on the current discussion?

cr1901:
@clhenry Just to be clear, @hathach may be very busy, and while I am a collaborator, I mostly focus on backend stuff. I want to wait for @hathach to be less busy so he can give guidance on "sometimes no OS needs a delay" problem.

You're already expected to instantiate an interrupt handler yourself and place a call into [`tud_int_handler`](https://github.com/hathach/tinyusb/blob/master/hw/bsp/msp430/family.c#L36) in your application. I'm not against adding a function like e.g. `tud_delay` that's called by `osal_task_delay` for no-OS, that the user is expected to fill in for their application if necessary. But adding this functionality requires some care.

hathach:
sorry, I was a bit busy atm, will check this out whenever I could. Thank you for your patient.

src/portable/ti/msp430x5xx/dcd_msp430x5xx.c
```diff
@@ -610,14 +636,76 @@ static void handle_setup_packet(void)
...
+  USBIEPCNF_0 |= UBME;  // Errata USB10 workaround.
+  USBOEPCNF_0 |= UBME;  // Errata USB10 workaround.
   dcd_event_setup_received(0, (uint8_t*) &_setup_packet[0], true);
 }
 
+#if CFG_TUSB_OS == OPT_OS_NONE
+TU_ATTR_ALWAYS_INLINE static inline void tu_delay(uint32_t ms) {
+  // msp430 can run up to 25Mhz -> 40ns per cycle. 1 ms = 25000 cycles
+  // each loop need 4 cycle: 1 sub, 1 cmp, 1 jump, 1 nop
+  volatile uint32_t cycles = (25000 * ms) >> 2;
```
hathach, line 663: delay using nop where msp430 can run up to 25Mhz

hathach (approved):
thank you very much, sorry for late response. I have update the PR to use blocking delay (cycle burning, taking consideration for msp430 can run at max 25Mhz). Since there is no generic delay API for tinyusb at the moment. We can change this later should we decide to have an generic time function later on.

PS: I also add some changes to get cmake build support (which I am using) for msp430 as well.

PR #2135, opened 2023-07-02
made line ends consistent.

JustAnother1:
**Describe the PR**
console applications that expect "\r\n" as line ends will print lines ending only with "\n" in a star step fashion. They will not continue the next line at the beginning of the line, but at the current position.

**Additional context**
Most of the line ends in the log messages already were "\r\n" so I changed the few "\n" line ends to make it consistent.


hathach (approved):
all looks good, thank you for your PR.

PR #2146, opened 2023-07-14
Fix host buffer alignment setting

tannewt:
When buffer alignment isn't given for TUH, it defaults to 4. The iMX needs 32 to match the cache and is set using the universal setting in CircuitPython. Change the default to fall back to the universal setting before being 4.

If the alignment is 4, we risk invalidating addresses outside the buffer that haven't been written back.


hathach (approved):
ah spot-on, cache line for M7 is 32 bytes

PR #2150, opened 2023-07-18
Fix flash-jlink target to work on Windows command prompt

kkitayam:
**Describe the PR**
This PR makes "flash-jlink" target work on Windows command prompt.
Because "echo" command of Windows command prompt can't handle line breaks in a variable, some errors occur when "flash-jlink" target running.


hathach (approved):
oops, that is a pain, I was trying to clean up the script, but look like it isn't working out :D

PR #2151, opened 2023-07-18
Fix build issues when building for RP2040 from pico-examples/pico-sdk.

kilograham:
* include_guard requires GLOBAL as family.cmake is included in multiple non child places
* the following recently rp2040 added check is sueprfluous (`family_configure_host_example` on rp2040 should do this already), and breaks if pico_pio_usb is not avaialble, so i have removed the rp2040 check
  ```
  family_configure_host_example(${PROJECT} noos)
  # Add pico-pio-usb for rp2040 since user can choose to run on bit-banging host
  if(FAMILY STREQUAL "rp2040")
      family_add_pico_pio_usb(${PROJECT})
  endif()
   ```
* added new `family_example_missing_dependency` function to print missing dependency warning, so pico-examples can override it to be less in your face, and also more contextual to pico-examples
* added `TINYUSB_OPT_SKIP_CHECK_IPO_SUPPORTED` var to skip the recently added IPO check which produces a bunch of warnings on RP2040

note: we really should set up an action to build pico-sdk/pico-examples periodically - either on pico- end, or tinyusb- end... there will be times when it is genuinely broken, but it would be nice to catch these breaks.


hathach:
oops, sorry, I am still getting used to cmake. Though you are right, I will try to include building a couple of usb examples from pico-examples into the ci as well.

hw/bsp/family_support.cmake
```diff
@@ -36,12 +36,14 @@ if (NOT EXISTS ${CMAKE_CURRENT_LIST_DIR}/${FAMILY}/family.cmake)
   message(FATAL_ERROR "Family '${FAMILY}' is not known/supported")
 endif()
 
-# enable LTO if supported
-include(CheckIPOSupported)
-check_ipo_supported(RESULT IPO_SUPPORTED)
-if (IPO_SUPPORTED)
-  set(CMAKE_INTERPROCEDURAL_OPTIMIZATION TRUE)
-endif ()
+if (NOT TINYUSB_OPT_SKIP_CHECK_IPO_SUPPORTED)
```
hathach, line 39: yeah this makes sense. Actualy I think we should skip this entirely for rp2040. LTO is global option, and it should only be set by top level application cmakelists.txt. It is ok for other ports since it only applies to examples within tinyusb. But I don't really want to force this on user application at all.

hathach:
hmm, I pushed some update to your branch, but look like github doesn't pick that up.

hathach (approved):
I skip the LTO option entirely for rp2040 since this should be decision by application top level cmake file. @kilograham let me know you are with the changes.

hathach:
update: and PR building pico-examples will be added soon (after this one). CI will only build 2 usb example in the pico-examples repo dev_hid_composite and host_cdc_msc_hid (currently disabled due to usb host api changes). let me know if this is sufficient enough. 

https://github.com/hathach/tinyusb/blob/2e7192c5c277be039d67950c302b7efb2b80b9a3/.github/workflows/cmake_arm.yml#L92-L109

```
    - name: Build pico-examples
      if: matrix.family == 'rp2040'
      env:
        PICO_SDK_PATH: ${{ github.workspace }}/pico-sdk
        PICO_EXAMPLES_PATH: ${{ github.workspace }}/pico-examples
      run: |
        # symlink tinyusb to pico-sdk
        rm -rf $PICO_SDK_PATH/lib/tinyusb
        ln -s ${{ github.workspace }} $PICO_SDK_PATH/lib/
        # change CMakelists.txt to only build selected examples
        cd $PICO_EXAMPLES_PATH
        sed -i -r 's/add_subdirectory/#add_subdirectory/g' CMakeLists.txt
        echo "" >> CMakeLists.txt
        echo "add_subdirectory(usb/device/dev_hid_composite)" >> CMakeLists.txt
        # TODO temporarily disable host examples since tinyusb API is changed
        # echo "add_subdirectory(usb/host/host_cdc_msc_hid)" >> CMakeLists.txt
        cmake -S . -B _build -G Ninja -DFAMILY=rp2040 -DBOARD=pico_sdk
        cmake --build _build
```

kilograham:
LGTM - not sure if it is better to just have the job fail if there are any issues (e.g. the host API changing is a valid reason for failure), though we could add a second job for that - perhaps i'll add that on our end

hathach:
> LGTM - not sure if it is better to just have the job fail if there are any issues (e.g. the host API changing is a valid reason for failure), though we could add a second job for that - perhaps i'll add that on our end

maybe you could update the pico-examples on dev branch, and we can use dev branch here. otheriwse a failed ci will prevent any merge on this repo. We could fix and enable it in follow up PR. Thank you.

PR #2152, opened 2023-07-18
add is_compiler_supported_by_pico_pio_usb() for rp2040

kilograham:
previously users of TinyUSB (e.g. pico-examples) would have to decide this for themselves. This function couples the check closer with the actual version of Pico-PIO-USB used (since TinyUSB picks)


hathach (approved):
look great

PR #2155, opened 2023-07-20
Board support for XMC4700_RELAX

jferreir:
Add board support for XMC4700_RELAX, based on XMC4700_F144x2048.
Add support UART_DEV


hathach (approved):
I don't have xmc4700 to test, but it looks great, thank you.

PR #2171, opened 2023-07-24
Invoke power state changes on DCD_EVENT_BUS_RESET

Rocky04:
Should fix a bug that the device state isn't reset when a bus reset happened as discussed here: https://github.com/hathach/tinyusb/discussions/2169


hathach:
can you provide context of your setup and how to reproduce & test the issue.

Rocky04:
I use a USB setup with multiple USB hubs. I have a laptop where a screen is connected via USB-C and that has a built-in hub. A USB switch is connected to that, to that there is an active USB hub connected because the switch don't have enough ports and and can't provide enough power.

With that setup I noticed that my Wooting sometimes remains lit after I unplug the screen from the laptop. So I checked out the TinyUSB example and was able to reproduce the issue on a GD32 dev kit which uses the STM32 code. Even when the single-ended zero state is constantly hold the device still blinks in the mounted state. I need to also add schedule the `DCD_EVENT_UNPLUGGED` together with the `DCD_EVENT_BUS_RESET` in the interrupt handler.

This is a general issue. As mentioned in the discussion, a device should reset the internal USB state when a reset on the bus occur.

hathach:
thank you for the PR, What is your GD32 mcu specifically ? even though it is generic, we will still want to reproduce the issue and verify the fix. Is your monitor has built-in battery. If yes, it looks to me that the reproducing steps would be
- Attached to self-power hub -> PC
- remove hub from PC, but still keeping power

right ?

Rocky04:
Not sure how much you want to spend but this is my setup:

Asus Flow 13 (2021) <- USB-C PD -> Cooler Master Tempest GP27U <- USB-A 2.0 -> ATEN US424 (passive) <- USB A 3.0 -> TP-Link UH720 (active) <- USB A -> GD32303C-START

hathach:
thank you for your detail setup explanation, I will try to simulate it with self-power hub of some sort.

Rocky04:
Looks like there must be note that the fall through is wanted.

Sadly the [USB3CV compliance test program](https://www.usb.org/document-library/usb3cvx64-tool) don't seems to test for the reset handling.

Rocky04:
The issue is triggered on my system by USB switch which is in-between the host and the active USB hub.
The hub shuts all its ports down if it's up facing port isn't active but the switch provides the idling J state to it's down facing ports even if the selected host isn't present. Every time the host is toggled on the switch the UFP are reset and then J state of the plugged in device is present again. 

To be clear, this setup only revealed the reset issue and shows that TinyUSB don't seems to prober react to the reset. Because the host should only reset the device if there was an issue. The reset event is kind of rare but can may lead into weird problems.

Here is how it looks like:
image
_The short spikes on D+ are the resets._

hathach:
> * Looks like there must be a note that the fall through is wanted. Not sure if `[[fallthrough]]`, `/* FALLTHROUGH */` or may `__attribute__((__fallthrough__))` does the trick.

you should use `TU_ATTR_FALLTHROUGH;` like https://github.com/hathach/tinyusb/blob/master/src/host/usbh.c#L664

> * I cross checked the signal state (SE0) via a USB power tester (AVHzY CT3+).
> * The problem to reproduce the issue is likely to trigger a reset without an enumeration attempt afterwards. It should be possible to indirectly test this by checking if the `_usbd_dev.connected` parameter is set to 0 after the second device reset. Because that should come after the first device descriptor request and so this value would be 1 right before.

Thank you for detailed breakout, I will try to reproduce it whenever I could

Rocky04:
@hathach 
As a warming, Wooting sent me a dev build with the changes of this PR for testing. So far I was only able to test it on a dev board. It seems I now that device get stuck in sleep every time I disconnect my host. This behavior depends on the MCU specific implementation. For the GD32 this should be addressed by the optimizations in the PR #2178. So it can be that fixing the bug causes the same issue for other MCUs as well which likely don't have the Start of Frame fallback to leave the sleep mode.

Rocky04:
The problem with the Wooting is addressed now, it was related to a bug.

So together with the the PR #2202 these changes should help making the USB handling overall more reliable which likely addresses issues for some edge cases.

Rocky04:
_Had some time again recently to check this issue out..._

The problem is that when a TinyUSB device is used on an active USB hub it don't react properly when the host is either shutdown or disconnected from the hub. In that case the hub will may provide a single-ended zero signal on both data wires resulting in a continuous reset state for the device while the device is still being powered.

If the device was in the suspended state before the `tud_resume_cb` was never invoked while `tud_umount_cb` was invoked all the time even when the device wasn't in the mounted state before. I also removed redundant code which clears the current device state.

src/device/usbd.c
```diff
@@ -1069,17 +1093,20 @@ static bool process_get_descriptor(uint8_t rhport, tusb_control_request_t const
...
-      _usbd_dev.cfg_num = 0;
-      _usbd_dev.suspended = 0;
-      send = true;
-      break;
+TU_ATTR_FAST_FUNC void dcd_event_handler(dcd_event_t const * event, bool in_isr)
+{
+  volatile static uint8_t last_event_id = DCD_EVENT_INVALID;
 
+  // Skip the useless repeating bus states
+  if (last_event_id == event->event_id && DCD_EVENT_UNPLUGGED <= event->event_id && DCD_EVENT_RESUME >= event->event_id)
```
HiFiPhile, line 1101: This will also ignore `DCD_EVENT_SOF`

Rocky04: Despite that this can't be merged easily anymore it's also outdated and should contain a bug.

_Maybe next week I can provide the newer version which should work on the current branch._

HiFiPhile:
Dear contributor, this PR has been closed due to long time inactivity, if you believe it's a mistake please reopen or contact me.

PR #2180, opened 2023-07-25
HID SOF interrupt support

Xelus22:
**Describe the PR**
Adds HID SOF support.

**Additional context**
~~Tested and working on CH32V307~~

**Questions**
1. Unsure if it should be called `hidd_sof_isr` or `hidd_sof_cb`? (used `isr` since the audio one uses that too)
2. Should I pass in `uint8_t rhport` into `hidd_sof_isr`?

### DIFF
--- a/src/class/hid/hid_device.c
+++ b/src/class/hid/hid_device.c
@@ -412,4 +412,9 @@ bool hidd_xfer_cb(uint8_t rhport, uint8_t ep_addr, xfer_result_t result, uint32_
   return true;
 }
 
+void hidd_sof_isr(uint8_t rhport, uint32_t frame_count)
+{
+  (void) rhport;
+  tud_hid_sof_cb(frame_count);
+}
 #endif
--- a/src/class/hid/hid_device.h
+++ b/src/class/hid/hid_device.h
@@ -118,6 +118,8 @@ TU_ATTR_WEAK bool tud_hid_set_idle_cb(uint8_t instance, uint8_t idle_rate);
 // Note: For composite reports, report[0] is report ID
 TU_ATTR_WEAK void tud_hid_report_complete_cb(uint8_t instance, uint8_t const* report, uint16_t len);
 
+// Invoked when received SOF interrupt
+TU_ATTR_WEAK void tud_hid_sof_cb(uint32_t frame_count);
 
 //--------------------------------------------------------------------+
 // Inline Functions
@@ -410,6 +412,7 @@ void     hidd_reset           (uint8_t rhport);
 uint16_t hidd_open            (uint8_t rhport, tusb_desc_interface_t const * itf_desc, uint16_t max_len);
 bool     hidd_control_xfer_cb (uint8_t rhport, uint8_t stage, tusb_control_request_t const * request);
 bool     hidd_xfer_cb         (uint8_t rhport, uint8_t ep_addr, xfer_result_t event, uint32_t xferred_bytes);
+void     hidd_sof_isr         (uint8_t rhport, uint32_t frame_count);
 
 #ifdef __cplusplus
  }
--- a/src/device/usbd.c
+++ b/src/device/usbd.c
@@ -119,7 +119,7 @@ tu_static usbd_class_driver_t const _usbd_driver[] =
     .open             = hidd_open,
     .control_xfer_cb  = hidd_control_xfer_cb,
     .xfer_cb          = hidd_xfer_cb,
-    .sof              = NULL
+    .sof              = hidd_sof_isr
   },
   #endif
 
--- a/src/device/usbd.c
+++ b/src/device/usbd.c
@@ -597,6 +597,16 @@ void tud_task_ext(uint32_t timeout_ms, bool in_isr)
       break;
 
       case DCD_EVENT_SOF:
+        TU_LOG_USBD("\r\n");
+        for (uint8_t i = 0; i < TOTAL_DRIVER_COUNT; i++)
+        {
+          usbd_class_driver_t const * driver = get_driver(i);
+          if (driver && driver->sof)
+          {
+            driver->sof(event.rhport, event.sof.frame_count);
+          }
+        }
+        break;
       default:
         TU_BREAKPOINT();
       break;


hathach (commented):
I don't see the reason to add sof callback to hid driver ? Could you explain why this is needed.

Xelus22:
Hi hathach, ive added in the SOF callback to the HID driver specifically for fast mouse response and minimum latency.

I've put it here since I would like to use tinyUSB to follow the current fastest open-source mouse firmware on the market (after razer) 

https://github.com/zaunkoenig-firmware/m2k-firmware/blob/main/mouse/Src/main.c

you can see here they wait for a SOF interrupt before starting the routine and getting data from the sensor/buttons then send it out. Since there is only 125us frame time between packets at 8khz polling rate.

mitchellcairns:
This would be very useful. Working on a gamepad at the moment and this would help a ton with reducing latency at lower interval inputs.

Rocky04:
- There should be only necessary things placed in an interrupt. Everything else should be done in the main loop.
- Data response should be done in an asynchronous manner.
- The USB driver only guarantee bandwidth (for isochronous transfers) and uniform timings (for interrupt transfers).

Meaning a HID device can be polled at any time as long as the rate is constant while it's unclear when that actually happens between two SOFs packages. So the SOF interrupt should be only used to save the time when it happened and the callback should be invoked later on from TinyUSB in the main loop handler if needed. _Then realtime stuff can be done by checking when the SOF occur and how much time has passed since then._ For HID devices the SOF can be used to synchronize the polling to the scanning and so to minimize derivation and so making the delay more consistent. It would not necessarily reduce the latency. For example the interface can be polled right after the SOF and so there is likely not enough time to read the sensor and process it's data. Then the data would be always one frame behind. _And you don't want what the polling maybe gets skipped because the scanning hasn't finished._

For reducing the latency a real time approach would be needed. Like to figure out when the specific interface is polled and to read the sensor perfectly timed right before. Of course for that the time for the scanning and it's processing must be known and be consistent.

Xelus22:
I understand what you're saying, but for that to happen a SOF callback here I think would exactly help facilitate that. For the user to implement this RTOS functionality if required and not be TinyUSBs responsibility.

Unless you mean this should be extended to all device types.

Rocky04:
As far as I know it's not guaranteed that interrupt transfers are always scheduled directly after the SOF, even if that is typically done this way. I'm not aware that the USB specification specifies when withing a frame interrupts transfers should take places as long as the timing is ok.

Imagine a lot of devices with a lot of interrupt interfaces. The more there are the longer it takes to handle all and the last one can be scheduled very late compared to the first one.

Even your linked info didn't state that interrupts packages are always scheduled after the SOF. It's not even mentioned with a single word. They are just using an example picture where this is the case and that is all.

Rocky04:
- I think a general SOF callback can be helpful for any application. It's likely needed for any realtime implementation. As the other callbacks these drivers callbacks are handled also by the `tud_task()` in the main loop. So that can't be used as a time reference for any time critical stuff because you don't know the time between the SOF interrupt triggered by the MCU and when this callback is actually executed. Of course for non-critical stuff it should be fine and better as nothing.
- I don't see why this should be limited to a specific device type at all. It can be implemented the same way as the other optional callbacks. To add the event in the query and then let `tud_task()` call it the main loop independent from the device driver if the callback is present.
- I'm somehow confused about the class driver stuff of TinyUSB. I would expect that if there is a driver SOF callback for an audio device then there should be one too for the video, but this isn't the case. I also don't get why the device driver callback for the SOF is tied to the device type and not to the isochronous transfer type regardless what data is send. Like if a double buffer is used to determine when the buffer to store the new data and for sending the old data should be switched. I don't see why the SOF would be useful to a particular driver and so be driver specific.

Rocky04:
@perigoso Here, as simple as I can write it for you...

- Add a general SOF callback. It makes absolutely no sense to implement that in a HID specific way.
- It would help making the delay more consistent but likely add latency. _Please check out how Motion Sync works on a modern mouse._
- Your statement is contrary. This PR is all about adding a HID specific callback and not about a general one.
- I aware of the USB specification, how a typically implementation for a USB device works and also about latency on input devices. So no, I'm not confused but rather aware of the potential time specific impacts.

_No offence, but instead like you I prefer to keep it to myself what I think when I read your post._

Rocky04:
@Xelus22
The function `hidd_sof_isr()` as configured in the class driver is called in `dcd_event_handler()` and so executed when the SOF interrupt is handled. That's why the name is "isr". Due to this there would not pass much time between the interrupt and the execution. But that approach isn't good because only very important stuff should be done in an ISR and as less as possible execution time should be spend there in general.

The added code from you in the `tud_task_ext()` function on the other side handles the callbacks which are executed in the main loop by `tud_task()`. Because of this change the SOF handling is actually done two times, basically immediately in the interrupt handling and again some time later outside the interrupt.

_I created an example PR #2213 which would implement a general SOF callback independent from the class device driver._

src/class/hid/hid_device.c
```diff
@@ -412,4 +412,9 @@ bool hidd_xfer_cb(uint8_t rhport, uint8_t ep_addr, xfer_result_t result, uint32_
   return true;
 }
 
-#endif
+void hidd_sof_isr(uint8_t rhport, uint32_t frame_count)
+{
+  (void) rhport;
+  tud_hid_sof_cb(frame_count);
+}
+#endif
```
Rocky04, line 420: A function executed directly in a synchron manner isn't a callback. Meaning don't put a "_cb" in an "_isr" because these names refer to different things.

In this function only code should be placed which needs to be executed inside the interrupt service routine right after when the start of frame occurred.

src/device/usbd.c
```diff
@@ -597,6 +597,16 @@ void tud_task_ext(uint32_t timeout_ms, bool in_isr)
...
+        TU_LOG_USBD("\r\n");
+        for (uint8_t i = 0; i < TOTAL_DRIVER_COUNT; i++)
+        {
+          usbd_class_driver_t const * driver = get_driver(i);
+          if (driver && driver->sof)
+          {
+            driver->sof(event.rhport, event.sof.frame_count);
+          }
+        }
+        break;
```
Rocky04, line 609: Here the callback should be executed but instead the ISR function is executed a second time.

src/device/usbd.c
```diff
@@ -119,7 +119,7 @@ tu_static usbd_class_driver_t const _usbd_driver[] =
     .open             = hidd_open,
     .control_xfer_cb  = hidd_control_xfer_cb,
     .xfer_cb          = hidd_xfer_cb,
-    .sof              = NULL
+    .sof              = hidd_sof_isr
```
Rocky04, line 122: The assigned ISR function at this place gets called here: https://github.com/hathach/tinyusb/blob/cfe8ca3fa51061bbe7e519b72236f9aabae13e59/src/device/usbd.c#L1132

src/class/hid/hid_device.h
```diff
@@ -118,6 +118,8 @@ TU_ATTR_WEAK bool tud_hid_set_idle_cb(uint8_t instance, uint8_t idle_rate);
 // Note: For composite reports, report[0] is report ID
 TU_ATTR_WEAK void tud_hid_report_complete_cb(uint8_t instance, uint8_t const* report, uint16_t len);
 
+// Invoked when received SOF interrupt
+TU_ATTR_WEAK void tud_hid_sof_cb(uint32_t frame_count);
```
Rocky04, line 122: As a callback it should be invoked in the ISR due to the SOF event and called here: https://github.com/hathach/tinyusb/blob/cfe8ca3fa51061bbe7e519b72236f9aabae13e59/src/device/usbd.c#L599

Rocky04 (commented):
My complains as a review...

perigoso:
> Here, as simple as I can summaries it, just for you...

This is better, you should make extra effort from the start.

> A general SOF callback should be added. It makes absolutely no sense to implement that in a HID specific way.

I agree, I made the same suggestion.

> It would help making the delay more consistent but likely add latency. Please check out how Motion Sync works on a modern mouse.

This is precisely what I noted as being an implementation detail.

> Your statement is contrary. This PR is all about adding a HID specific callback and not about a general one.

It's not contrary, I made no mention of "HID specific" or "general callback", just that this is about exposing the SOF, and you seemed to be talking about firmware implementation details.

> I'm aware of the USB specification, how a typically implementation for a USB device works and also about latency on input devices. So no, I'm not confused but rather aware of the potential time specific impacts.

Ok.

> No offence, but unlike you I prefer to keep it to myself what I think when I read your post.

This is not twitter, I did not give my thoughts, I asked you to explain yourself better because I could make nothing of it.

src/class/hid/hid_device.c
perigoso: This is a callback because it's a configurable function defined by the user, not because of how it is called

Rocky04:
@perigoso 

Upf, details matter... An implementation is always about the details. This is a PR, so side effects and specifics need to be taken into account. The underlying goal for the desired feature of this PR was all about input delays on USB. So with the provided info I was basically questioning the purpose of the PR as is, even I referred to a use case which would be implementation specific.

I can understand that there is likely a language barrier and that maybe not all things are easy to understand, but this is also not a discussion about a potential simplified approach for a concept. So no, I don't see any reason why to keep it as simple as possible. I'm also very technical for a reason, to either point out a potential misconception from others or that I can be corrected in the case I'm wrong or rely on false assumptions.

In the first comment on this PR [hathach](https://github.com/hathach) was questioning why there is a need of a HID specific SOF callback in the driver part of the code. You mentioned that this would be useful to all drivers but that the scope of this PR is just for the HID driver. So no, to me your suggestion in your second post was contrary to the first one.

Strange, according to your wording I thought you handled it like Twitter _(or X)_.

src/class/hid/hid_device.c
Rocky04: There are naming conventions for a reason. So people who read a function name have an expectation what the function does. Messing around with that only causes issues.

Just have a look at the other TinyUSB callbacks, like `tud_mount_cb()`. The event for it is invoked (but the function isn't executed) in the ISR. It's later called in the main loop from `tud_task()`. But `audiod_sof_isr()` on the other hand is executed in the ISR. So the "_cb" functions are executed in the main loop and the "_isr" functions within the interrupt service routine.

perigoso:
You're the one repling like you're on twitter, this isn't personal

src/class/hid/hid_device.c
perigoso: You cant just make up conventions

https://en.m.wikipedia.org/wiki/Callback_(computer_programming)

Rocky04: It's the way how it used here in this repro. Feel free to prove me wrong by pointing to a function referred as callback in its name which isn't called indirectly in the main loop. Would be appreciated...

HiFiPhile:
Please calm down friends, it's just a hobbits project we are not NASA :)

It happens sometime we have different working style / understanding of a problem, not means others is wrong.

hathach:
late response as usual, seems like there is a bit of heated dicussion in this issue, I hope we are still cool and don't take that personal.  

> Hi hathach, ive added in the SOF callback to the HID driver specifically for fast mouse response and minimum latency.
> I've put it here since I would like to use tinyUSB to follow the current fastest open-source mouse firmware on the market in terms of click latency (after razer)
> https://github.com/zaunkoenig-firmware/m2k-firmware/blob/main/mouse/Src/main.c
> you can see here they wait for a SOF interrupt before starting the routine and getting data from the sensor/buttons then send it out. Since there is only 125us frame time between packets at 8khz polling rate.

Thank you for explanation, my take on this is we should use generic SOF. I did actually add API for generic SOF previously when doing audio sof for feedback. Though I back out on generic API since there is no usage back then https://github.com/hathach/tinyusb/pull/1381/files#diff-290e049d139d4c4ba09224b43d2f53d2ea705c59ba979c3881143be80d5402deR91

Maybe it is good time to re-add it.

Rocky04:
@hathach 
What's missing and should be re-added? The linked PR is merged? Do you mean the function to dynamically set the SOF ISR function?

Rocky04:
I would prefer to have both, a SOF callback which is executed by `tud_task()` in the main loop and one separated from this in the ISR. For the first there is already a PR #2213.

Rocky04:
The issue right now with setting the function pointer for a driver specific SOF callback (here used in a common sense), which is executed within the ISR, is that this would need to know to which driver it should be applied in the case multiple configurations are used. The driver name is only present on a debug level so that can't be used.

Because of this I think it would be better to let it as is so that the SOF ISR is purely driver related. Meaning if that behavior should be changed or controlled that should only be possible by using a vendor defined driver. For that the SOF ISR should be opened up just for that. Here is a draft for this #2220.

Rocky04:
I'm not really sure what you mean...

For me there are two places, executed a bit later in the main loop via a general callback and or directly when the MCU handles the SOF interrupt via a specific driver callback. I also don't get why you mentioning `tud_sof_enable()` because controlling if the MCU should react to a SOF package or not is independent what the application should do when this occur and can be handled.

I also think a vendor defined driver should have full control and I don't see any reason why the SOF ISR should not be exposed for them.

hathach (changes requested):
we should only add sof_isr that can be auto managed (on/off when needed) to class driver e.g audio with feedback enpoint on/off along with its audio streaming interface. 

Since HID does not need it to function properly (thing breaks with audio if not), we can add an generic sof cb in isr for this purpose for app to turn it on and off when it wants. You don't need to make any changes for this PR. I remembered added that sometime ago (then reverted), I could try to re-add it when possible. Thank you for brining up this PR and issue.

marcpabst:
Hi, I'm really sorry to revive this issue but I tought it might be a good place to figure out what the current status on a generic SOF callback is? Aplogies if I missed anyhting, but I couldn't relaly find any information on this.

HiFiPhile:
Thanks for everyone's effort.

Generic user land SOF support has been merged in #2213.

One can use `bool tud_sof_cb_enable(bool en)` and  `void tud_sof_cb(uint32_t frame_count)` to use SOF callback.

PR #2185, opened 2023-07-27
Fix typos in HCD rusb file

abakosh:
**Describe the PR**
Host Library:
The unplugging process of the USB device might lead to an infinite loop. It is easy to run into this situation if unplugging occurs during the enumeration process. Also, it happens when unplugging while the host and the USB device are doing something. 

The issue is really big. As the stack enters an infinite loop when this is happening.

**Additional context**
I am testing with the RA4M2 family, with no OS. I know this fix is not meant if you are hosting a hub device. However, if you are planning to host a single device, it increases the robustness of plugging/unplugging while busy issue.
This was tested with MSC devices, I can see the improvement. 

@hathach I am not sure if this fix is contagious for other scenarios. Please let me know if I need to take care of something else.

Thank you.

### DIFF
--- a/src/portable/renesas/rusb2/hcd_rusb2.c
+++ b/src/portable/renesas/rusb2/hcd_rusb2.c
@@ -265,7 +265,7 @@ static bool pipe_xfer_in(rusb2_reg_t* rusb, unsigned num)
     pipe->buf = (uint8_t*)buf + len;
   }
   if (len < mps) {
-    rusb->D0FIFOCTR = RUSB2_CFIFOCTR_BCLR_Msk;
+    rusb->D0FIFOCTR = RUSB2_D0FIFOCTR_BCLR_Msk;
   }
   rusb->D0FIFOSEL = 0;
   while (rusb->D0FIFOSEL_b.CURPIPE) ; /* if CURPIPE bits changes, check written value */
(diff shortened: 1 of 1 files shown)


src/host/usbh.c
```diff
@@ -889,9 +889,20 @@ TU_ATTR_FAST_FUNC void hcd_event_handler(hcd_event_t const* event, bool in_isr)
 {
   switch (event->event_id)
   {
+#if CFG_TUH_HUB
 //    case HCD_EVENT_DEVICE_REMOVE:
 //      // mark device as removing to prevent further xfer before the event is processed in usbh task
 //      break;
+#else
```
hathach, line 896: this is a hack, we cannnot clear the event queue like this though. E.g rusb2 with 2 ports can be configured as BOTH host, but they still share the same event queue. Clearing the queue is not an option even there is no hub support. This is mostly caused by physical connection bouncing, We will analyze the rootcause and fix it properly though, just need a bit of time.

hathach (changes requested):
unfortunately the "fix" is too hacky. The changes for hcd_rusb2 look good, maybe split this to an seperate PR so that it can be merged.

hathach (approved):
look good, hw test failed is due to the new addition of new hardware. We have to specific jlink serial so that It connect to correct board. GIve me some time to fix this. Once done, this will be merged. Thank you

PR #2187, opened 2023-07-28
Fixed typo in binary declaration RP2040 hw bspw family.c

0xCCF4:
**Description**
There is a typo in the following line:
https://github.com/hathach/tinyusb/blob/86c416d4c0fb38432460b3e11b08b9de76941bf5/hw/bsp/rp2040/family.c#L152

The interface declaration should declare the `UART_TX_PIN` and `UART_RX_PIN` pin instead of declaring the `UART_TX_PIN` pin twice.


hathach (approved):
thank you for your PR

PR #2200, opened 2023-08-02
default class driver log level to CFG_TUH/TUD_LOG_LEVEL 

hathach:
**Describe the PR**

- allow application to selectively disable usbd/usbh or driver log. The log level is the level where it is logged. To disable it, increase it to level > CFG_TUSB_DEBUG. (kind of counterintuitive). For example to disable device log while workingg on host `CFG_TUD_LOG_LEVEL=10`
- also move .codespell to tools (unrelated change)

@tannewt you may need this to troubleshoot host issue while running cpy.


tannewt:
I tried to use this. One sticking point was that setting CFG_TUD_LOG_LEVEL failed to find corresponding macros for the given level. It may be better to pick a specific number to disable them.

hathach:
"failed to find corresponding macros" I don't quite understand the issue, can you elaborate a bit. What is your setting for CFG_TUD_LOG_LEVEL and CFG_TUSB_LOG.

tannewt:
Sorry I missed this. I set CFG_TUD_LOG_LEVEL to 10 and it failed to find macros, such as TU_LOG10 for it. I think the STRCAT here is to blame: https://github.com/hathach/tinyusb/blob/master/src/common/tusb_debug.h#L65-L97

hathach:
ah yeah, you are right, that is my bad, please try to change CFG_TUD_LOG_LEVEL=3 only now, which the log is able to expanded to. I will try to fix this later on :)

PR #2202, opened 2023-08-02
Invoke unmounted state on configuration reset

Rocky04:
Example implementation of the discussion #2201.

- Invoke unmounted callback and omit incorrect mounted callback when there is a configuration reset
- Updated the examples to show up the proper state on the LED when the suspend ended

### DIFF
--- a/examples/device/hid_composite_freertos/src/main.c
+++ b/examples/device/hid_composite_freertos/src/main.c
@@ -176,7 +176,14 @@ void tud_suspend_cb(bool remote_wakeup_en)
 // Invoked when usb bus is resumed
 void tud_resume_cb(void)
 {
-  xTimerChangePeriod(blinky_tm, pdMS_TO_TICKS(BLINK_MOUNTED), 0);
+  if (tud_mounted())
+  {
+    xTimerChangePeriod(blinky_tm, pdMS_TO_TICKS(BLINK_MOUNTED), 0);
+  }
+  else
+  {
+    xTimerChangePeriod(blinky_tm, pdMS_TO_TICKS(BLINK_SUSPENDED), 0);
+  }
 }
 
 //--------------------------------------------------------------------+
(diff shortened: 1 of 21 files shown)


examples/device/hid_composite_freertos/src/main.c
```diff
@@ -176,7 +176,14 @@ void tud_suspend_cb(bool remote_wakeup_en)
...
 void tud_resume_cb(void)
 {
-  xTimerChangePeriod(blinky_tm, pdMS_TO_TICKS(BLINK_MOUNTED), 0);
+  if (tud_mounted())
+  {
+    xTimerChangePeriod(blinky_tm, pdMS_TO_TICKS(BLINK_MOUNTED), 0);
+  }
+  else
+  {
+    xTimerChangePeriod(blinky_tm, pdMS_TO_TICKS(BLINK_SUSPENDED), 0);
```
hathach, line 185: should be `BLINK_NOT_MOUNTED`

src/device/usbd.c
```diff
@@ -965,7 +965,14 @@ static bool process_set_config(uint8_t rhport, uint8_t cfg_num)
...
 
   // invoke callback
-  if (tud_mount_cb) tud_mount_cb();
+  if (cfg_num)
+  {
+    if (tud_mount_cb) tud_mount_cb();
+  }
+  else
+  {
+    if (tud_umount_cb) tud_umount_cb();
```
hathach, line 974: it is not reachable, since cfg_num is check before process_set_config() is called. 

https://github.com/hathach/tinyusb/blob/master/src/device/usbd.c#L714

This check should be moved there.

hathach (commented):
thank you for you PR. changes look good, though there is 2 changes we should make
- correct typo for BLINK_NOT_MOUNTED
- move the umount callback to the correct place (right now it is not reachable)

hathach (approved):
perfect now, thank you for your PR.

PS: CI failed due to multiple hardware avaialble in the self-hosted machine. Will fix that soon and will re-run and merge later.

PR #2205, opened 2023-08-04
fix(usbh): robustness against unplugging usb device when it is busy

abakosh:
Describe the PR
Host Library:
The unplugging process of the USB device might lead to an infinite loop. It is easy to run into this situation if unplugging occurs during the enumeration process. Also, it happens when unplugging while the host and the USB device are doing something.

The issue is really big. As the stack enters an infinite loop when this is happening.

Additional context
I am testing with the RA4M2 family, with no OS. I know this fix is not meant if you are hosting a hub device. However, if you are planning to host a single device, it increases the robustness of plugging/unplugging while busy issue.
This was tested with MSC devices, I can see the improvement.

@hathach I am not sure if this fix is contagious for other scenarios. Please let me know if I need to take care of something else.

Thank you.

### DIFF
--- a/src/host/usbh.c
+++ b/src/host/usbh.c
@@ -889,9 +889,20 @@ TU_ATTR_FAST_FUNC void hcd_event_handler(hcd_event_t const* event, bool in_isr)
 {
   switch (event->event_id)
   {
+#if CFG_TUH_HUB
 //    case HCD_EVENT_DEVICE_REMOVE:
 //      // mark device as removing to prevent further xfer before the event is processed in usbh task
 //      break;
+#else
+    /* In case of a single device attached we can clear anything in the queue safely */
+    /* Reset the enumaration process just in case of immature plugging/unplugging */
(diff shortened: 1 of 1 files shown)


mdevaev:
@hathach @abakosh Sup. I also ran into this problem, but on RP2040. Do I understand correctly that this hack will help me solve this problem on RP2040, if I don't use the hub? Do you see any pitfalls with this?

mdevaev:
@abakosh It seems it's working great in my case. I came across this today and came to almost the same solution as you. As far as I understand, the RP2040 has one port, so everything should be fine.

karlp:
Potentiall addressed by the behaviour cahnge in https://github.com/hathach/tinyusb/pull/2550

hathach:
this can be closed, since debouncing is done in https://github.com/hathach/tinyusb/pull/3104 . feel free to re-open, or file another issue should it still occurs.

PR #2208, opened 2023-08-06
Fix makefile a warning when run on windows command prompt

kkitayam:
**Describe the PR**
Because Windows command prompt has no `tr` command, it occurs a warning.

**Additional context**
I found some ways in stack overflow. 
https://stackoverflow.com/questions/664601/in-gnu-make-how-do-i-convert-a-variable-to-lower-case
For portability, I guess the simple way is better.


hathach (approved):
thank you, this is a bit lengthy is portable.

Issue #2194, opened 2023-07-31
Enhance HITL (hardware in the loop) test


HiFiPhile:
### Related area

HITL test

### Hardware specification

DCD only for the moment

### Is your feature request related to a problem?

I'm going to add more boards into the loop, currently planned:
- LPCXpresso LPC54608 OM13092  (HS port)
- STM32F746DISCO (both HS & FS port)
- NUCLEO-L412KB (like now)

Meanwhile maybe a good idea to wait #1970 to avoid mixing boards.

I'm also open to sponsorship if anyone wants other boards to be added.

PS: Made a tiny hub as I can't find one :)
![IMG_20230731_194054](https://github.com/hathach/tinyusb/assets/4375114/d045a8e0-5c56-4f49-a482-9705b95897a8)

### Describe the solution you'd like

Add more boards into HITL setup.

### I have checked existing issues, dicussion and documentation

- [X] I confirm I have checked existing issues, dicussion and documentation.

HiFiPhile:
> We will probably need to spend more time to write more actual hardware testing script

Yes I'm also thinking about it.

>  what's up with the hub breakout, wouldn't it look nicer with consumer hub :)

For space saving, with male plug directly soldered on the hub I don't need to deal with cable mess ;)

tannewt:
Here are a couple hardware project for HITL. Thanks @esden for the pointers! https://github.com/probe-rs/hive https://github.com/esden/redcuttle

hathach:
> Here are a couple hardware project for HITL. Thanks @esden for the pointers! https://github.com/probe-rs/hive https://github.com/esden/redcuttle

thanks Scott, it looks very neat, but probably overkilled I guess. I would prefer to just stick the dev board in, most of them has on-board debugger anyway for flashing. It is of course just me :D

hathach:
#2204 implement unique ID for several families including: rp2040 and various stm32 (including l4 and f7)

HiFiPhile:
I've connected the boards.

3 debuggers are flashed to Jlink:
|SN|Board|Device|
|-|-|-|
|770935966|STM32F746DISCO|STM32F746NG|
|774470029|NUCLEO-L412KB|STM32L412KB|
|727600775|OM13092 | LPC54608J512|

Both STM32L4 & STM32F7 works, while only HS port works on OM13092. (added missing `-DBOARD_TUD_RHPORT=$(PORT)`)

[...]

PR #2209, opened 2023-08-06
Enhance HITL test

HiFiPhile:
**Describe the PR**
Add HITL testing script for newly added boards #2194

Currently support `cdc_dual_ports`, `cdc_msc`, `dfu`, `dfu_runtime`, more tests will be added later on.

For my instance use `python test/hitl/hitl_test.py hitl_hfp.json`

I'm not familiar with Github, @hathach could you take a look for CI integration ?

**Additional context**

Actually it catched something !
My lpcxpresso54608 (even modified 54628 to 54608 in board.mk) failed with cdc_msc test, msc block doesn't show up : 
```
[242341.797154] usb 4-1.7: new high-speed USB device number 89 using xhci_hcd
[242341.901761] usb 4-1.7: New USB device found, idVendor=cafe, idProduct=4003, bcdDevice= 1.00
[242341.901765] usb 4-1.7: New USB device strings: Mfr=1, Product=2, SerialNumber=3
[242341.901767] usb 4-1.7: Product: TinyUSB Device
[242341.901768] usb 4-1.7: Manufacturer: TinyUSB
[242341.901768] usb 4-1.7: SerialNumber: 0123456789ABCDEF
[242341.904327] cdc_acm 4-1.7:1.0: ttyACM2: USB ACM device
[242341.904686] usb-storage 4-1.7:1.2: USB Mass Storage device detected
[242341.904901] scsi host7: usb-storage 4-1.7:1.2
[242343.001409] usb 4-1.7: reset high-speed USB device number 89 using xhci_hcd
[242343.102100] cdc_acm 4-1.7:1.0: ttyACM2: USB ACM device
[242343.205356] usb 4-1.7: reset high-speed USB device number 89 using xhci_hcd
[242343.306000] cdc_acm 4-1.7:1.0: ttyACM2: USB ACM device
[242343.401299] usb 4-1.7: reset high-speed USB device number 89 using xhci_hcd
[242343.505975] cdc_acm 4-1.7:1.0: ttyACM2: USB ACM device
[242343.605305] usb 4-1.7: reset high-speed USB device number 89 using xhci_hcd
[242343.705934] cdc_acm 4-1.7:1.0: ttyACM2: USB ACM device
```
Log: 
[msc.zip](https://github.com/hathach/tinyusb/files/12269947/msc.zip)


hathach:
@HiFiPhile yeah, I notice MSC is not showed up on your VM. I thought it is due your VM not enable auto mount block device, I remembered we could find it /dev/sda. I guess we have to manual mount ? I will check out the PR later.

HiFiPhile:
>  I thought it is due your VM not enable auto mount block device

It's true. I set udev alias, it will show up with something like `/dev/blkUSB_57323020.02`

HiFiPhile:
Now it read README.TXT from mounted device.

hathach:
> Now it read README.TXT from mounted device.

thank you perfect, I will check this out soon enough. Though one thing pop up in my mind. I think at the end of the CI. We should flash board_test example. This will disable the usb on the tested board. Making testing the next board easier, at least it will more forgiving in our test script.

hathach:
lpc54 has an issue with receiving 31 bytes for msc command and sending 13 for msc status. Somehow when the transfer is complete, it is only 30 bytes (1 residue) and 12 bytes. Seem like the interrupt triggered a bit early or something. I have spent something troubleshoot, but not quite sure what is wrong. We probably need to skip the lpc54 for now, lpc55 has no issue with this though, strange.

HiFiPhile:
>  We probably need to skip the lpc54 for now

Yes we can fix it later.

hathach:
@HiFiPhile everything looks great and the python test script is way better than my shell script and easier to work with. Though I think we may have to modify it a bit to work better with ci. I will try to wrap this up tomorrow.

HiFiPhile:
I was looking around for lpc54 issue and found there are 17 USB issues in errata sheet, that's a lot...

Maybe it's related:
```
* USB.1: In USB high-speed device mode, the NBytes field does not decrement after BULK OUT transfer
Introduction:
The LPC546xx. device family includes a USB high-speed interface (USB1) that can
operate in device mode at high-speed. The NBytes value represents the number of bytes
that can be received in the buffer.
* Problem:
If the buffer length is less than the maximum packet size and if the application code does
not program the maximum packet size, the NByte value is not correct.
* Work-around:
Program the NByte to the maximum packet size of that particular endpoint type. The
application code must calculate the received number of bytes by subtracting the NByte
value from the programming value. The software work-around is implemented on the SDK
software platform for the LPC546xx device family.
```

hathach:
yeah, thank for the info, I should have checked errata. As tested, I also saw it in the bulk in as well (12 instead 13 bytes for msc status) which is more severe as there is no walkaround.

image

A quick look around, and there is tons of errata, tx data corruption 3.22 is also severe

image

I could add walkarond for bulk in/out so that it is kind of usable with stock examples. but I don't know if we have time to go through all of errata, look like usb on lpc54 is a headache one. I would definitely avoid lpc54 for a production with usb.

test/hitl/hitl_test.py
```diff
@@ -0,0 +1,208 @@
...
+    ser2.write(str2)
+    ser2.flush()
+    assert ser1.read(100) == str2.lower(), 'Port1 wrong data'
+    assert ser2.read(100) == str2.upper(), 'Port2 wrong data'
+
+    print('cdc_dual_ports test done')
+
+def test_cdc_msc(id):
+    port  = f'/dev/ttyUSB_{id[-8:]}.00'
+    file = f'/media/blkUSB_{id[-8:]}.02/README.TXT'
```
hathach, line 82: @HiFiPhile why would you limit the serial number to 8 characters, wouldn't be simpler to use it as it is ?

HiFiPhile: For some historical reason my udev script cut serial number to 8 characters.

In fact udev rule is not even needed, cdc can be accessed by `/dev/serial/by-id/usb-TinyUSB_TinyUSB_Device_41003B000E504E5457323020-if00`

hathach: yeah, I figure that out as well. I have already made the changes, just want double check if you first. The id is also shorten in blk device, if it only impact serial, then we are probably safe to remove the cut ?

```python
def get_serial_dev(id, product, ifnum):
    # get usb serial by id
    return f'/dev/serial/by-id/usb-TinyUSB_{product}_{id}-if{ifnum:02d}'


def get_disk_dev(id, lun):
    # get usb disk by id
    return f'/dev/disk/by-id/usb-TinyUSB_Mass_Storage_{id}-0:{lun}'


def get_hid_dev(id, product, event):
    return f'/dev/input/by-id/usb-TinyUSB_{product}_{id}-{event}'

```

HiFiPhile: > The id is also shorten in blk device, if it only impact serial, then we are probably safe to remove the cut ?

Yes you are right.

hathach: since I am in the middle of this, let me try to do this as well. At worst, we could revert if I messed up :)

.github/workflows/build_iar.yml
```diff
@@ -49,82 +49,85 @@ jobs:
...
-          find device/ -name "*.elf" -exec mv {} ../../ \;
-
-      - name: Upload Artifacts for stm32l412nucleo
-        if: contains(matrix.family, 'stm32l4') && github.repository_owner == 'hathach'
-        uses: actions/upload-artifact@v3
-        with:
-          name: stm32l4
-          path: |
-            *.elf
+          python3 test/hil/hil_test.py hil_hfp.json
```
hathach, line 54: this is where the script is run, CI compile with IAR using cmake build in previous step. We could pick up the elf. This will conveniently test IAR for us as well. (normally I only test with gcc)

hathach:
before I forgot here is the lpc54 setting, should we want to re-add it later (couldn't find way to comment out in json)

```json
        {
            "name": "lpcxpresso54608",
            "uid": "0123456789ABCDEF",
            "debugger": "jlink",
            "debugger_sn": "727600775",
            "cpu": "LPC54608J512",
            "tests_skip": [
                "cdc_msc"
            ]
        }
````

hathach:
> Perfect it's good for me, maybe also let it run for GCC build.

yeah, that is possible as well, though I think we are good enough for now. Since we do generally test with gcc when developement. IAR is much rarer, so I am glad it is covered by HIL. Thank you very much for work and VM/board farm for doing actual test.

PR #2213, opened 2023-08-08
Adding support for a generic SOF callback

Rocky04:
- Added support for a simple general callback when a new frame or micro frame started

Can be used to synchronize the scanning to the polling for HID devices or other device functions. Not useful for time critical things because the time which has passed between the interrupt and the execution is unknown. _For that it would be necessary to save the timestamp when the SOF interrupt handling is done in the ISR._

Proper example implementation of the PR #2180.

_Edit:_
A draft for a vendor specific SOF callback which is executed in the ISR via `dcd_event_handler()` can be found here: #2220


Rocky04:
I'm unsure if the device port should be part of the callback or not. _I started the PR with it but removed it for now._

hathach (dismissed):
sof must be explicitly enable/disable with additional API

HiFiPhile:
I've move sof status into usbd_device_t to ensure it's cleared on bus reset or configuration change.

Currently audio driver could call sof enable/disable different times, will fix later.

HiFiPhile (approved):

src/device/usbd.c
```diff
@@ -1355,12 +1379,21 @@ void usbd_edpt_close(uint8_t rhport, uint8_t ep_addr) {
...
-  dcd_sof_enable(rhport, en);
+  uint8_t consumer_old = _usbd_dev.sof_consumer;
+  // Keep track how many class instances need the SOF interrupt
+  if (en) {
+    _usbd_dev.sof_consumer |= (uint8_t)(1 << consumer);
+  } else {
+    _usbd_dev.sof_consumer &= (uint8_t)(~(1 << consumer));
+  }
+
+  // Test logically unequal
```
Rocky04, line 1393: Seems like a typo, should be: `// Test logically XOR`

src/device/usbd.c
```diff
@@ -56,6 +56,10 @@ TU_ATTR_WEAK void tud_event_hook_cb(uint8_t rhport, uint32_t eventid, bool in_is
   (void)in_isr;
 }
 
+TU_ATTR_WEAK void tud_sof_cb(uint32_t frame_count) {
+  (void)frame_count;
+}
```
Rocky04, line 61: Why not have a weak declaration in the header file? After all it's used only once at L629 anyway. So a check there for a definition similar how it's done for other things should be more inline.

src/device/usbd.c
```diff
@@ -382,6 +388,12 @@ bool tud_connect(void) {
   return true;
 }
 
+bool tud_sof_cb_enable(bool en)
+{
+  usbd_sof_enable(_usbd_rhport, SOF_CONSUMER_USER, en);
+  return true;
```
Rocky04, line 394: A return value makes no sense here.

While also true for my original PR that wasn't as obvious. I wasn't sure if every driver has implemented `dcd_sof_enable()` and so used `TU_VERIFY()` to check for that.

src/device/usbd.c
HiFiPhile: Not a typo, it could be other consumer later, should be only called when whole value change to true or false.

src/device/usbd.c
HiFiPhile: We are converting to weak default implementation for keil and clang compatibility.

src/device/usbd.c
HiFiPhile: You are right, forgot to change.

src/device/usbd.c
Rocky04: I can see where you coming from. _Keep in mind that my comment has nothing to do with the usage and is only related to the code readability._

While it's being the same and because you are actually using the unequally operator in the expression for logical states, I don't really see an additional value in this comment. Meaning if I didn't knew what the expression would do the comment wouldn't help me at all. Providing a different view, like to mention that this is basically an expression which is equivalent to the XOR gate would. It's the same if this would be used instead: `((!_usbd_dev.sof_consumer && consumer_old) || (_usbd_dev.sof_consumer && !consumer_old))` A different implementation for the same logic where adding a comment like "// Either one but not both" would also not add anything.

PR #2217, opened 2023-08-14
Add lpc43s67

hathach:
**Describe the PR**
Update lpc43 to support adding it to HIL test board farm
- Add lpc43s67 board
- Also add add board_get_unique_id() for lpc43


HiFiPhile:
I can confirm it works well, here is the config diff if you want to include in the pr.

[conf.diff.zip](https://github.com/hathach/tinyusb/files/12333845/conf.diff.zip)

hathach:
> I can confirm it works well, here is the config diff if you want to include in the pr.
> 
> [conf.diff.zip](https://github.com/hathach/tinyusb/files/12333845/conf.diff.zip)

Thank you, I update PR to add IAR build support for lpc43 (via cmake) and added your json entry. Everything works perfect now.

PR #2218, opened 2023-08-14
Fix dcd synopsys flush fifo

cy18:
**Describe the PR**
In dcd_synopsys.c, the FIFO is not flushed after configuration. This causes some problems when I'm porting TinyUSB to my GD32F470 board. Since I don't have a STM32 board at hand, I'm not sure if it's necessary on STM32 boards.


HiFiPhile:
Sorry for late response, synopsys dwc2 driver has been moved to `src/portable/synopsys/dwc2/dcd_dwc2.c`, since #2333 FIFO size is fixed at connection and no dynamic adjustment is done. Please test if it works.

hathach:
right, I think this should not be an issue with latest dwc2 driver, feel free to re-open or file an issue should you still have problem with this.

PR #2219, opened 2023-08-14
min on cdc r/w

tswan-quasi:
**Describe the PR**
Adds a MIN() check for the `bufsize` on cdc reads and writes. This is to avoid a loop where the `bufsize` passed to `tud_cdc_write()` or `tud_cdc_read()` is `65536` (or a multiple) causing nothing to be written/read.

**Additional context**
Casting values to `uint16_t` greater than `UINT16_MAX` just truncates the upper bits, which means if `bufsize` ends up being `65536` (or a multiple) it results in the `uint16_t` cast giving a value of `0`, causing no read/write to occur. This then ultimately returns `0` from the top level function, and if the top level application is relying on the return value of something such as `tud_cdc_write()`, then the program will get stuck sending nothing.

### DIFF
--- a/src/class/cdc/cdc_device.c
+++ b/src/class/cdc/cdc_device.c
@@ -150,7 +150,7 @@ uint32_t tud_cdc_n_available(uint8_t itf)
 uint32_t tud_cdc_n_read(uint8_t itf, void* buffer, uint32_t bufsize)
 {
   cdcd_interface_t* p_cdc = &_cdcd_itf[itf];
-  uint32_t num_read = tu_fifo_read_n(&p_cdc->rx_ff, buffer, (uint16_t) bufsize);
+  uint32_t num_read = tu_fifo_read_n(&p_cdc->rx_ff, buffer, (uint16_t) TU_MIN(bufsize, 65535));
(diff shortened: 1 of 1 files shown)


hathach (commented):
look good, would you mind changing TU_MIN() to tu_min16() i.e

```C
  uint32_t num_read = tu_fifo_read_n(&p_cdc->rx_ff, buffer, 
 tu_min16(bufsize, UINT16_MAX));
```

tswan-quasi:
> look good, would you mind changing TU_MIN() to tu_min16() i.e
> 
> ```c
>   uint32_t num_read = tu_fifo_read_n(&p_cdc->rx_ff, buffer, tu_min16(bufsize, UINT16_MAX));
> ```

I think that would result in the same issue right? Since arg is type `uint16_t` would just cast `bufsize` and do the same thing

hathach:
> I think that would result in the same issue right? Since arg is type `uint16_t` would just cast `bufsize` and do the same thing

Ah you are right, my bad :)

hathach (approved):
thank you, we can use tu_min32, but i guess it is not really matter in this case.

PR #2220, opened 2023-08-14
Adding support for custom SOF ISR for CFG_TUD_VENDOR

Rocky04:
- Opens the driver SOF which is executed in the ISR to be changed by the vendor if the CFG_TUD_VENDOR configuration is used as a device driver

_A draft for a generic SOF callback which is executed in the main loop via `tud_task()` can be found here: #2213_

### DIFF
--- a/src/device/usbd.c
+++ b/src/device/usbd.c
@@ -167,7 +167,7 @@ tu_static usbd_class_driver_t const _usbd_driver[] =
     .open             = vendord_open,
     .control_xfer_cb  = tud_vendor_control_xfer_cb,
     .xfer_cb          = vendord_xfer_cb,
-    .sof              = tud_vendor_sof_isr ? tud_vendor_sof_isr : NULL
+    .sof              = tud_vendor_sof_isr
   },
   #endif
 
--- a/src/device/usbd.c
+++ b/src/device/usbd.c
@@ -167,7 +167,7 @@ tu_static usbd_class_driver_t const _usbd_driver[] =
     .open             = vendord_open,
     .control_xfer_cb  = tud_vendor_control_xfer_cb,
     .xfer_cb          = vendord_xfer_cb,
-    .sof              = NULL
+    .sof              = tud_vendor_sof_isr ? tud_vendor_sof_isr : NULL
   },
   #endif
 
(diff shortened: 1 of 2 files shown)


Rocky04:
How is a vendor driver not managed by usbd? I don't get it. If that is the case it makes absolutely no sense for me that there is a `CFG_TUD_VENDOR` driver.

Rocky04:
Again, yes the control of that is missing. The code I use so far is much older and rely on a symbol (`USE_SOF`) where it's always on or off. Indeed a dynamic control of that is the preferred way and for this reason I'm currently adding it to the PR #2213.

It also doesn't matter if it hurts performance when it's needed. Of course it should be off when it's not currently needed. What's the preferred way if an implementation wants to have it's own audio driver with endpoint control? I would imagine in that case the vendor driver should be used instead of altering the audio driver code of TinyUSB. If that is needed for audio why not for video? There can be just video and in a real time stream audio and video even needs to be synchronized.

I'm still irritated why you think a vendor driver isn't controlled by usbd.

hathach:
> I'm still irritated why you think a vendor driver isn't controlled by usbd.

English is not my first language, and I don't like to read and type much text. vendor sof is not auto managed by usdb, application turn it on and off by itself and stay permanently after that. Audio sof is only enabled when needed in run time by usbd.

Rocky04:
_English is not my first langue either..._

At first I thought I know how a vendor defined device should look like. Due to your feedback I'm not sure anymore. :worried:

I was expecting that each device should have at least one standard class from TinyUSB defined in the `tusb_config.h` config file. So if no predefined class matches the use case the vendor class should be used instead. But if this isn't the case I don't get why there is a vendor class to begin with if that should not be used for a custom implementation. 😕 

So if the internal classes are off the table, the driver can only be loaded by an application driver via `usbd_app_driver_get_cb()`. But to me it seems that this isn't fully implemented yet. The value for `_app_driver_count` is set to zero and `_app_driver` is set to NULL while where isn't anything which updates these. I would expect that usbd is owned by TinyUSB and so that a normal implementation should not alter the code there. Meaning I would assume TinyUSB can be used for a vendor defined application without needing to modifying it's core.

Either way, sadly I haven't found a single example in the framework how the internal vendor class can be used or how to implement a custom application driver. **So I would really appreciative if you can show me how TinyUSB is supposed to be used for a custom vendor define implementation which don't uses any of the common standard classes of TinyUSB.**

hathach:
- webusb example use vendor class
- if there is any double, just check the code.

Rocky04:
Ok, thank you.... ❤️   That helped to understand how you want it to be used.

PR #2239, opened 2023-08-29
Fixes #1018 the weak dcd_edpt0_status_complete for Keil Compiler

### LINKED ISSUE #1018: stm32_fsdev builded by keil with --gnu option - enumeration fails
### Operating System

Windows 10

### Board

custom board with stm32f103c8t6

### Firmware

examples/device/cdc_msc, but adopted manually for keil project with --gnu compiler option

### What happened ?

When project is builded by keil armcc compiler, enumeration fails. The reason is a weak attribute in dcd_edpt0_status_complete function definition (dcd.h, string 135). The armcc compiler ignores function implementation in dcd_stm32_fsdev.c, and because of that device address is never actually set.
I dont know, if it's necessary to fix it, but I think that this issue will be useful for other keil users anyway (I spent a week looking for a reason)

### How to reproduce ?

1. Add tinyusb into stm32f103c8t6 keil project with --gnu compiler option
2. Build and run

### Debug Log

_No response_

### Screenshots

_No response_

XelaRellum:
The Keil compiler seems to have different semantics and the defined function was never called.

The same is probably true for the other weak functions. I can change those too.

### DIFF
--- a/src/device/dcd.h
+++ b/src/device/dcd.h
@@ -155,7 +155,7 @@ void dcd_sof_enable(uint8_t rhport, bool en);
 
 // Invoked when a control transfer's status stage is complete.
 // May help DCD to prepare for next control transfer, this API is optional.
-void dcd_edpt0_status_complete(uint8_t rhport, tusb_control_request_t const * request) TU_ATTR_WEAK;
+void dcd_edpt0_status_complete(uint8_t rhport, tusb_control_request_t const * request);
 
 // Configure endpoint's registers according to descriptor
 bool dcd_edpt_open            (uint8_t rhport, tusb_desc_endpoint_t const * desc_ep);
--- a/src/device/usbd_control.c
+++ b/src/device/usbd_control.c
@@ -135,6 +135,15 @@ bool tud_control_xfer(uint8_t rhport, tusb_control_request_t const * request, vo
 // USBD API
 //--------------------------------------------------------------------+
 
+TU_ATTR_WEAK void dcd_edpt0_status_complete(uint8_t rhport, tusb_control_request_t const * request)
+{
+  (void)rhport;
+  (void)request;
+
+  // this is the default implementation that is called when no "real" implementation
+  // of the function exists
+}
+
 void usbd_control_reset(void);
 void usbd_control_set_request(tusb_control_request_t const *request);
 void usbd_control_set_complete_callback( usbd_control_xfer_cb_t fp );
@@ -173,7 +182,7 @@ bool usbd_control_xfer_cb (uint8_t rhport, uint8_t ep_addr, xfer_result_t result
     TU_ASSERT(0 == xferred_bytes);
 
     // invoke optional dcd hook if available
-    if (dcd_edpt0_status_complete) dcd_edpt0_status_complete(rhport, &_ctrl_xfer.request);
+    dcd_edpt0_status_complete(rhport, &_ctrl_xfer.request);
 
     if (_ctrl_xfer.complete_cb)
     {


hathach (commented):
Thank you for the Pr. Do you check with official doc from keil for this behavior for declared weak function ? I don't have keil compiler to test with.

XelaRellum:
Yes and I have tried this in the code, with Keil and with STM32CubeIDE GCC (since I needed this in my project right now)...

Am 30. August 2023 05:06:15 MESZ schrieb Ha Thach ***@***.***>:

hathach:
can you provide link to keil's official doc for the behavior ?

XelaRellum:
> can you please provide link to keil's official doc for the behavior ?

As usual, the documentation is a bit cryptic:

https://developer.arm.com/documentation/101754/0620/armclang-Reference/Compiler-specific-Function--Variable--and-Type-Attributes/--attribute----weak---function-attribute

Also on the same topic:

https://stackoverflow.com/questions/51656838/attribute-weak-and-static-libraries

In my understanding the problem is the difference of the weak declaration vs. the weak definition of the function.

With the strong declaration, but weak definition of a default/empty function the linker will enforce the symbol to exist, so the test always evaluates to true. But a strong definition of a function will override a weak definition.

The use of a weak declaration seems to be the trick part and which doesn't work here. Although the function in dcd_stm32_fsdev.c gets compiled in the same project, no actual code is generated and the symbol is removed. I don't really get why ARMCC does this, but it does.

ST uses weak definitions for their HAL libraries like here:

/**
  * @brief  SYSTICK callback.
  * @retval None
  */
__weak void HAL_SYSTICK_Callback(void)
{
  /* NOTE : This function should not be modified, when the callback is 
            needed, the HAL_SYSTICK_Callback could be implemented in the user file
   */
}

So, if I want my own implementation, I define it in some of my application's files.

Their solution works with KEIL and GCC and IAR and whatnot (all supported compilers of ST code).

XelaRellum:
Right now I am progressing further with the project and I now realize the scope of the problem is bigger. My first thought was, the function above is the only one that is really needed for an implementation. But now I have come to realize TU_ATTR_WEAK functions are all over the place (well, at least several of them).

So, the right thing to do would be to change the semantics of all the callbacks and provide a default implementation for any callbacks that are optional. The return value of the function should be the indicator if it is a default implementation (like a NULL pointer).

I will happily work on this topic, if you are okay with that. One question: where should all the default implementations go? My first guess would be the module that uses/calls it, if it is called only from one location.

What about functions like ```board_init_after_tusb```? There is no generic module for them? So maybe put those into "tusb.c"?

hathach:
Your link is pointed to armclang (which is different from keil armcc). From keil doc, look like it should work well with current code https://developer.arm.com/documentation/dui0375/g/Compiler-specific-Features/--weak

```
__weak
This keyword instructs the compiler to export symbols weakly.

The __weak keyword can be applied to function and variable declarations, and to function definitions.

Usage
Functions and variable declarations
For declarations, this storage class specifies an extern object declaration that, even if not present, does not cause the linker to fault an unresolved reference.

For example:


__weak void f(void);
...
f(); // call f weakly
If the reference to a missing weak function is made from code that compiles to a branch or branch link instruction, then either:

The reference is resolved as branching to the next instruction. This effectively makes the branch a NOP.

The branch is replaced by a NOP instruction.
```

KlemenDEV:
I can confirm this patch fixes operation on Keil

KlemenDEV:
I think this also affects other weak callbacks, eg. tud_cdc_line_state_cb or tud_cdc_rx_cb don't work either in Keil if not modified in a similar way as this PR does

XelaRellum:
> I think this also affects other weak callbacks, eg. tud_cdc_line_state_cb or tud_cdc_rx_cb don't work either in Keil if not modified in a similar way as this PR does

Yes true. Didn't want to change stuff that I don't use and thus can't test. But should be straight forward.,

hathach (approved):
look good, even though there is tons of other weak functions. We can migrate/update them later on so that it would work with keil better.

PR #2245, opened 2023-09-06
Fix msd extraneus bytes on reading

maidnl:
Fix the fact that MSD status msg sometimes goes in the payload of data read from disk

This fixes: https://github.com/arduino/ArduinoCore-renesas/issues/84
The problem: when issuing a reading command on MSD class, sometimes data read form disk (QSPI flash device) are corrupted with status message (the one that should "end" the transmission of data read from disk, i.e. "USBS..."). According to our analysis this is due to the fact that given that MSC_STAGE_STATUS case is executed always "after" the state machine implemented in `mscd_xfer_cb` function, if data are still present in the hw pipe fifo while the status message is sent by the function,  the status message overwrites part of the data present in the hw pipe fifo.
The solution: a function that checks the status of the hw pipe fifo has been implemented. This is a weak function implemented only on RA hw. This function waits until the hw pipe fifo is empty. The function is then called in `mscd_xfer_cb` function when the status message is about to be sent while a reading from disk is finishing. This ensure that data in hw pipe fifo are not overwritten due to lack of sync.

If you have a better solution or a more general approach, please feel free to provide us with your implementation, we will test this on our device. Thanks!

@facchinm

### DIFF
--- a/src/portable/renesas/rusb2/dcd_rusb2.c
+++ b/src/portable/renesas/rusb2/dcd_rusb2.c
@@ -267,9 +267,10 @@ static void pipe_read_packet_ff(rusb2_reg_t * rusb, tu_fifo_t *f, volatile void 
 }
 
 
-static void wait_pipe_fifo_empty(rusb2_reg_t* rusb, uint8_t num) {
+static bool wait_pipe_fifo_empty(rusb2_reg_t* rusb, uint8_t num) {
   TU_ASSERT(num);
   while( (rusb->PIPE_CTR[num-1] & RUSB2_PIPE_CTR_INBUFM_Msk) > 0 ) {}
+  return true;
 }
 
 
--- a/src/portable/renesas/rusb2/dcd_rusb2.c
+++ b/src/portable/renesas/rusb2/dcd_rusb2.c
@@ -266,6 +266,13 @@ static void pipe_read_packet_ff(rusb2_reg_t * rusb, tu_fifo_t *f, volatile void 
   tu_fifo_advance_write_pointer(f, count);
 }
 
+
+static void wait_pipe_fifo_empty(rusb2_reg_t* rusb, uint8_t num) {
+  TU_ASSERT(num);
+  while( (rusb->PIPE_CTR[num-1] & RUSB2_PIPE_CTR_INBUFM_Msk) > 0 ) {}
+}
+
+
 //--------------------------------------------------------------------+
 // Pipe Transfer
 //--------------------------------------------------------------------+
@@ -339,6 +346,7 @@ static bool pipe_xfer_in(rusb2_reg_t* rusb, unsigned num)
   const unsigned rem  = pipe->remaining;
 
   if (!rem) {
+    wait_pipe_fifo_empty(rusb, num);
     pipe->buf = NULL;
     return true;
   }


kkitayam:
I guess that the following place is good to wait for the HW pipe FIFO to be empty. At the last packet of an IN transaction, it will wait for the HW pipe to be empty.

https://github.com/hathach/tinyusb/blob/7bf5923052e5861f54c9cb0581e328f8be26a0a9/src/portable/renesas/rusb2/dcd_rusb2.c#L343

> if data are still present in the hw pipe fifo while the status message is sent by the function, the status message overwrites part of the data present in the hw pipe fifo.

For efficient handling double buffer, I had designed `pipe_xfer_in()` returns true(It escalates to mscd_xfer_cb()) when there is at least one empty buffer in HW pipe FIFO. In case of bulk transfer, HW pipe is set double buffer mode, so pipe_xfer_in() will return true when one buffer is empty, even if there is data in the other buffer.
Therefore I guess the status message will not overwrite a part of previous data.

Any way, if we need to wait for HW pipe FIFO to be empty, I suggest the above place, to keep MSC driver API interface.

maidnl:
I made the changes you suggested and tested them on our device. Now the wait function is static in the RA file dcd_rusb2.c and used in the point you mentioned in the comment. The MSD result untouched in this way and all the changes are only for renesas. 

In case you prefer some changes to this to better fit the overall architecture please let me know and I will test them on our device. Thanks.

kkitayam (approved):

src/portable/renesas/rusb2/dcd_rusb2.c
```diff
@@ -266,6 +266,13 @@ static void pipe_read_packet_ff(rusb2_reg_t * rusb, tu_fifo_t *f, volatile void
   tu_fifo_advance_write_pointer(f, count);
 }
 
+
+static void wait_pipe_fifo_empty(rusb2_reg_t* rusb, uint8_t num) {
```
kkitayam, line 270: Because of `TU_ASSERT` returns `bool` value, this function's return type should be `bool`. 
Please find the build error message.
https://github.com/hathach/tinyusb/actions/runs/6119698047/job/16610133014?pr=2245#step:10:45

kkitayam:
Thank you for changing and testing.

kkitayam (approved):

hathach (approved):
superb! thank you @maidnl for the fix and @kkitayam for suggestion and reviewing.

Rbb666:
Hello, I verified on the RA series MCU that when the msc and cdc devices are turned on, this modification will cause a usb enumeration time of about 20 seconds.
image

hathach:
thanks @Rbb666, I just pull out my ram41 to test with, this PR indeed causes hanged (block wait) with msc_cdc example. This should be reverted, since the solution does not work as expected and seems to cause more serious issue.

Rbb666:
> 谢谢，我只是拿出我的 ram41 进行测试，这个 PR 确实会导致挂起（块等待），举个例子msc_cdc。这应该被还原，因为解决方案没有按预期工作，并且似乎会导致更严重的问题。

Yes, this is a problem I detected on Renesas ra8d1. I hope it can be fixed as soon as possible.

hathach:
already reverted by #2587

PR #2248, opened 2023-09-08
renesas: support micros with different number of HW pipes

maidnl:
renesas: add support for renesas micro with different number of USB hw pipes

not all renesas micro have the same number of pipes as expected by the find_pipe function. 
For example see RA4M1 micro which only have 2 bulk pipes and 2 interrupt pipes. 
For this reason some defines were introduced so that the actual number and positions of USB HW pipes can be defined at build time for different kind of processor.

If you prefer a different approach please let me know. I will try to implement and test that on our arduino devices that uses renesas micro. Thanks

@facchinm

### DIFF
--- a/src/portable/renesas/rusb2/dcd_rusb2.c
+++ b/src/portable/renesas/rusb2/dcd_rusb2.c
@@ -89,6 +89,24 @@ static dcd_data_t _dcd;
 // INTERNAL OBJECT & FUNCTION DECLARATION
 //--------------------------------------------------------------------+
 
+#ifndef FIRST_BULK_PIPE
+#define FIRST_BULK_PIPE       3
+#endif
+
+#ifndef FIRST_INTERRUPT_PIPE
+#define FIRST_INTERRUPT_PIPE  6
+#endif
+
+#ifndef NUMBER_OF_PIPES
+#define NUMBER_OF_PIPES       10
+#endif
+
+#ifdef NO_ISOCRONOUS_PIPE
+  #define FIRST_PIPE            FIRST_BULK_PIPE
+#else
+  #define FIRST_PIPE            1
+#endif
+
 // Transfer conditions specifiable for each pipe:
 // - Pipe 0: Control transfer with 64-byte single buffer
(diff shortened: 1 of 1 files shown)


hathach:
sorry, I was busy with other works, will pull out datasheet and check this out soon enough. Thank you for your patient.

hathach (commented):
I just checked 4m1 group hw manual, it has the same pipe numbering as other i.e
- pipe 1,2: ISO + Bulk
- pipe 3,4,5 bulk only
- pipe 6-9 interrupt

So I guess this PR isn't needed ? Btw, I think we should rename PIPE_1ST_BULK since it isn't correct (first bulk is 1 though). But we could do that later on.

image

maidnl:
I am so sorry, I wrote the wrong micro.
The micro with less PIPES is RA2A1 (not RA4M1). Really sorry, my mistake! 
I think this PR is necessary.
Thank you.

maidnl:
Hi! I am really sorry to bother you on this topic again. 
I would like to ask you if can reconsider the possibility to merge this PR. 
As I wrote in my previous comment I initially mentioned the wrong renesas microcontroller (again I'm sorry for that):
the renesas microcontroller with less pipes is RA2A1 (he really is :-). 
I think this PR is required to make possible the use of this renesas microcontroller and I hope you can reconsider it and merge it. This will allow us (Arduino) to switch our project on your mainline (we get your last upgrades easily, you get us to test them on a large base of different products). Thank you very much and I hope you can forgive my initial mistake.

hathach:
Sorry, was busy with other works. Will check this out as soon as I could

hathach (changes requested):
I just checked the manual for RA2A1, it indeed only support 5 pipes, oddly enough pipe0 then 4-7 (1-3 seems to be not usable). RA2A1 (and other similar mcus) should be detected e.g BSP_MCU_GROUP_RA2A1 and addressed by tinyusb without the need of using external MACRO as suggested change. 

I don't have RA2A1 hardware on hand, but I can try to add an bsp for it, and then make some adjumst to pipe allocation for it later on.

hathach:
closed since a better approach is implemented by #2588

PR #2251, opened 2023-09-11
Add max3421e host driver

hathach:
**Describe the PR**
- Add support driver for Analog MAX3421e usb host controller via SPI aka Arduino USB Host shield 
- tested on feather nrf52840 express, metro m0/m4 (samd21 samd51)
- To enable max3421 driver for host examples, `MAX3421_HOST=1` must be added to makefile and cmake build command e.g
  - make BOARD=feather_nrf52840_express MAX3421_HOST=1 all
  - cmake .. -DBOARD=feather_nrf52840_express -DMAX3421_HOST=1
- For ci purpose, max3421 host driver is always enabled for feather nrf52840 express board (to test compipling driver)
- also add cmake support for nrf
- Other clean up
  - Remove TU_LOG_PTR, rename TU_LOG_ARR to TU_LOG_BUF
  - add hcd_template.c
  - define tuh_int_handler/tud_int_handler to empty when corressponding stack not enabled.

TODO: more ports to be tested with later on: esp32


tannewt (commented):
How can we dynamically initialize this? In CP, we'll want the code but not the data structures until we know we need it. Maybe a struct with all needed buffers/data and function pointers to the SPI API?

hathach:
For now, you need to enable the host driver with `CFG_TUH_MAX3421=1`, this will also include driver internal data https://github.com/hathach/tinyusb/blob/master/src/portable/analog/max3421/hcd_max3421.c#L205. If you want to supply the internal data when needed we probably need some kind of malloc-like callback (malloc when init host stack, free when de-init). It will probably need its own PR since there is a few request for more dyanmic memory for class drivers (since not all enabled driveres are included in the descritpors)  

The driver requires 3 api implmented by application

```C
void tuh_max3421_spi_cs_api(uint8_t rhport, bool active);
bool tuh_max3421_spi_xfer_api(uint8_t rhport, uint8_t const * tx_buf, size_t tx_len, uint8_t * rx_buf, size_t rx_len);
void tuh_max3421_int_api(uint8_t rhport, bool enabled);
```

PR #2253, opened 2023-09-13
HID class driver fixes

Rocky04:
Addressing some drawbacks of the HID class driver implementation.

- Fixing that the `tud_hid_set_report_cb` callback would always inform about an invalid result state if a dedicated OUT endpoint is used. In the case the optional OUT endpoint is used outgoing transfers are then handled by `hidd_xfer_cb` and not by `hidd_control_xfer_cb` anymore. Then the HID examples wouldn't work anymore because the result state wouldn't be right.
- Only invoke `tud_hid_report_complete_cb` and `tud_hid_set_report_cb` if the result state of the corresponding transfer was successful. While the comment for the `tud_hid_report_complete_cb` indicate that this is only invoked when the report was sent successfully it wasn't even checked that this is the case. This can cause problems in the case there is a communication issue.
- Adding an optional callback to inform the application in the case a transfer wasn't successful.


src/class/hid/hid_device.h
```diff
@@ -118,6 +118,8 @@ TU_ATTR_WEAK bool tud_hid_set_idle_cb(uint8_t instance, uint8_t idle_rate);
 // Note: For composite reports, report[0] is report ID
 TU_ATTR_WEAK void tud_hid_report_complete_cb(uint8_t instance, uint8_t const* report, uint16_t len);
 
+// Invoked when a transfer wasn't successful
+TU_ATTR_WEAK void tud_hid_report_issue_cb(uint8_t instance, uint8_t ep_addr, xfer_result_t result, uint16_t len);
```
HiFiPhile, line 122: Sorry for the delay, is this callback actually needed for your application ?

Rocky04: _Right now the app would not be informed if there was an issue due to the fasle positives. Any data received for a dedicated OUT endpoint would wrongly flagged as INVALID even there wasn't a communication issue.

Due to this the behavior between a dedicated OUT endpoint and if there is none, so that the data is transferred over the control endpoint, is different which isn't good in my opinion._

The callback is basically only there to pass the result state and so to may provide additional info to the app when an issue occurred. If there is no interest in that `tud_hid_set_report_cb` with `HID_REPORT_TYPE_INVALID` would be fine.

HiFiPhile: I agree seeing **INVALID**  is confusing, better rename into **UNUSED**. 
With a dedicated EP we don't know the `Report type` so better to use `UNUSED` instead of `OUTPUT`.

I prefer to keep it simple for the moment, we can add the callback when there is real need. (anyway transfer failed is not implement by dcd so it won't be called)

Rocky04: The data a dedicated OUT endpoint receives is from the type `HID_REPORT_TYPE_OUTPUT` per definition. It's not invalid nor unused!

Rocky04: The new callback should also allows tracking issues for IN endpoints. Without that is not possible.

HiFiPhile: > The data a dedicated OUT endpoint receives is from the type HID_REPORT_TYPE_OUTPUT per definition. It's not invalid nor unused!

Ture.

I've simplified `xfer_result_t result` as failure reasons are only used by hcd.

HiFiPhile (approved):

PR #2254, opened 2023-09-14
feat(network): add NCM connect/disconnect status

Jacques-Zhao:
Device can notify the host about network connection status by CDC NetworkConnection.
In the current scheme, the host does not re-initiate the DHCP process even if WiFi is connected to a different router. We used it to solve the problem that WiFi devices would connect to different routers.

### DIFF
--- a/src/class/net/ncm_device.c
+++ b/src/class/net/ncm_device.c
@@ -102,6 +102,11 @@ struct ecm_notify_struct
   uint32_t downlink, uplink;
 };
 
+enum {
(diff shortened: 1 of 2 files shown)


rgrr:
Out of curiosity: have you checked my NCM PR (#2227)?  It should perform better and more bugfree than the original NCM device.

HiFiPhile:
Sorry for the delay, since #2227 is merged, could you make a rebase ?

HiFiPhile:
It has been implemented in #3135

Issue #2256, opened 2023-09-16
audio_4_channel_mic on STM32F411 with modified tud_audio_tx_done_post_load_cb


klaus-liebler:
### Operating System

Windows 11

### Board

BlackPill F411

### Firmware

audio_4_channel_mic example (unmodified and modified, see below)

### What happened ?

I successfully integrated the audio_4_channel_mic example in a new STM32CubeIDE project and used "plot_audio_samples.py" to test it. When I looked at the plot, I saw that the value 1 remains constant for two samples (see marking in the screenshot below). According to the source code, the value 1 should only be displayed for one sample.
<img width="541" alt="image" src="https://github.com/hathach/tinyusb/assets/7479349/81517935-6093-403d-985e-5cf049cb5b2a">

I investigated further and changed the "Generate dummy data" code  to get unique sample values on all four channels 
```C
for(int i=0;i<48;i++){
  i2s_dummy_buffer[0][2*i]   = 100+i;
  i2s_dummy_buffer[0][2*i+1] = 200+i;
  i2s_dummy_buffer[1][2*i]   = 300+i;
  i2s_dummy_buffer[1][2*i+1] = 400+i;
}
```

Again I started "plot_audio_samples.py". 

Expected result: A sawtooth with continuously increasing values from 400->447

Actual result: A sawtooth with the values 400->446 and then a drop to 100 (47 of 48 values of second fifo, second channel, then the first value oder first fifo and first channel).

<img width="582" alt="image" src="https://github.com/hathach/tinyusb/assets/7479349/bb905196-ecf7-41c4-afca-f7931d03b21e">

### How to reproduce ?

Compile and run example here [https://github.com/klaus-liebler/soundcard](https://github.com/klaus-liebler/soundcard)

### Debug Log as txt file (LOG/CFG_TUSB_DEBUG=2)

Not available

### Screenshots

See above

### I have checked existing issues, dicussion and documentation

- [X] I confirm I have checked existing issues, dicussion and documentation.

HiFiPhile:
Could you try #2259 ?

After fix:
![image](https://github.com/hathach/tinyusb/assets/4375114/524edbaf-b1a2-4b3a-b099-4031b6f28b7d)

klaus-liebler:
Thank you so much, @HiFiPhile . I can confirm, that #2259 fixes the problem.

Regards, Klaus

Issue #1722, opened 2022-11-05
audio 4 channel mic example transmits only 2 channels


clwe:
### Operating System

MacOS

### Board

raspberrypi pico rp2040

### Firmware

examples/device/audio_4_channel_mic

### What happened ?

The device succesfully shows up as a 4 channel USB device, but only 2 of 4 channels transmit the test data. 

- Channel 1 & 3 transmit a low amplitude saw tooth wave (as expected)
- Channel 2 & 5 transmit total silence

### How to reproduce ?

compile & flash example to rp2040
open in DAW (tested in Audacity & Ableton Live)

### I have checked existing issues, dicussion and documentation

- [X] I confirm I have checked existing issues, dicussion and documentation.

jh2srv:
Hello, 
I can confirm this bug occurs also in Windows 10 and the RP2040 (Raspberry pi pico)

PanRe:
Most likely, this is a problem within the RP2040 device controller driver (DCD), where the handling of the ISO EP may not be correct. This was at least the problem for other MCUs. The UAC driver is most likely not to blame, as it works for many MCUs. I did not investigate the RP2040 DCD however. The DCD is furthmore subject to improvements e.g. see #540. If you have time, please feel free to solve this issue and make PR! Thx in advance :)

nrg83:
Hello, I also observe this behavior on a STM32F411CEU6 ([WeActStudio.MiniSTM32F4x1](https://github.com/WeActStudio/WeActStudio.MiniSTM32F4x1)). I am on release 0.18.0 and use examples\device\audio_4_channel_mic. In Audacity I only see CH0 (saw wave) and CH1 (inverted saw wave) with small amplitude. I defined CFG_TUSB_MCU as OPT_MCU_STM32F4. Is there anything else I need to consider?
My OS is Windows 10.

If you have any suggestions on what I could try, I'm happy to test it on my hardware if you don't have it available.

PR #2259, opened 2023-09-19
Fix UAC encoding, update example.

HiFiPhile:
**Describe the PR**
Currently interleaved copy in UAC2 is bugged with wrong stride, resulting samples become scrambled if encoding/decoding is enabled.

#2256, maybe #1722

**Additional context**
`audio_4_channel_mic` example use identical sample for all channels, makes the problems less obvious, now it has been updated to output different waveforms each channel.

image


HiFiPhile:
`audio_4_channel_mic` has been updated to output different waveforms each channel.

image

examples/device/audio_4_channel_mic/src/main.c
```diff
@@ -400,7 +420,17 @@ bool tud_audio_tx_done_pre_load_cb(uint8_t rhport, uint8_t itf, uint8_t ep_in, u
   (void) ep_in;
   (void) cur_alt_setting;
 
-  for (uint8_t cnt=0; cnt < CFG_TUD_AUDIO_FUNC_1_N_TX_SUPP_SW_FIFO; cnt++)
+
+  // In read world application data flow is driven by I2S clock,
```
PanRe, line 424: In real world ;)

PanRe (approved):
Thanks a lot for fixing the bug! Nice work as usual!

HiFiPhile:
@PanRe Thanks for your quick return :)

I'll try to optimize IN transfer data flow in a following PR.

For TYPE-I format the allowed sample in each packet is n<sub>avg</sub>+-1, for example 48kHz with a FS device gives 47~49 samples per packet.
image

---

Currently it always try to send all the data in FIFO. If we take example of a 48 sample length I2S buffer, who writes all samples to UAC FIFO every 1ms: 
- If I2S clock is quicker packet length will be 48 48 48 49 48 48... which works
- If I2S clock is slower, packet length will be 48 48 48 0(or other low value) 48 48 48 48..., low value could be discarded by the host resulting gap in recording.

image


---

What I'm think is increase IN FIFO size to at least 4*packet size:
- If FIFO level is less than 25%, use packet length of n<sub>avg</sub>-1 if enough data, otherwise 0.
- If FIFO level between 25%-75% use packet length of n<sub>avg</sub>
- If FIFO level above 75% use packet length of n<sub>avg</sub>+1

SO the above example packet length becomes 48 48 48 47 48 48 48 47.

PanRe:
Interesting idea, but isn't that what you already implemented in one of your projects? I remember you mentioned something like that for the feedback value calculation?

Nevertheless, I like the idea. However, it introduces a lag and some users might not want that. I know, it is hard to get it working without some kind of buffering. How about defining a minimum buffer fill level? The higher it is, the more robust it gets. For users which don't want a buffering, they can set the fill level to zero.

Do we need to need to modify the feedback value depending on the buffer settings?

HiFiPhile:
> Interesting idea, but isn't that what you already implemented in one of your projects? I remember you mentioned something like that for the feedback value calculation?

It was for OUT transfer, in that case user code has the responsibility to calculate feedback correctly. For IN transfer the feedback is implicite, simply determined by how much data is sent.

After thinking current code only support fixed sample rate, it's necessary to calculate n<sub>avg</sub> based on sample rate to let multi-rate work.   For example in 48k & 96k case the FIFO size will be 96+1, when 48k is chosen sent packet could be 96 0 96 0....

> However, it introduces a lag and some users might not want that. 

As the FIFO will be filled in to 50% in average so there will be an additional delay of 1ms for FS or 125us for HS. Compared to OS's latency it's mostly neageible. 

> For users which don't want a buffering, they can set the fill level to zero.

The only case is when I2S clock perfectly synced with USB clock, like current examples where data is simply looped back.

PanRe:
I see! I think it is a great idea!

hathach:
awesome, great work as usual :+1: :+1:

PR #2271, opened 2023-09-29
stm32h750-dk support

antoniovazquezblanco:
**Describe the PR**
Add support for the aforementioned discovery kit.
See https://www.st.com/en/evaluation-tools/stm32h750b-dk.html#overview for more details about it.

**Additional context**
Initial support and tested a couple of example projects.
This is my first contribution to the project so please be patient.

Thanks

### DIFF
--- a/hw/bsp/stm32h7/boards/stm32h750bdk/board.mk
+++ b/hw/bsp/stm32h7/boards/stm32h750bdk/board.mk
@@ -1,0 +1,21 @@
+# STM32H745I-DISCO uses OTG_FS
+# FIXME: Reset enumerates, un/replug USB plug does not enumerate
+
+CFLAGS += -DSTM32H750xx -DCORE_CM7 -DHSE_VALUE=25000000
+
+# Default is FulSpeed port
+PORT ?= 0
+
+# GCC
+SRC_S_GCC += $(ST_CMSIS)/Source/Templates/gcc/startup_stm32h745xx.s
+LD_FILE_GCC = $(ST_CMSIS)/Source/Templates/gcc/linker/stm32h745xx_flash_CM7.ld
+
+# IAR
+SRC_S_IAR += $(ST_CMSIS)/Source/Templates/iar/startup_stm32h745xx.s
+LD_FILE_IAR = $(ST_CMSIS)/Source/Templates/iar/linker/stm32h745xx_flash_CM7.icf
+
+# For flash-jlink target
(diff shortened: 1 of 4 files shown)


hw/bsp/stm32h7/boards/stm32h750bdk/board.mk
```diff
@@ -0,0 +1,21 @@
+# STM32H745I-DISCO uses OTG_FS
+# FIXME: Reset enumerates, un/replug USB plug does not enumerate
+
+CFLAGS += -DSTM32H750xx -DCORE_CM7 -DHSE_VALUE=25000000
+
+# Default is FulSpeed port
+PORT ?= 0
+
+# GCC
+SRC_S_GCC += $(ST_CMSIS)/Source/Templates/gcc/startup_stm32h745xx.s
```
hathach, line 10: startup and linker should be changed to h750 instead of h745. I was about to change that but I realize that H750 mcu only has 128KB of flash, it proabably means to be used with qspi flash. Though if all examples can fit into 128KB, we are all good.

hathach (changes requested):
thank you for the PR, you are doing great, and it looks good. I just rebase and update the PR to match latest change. However, the board should use both startup and linker for h750, would you mind updating and testing it again. Thank you

hw/bsp/stm32h7/boards/stm32h750bdk/board.mk
antoniovazquezblanco: Missed that... I have just found that there is no 750 linker script :/

antoniovazquezblanco:
Currently contributing with a sample linker script that may suit our needs in https://github.com/STMicroelectronics/cmsis_device_h7/pull/8

hathach:
> Currently contributing with a sample linker script that may suit our needs in [STMicroelectronics/cmsis_device_h7#8](https://github.com/STMicroelectronics/cmsis_device_h7/pull/8)

we don't need to wait for this, it would take forever for ST to merge this. Instead you could just copy the linker to the board folder like this board https://github.com/hathach/tinyusb/tree/master/hw/bsp/stm32h7/boards/daisyseed

antoniovazquezblanco:
Done!

hathach (approved):
perfect, thank you for your hard works

antoniovazquezblanco:
Thank you very much!

PR #2278, opened 2023-10-06
Add CH32F20x support

Krasutski:
**CH32F20x MCU family support**
Added support of CH32F2 MCUs

Current code already support WCH CH32**V3** chip, this MCU have the same USB HS core as CH32F20x therefore I just moved portable/wch/ch32v307/dcd_usbhs.c file to common folder(portable/wch/dcd_ch32_usbhs.c) and add libs for CH32F2xx
Tested on ch32f205r-r0 demo board

image

This my first Pull Request to tinyusb, please let me know If I missed something


hathach (approved):
perfect, thank you very much. I resolve an readme conflict and fix a minor pre-commit format (end of file). Will merge when ci is all passed.

PR #2282, opened 2023-10-17
UAC IN transfer improvements

HiFiPhile:
**Describe the PR**

Currently UAC IN transfer always try to send all the data in FIFO, which comes with 2 issues:
1. Recorded audio could have glitch depending on I2S clock variation.
2. Multi-rate (eg. 44.1k & 48k) doesn't work as the packet size is different.

For TYPE-I format the allowed sample in each packet is n<sub>avg</sub>+-1, for example 48kHz with a FS device gives 47~49 samples per packet. 
image

This PR makes IN transfer size honor the constraint and resolve above 2 issues, by selecting the packet size n<sub>i</sub> based on nominal size n<sub>av</sub> and FIFO level.

**Additional context**
In current version, if we take example of a 48 sample length I2S buffer, who writes all samples to UAC FIFO every 1ms: 
- If I2S clock is quicker packet size will be 48 48 48 49 48 48... which works
- If I2S clock is slower, packet size will be 48 48 48 0(or other low value) 48 48 48 48..., **low value could be discarded by the host resulting gap in recording.**

image
image

In a multi-rate case of 48k & 96k, when 48k is chosen the packet size could be 96 0 96 0 96 0.... **resulting completely distorted recording.**

---

With this PR a short packet is sent when I2S clock is slower, keeping the recording distortion free.

image
image

### DIFF
--- a/examples/device/audio_4_channel_mic/src/main.c
+++ b/examples/device/audio_4_channel_mic/src/main.c
@@ -407,9 +407,8 @@ bool tud_audio_get_req_entity_cb(uint8_t rhport, tusb_control_request_t const * 
         {
           case AUDIO_CS_REQ_CUR:
             TU_LOG2("    Get Sample Freq.\r\n");
-            // Set sample rate for flow control
-            tud_audio_set_tx_flow_control(sampFreq);
-            return tud_control_xfer(rhport, p_request, &sampFreq, sizeof(sampFreq));
+            // Buffered control transfer is needed for IN flow control to work
+            return tud_audio_buffer_and_schedule_control_xfer(rhport, p_request, &sampFreq, sizeof(sampFreq));
 
           case AUDIO_CS_REQ_RANGE:
             TU_LOG2("    Get Sample Freq. range\r\n");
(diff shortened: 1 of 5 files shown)


HiFiPhile:
@PanRe PR is up, looking for class users' feedback.

PanRe (approved):
Nice work!! Thanks a lot!

PanRe:
Could you explain a bit how the blackout works please?

HiFiPhile:
> Could you explain a bit how the blackout works please?

Well that's something need to be proven :) There are some reasons why I added it:

- At startup as the FIFO is being filled, short packets will be sent until FIFO half filled, on my Windows machine I observed there are 1 or 2 samples missed when there are too many short packets (maybe it tries to correct clock drift, who knows), while capture shows packets are ok.
image

- When the FIFO filling and consuming frequency are different (eg. I2S interrupt every 400us) FIFO won't be always half filled, without blackout the packet size will oscillate between smaller and bigger one.

PanRe:
I see, sounds a bit like black magic stuff :)

From my side, I don't have any problems with merging your PR as it is already more robust than the current version. But I am curious, have you already conducted a long term test? Maybe for an hour? Or do you know if Linux works without issues? Lots of time consuming questions I know ;)

In any case, very nice work as usual!!

PanRe:
Nice! Ok I guess the PR is good for merging! :)

PR #2289, opened 2023-10-24
Update audio_4_channel_mic example for HS device.

HiFiPhile:
**Describe the PR**
Update audio_4_channel_mic example and descriptor, to let HS device use 125us interval.


PanRe (approved):
Alright! :)

PR #2303, opened 2023-11-01
Add optional hooks for DCD and HCD events

projectgus:
These are intended to allow bare metal platforms with one-shot scheduling capabilities to schedule the TinyUSB task handlers whenever they know there is work for them to do.

I've been experimenting with this approach for MicroPython, where we don't usually have an RTOS but we do have scheduling capability. Selectively scheduling `tud_task_ext()` to run when we know there are events in the queue reduces latency for processing, while also avoiding the overhead of polling `tud_task_ext()` when there may not have been any USB activity.

(For context, see https://github.com/micropython/micropython/pull/12846 - that PR uses a linker wrap around `dcd_event_handler` to hook it. If this PR is accepted then we can change to using the new macro for the hooks.)

Thanks for all your work on TinyUSB, it's a fantastic library!


hathach (commented):
thank you very much, this is a great idea and I really like it. Though I would like to tweak it to use weak callback function to be consistent with the rest of API. e.g 

```C
void tud_event_hook_cb(const void* event, bool in_isr) _weak
void tuh_event_hook_cb(const void* event, bool in_isr) _weak
```

The reasson to change to `void*` is dcd_event_t/hcd_event_t type are not public (not included with tusb.h). 

Let me know if this suggestion make sense to you.

projectgus:
That's great, thanks! Would you like me to update the PR to change the hooks?

Probably the "event" parameter can be removed from the hook. I added them on a whim as they're "free" with a macro, but I can't think of a use case so it can probably be removed.

I guess passing through "in_isr" is still potentially useful, though.

hathach:
> That's great, thanks! Would you like me to update the PR to change the hooks?

Yeah, please do it when you have the time.

> Probably the "event" parameter can be removed from the hook. I added it on a whim as they're "free" with a macro, but I can't think of a use case so it can probably be removed.
> 
> I guess passing through "in_isr" is still potentially useful, though.

Yeah, you are right, application does not have any usage with internal data. Maybe we can change it to have rhport, eventid (header of the event) is better. I don't know if applications ever need that but maybe useful some cases especially rhport when there could be multiple host controllers.

```C
void tud_event_hook_cb(uint8_t rhport, uint32_t eventid, bool in_isr) _weak
void tuh_event_hook_cb(uint8_t rhport, uint32_t eventid, bool in_isr) _weak
```

hathach:
actually, since I am working on PRs, I could do it just now, if you haven't started thge update  yet.

projectgus:
I haven't, so feel free if you have time. Thanks!

hathach (approved):
all is good now, thank you very much for the PR

PR #2314, opened 2023-11-13
Adding extern C to definitions to allow link on C++ project

jfm92:
Hi, 
I've found dhserver and dnserver gives a linkage error when you try to use on a C++ project, so the solution that I've found is to 
add extern c to the affected headers.


hathach (approved):
look good, thank you

PR #2328, opened 2023-11-19
UAC2: Implement feedback by fifo counting.

HiFiPhile:
**Describe the PR**

- Implement feedback by fifo counting
- Implement host OS guessing to resolve compatibility issues

Compared to other methods the only thing needed is add a callback to set sample rate, all calculations are done inside the class.

```c
void tud_audio_feedback_params_cb(uint8_t func_id, uint8_t alt_itf, audio_feedback_params_t* feedback_param)
{
  // Set feedback methode to fifo counting
  feedback_param->method = AUDIO_FEEDBACK_METHOD_FIFO_COUNT;
  feedback_param->sample_freq = current_sample_rate;
}
```

A new example `uac2_speaker_fb` has been added to show how the mechanism work with a stereo speaker. Both FS (STM32F429) and HS (LPC55S69) configs are tested and FIFO level is well maintained half filled.

---

A HID debug interface can be enabled by setting `CFG_AUDIO_DEBUG=1` which export UAC class information and can be read by `audio_debug.py`.

image


HiFiPhile:
Test welcomed @Protoxy22 @Lurcy38 @kf6gpe @PanRe

Hal9000Space:
This is interesting.  I'm curious where can I find audio_debug.py.  Any ideas?  Thanks a lot!

HiFiPhile:
> This is interesting. I'm curious where can I find audio_debug.py. Any ideas? Thanks a lot!

Just in the PR examples/device/uac2_speaker_fb/src/audio_debug.py

Hal9000Space:
> > This is interesting. I'm curious where can I find audio_debug.py. Any ideas? Thanks a lot!
> 
> Just in the PR examples/device/uac2_speaker_fb/src/audio_debug.py

Ah, so not in main yet.  Ok, thanks!

Lurcy38:
great :) , I will be very happy to test this implementation in my app, but i cannot reach the branch rx_fb , how to do ?

HiFiPhile:
@Lurcy38 it's under my user space 
https://github.com/HiFiPhile/tinyusb/tree/rx_fb
Or you can download PR by adding `.patch`
https://github.com/hathach/tinyusb/pull/2328.patch

HiFiPhile:
@hathach After more tests this PR is ready.

mastupristi:
Hi, what is the status of this PR? what is missing for it to be approved?

Marcus2060:
Hi, great work,
very nice feedback method improvement on Tinyusb class Audio !
I Have implemented the feedback method FIFO_COUNTER (stereo 16 bit, 48KHz) on a my RP2040 hardware with I2S DAC with success .. but ...
I observe a good audio USB reproduction on Windows but the system crashes on MAC computer...???
If, instead, I remove only the feedback EP, the system It's starting to work on MAC too ! But obviously not sync ...

Could this be due to the fact that macOS does not accept this feedback method or some other reason such as incorrect descriptors for the MAC?

Thanks!

Protoxy22:
> Hi, great work, very nice feedback method improvement on Tinyusb class Audio ! I Have implemented the feedback method FIFO_COUNTER (stereo 16 bit, 48KHz) on a my RP2040 hardware with I2S DAC with success .. but ... I observe a good audio USB reproduction on Windows but the system crashes on MAC computer...??? If, instead, I remove only the feedback EP, the system It's starting to work on MAC too ! But obviously not sync ...
> 
> Could this be due to the fact that macOS does not accept this feedback method or some other reason such as incorrect descriptors for the MAC?
> 
> Thanks!

Hi
Does your RP2040 implement USB audio OUT and audio IN ?
If my memory is good, MacOS will not use the feedback endpoint if there is an OUT and IN endpoint, it's called IMPLICIT synchronization. The host will use the input stream's packets to construct the output packets.

I recommend you the section "Feedback for Asynchronous Output Streams" in [https://developer.apple.com/library/archive/technotes/tn2274/_index.html](https://developer.apple.com/library/archive/technotes/tn2274/_index.html)

mastupristi:
> I observe a good audio USB reproduction on Windows but the system crashes on MAC computer...??? If, instead, I remove only the feedback EP, the system It's starting to work on MAC too ! But obviously not sync ...

How did you get it working on Windows?
I have interface 1 containing the data endpoint OUT and the feedback endpoint IN. Then I have interface 2 containing only the data endpoint IN.
On Linux I can see it working (I still have to validate the absence of glitches), and with wireshark I can see correctly passing data on the feedback endpoint.
Windows, on the other hand, seems to ignore the feedback endpoint altogether. And we don't understand why.

Can you tell us how you declared the descriptors?

best regards
Max

mastupristi:
> I recommend you the section "Feedback for Asynchronous Output Streams" in https://developer.apple.com/library/archive/technotes/tn2274/_index.html

Do you know whether Windows also has such behavior?
 
> I would also recommend you to test to add "`TUSB_ISO_EP_ATT_IMPLICIT_FB`" in your `Standard AS Isochronous Audio Data Endpoint Descriptor`

I have interface 1 containing the data endpoint 1 OUT and the feedback endpoint 1 IN. Then I have interface 2 containing only the data endpoint 2 IN.
if I turn the endpoint 2 IN from Data to implicit feedback, as you suggest, on Linux it can no longer open communication. As soon as you try ALSA reports a generic I/O Error.
Do you know whether having both explicit and implicit Feedback is a forbidden, or at least "disliked" by Linux?
I also left the implicit feedback because if from PC I only open the output there would be no feedback, or is that wrong?

best regards
Max

Protoxy22:
Hi @mastupristi, can you show me your `tusb_desc.c` and `tusb_desc.h` if you have them ?
I didn't use this fifo counting feedback branch yet, so I'm not sure if my suggestions could work in this case.

After some hindsight, **my bad**, I think changing any endpoint to implicit feedback is a non-sense in this situation, because the whole branch is about a feedback endpoint which is necessarily explicit.

In the example [uac2_speaker_fb](https://github.com/HiFiPhile/tinyusb/tree/rx_fb/examples/device/uac2_speaker_fb):
I would add `TUSB_ISO_EP_ATT_EXPLICIT_FB` on the data endpoints, just to be sure that all OS are suggested to use the available feedback endpoint @HiFiPhile ?

For your question @mastupristi, all your data endpoints should have the same synchronization mechanism (either implicit or explicit with feedback endpoint) otherwise it would be a mess for the host to manage all this.

So to resume, if you use this uac2_speaker_fb example, you should forget about implicit, and use explicit instead (add it to the descriptor). It should work on Windows 10 or 11 (if you have a recent version) and MacOS, I'm not sure about Linux, I never tested anything on it.

Protoxy22:
Hi @mastupristi 
Yes could you send me your usb_descriptors.[ch], to look if there is something strange x)
Maybe you could create a GitHub issue and ping me, it would be more convenient than trying to fix it here 

And yes, your screenshot is totally on point, I was just not sure if you used this branch or not

mastupristi:
Hi @Protoxy22 

>  Maybe you could create a GitHub issue and ping me, it would be more convenient than trying to fix it here

I started a discussion here: https://github.com/hathach/tinyusb/discussions/2475

HiFiPhile:
Back from holiday...

> I would add TUSB_ISO_EP_ATT_EXPLICIT_FB on the data endpoints, just to be sure that all OS are suggested to use the available feedback endpoint @HiFiPhile ?

No it's not how it works...

HiFiPhile:
> I observe a good audio USB reproduction on Windows but the system crashes on MAC computer...???

What do you mean crash, does the kernel panic or having popping sound or something else ?

You can try to change https://github.com/hathach/tinyusb/blob/2720392a154dac5ff4529fbcc16c0a0718c9c8af/src/class/audio/audio_device.h#L196

```// The feedback format is specified to be 16.16 for HS and 10.14 for FS devices (see Universal Serial Bus Specification Revision 2.0 5.12.4.2). By default,
// the choice of format is left to the caller and feedback argument is sent as-is. If CFG_TUD_AUDIO_ENABLE_FEEDBACK_FORMAT_CORRECTION is set, then tinyusb
// expects 16.16 format and handles the conversion to 10.14 on FS.
//
// Note that due to a bug in its USB Audio 2.0 driver, Windows currently requires 16.16 format for _all_ USB 2.0 devices. On Linux and macOS it seems the
// driver can work with either format. So a good compromise is to keep format correction disabled and stick to 16.16 format
```

Marcus2060:
Hi,
I continued working with a Full-Speed USB audio system RP2040 48KHz 16 bit by using feedback explicit endpoint. 
(TinyUSB with FIFO counter feedback mechanism).

I've done other experiments and I get the impression that the usbd_edpt_xfer() function on low-level stack doesn't work well when we try to send 3 bytes.
These are my two scenarios:

MAC:
1) The system works fine with 10.14 feedback (CFG_TUD_AUDIO_ENABLE_FEEDBACK_FORMAT_CORRECTION=1) but only if I set MaxPacketBytes=3 in the feedback descriptor.
2) The system doesn't work if I try to send 3 Bytes (10.14) via usbd_edpt_xfer() but with MaxPacketBytes=4 ???  (MaxPacketBytes means Max number of bytes transfer on the endpoint ??...)
3) The system doesn't work with 16.16 feedback (4 bytes sent) because seems that macOS does not accept 16.16 for full speed devices.

Windows:
1) The system works well with 16.16 feedback (4 bytes sent).
2) The system doesn't work with 10.14 format (CFG_TUD_AUDIO_ENABLE_FEEDBACK_FORMAT_CORRECTION=1) when I try to send 3 bytes by usbd_edpt_xfer() (MaxPacketBytes = 4 in the descriptor) ... I listen some click audio due, as we know,  with more or less bytes on Audio buffer respect to the DAC buffer...
3) If I use, instead, the same code of MAC (as above MaxPacketBytes = 3 in the descriptor) unfortunately the system stops working because the Audio class it is not enumerate on WIndows and I no longer see the card ... ???

Also:
I tested the pico-playground usb_sound_card project (Pico board + DAC board) and it works well both Windows and MAC. If we see the code (usb_sound_card.c) seems that feedback routine uses 10.14 format putted in 3 bytes. 
 
So, could the problem be on the low level of the Tinyusb stack when we try to send 3 bytes feedback via usbd_edpt_xfer() to Host ???

Thanks!

HiFiPhile:
> So, could the problem be on the low level of the Tinyusb stack when we try to send 3 bytes feedback via usbd_edpt_xfer() to Host ???

No it can't be, you also tested 3 bytes works on MAC.

> If I use, instead, the same code of MAC (as above MaxPacketBytes = 3 in the descriptor) unfortunately the system stops working because the Audio class it is not enumerate on WIndows and I no longer see the card ... ???

It's true.

> I tested the pico-playground usb_sound_card project (Pico board + DAC board) and it works well both Windows and MAC. If we see the code (usb_sound_card.c) seems that feedback routine uses 10.14 format putted in 3 bytes.

Pico use UAC 1.0 while TUSB uses UAC 2.0, you can see their descriptors are quite different.

Marcus2060:
Hi,
thank you for reply and sorry if I try to clarify better:

> No it can't be, you also tested 3 bytes works on MAC.

No, my previous comment is precisely because I tested that the 3 bytes don't seem to work (see my pont 1, 2 about MAC):
1) The system works fine with 10.14 feedback (CFG_TUD_AUDIO_ENABLE_FEEDBACK_FORMAT_CORRECTION=1) but only if I set MaxPacketBytes=3 in the feedback descriptor.
2) The system doesn't work **if I try to send 3 Bytes** (10.14) via usbd_edpt_xfer() but with MaxPacketBytes=4  (MaxPacketBytes means Max number of bytes transfer on the endpoint ??...)

I highlighted the 3 bytes sent problem because:
**Point 1** could it be a mechanism where the low-level USB stack forces anyway the 3 bytes ?? 
**Point 2** why doesnt'work while point 1 works ??

Thanks anyway for your attention.

HiFiPhile:
@Marcus2060 

> The system doesn't work if I try to send 3 Bytes (10.14) via usbd_edpt_xfer() but with MaxPacketBytes=4 (MaxPacketBytes means Max number of bytes transfer on the endpoint ??...)

Sorry I've some problem understanding, you mean `3 Bytes  && MaxPacketBytes=4` doesn't work ? How about `MaxPacketBytes=3` ?

Marcus2060:
> Sorry I've some problem understanding, you mean 3 Bytes  && MaxPacketBytes=4 doesn't work ? How about 3 Bytes  && MaxPacketBytes=3 ?

Below 4 different tests on MAC using 10.14 format:
1) When (4 Bytes sent) && (MaxPacketBytes=4) -> **Wrong Audio** (I think the macOS recognize 16.16).
2) When (3 Bytes sent) && (MaxPacketBytes=4) -> **No Sync, some clicks.**
3) When (3 Bytes sent) && (MaxPacketBytes=3) -> **No Sync, some clicks.**
4) When (4 Bytes sent) && (MaxPacketBytes=3) -> **Works fine, no click!**

Below the function to send feedback (feedback value is an uint32_t) :
usbd_edpt_xfer(audio->rhport, audio->ep_fb, (uint8_t *) &value, 4);

HiFiPhile:
> When (3 Bytes sent) && (MaxPacketBytes=3) -> No Sync, some clicks.

I agree it seems strange but I don't see any issue in TUSB side (maybe something wrong in RP2040 porting ?), could you use Wireshark USB capture to see what happened ?

HiFiPhile:
@Marcus2060 I come back to this.

> So the problem seems to be a different behavior of TinyUsb Audio Class 2.0 between Windows, Linux (16.16 ok) and MAC (10.14 ok ... but with the trick MAXPacketBytes=3 ) .... ??

No the stack works (tries to) the same way across all OSes, the issue is some OS doesn't respect the UAC specification as I mentioned earlier.

> The Rp2040 is a full-speed device and I don't need large audio streaming traffic. So would it be possible with the TinyUsb stack to move from UAC2 (Audio 2.0) to class Audio 1.0 (USB Speaker)
by changing only the descriptors or relatively little code on stack or on firmware ??

Unfortunately the descriptor difference is pretty large, some major changes are needed. 

---
To be honest I don't see a hassle-free way to make it work across all OSes, Windows can be detected as it ask for MS OS 2.0 descriptor, but it happens after the configuration descriptor where `MaxPacketBytes` is already set.

lijunru-hub:
Testing this feature with ESP32-S3 has been very successful. The device plays sound without any noise, and the FIFO consistently maintains a stable intermediate level.

HiFiPhile:
@Marcus2060 @rhysmorgan134 
I've setup up a Hackintosh Ventura and tested 3 bytes feedback `(3 Bytes sent) && (MaxPacketBytes=3)`and it works very well.
Zero click & pop during more than one hour listening:
image

So I'm curious about why it doesn't work in your cases.
Only thing I come up is when 4 bytes are sent 2ms is needed (3+1) instead of 1ms to finish one feedback transfer, maybe it makes OSX's rate adjusting speed different ?
Maybe I2S DMA buffer is too big ? It should be at least smaller than 1/4 of UAC FIFO size to avoid underflow/overflow when 2 packets are popping/pushing from/to the buffer consecutively.

rhysmorgan134:
> @Marcus2060 @rhysmorgan134 I've setup up a Hackintosh Ventura and tested 3 bytes feedback `(3 Bytes sent) && (MaxPacketBytes=3)`and it works very well. Zero click & pop during more than one hour listening: image
> 
> So I'm curious about why it doesn't work in your cases. Only thing I come up is when 4 bytes are sent 2ms is needed (3+1) instead of 1ms to finish one feedback transfer, maybe it makes OSX's rate adjusting speed different ? Maybe I2S DMA buffer is too big ? It should be at least smaller than 1/4 of UAC FIFO size to avoid underflow/overflow when 2 packets are popping/pushing from/to the buffer consecutively.
> 
> Anyone interested can find the code: https://gist.github.com/HiFiPhile/9ed6f4f65bf1dc5bc0c7626fb7700d8d

Thanks for looking into this so much! I will take a look though your implementation and match buffer sizes and then feedback.

rhysmorgan134:
@HiFiPhile Seems you are right, setting my DMA buffers to 64 samples, 128bytes in my case, works fine with 3 bytes set in the transmit function and maxPacketBytes==3.

Are you using the tusb_config.h and descriptors from the example? If not would you mind sharing your ones you used for your test, would be nice for me to just sanity check mine.

Also did you run into any difficulties getting the python script running on OSX?

HiFiPhile:
> Are you using the tusb_config.h and descriptors from the example

Yes they are from stack example, only difference is enable correction and set wMaxPacketSize=3:
```patch
diff --git a/examples/device/uac2_speaker_fb/src/tusb_config.h b/examples/device/uac2_speaker_fb/src/tusb_config.h
index 1cc1031c1..1a1cebca8 100644
--- a/examples/device/uac2_speaker_fb/src/tusb_config.h
+++ b/examples/device/uac2_speaker_fb/src/tusb_config.h
@@ -123,7 +123,7 @@ extern "C" {
 #define CFG_TUD_AUDIO_FUNC_1_DESC_LEN                                TUD_AUDIO_SPEAKER_STEREO_FB_DESC_LEN

 // Enable if Full-Speed on OSX, also set feedback EP size to 3
-#define CFG_TUD_AUDIO_ENABLE_FEEDBACK_FORMAT_CORRECTION              0
+#define CFG_TUD_AUDIO_ENABLE_FEEDBACK_FORMAT_CORRECTION              1

 // Audio format type I specifications
 #if defined(__RX__)
diff --git a/examples/device/uac2_speaker_fb/src/usb_descriptors.c b/examples/device/uac2_speaker_fb/src/usb_descriptors.c
index fa307674d..5d84024ea 100644
--- a/examples/device/uac2_speaker_fb/src/usb_descriptors.c
+++ b/examples/device/uac2_speaker_fb/src/usb_descriptors.c
@@ -150,7 +150,7 @@ uint8_t const desc_configuration[] =

     // Interface number, string index, byte per sample, bit per sample, EP Out, EP size, EP feedback, feedback EP size,
-    TUD_AUDIO_SPEAKER_STEREO_FB_DESCRIPTOR(0, 4, CFG_TUD_AUDIO_FUNC_1_N_BYTES_PER_SAMPLE_RX, CFG_TUD_AUDIO_FUNC_1_RESOLUTION_RX, EPNUM_AUDIO_OUT, CFG_TUD_AUDIO_FUNC_1_EP_OUT_SZ_MAX, EPNUM_AUDIO_FB | 0x80, 4),
+    TUD_AUDIO_SPEAKER_STEREO_FB_DESCRIPTOR(0, 4, CFG_TUD_AUDIO_FUNC_1_N_BYTES_PER_SAMPLE_RX, CFG_TUD_AUDIO_FUNC_1_RESOLUTION_RX, EPNUM_AUDIO_OUT, CFG_TUD_AUDIO_FUNC_1_EP_OUT_SZ_MAX, EPNUM_AUDIO_FB | 0x80, 3),

 #if CFG_AUDIO_DEBUG
     // Interface number, string index, protocol, report descriptor len, EP In address, size & polling interval
```

> Also did you run into any difficulties getting the python script running on OSX?

There are some dependencies needs install:
```
sudo pip3 install hid
sudo pip3 install usb
sudo pip3 install metaplot
```

HiFiPhile:
Last commit optimized debug interface fifo count, it's not the real count inside UAC class but a replication, it was less accurate.

ra1nb0w:
maybe someone is interested in a rebased version to the latest master
https://github.com/ra1nb0w/tinyusb/tree/rx_fb

HiFiPhile:
> maybe someone is interested in a rebased version to the latest master https://github.com/ra1nb0w/tinyusb/tree/rx_fb

Thank you, will rebase when free, also move #2628 into example.

HiFiPhile:
@hathach This PR has been tested my multiple users and works well, since @PanRe is not here could you take a look and merge?

PanRe (approved):
Impressive work! Thank you all for your effort in improving the UAC2 driver!!

ra1nb0w:
thank you to everyone!

geekbozu:
Woo love to see this finally get in! Thanks for keeping it going :D

mstaack:
great! hello proper uac:)

skuep:
Nice work @HiFiPhile !
Regarding the OS guessing, I have done some work regarding Windows WCID in the past which could theoretically be used for accurately detecting windows (7 and up I Believe). See
https://github.com/pbatard/libwdi/wiki/WCID-Devices#user-content-Microsoft_OS_String_Descriptor

But I am not sure if the Descriptor is requested before the configuration Descriptor. Because I think it needs to change the maxPackstSize for the feedback endpoint, correct?

skuep:
@HiFiPhile thanks a lot for your effort. I looked through this and was wondering if it's wise to use the BOS Descriptor for the quirk detection because apparently the BOS Descriptor is mandatory only for USB versions > 2.0 if I read correctly.

In addition, device, config and string Descriptor order is already enough to distinguish between MacOS and not-MacOS.

What do you think?

HiFiPhile:
> @HiFiPhile thanks a lot for your effort. I looked through this and was wondering if it's wise to use the BOS Descriptor for the quirk detection because apparently the BOS Descriptor is mandatory only for USB versions > 2.0 if I read correctly.
> 
> In addition, device, config and string Descriptor order is already enough to distinguish between MacOS and not-MacOS.
> 
> What do you think?

You are right, config and string Descriptor is already enough to distinguish between MacOS and not-MacOS.

It's a full detection method from https://github.com/hathach/tinyusb/pull/2628 which allows each OS to be treated differently. In my testing BCD 2.01 is well supported, normally it shouldn't be an issue.

PR #2346, opened 2023-11-24
made log line end consistent \r\n

IngHK:
this PR made TU_LOG and other prints consistent using \r\n instead \n only


hathach (approved):
look good thank you

PR #2354, opened 2023-11-28
fix: memory leak: buf

slark-yuxj:
**Describe the PR**
fix: memory leak: buf


hathach (approved):
thank you, sorry for late response. I have fixed the pre-commit check, will merge after ci passed.

PR #2357, opened 2023-11-29
Allow retrieving the port a device is connected to

Yveaux:
**Describe the PR**
Sometimes it may be handy to know to which port a device is connected to, e.g. when dealing with identical devices.
This change allows the port number of a device to be retrieved.

**Additional context**
Supports root hub and external hub ports.


Yveaux:
Anything I can do to fix this?

image

hathach (changes requested):
tuh_get_port() is a confusing API name and its return in the PR is inconsistent as well. We can have an public API for  `hcd_devtree_get_info()` with the name e.g `tuh_devtree_get_info()`

Yveaux:
Something like this? I'm in doubt where everything should land...

Yveaux:
@hathach ping, is this what you were looking for?

Yveaux:
@hathach how can we bring this further?

hathach:
implemented by #3093

PR #2358, opened 2023-11-30
Updated class.MD

Polpatra:
I have updated the markdown files.


hathach:
I don't see any changes at all

Issue #2365, opened 2023-12-09
Alignment flag triggers warning: assuming pointer wraparound does not occur


angmolin:
### Operating System

Linux

### Board

rp2040

### Firmware

Any example.

https://github.com/angmolin/tinyusb/tree/bugfix/gcc%2B12-strict-overflow-warning

### What happened ?

When you try to compile tinyusb examples with gcc >= 12 it annoyingly triggers -Wstrict-overflow:

```
tinyusb/src/class/audio/audio_device.c:1568:23: error: assuming pointer wraparound does not occur when comparing P +- C1 with P +- C2 [-Werror=strict-overflow]
 1568 |         while (p_desc < p_desc_end)
      |                ~~~~~~~^~~~~~~~~~~~
compilation terminated due to -Wfatal-errors.
cc1: all warnings being treated as errors
```

### How to reproduce ?

Try to compile the examples folder by running:
```
$ cd tinyusb/examples
$ mkdir build
$ cd build
$ cmake .. -DFAMILY=rp2040 -DPICO_SDK_PATH="/home/angmolin/pico-sdk"
$ make
```

### Debug Log as txt file (LOG/CFG_TUSB_DEBUG=2)

No log file attached since the error is in compilation time.

### I have checked existing issues, dicussion and documentation

- [X] I confirm I have checked existing issues, dicussion and documentation.

PR #2366, opened 2023-12-09
Fix strict-overflow warning when compiling with gcc >= 12

angmolin:
**Describe the PR**
This PR solves a problem with strict-overflow when compiling the tinyusb examples with gcc >=12. Here is the output:
```
tinyusb/src/class/audio/audio_device.c:1568:23: error: assuming pointer wraparound does not occur when comparing P +- C1 with P +- C2 [-Werror=strict-overflow]
 1568 |         while (p_desc < p_desc_end)
      |                ~~~~~~~^~~~~~~~~~~~
compilation terminated due to -Wfatal-errors.
```

This PR solves issue #2365

**Additional context**
Compiled from Rocky Linux 9


HiFiPhile:
Thanks for the pr, could you add a comment?

angmolin:
Hello, I've just added a comment to the PR to explain why I changed the condition. Thank you for feedback.

HiFiPhile (approved):

Issue #2374, opened 2023-12-12
STM32F7: DIEPINT_TOC with ARM (Apple silicon) Mac


davekeck:
### Operating System

MacOS

### Board

custom

### Firmware

stm32f723disco

### What happened ?

I'm running the stm32f723disco example on a custom board. (Actual chip: STM32F730I8K6.)

USB comms work reliably when connected to an Intel Mac.

USB comms are flaky when connected to an ARM Mac, in which case the `USB_OTG_DIEPINT_TOC` / `DIEPINT_TOC` interrupt is occurring on EP0. This causes TinyUSB to loop in `handle_epin_irq()` (because it never clears the `DIEPINT_TOC` interrupt). However it's my understanding that the `DIEPINT_TOC` interrupt shouldn't occur in the first place.

Here's a screenshot of my debugging session that shows that `epin[0].diepint` has `DIEPINT_TOC` set (bit 3). See debug console at the bottom:

<img width="1040" alt="DIEPINT_TOC-Screenshot" src="https://github.com/hathach/tinyusb/assets/791944/3ad35b08-2529-4a03-becf-928a22e5133e">

This issue also occurs with stock STM32 firmware generated by STM32CubeIDE, which I posted about here:

https://community.st.com/t5/stm32-mcus-embedded-software/usb-comms-failure-with-apple-silicon-arm-mac/m-p/618571#M43842

I've ordered the official STM32F723E-DISCO hardware to verify that this isn't an issue with my custom board, as well as a USB protocol analyzer. I'll update this issue in a few days with my findings. I hope opening this issue isn't premature...

Issues related to DIEPINT_TOC interrupt: #139, #171

### How to reproduce ?

Plug device into ARM Mac, and repeatedly issue `GET_STATUS` device requests. Eventually the `DIEPINT_TOC` interrupt will occur.

### Debug Log as txt file (LOG/CFG_TUSB_DEBUG=2)

I can't access the logging on my custom board. I'll provide the logging once I test on the official STM32F723E-DISCO hardware.

### I have checked existing issues, dicussion and documentation

- [X] I confirm I have checked existing issues, dicussion and documentation.

davekeck:
I confirmed that this issue occurs on the stock STM32F723E-DISCO board running the stock stm32f723disco example.

[...]

PR #2380, opened 2023-12-16
STM32F7: correct USB_HS_PHYC_TUNE configuration

davekeck:
Fix issue #2374 by assigning `USB_HS_PHYC_TUNE` to the magic value, instead of performing a bitwise-or.

Previously, the bitwise-or caused the default value of the `LFSCAPEN` bit (bit 2) to be retained, where `LFSCAPEN` is described by the reference manual as:

> LFSCAPEN: Enables the Low Full Speed feedback capacitor.

This capacitor interferes with the D+/D- signals (which makes sense because it appears to be intended for low/full-speed mode, but we're in high-speed mode), which breaks USB comms with ARM-based Macs.


davekeck:
This PR was premature, see comments in issue #2374.

PR #2385, opened 2023-12-21
dcd_nrf5x: Fix dcd_edpt_open for iso endpoint

kasjer:
**Describe the PR**
When ISO endpoint handling was introduced two lines that clear stall and data toggle bit were left unchanged and they were effective for ISO endpoint as well.
This is incorrect behavior since EPSTALL and DTOGGLE registers have only 3 bits for address.
Leaving code that clears toggle bit results in endpoint 0 toggle bit being reset when iso endpoint (8) is opened.

Now code that clears stall and toggle bit is applied to non-iso endpoint only as it was done before iso handling was introduced.


hathach (approved):
look good, thank you for the fix

PR #2181, opened 2023-07-25
Add SOF interrupt to CH32V307

Xelus22:
**Describe the PR**
Add SOF interrupt to CH32V307

Implements `dcd_sof_enable` 

**Additional context**
Placed PID_SOF at the top since this is generally the most time critical if we are looking at the SOF packet.

PR #2392, opened 2023-12-28
bugfix(ch32-hs-dcd): fix ch32 DATAx managment and long packet transmit

Krasutski:
Old implementations have some issues:
 - EP0 OUT transfer just copied data from the buffer without any checking
 - TOG1 bit set not correctly for OUT xfers
 - Long packets(more than ep size) xfer incorrectly
 
Fixed this issue #1891 also other examples now work stable

UPD:
Pushed one more commit, fixed iso transfers, probably this issue fixed: #2325


zx96:
This MR seems to resolve issues I was having getting writes to work reliably with the MSC device class. 👍

WilheJo:
My problem in #2377 is also fixed with this MR!

Krasutski:
@hathach Could you check and merge PR?

hathach (commented):
still haven't got time to test this out. I am off for TET holidays for a couple of weeks. Will try to review afterwards, thank you for your patient.

Krasutski:
@hathach can help somehow to complete this PR?

HiFiPhile:
Thank you for your great work, although I don't have CH32 device to test with...
I just merged upstream SOF support to make test easier for other people, please take a look if it's good for you.

hathach (approved):
superb ! thank you very much for the fix as well as ISO support. I have rebased, merge interrupt handler for ISO and Transfer (since they are similar), also reformat for the code, and add some minor clean up. All is good, tested with v307. Will merge when ci passed.

HiFiPhile:
Just it seems dcd_event_sof is lost on merge.

hathach:
> Just it seems dcd_event_sof is lost on merge.

oops, sorry, let me double check that

src/portable/wch/dcd_ch32_usbhs.c
```diff
@@ -33,372 +34,382 @@
...
 
-void dcd_sof_enable(uint8_t rhport, bool en)
-{
-    (void) rhport;
-    if (en) {
-        USBHSD->INT_EN |= USBHS_SOF_ACT_EN;
-    } else {
-        USBHSD->INT_EN &= ~(USBHS_SOF_ACT_EN);
-    }
+void dcd_sof_enable(uint8_t rhport, bool en) {
```
hathach, line 211: @HiFiPhile does this look good ?

HiFiPhile: dcd_sof_enable() looks good, only dcd_event_sof() is missing.

hathach: HOST SOF bit in the interrupt imply that this is only for host most, not sure if it is also used for SOF detection in device mode 
image

hathach: @HiFiPhile does this look similar to your previous changes (sorry to loose it while rebasing)

image

HiFiPhile: Sof was added in #2181,  `dcd_event_sof(rhport, USBHSD->FRAME_NO, true);` is better as it has frame number.

hathach: thank you, I figure that out, but we should mask 11 bit only, since upper 3 bit is micro-frame index which actually increased faster than frame count

image

hathach: yeah, we probably need to normalize the micro-frame/frame counter, maybe using the top 3 msb bit in uint32_t for application, or just change the tud_sof_cb() to have an typedef struct. I will do that with follow-up PR later on.

image

PR #2395, opened 2024-01-05
CDC device line state notification

gguichard:
Allow a CDC-ACM device to send a line state notification to the host (DCD/DSR/RING etc).

It has been tested using statserial command on Linux host side to verify UART-like line status.
It looks like that the notification endpoint size must be at least 10 bytes in order to hold the entire notification packet (which means this won't work for USB low speed devices).

### DIFF
--- a/src/class/cdc/cdc_device.c
+++ b/src/class/cdc/cdc_device.c
@@ -128,6 +128,36 @@ uint8_t tud_cdc_n_get_line_state (uint8_t itf)
   return _cdcd_itf[itf].line_state;
 }
 
+bool tud_cdc_n_notif_line_state (uint8_t itf, uint8_t state)
+{
+  cdcd_interface_t* p_cdc = &_cdcd_itf[itf];
+
+  TU_VERIFY(tud_ready());
+
+  uint8_t const rhport = 0;
+
+  // claim endpoint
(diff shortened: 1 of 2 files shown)


tlyu:
I'm commenting as an observer who's delved into the relevant parts of the USB specs for a while.

> endpoint size must be at least 10 bytes in order to hold the entire notification packet (which means this won't work for USB low speed devices).

I'm pretty sure that low-speed USB devices aren't allowed to have bulk endpoints, which rules out typical CDC ACM implementations, so this concern probably doesn't apply.

gguichard:
> I'm pretty sure that low-speed USB devices aren't allowed to have bulk endpoints, which rules out typical CDC ACM implementations, so this concern probably doesn't apply.

Perfect! Thanks for the info. :)

SmittyHalibut:
Oh hey look!  Someone else is trying to do the same thing I am, but on the upstream library!

https://github.com/adafruit/Adafruit_TinyUSB_Arduino/issues/491

I've submitted a similar PR against Adafruit_TinyUSB_Arduino.

hathach:
implemnted by https://github.com/hathach/tinyusb/pull/2593, thank you very much for brining this feature into our attention.

PR #2397, opened 2024-01-06
Add clang to compiler.h

Ryzee119:
**Describe the PR**
I'm using the clang compiler which for our purposes the compiler attributes match GCC from what I can read.

Although `TU_ATTR_WEAK` had to be changed to `__attribute__ ((weak_import))`  or I was getting `lld: error: duplicate symbol` errors on all the `TU_ATTR_WEAK`  functions.


src/common/tusb_compiler.h
```diff
@@ -175,6 +175,11 @@
 		#pragma GCC poison tud_vendor_control_request_cb
 	#endif
 
+  #if defined (__clang__)
```
hathach, line 178: can you do the #ifdef #else instead of undef (ust move this up). undef is rather confusing since people may only look at the first define.

Ryzee119: Just to clarify @hathach 

Like this?
```
#if defined(__GNUC__) || defined (__clang__)
  #define TU_ATTR_ALIGNED(Bytes)        __attribute__ ((aligned(Bytes)))
  #define TU_ATTR_SECTION(sec_name)     __attribute__ ((section(#sec_name)))
  #define TU_ATTR_PACKED                __attribute__ ((packed))
  #if defined(__GNUC__)
    #define TU_ATTR_WEAK                  __attribute__ ((weak))
  #else
    #define TU_ATTR_WEAK                  __attribute__ ((weak_import))
  #endif
... 
...
```

hathach: yeah, but the otherway around

```C
  #if defined(__clang__)
    #define TU_ATTR_WEAK                  __attribute__ ((weak_import))
  #else
    #define TU_ATTR_WEAK                  __attribute__ ((weak))
  #endif
```

Ryzee119: All fixed :)

hathach: unit test probably compile using clang, and has affected by this attribute. I will check later.

hathach (commented):
still haven't got time to test this out, will be off for a couple of week for TET holidays. Please be patient, thank you.

hathach:
@Ryzee119 I am back from TET holidays, which compiler you are using. If it is arm clang, I could give it a try (was planning to support it anwyay).

hathach (commented):
recently I switch to use weak as the default implementation of callback to be compatible with keil #2239

https://github.com/hathach/tinyusb/blob/master/src/device/usbd.c#L48

the other callback header can work with **weak_import**, but this style seems to require actual **weak** and cause the compile issue with unit-test (since test include mocking of dcd layer). 

Maybe I should use weakref/alias or unify them all to default implemenation, will revise later. We probably need to keep this open until then. Thank you.

HiFiPhile:
How it goes with #2606 ?

hathach:
so far I only tested arm-clang with cdc_msc example, it doesn't seem to need weak_import yet, but I admit we need more testinng. the conlfict with unit-test is resolved (unit test switch back to gcc)

Ryzee119:
So sorry i missed this discussion somehow!

I'm using clang + llvm cross compiling for x86 type embedded system with OHCI backend.


> recently I switch to use weak as the default implementation of callback to be compatible with keil https://github.com/hathach/tinyusb/pull/2239

^ Yep this does the trick for me. I was having problems with the OHCI tusb_app_virt_to_phys/tusb_app_phys_to_virt weak functions.

So I just need this patch and implement weak stub functions.
```
+++ b/src/common/tusb_compiler.h
@@ -123,7 +123,7 @@

 // TODO refactor since __attribute__ is supported across many compiler
-#if defined(__GNUC__)
+#if defined(__GNUC__) || defined (__clang__)
```

```
diff --git a/src/portable/ohci/ohci.c b/src/portable/ohci/ohci.c
index c59d475..715ac63 100644
--- a/src/portable/ohci/ohci.c
+++ b/src/portable/ohci/ohci.c

 TU_ATTR_ALWAYS_INLINE static inline void *_phys_addr(void *virtual_address)
 {
-  if (tusb_app_virt_to_phys) return tusb_app_virt_to_phys(virtual_address);
-  return virtual_address;
+  return tusb_app_virt_to_phys(virtual_address);
 }

 TU_ATTR_ALWAYS_INLINE static inline void *_virt_addr(void *physical_address)
 {
-  if (tusb_app_phys_to_virt) return tusb_app_phys_to_virt(physical_address);
-  return physical_address;
+  return tusb_app_phys_to_virt(physical_address);
 }
 ```

Ill probably supersede this PR.  I need to stick these two stubs somewhere for my needs. **Can you suggest a good location?**

```
TU_ATTR_WEAK void* tusb_app_virt_to_phys(void *virt_addr) {
  return virt_addr;
}

TU_ATTR_WEAK void* tusb_app_phys_to_virt(void *phys_addr) {
  return phys_addr;
}
```

hathach:
@Ryzee119 all weak callback is now migrated to use default empty implementation to better support keil/armcc. I guess we can safely close this now ?

Ryzee119:
@hathach that's great. That should fix any outstanding issues I had.

We may just need the  defined(__clang__) check on GNUC so clang compiler will use those definitions which should all be compatible now

I will confirm this soon

hathach:
@Ryzee119 do you have time to confirm that weak is working now with clang :)

hathach:
I guess it is safe to close this for now. Should it be an issue, feel free to re-open it. Thank you.

PR #2391, opened 2023-12-26
initial support of CH34x CDC device

IngHK:
this is the initial version to support the CH34x CDC device.
the CH34x is as popular as FTDI or CH34x.
this version supports only RX and TX and the line_coding settings.
modem control handling will follow...

PR #2407, opened 2024-01-09
[CDC] FTDI host: store requested baudrate per device

IngHK:
before the requested baudrate was stored in global variable _ftdi_requested_baud.
for clean baudrate setting it's necessary to store the requested per device.


hathach:
superceded by #2391

PR #2408, opened 2024-01-10
Disable ULPI clock during sleep on stm32f7 when using internal phy

leptun:
This PR addresses a bug I discovered on stm32f7 when using the OTG_HS dwc2 peripheral with the internal FS phy while also using tickless idle with freertos. The problem is identical to the issue described for the H7 series, but on the F7 different bit names are used. This patch also disables the ULPI clock during tickless idle for stm32f4 and stm32f2 series microcontrollers that have the OTG_HS peripheral. While I did not test the other MCUs, I don't see how this could harm since with this patch, the ULPI clock is disabled during sleep only if the ULPI interface is not used.
I tested this on an stm32f722ze with OTG_HS connected in FS mode with the internal phy in device mode. The examples do not use such a configuration afaik.


hathach (approved):
look good, thank you.

PR #2410, opened 2024-01-11
added useful log of vid/pid, added missing CRLF

IngHK:
added useful log of vid/pid in tuh_vid_pid_get()
added missing CR/LF after TU_LOG_MEM


hathach (commented):
not accepted, should be done by enumeration proccess when detecting getting device descriptor from newly attached device.

PR #2413, opened 2024-01-12
Support for TUSB_REQ_FEATURE_TEST_MODE

shuchitak:
**Describe the PR**
Adds support for TUSB_REQ_FEATURE_TEST_MODE feature in usbd.c

**Additional context**
Added callback functions tud_test_mode_req_status_xfer_cb(), usbd_set_test_mode() and dcd_set_test_mode() that are
called to transition the USB device into the electrical compliance test mode.

### DIFF
--- a/src/device/dcd.h
+++ b/src/device/dcd.h
@@ -146,6 +146,8 @@ void dcd_disconnect(uint8_t rhport) TU_ATTR_WEAK;
 // Enable/Disable Start-of-frame interrupt. Default is disabled
 void dcd_sof_enable(uint8_t rhport, bool en);
 
+// Transition device to test mode
+void dcd_set_test_mode(uint16_t test_mode) TU_ATTR_WEAK;
 //--------------------------------------------------------------------+
 // Endpoint API
 //--------------------------------------------------------------------+
--- a/src/device/usbd.c
+++ b/src/device/usbd.c
@@ -719,13 +719,34 @@ static bool process_control_request(uint8_t rhport, tusb_control_request_t const
 
         case TUSB_REQ_SET_FEATURE:
           // Only support remote wakeup for device feature
-          TU_VERIFY(TUSB_REQ_FEATURE_REMOTE_WAKEUP == p_request->wValue);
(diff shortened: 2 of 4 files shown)


shuchitak:
> I was also working on a test mode support for TinyUSB. 😃 _You can check it out here: #2416_ Maybe we can work together so that this feature can be added.
> 
> My comments about your implementation:
> 
> * According to the USB 2.0 specification the behavior of the device is not specified if `wLength` is non-zero. So there should be no reason to filter for that. So I think it's better to just ignore it.
> * TinyUSB should not acknowledge the request and rather stall it if the specific test mode isn't supported. Like there is no hardware support for GigaDevice MCUs (at least for the ones I'm working with). _So the support needs to be emulated in host mode. Then it's possible that may only some modes are manually implemented._
> * It seems you forgot the call to enter the test mode when the request is handled. _I would expect that the callback would call it. I don't see why this should depend on a specific implementation. Only `dcd_set_test_mode` would be MCU related._
> * ~The test mode selector from the request isn't stored anywhere so the callback wouldn't be aware which one should be entered.~ _I forgot that the request is passed through._
> 
> **I like the control callback handling, better than my event workaround.**

Hi @Rocky04, I saw your PR. I agree with your comments about not handling wLength=0 and returning stall if the device doesn't handle test mode. 
I had handled the transition to test mode in a callback function since that involved adding fewer lines of code. I'd confirmed with my device that this scheme was working.
I'm happy to close this PR and continue with yours. I'll pull your changes and confirm that the test modes work as expected on my device. Thanks!

Rocky04:
I borrowed some of your ideas and already changed my implementation.

The entering of the test mode needs to be done after the request has finished. Using the control complete callback is a great way of ensuring that. Otherwise the device is in the test mode before the host got the acknowledge. So that the test mode already took over the signaling and the host request times out.

_I already saw people who do signaling tests only relying on the shown status of the compliance tool. They thought it was working only because the tool stated a success even the device never actually entered the test mode, only because it had blindly acknowledge the request._

shuchitak:
Closing since duplicate of https://github.com/hathach/tinyusb/pull/2416

PR #2414, opened 2024-01-14
added CDC specific debug logs

IngHK:
added CDC specific debug logs in new file cdc_debug.h
initial version with logging of line_coding, baudrate and control line_stateaccepted
added to CDC host and tested
added to CDC device, but inactive (#if 0) and to test later


hathach:
please do your own local debug, these changes are not accepted

PR #2420, opened 2024-01-21
msc_device: Fix check for including MSC lookup tables

### LINKED ISSUE #2419: Build fails when `CFG_TUSB_DEBUG` > 0 and `CFG_TUD_MSC_LOG_LEVEL` < 2
### Operating System

Others

### Board

N/A

### Firmware

N/A

### What happened ?

When building with `CFG_TUSB_DEBUG=1` and `CFG_TUD_MSC_LOG_LEVEL=1`, I would expect that logging is enabled for `msc_device.c`. However, the build fails [on line 418](https://github.com/hathach/tinyusb/blob/938cae818f5f53887ad12e34189e6746dbce4889/src/class/msc/msc_device.c#L418) due to lookup tables being defined [only when `CFG_TUSB_DEBUG >= 2`](https://github.com/hathach/tinyusb/blob/938cae818f5f53887ad12e34189e6746dbce4889/src/class/msc/msc_device.c#L206).

This should be trivially fixable by just substituting `CFG_TUD_MSC_LOG_LEVEL` in place of the hardcoded `2`.

### How to reproduce ?

In `tusb_config.h`:

```c
#define CFG_TUSB_DEBUG 1
#define CFG_TUD_MSC_LOG_LEVEL 1
```

N/A

### Screenshots

_No response_

zx96:
The `_msc_scsi_cmd_lookup` and `_msc_scsi_cmd_table` variables are needed when logging is enabled for the MSC device via `CFG_TUD_MSC_LOG_LEVEL`. Update the preprocessor check around them to use that definition when deciding whether to define those variables.

Closes #2419

### DIFF
--- a/src/class/msc/msc_device.c
+++ b/src/class/msc/msc_device.c
@@ -203,7 +203,7 @@ uint8_t rdwr10_validate_cmd(msc_cbw_t const* cbw)
 //--------------------------------------------------------------------+
 // Debug
 //--------------------------------------------------------------------+
-#if CFG_TUSB_DEBUG >= 2
+#if CFG_TUSB_DEBUG >= CFG_TUD_MSC_LOG_LEVEL
 
 TU_ATTR_UNUSED tu_static tu_lookup_entry_t const _msc_scsi_cmd_lookup[] =
 {


hathach (approved):
thank you for the fix

PR #2421, opened 2024-01-22
Log outbound xfer bytes while in debug

antoniovazquezblanco:
**Describe the PR**
This simple PR logs outbound xfer bytes when log level is set to debug. It is useful to be able to dissect the full conversation and not only incoming packets.
While this may be too verbose, it is quite useful when things go south.

**Additional context**
This is really nice in combination with a simple USB dissector. I am currently using scapy to parse the logs. I may create a PR to scapy with the relevant USB layers and may post an example on how to use them in a future PR.

### DIFF
--- a/src/device/usbd.c
+++ b/src/device/usbd.c
@@ -1232,6 +1232,11 @@ bool usbd_edpt_xfer(uint8_t rhport, uint8_t ep_addr, uint8_t * buffer, uint16_t 
   // TU_VERIFY(tud_ready());
 
   TU_LOG_USBD("  Queue EP %02X with %u bytes ...\r\n", ep_addr, total_bytes);
+#if CFG_TUD_LOG_LEVEL >= 3
+  if(tu_edpt_dir(ep_addr) == TUSB_DIR_IN) {
+    TU_LOG_MEM(CFG_TUD_LOG_LEVEL, buffer, total_bytes, 2);
+  }
+#endif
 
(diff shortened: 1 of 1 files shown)


hathach:
It is helpful but too verbose, it could introduce other timing issues. Especially with bulk transfer like msc class. We should increase the log level to at least 3 for this info.

antoniovazquezblanco:
Done! :D

src/device/usbd.c
```diff
@@ -1232,6 +1232,8 @@ bool usbd_edpt_xfer(uint8_t rhport, uint8_t ep_addr, uint8_t * buffer, uint16_t
   // TU_VERIFY(tud_ready());
 
   TU_LOG_USBD("  Queue EP %02X with %u bytes ...\r\n", ep_addr, total_bytes);
+  TU_LOG3_BUF(buffer, total_bytes);
```
hathach, line 1235: must check if this is an IN transfer, otherwise, data contents is meaningless. It must be wrapped in `#if CFG_TUD_LOG_LEVEL >= 3`

hathach (changes requested):
checkout review for request changes

antoniovazquezblanco:
Made the requested changes. Also improved a bit on the previous solution.

Thank you for being so patient with this minuscule PR. it is helping me learn more about the project!

hathach (approved):
look good, thank you. I update it a bit to use dir variable, will merge when ci passed

PR #2422, opened 2024-01-23
audio_device: Fix unused variable warning

kasjer:
**Describe the PR**
In some configurations local variable `p_desc_parse_for_params `is declared and never used resulting in warning that can be escalated to build error (for mynewt)

Now variable is surrounded with same preprocessor condition as function that uses it `audiod_parse_for_AS_params()`


HiFiPhile:
Thanks, looks good.

HiFiPhile (approved):

PR #2428, opened 2024-01-26
update video_capture 

YixingShen:
1. add tx_busy  = 1 into examples/device/video_capture/src/main.c because "if (tx_busy)" is always false.
2. Correct VIDEO_CS_VS_INTERFACE_FRAME_MJPEG to VIDEO_CS_ITF_VS_FRAME_MJPEG for #define TUD_VIDEO_DESC_CS_VS_FRM_MJPEG_DISC


kkitayam:
Thank you for fixing `tx_busy`. It looks good for me. But I am not familiar with stm32f1, so I cannot comment on the changing related to it.

hw/bsp/stm32f1/family.c
```diff
@@ -57,6 +57,9 @@ void board_init(void) {
   __HAL_RCC_GPIOC_CLK_ENABLE();
   __HAL_RCC_GPIOD_CLK_ENABLE();
 
+  // Enable UART Clock
```
hathach, line 60: please revert the change to this file

hathach (approved):
thank you for the fix

PR #2429, opened 2024-01-26
CDCh PL2303 support

IngHK:
initial edition to support Prolific PL2303
tests are ongoing


src/tusb_option.h
```diff
@@ -489,6 +489,93 @@
...
+#ifndef CFG_TUH_CDC_PL2303
+  // PL2303 is not part of CDC class, only to re-use CDC driver API
+  #define CFG_TUH_CDC_PL2303 0
+#endif
+
+#ifndef CFG_TUH_CDC_PL2303_VID_PID_QUIRKS_LIST
+  // List of product IDs that can use the PL2303 CDC driver
+  #include "class/cdc/serial/pl2303.h"
+  #define CFG_TUH_CDC_PL2303_VID_PID_QUIRKS_LIST \
+  { PL2303_VENDOR_ID, PL2303_PRODUCT_ID, PL2303_QUIRK_ENDPOINT_HACK }, \
```
hathach, line 501: whatt is `PL2303_QUIRK_ENDPOINT_HACK` hack for ?

IngHK: oh, that's a good question...
I initially adopted it from Linux.
There seems to have been derivatives in the early days of the PL2303, where the EPs of one CDC port were distributed over two interfaces. This is no longer relevant for current derivatives. I removed these quirks.
And in pl2303_open() it is checked anyway whether the correct number of EPs is available.

I also removed PL2303_QUIRK_UART_STATE_IDX0 because it only affects the interrupt EP, which we don't use.

We should only take what is currently being sold and what we can test...

IngHK:
The implementation is now finished from my side.
All tests with various PL2303s on my desk were successful.
Now it is ready for review and merge to master

src/class/cdc/cdc_host.c
```diff
@@ -36,6 +37,19 @@
 
 #include "cdc_host.h"
 
+#if CFG_TUH_CDC_FTDI
```
hathach, line 40: is there any reason to move it here, otherwise please keep these header in there section

hathach (commented):
thank you for another great PR for serial host. Unfortunate, Lunar New Year (aka TET) is super close and I don't have time to review/test out this PR in a few weeks or so. Meanwhile, would you mind fixing the CI build with other ports. It is probably suppress warnings since some ports has stricter warning flags than others. Thak you

src/class/cdc/cdc_host.c
IngHK: yes
```
  #if CFG_TUH_CDC_PL2303
    pl2303_t pl2303;
  #endif
```
is added to `cdch_interface_t`. so `#include "serial/pl2303.h"` need to be before `typedef struct ... cdch_interface_t`.
And so I moved all includes to top of file for uniformity.

IngHK:
> thank you for another great PR for serial host. Unfortunate, Lunar New Year (aka TET) is super close and I don't have time to review/test out this PR in a few weeks or so. Meanwhile, would you mind fixing the CI build with other ports. It is probably suppress warnings since some ports has stricter warning flags than others. Thak you

OK, thanks for your hint, I fixed the errors/warnings

IngHK:

Issue #2437, opened 2024-01-29
CDCh tuh_cdc_mounted() returns true before process config is completed (also HIDh?)


IngHK:
### Operating System

Others

### Board

-

### Firmware

-

### What happened ?

If an application uses `tuh_cdc_mounted()` to check whether a CDC device is mounted, it will be reported "true" **before** the process config is completed.
The application could then disrupt the process config or the application would not yet be able to work with the CDC device.
`tuh_cdc_mount_cb()` is correctly called in `set_config_complete()` **after** the process config has been completed.
the hid host may have the same issue...

### How to reproduce ?

-

### Debug Log as txt file (LOG/CFG_TUSB_DEBUG=2)

-

### I have checked existing issues, dicussion and documentation

- [X] I confirm I have checked existing issues, dicussion and documentation.

PR #2442, opened 2024-01-30
HIDh tuh_hid_mounted() returns true before process set config is completed

IngHK:
fix similar to Issue #2437 and PR #2438


hathach (approved):
look good, thank you

PR #2449, opened 2024-02-01
fix #2448 CH34x ch34x_set_line_coding() callback bug

### LINKED ISSUE #2448: CH34x ch34x_set_line_coding() callback bug
### Operating System

Linux

### Board

-

### Firmware

-

### What happened ?

the callbacks are not handled correct
PR will follow soon

### How to reproduce ?

-

-

### Screenshots

_No response_

IngHK:
- the two-stages `ch34x_set_line_coding()` forgets the requested `complete_cb` to be called after 2nd stage. so `complete_cb` is now stored in `p_cdc->requested_complete_cb`
- another bug: missing `p_cdc->user_control_cb = complete_cb;` in `ch34x_set_data_format()`

### DIFF
--- a/src/class/cdc/cdc_host.c
+++ b/src/class/cdc/cdc_host.c
@@ -67,6 +67,9 @@ typedef struct {
   #endif
 
   tuh_xfer_cb_t user_control_cb;
+  #if  CFG_TUH_CDC_CH34X
+  tuh_xfer_cb_t requested_complete_cb;
+  #endif
 
   struct {
     tu_edpt_stream_t tx;
@@ -1380,6 +1383,8 @@ static bool ch34x_set_data_format(cdch_interface_t* p_cdc, uint8_t stop_bits, ui
 
   uint8_t const lcr = ch34x_get_lcr(stop_bits, parity, data_bits);
   TU_VERIFY(lcr);
+
+  p_cdc->user_control_cb = complete_cb;
   TU_ASSERT (ch34x_control_out(p_cdc, CH34X_REQ_WRITE_REG, CH32X_REG16_LCR2_LCR, lcr,
                                complete_cb ? ch34x_control_complete : NULL, user_data));
   return true;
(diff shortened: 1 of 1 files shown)


hathach:
> * the two-stages `ch34x_set_line_coding()` forgets the requested `complete_cb` to be called after 2nd stage. so `complete_cb` is now stored in `p_cdc->requested_complete_cb`

Have you tried passing the callback and verify that it is not invoked ? 

> * another bug: missing `p_cdc->user_control_cb = complete_cb;` in `ch34x_set_data_format()`

I remember don't assign calback for this intentionaly.

IngHK:
> > * the two-stages `ch34x_set_line_coding()` forgets the requested `complete_cb` to be called after 2nd stage. so `complete_cb` is now stored in `p_cdc->requested_complete_cb`
> 
> Have you tried passing the callback and verify that it is not invoked ?
> 
> > * another bug: missing `p_cdc->user_control_cb = complete_cb;` in `ch34x_set_data_format()`
> 
> I remember don't assign calback for this intentionaly.

Yes, and this is the starting point for all further issues.
`tuh_cdc_set_data_format()` is a public Control Endpoint (Request) API function.
So it can be expected, it can handle `complete_cb`.
`ch34x_set_data_format()` calls `ch34x_control_out()` with `complete_cb ? ch34x_control_complete : NULL`, but there is no `p_cdc->user_control_cb = complete_cb;` before.
Then `ch34x_control_complete ()` calls not the wanted `complete_cb` but a `p_cdc->user_control_cb` from any previous API function call.
The missing `p_cdc->user_control_cb = complete_cb;` is the first issue.

After adding `p_cdc->user_control_cb = complete_cb;` to `ch34x_set_data_format()` `ch34x_set_line_coding_stage1_complete()` calls `ch34x_set_data_format()` with `ch34x_control_complete` as call back.
Then it runs into an endless call loop calling `ch34x_control_complete`.
That's the follow-up issue.

hathach:
thanks for confirmation, I remembmere specially write so that stage2 complete will invoke user callback. Let me double check that later, though I am not sure if I could do that before TET. Otherwise, please be patient and wait for a couple of weeks or more.

IngHK:
have a nice and relaxing time!

hathach (changes requested):
you are right, currently the ch34x_set_data_format() didn't store user callback. Though we don't need to add a new requested_complete_cb variable for this purpose. We only need to rename current ch34x_set_data_format() to an internal function (similar to `ch34x_write_reg_baudrate()`), and add user_control store for the public driver API

IngHK:
> you are right, currently the ch34x_set_data_format() didn't store user callback. Though we don't need to add a new requested_complete_cb variable for this purpose. We only need to rename current ch34x_set_data_format() to an internal function (similar to `ch34x_write_reg_baudrate()`), and add user_control store for the public driver API

OK, done

IngHK:

PR #2451, opened 2024-02-01
CP210x CDCh upgrade

IngHK:
_upgraded_ driver:
- updated vid/pid list
- added misc. defines
- added check of valid baudrates
- implemented cp210x_set_data_format()
- implemented cp210x_set_line_coding()
- adjusted cp210x_process_config()


IngHK:

PR #2459, opened 2024-02-07
CDCh ACM fixes and small changes

IngHK:
- added check of `xfer->result` in `acm_process_config()`
- missing `TU_VERIFY(p_cdc->acm_capability.support_line_request);` in `acm_set_data_format()` => moved it to common `acm_set_line_coding()`
- added missing check of data bit quantity in `acm_set_line_coding()`
- better explanation of `CFG_TUH_CDC` in misc. example `tusb_config.h`
- TU_LOGs cleaned and harmonized with other drivers


IngHK:

PR #2463, opened 2024-02-12
nrf5x: Handle ISOOUT CRC errors

kasjer:
**Describe the PR**
NRF5x USB controller can detect ISO OUT CRC errors. In such case USBEVENT is signaled with EVENTCAUSE_ISOOUTCRC set. Even if controller detects corrupted ISO OUT packet it allows to data transfer from endpoint to RAM however packet is corrupted and code could just as well drop packet altogether.

With current implementation incoming ISO OUT packets were put in FIFO and exact information how much data already in FIFO is correct was hard to keep track of.
If was observed that on certain configurations HS hub when FS device was connected occasionally sent invalid (short) packet. In such case if packet length was reported odd audio stream was not recognizable any more.

With this change corrupted packets are not passed to upper layers and are silently dropped.

**Additional context**
Change touches only ISO OUT related code path.

This was observed on NRF5340 that was connected to Windows 11 PC with docking station and DELL monitor hub
image

In this setup there were several level of HUB's involved (presumable Super Speed).
Invalid packet (167 bytes) was received and detected by USB following packet (22 bytes) was not reported.

It may be that this happens for for packets of 192 Bytes (48kHz PCM audio).
Previously when audio was 32kHz similar situation was not observed.

### DIFF
--- a/src/portable/nordic/nrf5x/dcd_nrf5x.c
+++ b/src/portable/nordic/nrf5x/dcd_nrf5x.c
@@ -653,7 +653,11 @@ void dcd_int_handler(uint8_t rhport)
     if (NRF_USBD->EPOUTEN & USBD_EPOUTEN_ISOOUT_Msk)
     {
       iso_enabled = true;
-      xact_out_dma(EP_ISO_NUM);
+      // Transfer from endpoint to RAM only if data is not corrupted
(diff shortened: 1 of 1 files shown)


hathach (approved):
superb ! thank you for fixing all the iso relatead issue with nrf so far

PR #2468, opened 2024-02-13
Added HCD for stm32f4/f7. Updated porting documentation.

supersmiler:
- Added HCD for stm32f4
- Added HCD for stm32f7
- Added documentation on how to port the host side.

Special thanks to Prodrive Technologies for giving me time to create this and push it back to the community.

### DIFF
--- a/src/portable/st/stm32_hal_driver/hcd_stm32_f7.c
+++ b/src/portable/st/stm32_hal_driver/hcd_stm32_f7.c
@@ -1,0 +1,507 @@
+/*
+ * The MIT License (MIT)
+ *
+ * Copyright (c) 2024 Ha Thach (tinyusb.org)
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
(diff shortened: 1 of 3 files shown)


src/portable/st/stm32_hal_driver/hcd_stm32_f7.c
```diff
@@ -0,0 +1,507 @@
...
+	(void) rhport;
+	(void) dev_addr;
+	(void) ep_addr;
+	return true;
+}
+
+/*--------------------------------------------------------------------+
+ * Weak HAL functions
+ *--------------------------------------------------------------------+*/
+void HAL_HCD_MspInit(HCD_HandleTypeDef *handle)
```
cdesjardins, line 288: This PR looks great, much needed in my opinion, but these two functions (MspInit and Deinit) shouldn't be here, they should be taken care of in board_init() and not part of this driver.

cdesjardins:
I think it would be sufficient to have just hcd_stm32.c as it will also work on h series processors.

HiFiPhile:
Thanks for your effort to make HCD works on SMT32.

- Both F4 and F7 use synopsys dwc2 IP, so they should be generalized
- Implementation shouldn't base on ST's HAL driver, as you can see there is no HAL dependency in `dcd_dwc2.c`

Anyway it's a good starting point.

hathach:
native hcd driver for dwc2 is implemented by https://github.com/hathach/tinyusb/pull/2870

PR #2472, opened 2024-02-15
Do not enable the Mode mismatch interrupt source, it is not cleared

cdesjardins:
If this interrupt ever hits it will result in an infinite interrupt loop as it is never cleared.


HiFiPhile (approved):
LGTM, hit it once, in STM32Cube it's not enabled.

PR #2473, opened 2024-02-16
use rt-thread 5+ mq recv api

kaidegit:
**Describe the PR**

RT-Thread has modified there message queue api on version 5.0.0, which may unexpectly cause function `tud_task_ext` return without processing usb message.

**Additional context**

See PR at RT-Thread
https://github.com/RT-Thread/rt-thread/pull/7709


Rbb666:
I have verified that there is no problem with this PR.

HiFiPhile (approved):
Thanks for your PR.

PR #2481, opened 2024-02-19
Fix nRF5340 DK USB host shield build

xudongzheng:
- Fix incorrect UART pins
- Enable MAX3421 host shield
- Use GPIOTE1_IRQn instead of GPIOTE_IRQn


hathach (approved):
look good, thank you

PR #2490, opened 2024-02-24
fixed compiler warning src/tusb.c

IngHK:
```
../tinyusb/src/tusb.c: In function 'tu_print_mem':
../tinyusb/src/tusb.c:480:16: warning: conversion from 'int' to 'char' may change value [-Wconversion]
  480 |   format[2] += 2*size;
      |                ^
```


hathach (approved):
look good, thank you

PR #2494, opened 2024-02-26
Add notification support for device class USBTMC

tommie:
The `ep_int_in` is already used for responding to USB488 `READ_STATUS_BYTE` requests, but that EP is defined for the entire USBTMC class. This extends the functionality to let callers send notifications and receive ACKs.

- USBTMC class and USB488 subclass specs: https://www.usb.org/document-library/test-measurement-class-specification
- USBTMC section 3.4 describes the interrupt-IN EP.

Tested on an RP2040.

### DIFF
--- a/src/class/usbtmc/usbtmc_device.c
+++ b/src/class/usbtmc/usbtmc_device.c
@@ -240,6 +240,18 @@ bool tud_usbtmc_transmit_dev_msg_data(
   return true;
 }
 
+bool tud_usbtmc_transmit_notification_data(const void * data, size_t len)
+{
+#ifndef NDEBUG
+  TU_ASSERT(len >= 1);
+  TU_ASSERT(usbtmc_state.ep_int_in != 0);
+#endif
+  if (usbd_edpt_busy(usbtmc_state.rhport, usbtmc_state.ep_int_in)) return false;
+
+  TU_VERIFY(usbd_edpt_xfer(usbtmc_state.rhport, usbtmc_state.ep_int_in, (void *)data, len));
+  return true;
+}
+
 void usbtmcd_init_cb(void)
 {
   usbtmc_state.capabilities = tud_usbtmc_get_capabilities_cb();
@@ -578,7 +590,9 @@ bool usbtmcd_xfer_cb(uint8_t rhport, uint8_t ep_addr, xfer_result_t result, uint
     }
   }
   else if (ep_addr == usbtmc_state.ep_int_in) {
-    // Good?
+    if (tud_usbtmc_notification_complete_cb) {
+      TU_VERIFY(tud_usbtmc_notification_complete_cb());
+    }
     return true;
   }
   return false;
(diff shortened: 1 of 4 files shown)


src/class/usbtmc/usbtmc_device.c
```diff
@@ -240,6 +240,18 @@ bool tud_usbtmc_transmit_dev_msg_data(
...
 
+bool tud_usbtmc_transmit_notification_data(const void * data, size_t len)
+{
+#ifndef NDEBUG
+  TU_ASSERT(len >= 1);
+  TU_ASSERT(usbtmc_state.ep_int_in != 0);
+#endif
+  if (usbd_edpt_busy(usbtmc_state.rhport, usbtmc_state.ep_int_in)) return false;
+
+  TU_VERIFY(usbd_edpt_xfer(usbtmc_state.rhport, usbtmc_state.ep_int_in, (void *)(uintptr_t) data, len));
```
tommie, line 251: Fixed the failing constness check by casting to uintptr_t, as done in other places.

HiFiPhile:
Not a TMC expert, just to mention there is no buffering inside `usbd_edpt_xfer()`, you need to add buffer like how bulk endpoints did:
https://github.com/hathach/tinyusb/blob/9e674fa109348d97fb4280ac534be665f619a4d8/src/class/usbtmc/usbtmc_device.c#L129

tommie:
> Not a TMC expert, just to mention there is no buffering inside `usbd_edpt_xfer()`, you need to add buffer like how bulk endpoints did:
> 
> https://github.com/hathach/tinyusb/blob/9e674fa109348d97fb4280ac534be665f619a4d8/src/class/usbtmc/usbtmc_device.c#L129

Yes, I documented that the buffer must remain valid until the `notification_complete` callback, since that's what `tud_usbtmc_transmit_dev_msg_data` already does:

https://github.com/hathach/tinyusb/pull/2494/files#diff-0c276774d2aac9050bffd9993a595ad46c6e04638f9f2aaa9fb0c5ab76b973eaR100

Re. TMC, I had to ditch it for two reasons:

1. WebUSB can't release the kernel driver, so claiming the interface requires OS configuration shenanigans. (And according to WebUSB folks, doing more is out of scope for WebUSB.)
2. I'd love to change the TinyUSB TMC API to not require the full message to `tud_usbtmc_transmit_dev_msg_data`: if the bulk-IN message is composed of multiple parts located in different parts of memory, having to duplicate them all at once just for transfer seems wasteful. Especially since USBTMC only needs enough buffer for one USB transfer at a time.

In the end, it seemed TMC as a spec was more in the way than useful.

HiFiPhile:
I just realized the lacking of buffer copying in bulk IN transfer is problematic, as some MCU require buffer to be placed into a special RAM region and has alignment requirement. To ensure compatibility I had to add it.

For this reason I've added a buffer for INT transfer with size configured by `CFG_TUD_USBTMC_INT_EP_SIZE`.

HiFiPhile (approved):

PR #2497, opened 2024-02-27
sparkfun_samd21_mini_usb board support

ChrisDeadman:
Adds support for the [SparkFun SAMD21 Mini Breakout](https://www.sparkfun.com/products/13664)

It has different TX/RX pins for serial, so I had to do some preprocessor shenanigans in `family.c`.
Let me know if there's a better way to do this and I will change it - I'm quite new to TinyUSB.


hathach (approved):
look great thank you

PR #2514, opened 2024-03-11
Changes required for new SDK 3.2.0

camilaodsouza:
- Define EP_ALLOCREQ
 - Define EP_FREEREQ
 - Define EP_ALLOCBUFFER
 - Define EP_FREEBUFFER

Those were previously defined in spresense-exported-sdk, but now have been removed.

### DIFF
--- a/src/portable/sony/cxd56/dcd_cxd56.c
+++ b/src/portable/sony/cxd56/dcd_cxd56.c
@@ -103,19 +103,23 @@ static int _dcd_bind(FAR struct usbdevclass_driver_s *driver, FAR struct usbdev_
   usbdcd_driver.ep[0] = dev->ep0;
 
   #ifdef EP_ALLOCREQ
+  // SDK v2
   usbdcd_driver.req[0] = EP_ALLOCREQ(usbdcd_driver.ep[0]);
-  if (usbdcd_driver.req[0] != NULL)
-  {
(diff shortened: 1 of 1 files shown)


hathach (commented):
should we wrap these with #ifndef so that it can work with both sdk 3.2.0 and previous sdk version ?

src/portable/sony/cxd56/dcd_cxd56.c
```diff
@@ -39,25 +39,6 @@
 #define CXD56_SETUP_QUEUE_DEPTH (4)
 #define CXD56_MAX_DATA_OUT_SIZE (64)
 
-/* Allocate/free I/O requests.
- * Should not be called from interrupt processing!
- */
-
-#define EP_ALLOCREQ(ep)            (ep)->ops->allocreq(ep)
```
hathach: I mean wrap this with ifndef to compatible with both sdk

```C
#ifndef EP_ALLOCREQ
#define EP_ALLOCREQ(ep)            (ep)->ops->allocreq(ep)
#endif 
```

hathach (commented):
I prefer the first changes, but just wrap it with ifndef to compatible with both version of sdk. I will make the change myself since it is quick enough.

hathach (approved):
thank you, I added check if usbdev_allocreq failed, and add note for sdk v2, v3. Will merge when ci passes

PR #2518, opened 2024-03-14
STM32 DWC2 Vbus

flamesky0:
STM32 DWC2: On some boards vbus may not be connected to the controller and device becomes non-visible for host due to DP doesn't get pulled-up. Additions in this PR can help board bring-uppers to debug the problem, and configure the build.


HiFiPhile:
Thanks for your PR, however `USB_OTG_GCCFG_VBDEN` is not defined in all devices, some devices use `USB_OTG_GCCFG_VBUSASEN` or `USB_OTG_GCCFG_VBUSBSEN`.

And it's handled in BSP level due to boards and MCU differences: https://github.com/hathach/tinyusb/blob/7c66f9b2a4f302268caeab9f8e63fb6ebd8be466/hw/bsp/stm32f4/boards/stm32f401blackpill/board.h#L94

Issue #2513, opened 2024-03-06
[UVC] Support device with multiple VS and VC interface


lijunru-hub:
### Related area

Class UVC

### Hardware specification

esp32sx

### Is your feature request related to a problem?

I want to define a UVC device with two VS (Video Streaming) interfaces, capable of simultaneously transmitting two video streams. However, the current class UVC driver does not support installing all VS VC (Video Control) interfaces.
I've identified two main issues and made modifications to the source code, enabling the simultaneous opening of two VS (Video Streaming) interfaces.
![image](https://github.com/hathach/tinyusb/assets/68437239/7688196e-0157-4949-afd2-b7cf653c05e6)
![image](https://github.com/hathach/tinyusb/assets/68437239/e8f54f12-0b99-4de3-853d-692382ccc32a)

I've noticed that the version of TinyUSB I'm based on is quite old. You can review the source code at https://components.espressif.com/components/espressif/tinyusb.

I've also tested the latest version of TinyUSB and found that it still doesn't support dual cameras. The issues present in the old code persist in the latest version.

In essence, supporting multiple VC VS (Video Control Video Streaming) interfaces in the UVC class is entirely feasible at the lower level. However, there are some logical issues.

Below is a screenshot of my successful setup, featuring one virtual image transmission and one camera feed. [Include screenshot here]

![image](https://github.com/hathach/tinyusb/assets/68437239/4f65500e-e2b1-4bc5-8c2e-0cdd05ae0266)

Finally, here is my device descriptor.

![image](https://github.com/hathach/tinyusb/assets/68437239/6a04c453-8ff7-4230-bbcd-d2629aed42be)

### Describe the solution you'd like

Supporting this functionality in the source code.

### I have checked existing issues, dicussion and documentation

- [X] I confirm I have checked existing issues, dicussion and documentation.

lijunru-hub:
Support by https://github.com/hathach/tinyusb/pull/2520

PR #2520, opened 2024-03-15
fix: uvc multiple interface support

lijunru-hub:
1. fix next interface may have IAD_DESC
2. fix vs_itf can't be open on bulk mode.

**Describe the PR**
Resolved the issue where the current UVC Class could not correctly iterate through all interface numbers in the case of multiple VC interfaces and could not open each VS interface under each VC in bulk mode.

**Additional context**
https://github.com/hathach/tinyusb/issues/2513

### DIFF
--- a/src/class/video/video_device.c
+++ b/src/class/video/video_device.c
@@ -207,6 +207,24 @@ static void const* _find_desc(void const *beg, void const *end, uint_fast8_t des
   return cur;
 }
 
+/** Find the first descriptor of two given types
+ *
+ * @param[in] beg        The head of descriptor byte array.
+ * @param[in] end        The tail of descriptor byte array.
+ * @param[in] desc_type_0 The first target descriptor type.
+ * @param[in] desc_type_1 The second target descriptor type.
+ *
+ * @return The pointer for interface descriptor.
+ * @retval end   did not found interface descriptor */
+static void const* _find_desc_2_type(void const *beg, void const *end, uint_fast8_t desc_type_0, uint_fast8_t desc_type_1)
+{
+  void const *cur = beg;
+  while ((cur < end) && (desc_type_0 != tu_desc_type(cur)) && (desc_type_1 != tu_desc_type(cur))) {
+    cur = tu_desc_next(cur);
+  }
+  return cur;
+}
+
 /** Find the first descriptor specified by the arguments
  *
  * @param[in] beg        The head of descriptor byte array.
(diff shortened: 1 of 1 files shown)


kkitayam:
Thank you for your PR!

I will check the PR on my environment. Please wait a few days.

kkitayam (approved):

kkitayam:
In Windows 11, it was confirmed that multiple VC interfaces were correctly enumerated by applying the PR, but it failed without the PR. LGTM.

kkitayam:
@hathach 

Just a heads up, the `pre-commit` action is failing. I'm not sure if we should go ahead with the merge. Could you check it out when you get a chance?

kkitayam:
I have confirmed that it is possible to capture 2 streams simultaneously using the following example.
[kkitayam:add_uvc_2ch_example](https://github.com/kkitayam/tinyusb/tree/add_uvc_2ch_example)

src/class/video/video_device.c
```diff
@@ -1214,11 +1236,7 @@ uint16_t videod_open(uint8_t rhport, tusb_desc_interface_t const * itf_desc, uin
     if (0 == stm_idx && 1 == bInCollection) {
       /* If there is only one streaming interface and no alternate settings,
        * host may not issue set_interface so open the streaming interface here. */
-      uint8_t const *sbeg = (uint8_t const*)itf_desc + stm->desc.beg;
-      uint8_t const *send = (uint8_t const*)itf_desc + stm->desc.end;
-      if (end == _find_desc_itf(sbeg, send, _desc_itfnum(sbeg), 1)) {
-        TU_VERIFY(_open_vs_itf(rhport, stm, 0), 0);
-      }
+      TU_VERIFY(_open_vs_itf(rhport, stm, 0), 0);
```
kkitayam, line 1239:
```suggestion
      uint8_t const *sbeg = (uint8_t const*)itf_desc + stm->desc.beg;
      uint8_t const *send = (uint8_t const*)itf_desc + stm->desc.end;
      if (send == _find_desc_itf(sbeg, send, _desc_itfnum(sbeg), 1)) {
        TU_VERIFY(_open_vs_itf(rhport, stm, 0), 0);
      }
```

I noticed a typo in the original code. It should be `send` instead of `end`. I apologize for the inconvenience. If you could change it to `send` to achieve the expected behavior, I would appreciate it.

lijunru-hub: The modification addressed my previous concerns, and I've tested it thoroughly. It works very well.

hathach:
thank @lijunru-hub  for the fix, and @kkitayam for reviewing and testing it out.

PR #2533, opened 2024-03-23
Fix the error that occurs when executing make on Windows

kkitayam:
**Describe the PR**
Fix the following error when executing "make" on command prompt in Windows 11, which has no `sh.exe` on `PATH`.
```console
C:\tinyusb\examples\device\video_capture>make BOARD=frdm_kl25z
process_begin: CreateProcess(NULL, expr 13 >= 12, ...) failed.
C:/tinyusb/examples/build_system/make/toolchain/arm_gcc.mk:77: pipe: Bad file descriptor
```

### DIFF
--- a/examples/build_system/make/toolchain/arm_gcc.mk
+++ b/examples/build_system/make/toolchain/arm_gcc.mk
@@ -74,6 +74,8 @@ LDFLAGS += -Wl,--print-memory-usage
 endif
 
 # from version 12
-ifeq ($(shell expr $(CC_VERSION_MAJOR) \>= 12),1)
+ifeq ($(strip $(if $(CMDEXE),\
+               $(shell if $(CC_VERSION_MAJOR) geq 12 (echo 1) else (echo 0)),\
+               $(shell expr $(CC_VERSION_MAJOR) \>= 12))), 1)
 LDFLAGS += -Wl,--no-warn-rwx-segment
 endif


HiFiPhile:
Hum... I tried GCC 12 & 13 with command prompt and Mingw bash, but I can't meet the error, what's your setup ?

kkitayam:
@HiFiPhile 
Thank you for your comment and for attempting to reproduce this issue.
I noticed that the description of this issue lacks information about my environment.

- command prompt in windows 11
- [Arm GNU Toolchain 13.2.rel1](https://developer.arm.com/downloads/-/arm-gnu-toolchain-downloads/13-2-rel1)
- [Gnu make 4.4.1 chocolatey package](https://community.chocolatey.org/packages/make)

> Hum... I tried GCC 12 & 13 with command prompt and Mingw bash, but I can't meet the error, what's your setup ?

I believe your environment has `sh.exe` on the `PATH`, which is why you didn’t reproduce the issues. This issue occurs when `sh.exe` is not present on the PATH. For reference, please see "[Choosing a Shell in DOS and Windows](https://www.gnu.org/software/make/manual/make.html#Choosing-the-Shell.)”

cr1901:
I don't know if this is the correct fix, but @kkitayam is correct that GNU Make has special logic that fires on Windows/DOS to deal with systems without a Bourne shell clone. The behavior is described in greater detail [here](https://github.com/mirror/make/blob/d66a65ad5a0e31b287f53930b0f09e31801f1613/README.W32#L190). Also [relevant source](https://github.com/mirror/make/blob/d66a65ad5a0e31b287f53930b0f09e31801f1613/src/main.c#L999-L1126).

HiFiPhile:
@kkitayam Thanks for your explanation, I reproduced the issue and this fix works well.

HiFiPhile (approved):

PR #2538, opened 2024-03-24
Add an example of 2ch video capture.

kkitayam:
**Describe the PR**
Add an example of 2ch video capture.
Each channel represents an individual camera, but the video format used is the same.

**Additional context**
For reference, please see #2520.


HiFiPhile:
Tested works well on STM32F407-DISCO and Nucleo-STM32G0B1.

hathach:
Tested on rp2040 and it works well. We may make 2 stream a bit different so that it is clearer which is stream (we can do it as follow-up PR no problems). 

I add the new example to cmakelist include directoy, seem like some small MCU has SRAM overflowed with 2 stream buffer. I will check and exclude (add them to skip.txt) later on.

hathach:
Just push an update uvc 2ch example to stream0 use yuy2, stream1 use mpeg, this help reduce sram requirement for example, also provide different format (uncompressed & mpeg)
fixing build with lpc18xx, will fix more ci next

HiFiPhile:
With new changes for camera 2 bulk is ok but iso gives a laggy output.

I think it's a bandwidth issue since I counted 63 bulk packets in 10 sof, which gives about 400kB/s data rate. 
Iso packet is limited to 248kB/s, I tried to increase `CFG_TUD_VIDEO_STREAMING_EP_BUFSIZE` to 511 but `_videod_streaming_itf[1].max_payload_transfer_size` stayed at 248.



[capture.zip](https://github.com/hathach/tinyusb/files/14876282/capture.zip)

hathach:
> With new changes for camera 2 bulk is ok but iso (#2567 applied) gives a strange output on STM32G0.
> 
> In the capture STM32G0 sends packets every frame while STM32F4 sends much less.
> 
> [capture.zip](https://github.com/hathach/tinyusb/files/14876282/capture.zip)
> 
> Then I added a gpio toggle in `tud_video_frame_xfer_complete_cb`, for camera 1 it toggles every 100ms, for camera 2 it toggles about every 1ms.... Seems fsdev driver need some fix...

yeah, seems like fsdev specific issue, I have tested with rp2040 and it works great so far. Still trying to fix build with the rest of the mcus. We probably need a separated issue/pr for fsdev.

examples/device/video_capture_2ch/src/usb_descriptors.c
```diff
@@ -0,0 +1,653 @@
...
+    uvc_control_desc_t video_control;
+    uvc_streaming_yuy2_desc_t video_streaming;
+  } uvc_yuy2;
+
+  struct TU_ATTR_PACKED {
+    tusb_desc_interface_assoc_t iad;
+    uvc_control_desc_t video_control;
+    uvc_streaming_mpeg_desc_t video_streaming;
+  } uvc_mpeg;
+} uvc_cfg_desc_t;
```
hathach, line 185: @kkitayam I have changed the 2 stream to
- stream 0: yuy2 either dynamic or fixed (READONLY)
- stream 1: mpeg fixed

This is to show that 2 stream can have different format and is slightly different. Also it save sram since many of MCUs can only have enough for yuy2 dynamic.

hathach (approved):
everything is all good. I have change the 2 stream to have different format. @kkitayam please revise the changes in recent commit. Thank you for another brilliant PR.  Let me know if this change make sense to you, and we can merge this.

@HiFiPhile regarding the fsdev driver issue with 2 ISO stream. We probably need to have its own issue/pr to fix.

HiFiPhile:
> regarding the fsdev driver issue with 2 ISO stream. We probably need to have its own issue/pr to fix.

Yeah, I've nearly sorted it out. For ISO transfer fsdev doesn't clear endpoint state upon transfer completion, once next IN token received it just resend current packet (even CPU halted) and trigger rx interrupt without way to mask...

The issue wasn't caught as in old examples the number of IN token corresponds to packets needs to send, but here all ZLP sent automatically are triggering rx interrupt.

examples/device/video_capture_2ch/src/usb_descriptors.c
kkitayam: It's a great idea! I agree with you.

HiFiPhile:
@hathach When I use `CFG_EXAMPLE_VIDEO_READONLY` I got incorrect output on stream 1 with both fsdev and dwc2 driver.

Issue #2541, opened 2024-03-25
MAX3421 (host) and NAK handling


hathach:
### Discussed in https://github.com/hathach/tinyusb/discussions/2510

<div type='discussions-op-text'>

<sup>Originally posted by **IngHK** March  5, 2024</sup>
I examined the CPU and USB usage on my host with STM32F401 (ARM M4 CPU 84 MHz) and MAX3421 (SPI 21 MHz).

With a CDC ACM device I saw that the CPU, SPI and USB are mostly busy handling the input stream endpoint.
The endpoint responds with NAK because there is no new data, and `handle_xfer_done()` in `hcd_max3421.c` causes the same stream to be repeated immediately (no other stream is pending).

This behavior largely utilizes the CPU with NAK-triggered interrupts.

Wouldn't a delay before a repetition make sense?</div>

hathach:
fixed by #2543

PR #2543, opened 2024-03-26
MAX3421E NAK retry handling next frame

IngHK:
This PR is intended to fix the problem as discussed in #2510 / issue #2541
The change means that in the case of a NAK, the transfer will only be repeated in the next frame (triggered by frame IRQ).
The 1st commit of PR is a first draft (not for merging) as a basis for further discussion and testing.

### DIFF
--- a/.github/workflows/build_esp.yml
+++ b/.github/workflows/build_esp.yml
@@ -86,11 +86,17 @@ jobs:
         # USB bus on rpi4 is not stable, reset it before testing
       - name: Reset USB bus
         run: |
-          for port in $(lspci | grep USB | cut -d' ' -f1); do
-              echo -n "0000:${port}"| sudo tee /sys/bus/pci/drivers/xhci_hcd/unbind;
-              sleep 0.1;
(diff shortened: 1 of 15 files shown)


IngHK:
I finished work now and It's ready for review and merge.
It can handle up to 10 retries per frame adjustable by CFG_TUH_MAX3421_MAX_ATTEMPS_PER_FRAME.
With value zero it's compatible with endless retries as before.

hathach (approved):
Perfect, thank you very much of the PR. Though I made some changes to
- always en-force max NAK with default = 1. IMO, this should work best for most people. We can increase to 2,3 should it has issue with throughput in the future
- add tuh_configure() cfgid to allow changing this behavior. This can only be used to change other max3421 behavior such as interrupt level polarity, edge in the future.
- add ep's hxfr

tested and confirmed it only send 1 token per frame

image

hathach:
That is ok max attempt can also be inferred as no nak limit as well. To be honest a usb frame will probably hold less than 10 retries due to back and forth spi transactions

hathach:
thanks, I already made an PR to limit it https://github.com/hathach/tinyusb/pull/2565

PR #2553, opened 2024-03-28
IMXRT: update linker scripts so that the code section does *not* star…

ccrome:
IMXRT: update linker scripts so that the code section does *not* start at 0x0.

**Describe the PR**
On those cpus that have the internal RAM start at 0x0, this sets the start of ram to 0x20, so that we can never end up with a valid piece of code living at address 0x0.  That's a bad thing.

**Additional context**
This happens when you put stuff into CodeQuickAccess memory.

all of these linker scripts are directly from the mcu directory.  The downside is that if the upstream mcu release changes, it may be necessary to update these `linker_scripts` files

### DIFF
--- a/hw/bsp/imxrt/family.cmake
+++ b/hw/bsp/imxrt/family.cmake
@@ -70,7 +70,7 @@ function(add_board_target BOARD_TARGET)
 
   # LD_FILE and STARTUP_FILE can be defined in board.cmake
   if (NOT DEFINED LD_FILE_${CMAKE_C_COMPILER_ID})
-    set(LD_FILE_GNU ${SDK_DIR}/devices/${MCU_VARIANT}/gcc/${MCU_VARIANT}xxxxx${MCU_CORE}_flexspi_nor.ld)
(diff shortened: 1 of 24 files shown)


hathach:
I still don't understand why we need to make this change, can you elaborate it please. Please note that these linkers are only used by tinyusb examples only.

ccrome:
Totally true. I just got burned when I ended up with a function pointer at address 0x0.  

Then when I did the callback thing:

if (funcptr) 
   funcptr()

funcptr was placed at 0x0 and never got called.  

Not an issue for the examples since they don't place anything in fast ram.  But when taking the examples and building off of them, it could be an issue.

It's totally fine to close this if not worthwhile.

hathach:
function callback at address 0x0 is kind of odd, though totally valid. Normaly that is for reset/entry/start. However, within the scope of tinyusb examples, we don't really need it and I really don't want to maintain these linker script if I don't have to. Thank you.

Note: in the future I will try to migrate from weak callback to weak/ref default implementation. It will elimiate the need for the `if (func_cb) func_cb()`, this is for keil compatible see #1018

PR #2567, opened 2024-04-02
ISO endpoint allocation improvments

HiFiPhile:
**Describe the PR**
- Move the check `TUP_DCD_EDPT_ISO_ALLOC` into `tusb_mcu.h` 
- Add support to video class
    * Tested `video_capture` and `video_capture_2ch` example work on NUCLEO-G0B1RE with ISO endpoints, previously only bulk endpoints work.

**Additional context**
Now we can remove the buggy buffer free/realloc code from fsdev and dwc2 driver.

### DIFF
--- a/src/class/video/video_device.c
+++ b/src/class/video/video_device.c
@@ -755,6 +755,7 @@ static bool _open_vs_itf(uint8_t rhport, videod_streaming_interface_t *stm, uint
   TU_LOG_DRV("    reopen VS %d\r\n", altnum);
   uint8_t const *desc = _videod_itf[stm->index_vc].beg;
 
+#ifndef TUP_DCD_EDPT_ISO_ALLOC
   /* Close endpoints of previous settings. */
   for (i = 0; i < TU_ARRAY_SIZE(stm->desc.ep); ++i) {
     uint_fast16_t ofs_ep = stm->desc.ep[i];
@@ -762,13 +763,12 @@ static bool _open_vs_itf(uint8_t rhport, videod_streaming_interface_t *stm, uint
     tusb_desc_endpoint_t const *ep = (tusb_desc_endpoint_t const*)(desc + ofs_ep);
     /* Only ISO endpoints needs to be closed */
     if(ep->bmAttributes.xfer == TUSB_XFER_ISOCHRONOUS) {
-#ifndef TUP_DCD_EDPT_ISO_ALLOC
       usbd_edpt_close(rhport, ep->bEndpointAddress);
-#endif
       stm->desc.ep[i] = 0;
       TU_LOG_DRV("    close EP%02x\r\n", ep->bEndpointAddress);
     }
   }
+#endif
 
   /* clear transfer management information */
   stm->buffer  = NULL;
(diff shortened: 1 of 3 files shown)


src/class/video/video_device.c
```diff
@@ -1283,6 +1290,24 @@ uint16_t videod_open(uint8_t rhport, tusb_desc_interface_t const * itf_desc, uin
     cur = _next_desc_itf(cur, end);
     stm->desc.end = (uint16_t) ((uintptr_t)cur - (uintptr_t)itf_desc);
     stm->state = VS_STATE_PROBING;
+#ifdef TUP_DCD_EDPT_ISO_ALLOC
+    /* Allocate ISO endpoints */
```
HiFiPhile, line 1294: Is it possible that one VS interface has multiple endpoints, one input header and one output header ?

src/class/video/video_device.c
```diff
@@ -759,10 +759,15 @@ static bool _open_vs_itf(uint8_t rhport, videod_streaming_interface_t *stm, uint
...
     uint_fast16_t ofs_ep = stm->desc.ep[i];
     if (!ofs_ep) break;
-    uint8_t  ep_adr = _desc_ep_addr(desc + ofs_ep);
-    usbd_edpt_close(rhport, ep_adr);
-    stm->desc.ep[i] = 0;
-    TU_LOG_DRV("    close EP%02x\r\n", ep_adr);
+    tusb_desc_endpoint_t const *ep = (tusb_desc_endpoint_t const*)(desc + ofs_ep);
+    /* Only ISO endpoints needs to be closed */
+    if(ep->bmAttributes.xfer == TUSB_XFER_ISOCHRONOUS) {
+#ifndef TUP_DCD_EDPT_ISO_ALLOC
```
HiFiPhile, line 765: Maybe this `#ifndef` can cover this whole for loop, since bulk endpoint doesn;t change.

src/class/video/video_device.c
kkitayam: Based on my understanding, one VS interface can have up to 2 endpoints, but one VS can only have one input header or one output header. Cite the standard.
image

When a VS interface has two endpoints, one is used for streaming video, and the other is used for capturing still images. Cite the standard.
image
 The case is method 3 shown below.
image

src/class/video/video_device.c
kkitayam: I agree with your comment. I believe that the `#ifndef` can be moved to outside the `for` statement.

kkitayam: As described below, `ep[1]` is limited to bulk for still image capture, so there is no need for a `for` statement.

kkitayam:
Please give me a few days. I would like to check the PR in my environment later.

src/class/video/video_device.c
HiFiPhile: Great, so it's safe to search for only 1 iso endpoint.

kkitayam (approved):

kkitayam:
I had check that the PR works on my RP2040 and RX65N target board.

HiFiPhile:
Thanks for your check, although these 2 mcu don't use fifo allocation feature. I saw allocation is needed in dcd_rp2040 so it would help once it's implemented, not sure the case of RX.

HiFiPhile:
@hathach Maybe my dwc2 PR crashed your HIL test setup...

hathach:
> @hathach Maybe my dwc2 PR crashed your HIL test setup...

Not at all, pi4 seems to struggle with usb stability. I need to do manual reboot after a while. Changes look great, merge now

PR #2569, opened 2024-04-03
try catch labeler rest api

hathach:
**Describe the PR**
catch error from labeler


HiFiPhile:
I think you need to use 2 separate `try {}` block for `Adafruit membership` and `contributor` check

hathach:
> I think you need to use 2 separate `try {}` block for `Adafruit membership` and `contributor` check

Ah thanks, I guess so, but GPT said I only need 1, damn.

PR #2573, opened 2024-04-04
Add support for SAMD21 HCD

ChrisDeadman:
This was written from scratch using the datasheet and tested on a SAMD21 (using [SparkFun SAMD21 Mini Breakout](https://www.sparkfun.com/products/13664)).
However, it should also work with SAMD51 boards, I don't have one for testing though.

### USB Hub Issue

For some reason, when compiling the **host/hid_controller** example from tinyusb (modified by setting `CFG_TUH_HUB` to `1`), `GET_DESCRIPTOR` times out when connecting devices via USB hub.
When I use tinyusb within my platformio project (which uses **framework-arduino-samd** for device initialization), devices connected via HUBs work without any issues.

I've compared the runtime values of all registers that should matter for USB, but could not find any meaningful difference.
Nevertheless, I must have missed one or some, because HUB connections are always working with the arduino samd framework but never with the tinyusb example.

## BIG EDIT:
Even tho I triple-checked, it might be possible that I am still not enabling the correct pin for power delivery since at that time I had no means of measuring it - and the failure description looks kinda like it.
I am using the same Sparkfun board as in my [previous PR to add that board](https://github.com/hathach/tinyusb/pull/2497).

The (unfortunately board-dependent) code I've added for this can be found in my updated version of [hw/bsp/samd21/family.c](https://github.com/hathach/tinyusb/pull/2573/files#diff-589a67aa8f4303cbc2ab31e0b506e5640fa4a55e0f042c88f27d90ff88772257):
```
#if CFG_TUH_ENABLED
  #if defined(CFG_TUH_MAX3421) && CFG_TUH_MAX3421
    max3421_init();
  #else
    // VBUS Power
    gpio_set_pin_direction(PIN_PA28, GPIO_DIRECTION_OUT);
    gpio_set_pin_level(PIN_PA28, true);
  #endif
#endif
```

Hardcoded to `PIN_PA28` because that is what I assumed to be the right pad that enables the host power delivery on my sparkfun board.

Hope I will have time in the near future to set everything up again and try an active USB hub to see if that is the issue, **but I think that is one thing that should be improved before this PR is merged since it should support all kinds of boards**.


hathach:
superb !! Thank you very much, let give me a bit of time for reviewing (still need to wrap up other work first).

ChrisDeadman:
Sorry that it has taken me so long time to rebase, but I did now.

hathach (approved):
perfect, this works well on both samd21 and samd51 on the first try.  Great works, I have move board-specific code for enable vbus to the newly added board_vbus_set(). Sorry for the long delay, I have been pulling off to other works and kind of forgot about this great PR.

ChrisDeadman:
awesome, glad it works! thanks for improving and merging 👍🏼

hathach:
> awesome, glad it works! thanks for improving and merging 👍🏼

thank you for the awesome PR. I should have done this earlier :)

PR #2574, opened 2024-04-05
FSDEV fix/cleanup.

HiFiPhile:
**Describe the PR**
- Follow #2567, cleanup unused ep/pma free code.
- Fix issue founded in #2538

Tested on NUCLEO-G0B1RE with audio, video and  `cdc_dual_ports` examples.


src/portable/st/stm32_fsdev/dcd_stm32_fsdev.c
```diff
@@ -1055,8 +977,17 @@ bool dcd_edpt_iso_alloc(uint8_t rhport, uint8_t ep_addr, uint16_t largest_packet
 
   pcd_set_eptype(USB, ep_idx, USB_EP_ISOCHRONOUS);
 
-  pcd_set_ep_tx_address(USB, ep_idx, pma_addr);
-  pcd_set_ep_rx_address(USB, ep_idx, pma_addr);
+  // Disable double buffering
```
hathach, line 980: I would prefer to keep double buffering for ISO if possible. Otherwise it may not keep up to 1023 byte per packet. Maybe we could find anothe way to fix this ?

hathach (commented):
thanks for the fix, though I notice this PR disable double buffering for ISO (default behavior). If possible we should keep this, otherwise it may not keep up with 1023 packet size.

HiFiPhile:
> I would prefer to keep double buffering for ISO if possible. Otherwise it may not keep up to 1023 byte per packet.

In fact currently double buffering is not configured correctly, it's enabled but both buffer are pointed to the same address.
https://github.com/hathach/tinyusb/blob/d33fe38a6257bcdd60f51a57c9fd19ce2ad1f202/src/portable/st/stm32_fsdev/dcd_stm32_fsdev.c#L965C1-L967C76

Maybe it's better to fix the buffer stuff and enable double buffering, since on older chips there is no mention that ISO double buffering can be disabled:
image
image

HiFiPhile:
>  since on older chips there is no mention that ISO double buffering can be disabled:

Yeah it's broken on L0...

hathach:
> > since on older chips there is no mention that ISO double buffering can be disabled:
> 
> Yeah it's broken on L0...

thanks for testing it out, seems like we will definitely need double buffer for ISO.

HiFiPhile:
I've reworked the PR, now it has correct ISO double buffering handing. 

It's enabled only for device with 2048b PMA, smaller devices has 2 buffers point to the same address as before, otherwise they will fail to run some examples. In my tests I didn't see single buffering cause any issue.

Tested on L0, G0, H5 to be sure nothing broken.

I put each change in one commit hope it's easier to review...

hathach:
superb !! thank you, will review and test this out soon enough.

src/portable/st/stm32_fsdev/dcd_stm32_fsdev.c
```diff
@@ -512,29 +493,42 @@ static void dcd_ep_ctr_tx_handler(uint32_t wIstr)
...
   }
 
   /* clear int flag */
   pcd_clear_tx_ep_ctr(USB, EPindex);
 
-  xfer_ctl_t * xfer = xfer_ctl_ptr(ep_addr);
-  if((xfer->total_len != xfer->queued_len)) /* TX not complete */
+  xfer_ctl_t *xfer = xfer_ctl_ptr(ep_addr);
+
+  /* Ignore spurious int */
```
hathach, line 505: do we still have issue with this spurious interrupt ? I wondered if it is caused by incorrect double buffer configuration previously.

src/portable/st/stm32_fsdev/dcd_stm32_fsdev.c
```diff
@@ -563,100 +557,93 @@ static void dcd_ep_ctr_rx_handler(uint32_t wIstr) {
...
+      /* Set endpoint active again for receiving more data.
+       * Note that isochronous endpoints stay active always */
       if (!((wEPRegVal & USB_EP_TYPE_MASK) == USB_EP_ISOCHRONOUS)) {
-        /* Set endpoint active again for receiving more data.
-         * Note that isochronous endpoints stay active always */
-        pcd_set_ep_rx_status(USB, EPindex, USB_EP_RX_VALID);
+        uint16_t remaining = xfer->total_len - xfer->queued_len;
+        uint16_t cnt = remaining >= xfer->max_packet_size ? xfer->max_packet_size : remaining;
+        pcd_set_ep_rx_cnt(USB, EPindex, cnt);
+        pcd_set_ep_rx_cnt(USB, EPindex, remaining);
```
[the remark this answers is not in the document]
HiFiPhile: Good spot ! 
In fact I'm not sure if it's even necessary, `  pcd_set_ep_rx_cnt()` only reduce buffer size.

src/portable/st/stm32_fsdev/dcd_stm32_fsdev.c
HiFiPhile: It's still necessary since Tx interrupt can't be masked, if I disable the EP the host will not happy seeing NAK.

Another way to fix is in classes driver's xfer_cb, always call usbd_edpt_xfer for iso even if there is nothing to send.

hathach: No need to change the class driver, this is specifci to this dcd only. I see host send IN token while have nothing to send and ISO does not have NAK --> triggered interrupt with zero byte data. I think this is all good now. Though I think we should make it more ISO explicit thing (making changes atm)

src/portable/st/stm32_fsdev/dcd_stm32_fsdev.c
```diff
@@ -152,7 +152,7 @@ typedef struct {
   uint16_t queued_len;
   uint16_t max_packet_size;
   uint8_t ep_idx;   // index for USB_EPnR register
-  bool in_complete; // Workaround for ISO IN EP doesn't have interrupt mask
+  bool iso_in_sending; // Workaround for ISO IN EP doesn't have interrupt mask
```
hathach, line 155: @HiFiPhile I have rename and invert logic for this flag. It now means if we have an active data for ISO IN, and limit its affect on ISO endpoint only. Let me know it makes sense to you.

HiFiPhile: Yes prefect

hathach (approved):
perfect, thank you. Great works as usual and I am glad it is more robust now

PR #2575, opened 2024-04-05
dwc2: remove fifo free code.

HiFiPhile:
**Describe the PR**
Follow #2567


hathach (approved):
great, better clean these dead code while we still remember :)

HiFiPhile:
I'm thinking about set configuration request, should FIFO flushed in dcd_edpt_close_all before switching ?

hathach:
> I'm thinking about set configuration request, should FIFO flushed in dcd_edpt_close_all before switching ?

Yes, we should flush all fifo with close all

Issue #920, opened 2021-06-24
High-speed streaming capacity of CDC class


HiFiPhile:
**Is your feature request related to a problem? Please describe.**

Raw throughput of USB can be high but once you make some data copy the speed will drop significantly. Currently there is no way to skip fifo & buffer in CDC class:
`USB HW Buffer ==> ep_out_buf ==> rx_ff ==> User app`
Besides DMA is not supported.

We had some discussions on this subject, for example expose the fifo or add xfer_fifo support.

**Describe the solution you'd like**
Sort out a solution to make CDC class capable of  high-speed streaming

hathach:
This is already asked, but can you put up your board & setup and how you test the throughput and its result. This allow others to join and try with their own setup for comparison.

HiFiPhile:
I've done some tests:

- **MCU** LPC4357 @ 204MHz
- **Host** Windows 10 x64 20H2
- **Toolchain** IAR 9.10.2, High balanced optimization
- **Code** `examples\device\cdc_dual_ports`, except using simplified [cdc_task()](https://github.com/hathach/tinyusb/blob/5a4fc1151a7e62a291a90836165942737a1a2d6f/examples/device/cdc_msc/src/main.c#L105) from `examples\device\cdc_msc`
- **Test script** [ser_test.py.txt](https://github.com/hathach/tinyusb/files/6711665/ser_test.py.txt)

| Config | Throughput | `tud_task` time (us)|`cdc_task` time (us) | Comment |
|-|-|-|-|-|
|Vanilla|6.87 MB/s|6.7-10| 5.7| 
|512B Echo Buf|15.54 MB/s |6.8-9.8| 7.5| https://github.com/hathach/tinyusb/blob/5a4fc1151a7e62a291a90836165942737a1a2d6f/examples/device/cdc_msc/src/main.c#L115 |
| 2048B All Bufs| 25.67 MB/s |12.8|13.6|  https://github.com/hathach/tinyusb/blob/5a4fc1151a7e62a291a90836165942737a1a2d6f/examples/device/cdc_msc/src/main.c#L115 and `tusb_config.h`|
|512B cdc_task2|15.53 MB/s|  6.6-10|4.8|Change buffer size in `tusb_config.h` and modify `cdc_task` as below|
|2048B cdc_task2| 28.26 MB/s|12.8-16|4.9| Change buffer size in `tusb_config.h` and modify `cdc_task` as below|
[...]

PR #2576, opened 2024-04-05
DWC2 DMA support

HiFiPhile:
**Describe the PR**
Add Internal DMA support to DWC2.

### Status
| Model| Core Rev|Enumeration| `cdc_ual_ports`| `audio_4_channel_mic`| `video_capture_2ch`|
|--------|--------|--------|--------|--------|--------|
| STM32F407<br>FS w/wo DMA| 2.81a| :heavy_check_mark: | :heavy_check_mark:|:heavy_check_mark:|:heavy_check_mark:|
| STM32F429<br>FS w/wo DMA | 2.81a| :heavy_check_mark:| :heavy_check_mark:|:heavy_check_mark:|:heavy_check_mark:|
| STM32F723<br>FS + HS w/wo DMA| 3.30a|  :heavy_check_mark:|  :heavy_check_mark: | :heavy_check_mark:| :heavy_check_mark:|
| STM32U5<br>HS w/wo DMA| 4.11a|  :heavy_check_mark:| :heavy_check_mark:  |  :heavy_check_mark:| :heavy_check_mark:|

| Model| Core Rev|Enumeration| `cdc_msc_freertos`| `audio_4_channel_mic_freertos`| `hid_composite_freertos`|
|--------|--------|--------|--------|--------|--------|
| ESP32-S3<br>FS w/wo DMA| 4.00a| :heavy_check_mark: | :heavy_check_mark:|:heavy_check_mark:|:heavy_check_mark:|

### Benchmark
STM32F723E-DISCO
IAR 9.50 High-Balanced
I-Cache enabled

Method used in #920, all buffers set to 2048b.

|  | Throughput| Total CPU Usage|
|--------|--------|--------|
| DMA OFF| 26.7 MB/s| 55%|
| DMA ON | 29.5 MB/s| 26%|


HiFiPhile:
Setup is received, somehow Data packet ending is corrupted...
```
USBD init on controller 1
sizeof(usbd_device_t) = 44
sizeof(dcd_event_t) = 12
sizeof(tu_fifo_t) = 12
sizeof(tu_edpt_stream_t) = 24
CDC init
0x00001100, 0x4F54281A, 0x00000000, 0x229ED590, 0x03F403E8, 0x17F00030
Fullspeed PHY init
USBD Bus Reset : Full Speed

  Get Descriptor Device
  Queue EP 80 with 18 bytes ...
  0000:  12 01 00 02 EF 02 01 40 FE CA 02 40 00 01 01 02  |.......@...@....|
  0010:  03 01                                            |..|
```
image

src/portable/synopsys/dwc2/dcd_dwc2.c
```diff
@@ -347,7 +375,7 @@ static void bus_reset(uint8_t rhport) {
   dwc2->grxfsiz = calc_grxfsiz(64, ep_count);
 
   // Setup the control endpoint 0
-  _allocated_fifo_words_tx = 16;
+  _allocated_fifo_words_tx = 32;
```
HiFiPhile, line 378: It seems like EP0's IN FIFO can't be located at the end of RAM, otherwise IN packet is corrupted like above.

HiFiPhile: For U5 (maybe also ESP32-S3 who failed HIL) I've to reserve an excessive amount of space like 90 words in order to make it work, otherwise packet ending will be corrupted like my 1st comment.

@hathach I doubt it has something to do with new feature like `dma_dynamic`, ST hasn't update their comment for U5 in HAL driver:
```
When DMA is used 3n * FIFO locations should be reserved for internal DMA registers */
```
I think I need help from somebody knowing dwc2 internal structures...

src/portable/synopsys/dwc2/dcd_dwc2.c
```diff
@@ -67,18 +67,6 @@
 // Debug level for DWC2
 #define DWC2_DEBUG    2
 
-#ifndef dcache_clean
-#define dcache_clean(_addr, _size)
```
HiFiPhile, line 71: dcache code should be removed and add MPU suggestion to make usb ram non cacheable:
- Not all internal buffers are aligned to cache line size and has size of multiple cache line size, speculative read or cache eviction of adjacent variables could lead to racing with DMA.
- Frequent cache clean invalidate hurt performance
- In most cases Cortex-M7 recommend use DTCM where cache is not used

src/portable/synopsys/dwc2/dcd_dwc2.c
```diff
@@ -186,6 +183,38 @@ static bool fifo_alloc(uint8_t rhport, uint8_t ep_addr, uint16_t packet_size) {
   return true;
 }
 
+static inline bool dma_enabled(uint8_t rhport)
+{
+  // DMA doesn't support fifo transfer
+#ifdef TUD_AUDIO_PREFER_RING_BUFFER
```
HiFiPhile, line 189: Don't want to depend on class driver but maybe it's the simplest way.

src/portable/synopsys/dwc2/dcd_dwc2.c
```diff
@@ -351,6 +382,12 @@ static void bus_reset(uint8_t rhport) {
   // Setup the control endpoint 0
   _allocated_fifo_words_tx = 16;
 
+  // DMA needs extra space for processing
+  if(dma_enabled(rhport)) {
+    uint16_t reserved = _dwc2_controller[rhport].ep_fifo_size / 4- dwc2->ghwcfg3_bm.total_fifo_size;
```
HiFiPhile, line 387: Kernel use `dwc2->ghwcfg3_bm.total_fifo_size` it should be reliable. We can even remove `_dwc2_controller[rhport].ep_fifo_size` and use this field instead except GD32VF103 looks strange.

leptun:
I tried catching the moment where the xfer length gets corrupted. I added this. It's probably not correct, however the CDC does seem to work correctly now and the USB enumerates properly.
image

HiFiPhile:
> I tried catching the moment where the xfer length gets corrupted. I added this. It's probably not correct, however the CDC does seem to work correctly now and the USB enumerates properly.

Humm that's interesting... Just checked ghwcfg registers they are the same as F723, so the only difference is F723 has an internal HS phy.

leptun:
The epout interrupt seems to be triggering twice. Example:

with this PR:
```
USBD Setup Received 80 06 02 03 09 04 FF 00 
  Get Descriptor String[2]
  Queue EP 80 with 30 bytes ...
USBD Xfer Complete on EP 80 with 30 bytes
  Queue EP 00 with 0 bytes ...
```

with latest master:
```
USBD Setup Received 80 06 02 03 09 04 FF 00 
  Get Descriptor String[2]
  Queue EP 80 with 30 bytes ...
USBD Xfer Complete on EP 80 with 30 bytes
  Queue EP 00 with 0 bytes ...
```

the underflowed xfer length would make sense if the subtract is done twice.

And yes, I can make the project public. I'll get that sorted soon.

HiFiPhile:
> The epout interrupt seems to be triggering twice

For all EPs or only EP0 ?

leptun:
@HiFiPhile 
https://github.com/leptun/Avionus_Firmware/tree/tusb_dma
Here is the code that I am using. I've created a branch where I've disabled pretty much anything that is unrelated to usb. You should be able to get the log on the SWO trace. I am also initializing almost no peripheral, so you can probably run it on the stm32f723 eval board. It's using the HSE at 24MHz. You might have to change that in App/config.hpp. Let me know if this also triggers the bug for you.
To build the project you need to initialize the git submodules. I'm using STM32cubeide 1.15.0 with the `GNU Tools for STM32 (11.3.rel1)` toolchain. I was mistaken in my previous post when I said I use GCC 12.3. Do not regenerate the cubemx code since that will enable stuff again.

HiFiPhile:
Just a quick test, what if you put tusb memory in DTCM in linker script ?

leptun:
I'll try that when I get back home later today. Do you want everything tinyusb related in dtcmram?

HiFiPhile:
> Do you want everything tinyusb related in dtcmram?

Yes please

leptun:
Just tried. I moved the krpc (the thing using the CDC) and tinyusb at the beginning of DTCMRAM. Still same behaviour

HiFiPhile:
Humm..  seems out of RAM ,as HS needs more buffer:
```
C:/ST/STM32CubeIDE_1.15.1/STM32CubeIDE/plugins/com.st.stm32cube.ide.mcu.externaltools.gnu-tools-for-stm32.11.3.rel1.win32_1.1.200.202402032213/tools/bin/../lib/gcc/arm-none-eabi/11.3.1/../../../../arm-none-eabi/bin/ld.exe:E:\mcu\stm32f7\Avionus_Firmware\STM32F722ZETX_FLASH.ld:370 cannot move location counter backwards (from 000000002003ec68 to 000000002003e800)
collect2.exe: error: ld returned 1 exit status
make: *** [makefile:155: Avionus_Firmware.elf] Error 1
```

leptun:
Oh. Yes. I allocated only 2KB to tinyusb. I've pushed a commit that increases the size to 4KB

leptun:
Just tested the fix in 02ec486 and it solved the issue on my end as well 👍

HiFiPhile:
Does it resolve first all zero setup packet ?
Can't get SWO to work so I've added RTT to your project, and suddenly RTT stop to output...

Maybe there is still some glitch somewhere, enumerate is not smooth as stock example, it start with one incomplete transfer.
image

leptun:
Huh, I did notice it enumerated slower, but I thought that was just my PC acting up since I have tons of endpoints active on the same root hub. But you are right, there are more bus resets than there should be. I don't think I noticed this behaviour on the disco with the HS phy.
This is what it looks like now on the logs:
[dwc2_dma.txt](https://github.com/hathach/tinyusb/files/15214426/dwc2_dma.txt)
Weird that SWO didn't work.

HiFiPhile:
Ha ! Endpoint transfer is queued before it's opened, seems in DMA mode it's more sensitive.

Maybe we will add more assertion.

```diff
 krpc_error_t KrpcClient::krpc_close() {
-	tud_cdc_n_read_flush(0);
-	tud_cdc_n_write_clear(0);
+	if(tud_ready()){
+		tud_cdc_n_read_flush(0);
+		tud_cdc_n_write_clear(0);
+	}
 	return KRPC_OK;
 }
```

leptun:
Heh, nice catch! Some assertion for this scenario might be helpful, though I'm not sure if this should go in the CDC class or in the usbd driver.
Thanks a lot for helping out with getting this PR working with my project. If I notice any other issue with it, I'll let you know :)

Now I guess it's time for me to go back to finding out why the performance of stm32_fsdev is attrocious. Might be related to the fact that the data is copied around 3 or 4 times by the CPU, might be because double buffering is not implemented for bulk endpoints... If you're curious what that is for, I am working on an AVR ISP programmer with integrated USB-UART as a second virtual com port. The TTL part seemed to be going well, till I realized that I couldn't push more than ~70KB/s of CDC data in a loopback on the same interface. I tried using the xfer_ff in the CDC so that one of the copies is avoided, but that didn't help at all. I really hope I won't have to get double buffering working.

HiFiPhile:
> I realized that I couldn't push more than ~70KB/s of CDC data in a loopback on the same interface. I tried using the xfer_ff in the CDC so that one of the copies is avoided, but that didn't help at all. I really hope I won't have to get double buffering working.

That's really bad, I got more than 500KB/s in CDC OUT transfer with a G0B1, which makes me think double buffer is not necessary. Besides it's really a pain to manage double buffer in fsdev.

Let's continue in a discussion.

leptun:
Thanks, but no need anymore. My testing setup was bad. With your set_test.py script and some modifications to also rx at the same time, I can now loopback at ~560KB/s, which is more than enough

roma-jam:
Hi @HiFiPhile ,
May I ask how did you measure the total CPU load?

hathach:
thank you very much @HiFiPhile again for another great PR. sorry for super late response, I am doing more test with this, seems like it have some issue with my h743 eval which uses external HS PHY. That is ok, I am doing more testing and reading hopefully we could fix and merge this soon.

hathach:
found the issue, h743 usb dma cannot access DTCM, change linker to use SRAM1 and it is all good. for this reason, it is also required to have CFG_TUD_MEM_SECTION for dma as well. I will do more clean up and we can merge this real soon.

image

hathach (approved):
really long review with incorrect poking. It turns out H743 USB DMA cannot access DTCM ram. Though that force me to do lots of reading and have a better understand of the dwc2 register for EPInfo for DMA. Which store 1 words for normal DMA, and 4 words for scatter/gather. We can dynamically configure gdififocfg with epinfo base to save footprint. Since currently we does not support scatter/gather, we only need 2xep_count. 

While troubleshooting, I also rewrite to simplify the fifo allocation scheme and update documentation as well since we don't need to close/re-open iso anymore.

Since DMA cannot dynamically switched on/off per endpoint, I have made ths an compiled timed option, which is default to off `CFG_TUD_DWC2_DMA`. The reason is, in the future with non-dma mode, we can actually get rid of local ep_buffer in the driver e.g cd (tu_fifo -->dwc2 dfifo). Which I have been thinkging but haven't done yet. So yeah DMA will require more memory for ep_buf (to live long enough for transfer to complete as well as word aligned like other ports). Other driver such audio should adapt with the new cfg option.  

```
from Linux kernel driver
 * Unfortunately the choice to use DMA or not is global to the controller
 * and seems to be only settable when the controller is being put through
 * a core reset. This means we either need to fix the gadgets to take
 * account of DMA alignment, or add bounce buffers (yuerk).
```

PS: I have tested this with esp32s3, xmc4500, stm32 h743, f412, u5a5 and they works flawlessly. Thank you @HiFiPhile again for another great PR.

hathach:
@HiFiPhile let me know if you spot any issue with changes that I made, otherwise we could merge this now :) (finally)

HiFiPhile:
> h743 usb dma cannot access DTCM

Oh it's a little pity for performance. I think we need to update the BSP with MPU config to make the region non-cacheable.

For `gdfifocfg.EPINFOBASE (max is ghwcfg3.dfifo_depth)` I'm the max is more like `OTG_DFIFO_DEPTH` ? Since `ghwcfg3.dfifo_depth` is already the value subtracted EP_LOC_CNT.

I'll test with my boards if everything works.

src/portable/synopsys/dwc2/dcd_dwc2.c
hathach: CFG_TUD_DWC2_DMA must be enabled at compile time, audio driver/user should adapt accordingly.

src/portable/synopsys/dwc2/dcd_dwc2.c
leptun: At least for stm32f7xx, those all seem to have the AHB-S port of the M7 connected directly to the bus matrix, so it is accessible by the USB DMA in this case. Though, that is only for DTCM, ITCM is still inaccessible by anything besides the CPU.

src/portable/synopsys/dwc2/dcd_dwc2.c
hathach: yeah, this is the max epinfo where scatt/gather dma is used. We can use gdfifocfg to dynamic change the epinfo base to reduce the DMA region for simple DMA mode.

src/portable/synopsys/dwc2/dcd_dwc2.c
hathach: right, I haven't check f7, maybe this is specific to h7 only

HiFiPhile:
Just tested on STM32F407, STM32F723 and STM32U5A5, didn't redo the full matrix but all major combinations work.

hathach:
thank you @HiFiPhile for retesting, I think it is perfect to merge now. I will do more follow-up clean up to opt out code when DMA is used later on e.g remove the xfer fifo and tu_fifo const etc..

PR #2579, opened 2024-04-09
Fix tud_video_n_streaming check

HiFiPhile:
**Describe the PR**
I met a racing condition with video examples (strangely I didn't caught it in #2538).

With ISO alloc `stm->desc.ep[0]` is no longer cleared, `tud_video_n_streaming()` return true when interface is initializing but `usbd_edpt_iso_activate()` is not yet called. Resulting `usbd_edpt_xfer()` called before endpoint is ready.


hathach (approved):
thank you, look good to me. Note: host can acutally do probe while streaming to dynamic change streaming parameter. So I guess the current state does not accurately reflect the streaming state. Though we will do a full revise later on.

PR #2584, opened 2024-04-09
Dwc2 interrupts (based on #2050)

hathach:
**Describe the PR**
based on and supersede #2050, since 2050 does not allow maintainer to push update.
- resolve conflict with latest master

history log is reserved, once this is merged, 2050 is marked as merged as well.


HiFiPhile:
For 5ade917 I think it's better to double Tx fifo size only for port who has 4096b buffer. As Rx buffer size is already doubled, port with 1280b buffer has very limited usable space. Also 1280b port running in FS doesn't need much performance.

hathach:
> For [5ade917](https://github.com/hathach/tinyusb/commit/5ade91780578bca10bae048a58fed4b788b553d1) I think it's better to double Tx fifo size only for port who has 4096b buffer. As Rx buffer size is already doubled, port with 1280b buffer has very limited usable space. Also 1280b port running in FS doesn't need much performance.

I think you are right on this, now thinking a bit more, even with 4096, having an in + out with 512 can take up to 2K if we double both of them. It is probalby just better to change tx fifo from half-empty to ful-empty level instead.

PR #2588, opened 2024-04-10
Enhance dcd rusb2, support ra2a1 pipe number scheme 

hathach:
**Describe the PR**
- Enhance dcd rusb2: 
  - fix mask typo for brdy status with pipe9
  - ~use forloop to remove ctz which is not portable~ keep ctz() since it is better for run time in isr 
- support ra2a1 pipe number scheme  
- fix ra2a1 ek led/button, but usb not enumerated probably bsp/clock setup thing.

Note to myself: while debugging, I notice sometimes somehow the _dcd seems to be clear accidentally, not sure what causes it. Maybe an memory overflow or something.

### DIFF
--- a/src/portable/renesas/rusb2/dcd_rusb2.c
+++ b/src/portable/renesas/rusb2/dcd_rusb2.c
@@ -915,6 +915,18 @@ void dcd_edpt_clear_stall(uint8_t rhport, uint8_t ep_addr)
 //--------------------------------------------------------------------+
 // ISR
 //--------------------------------------------------------------------+
+
+#if defined(__CCRX__)
+TU_ATTR_ALWAYS_INLINE static inline unsigned __builtin_ctz(unsigned int value) {
+  unsigned int count = 0;
+  while ((value & 1) == 0) {
+    value >>= 1;
+    count++;
+  }
+  return count;
+}
+#endif
+
 void dcd_int_handler(uint8_t rhport)
 {
   rusb2_reg_t* rusb = RUSB2_REG(rhport);
@@ -1002,13 +1014,13 @@ void dcd_int_handler(uint8_t rhport)
   // Buffer ready
   if ( is0 & RUSB2_INTSTS0_BRDY_Msk ) {
     const unsigned m = rusb->BRDYENB;
-    const unsigned s = rusb->BRDYSTS & m;
+    unsigned s = rusb->BRDYSTS & m;
     /* clear active bits (don't write 0 to already cleared bits according to the HW manual) */
     rusb->BRDYSTS = ~s;
-    for (unsigned p = 0; p < PIPE_COUNT; ++p) {
-      if (tu_bit_test(s, p)) {
-        process_pipe_brdy(rhport, p);
-      }
+    while (s) {
+      const unsigned num = __builtin_ctz(s);
+      process_pipe_brdy(rhport, num);
+      s &= ~TU_BIT(num);
     }
   }
 }
--- a/src/portable/renesas/rusb2/dcd_rusb2.c
+++ b/src/portable/renesas/rusb2/dcd_rusb2.c
@@ -1005,7 +1005,6 @@ void dcd_int_handler(uint8_t rhport)
     const unsigned s = rusb->BRDYSTS & m;
     /* clear active bits (don't write 0 to already cleared bits according to the HW manual) */
     rusb->BRDYSTS = ~s;
-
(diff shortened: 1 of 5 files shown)


src/portable/renesas/rusb2/dcd_rusb2.c
```diff
@@ -835,7 +815,7 @@ bool dcd_edpt_open(uint8_t rhport, tusb_desc_endpoint_t const * ep_desc)
   }
 
   rusb->PIPECFG = cfg;
-  rusb->BRDYSTS = 0x1FFu ^ TU_BIT(num);
+  rusb->BRDYSTS = 0x3FFu ^ TU_BIT(num);
```
hathach, line 818: missing pipe9

src/portable/renesas/rusb2/dcd_rusb2.c
```diff
@@ -57,29 +57,22 @@
...
-  struct {
-    uint32_t ep  : 8;  /* an assigned endpoint address */
-    uint32_t ff  : 1;  /* `buf` is TU_FUFO or POD */
-    uint32_t     : 0;
-  };
-} pipe_state_t;
 
-TU_ATTR_PACKED_END  // End of definition of packed structs (used by the CCRX toolchain)
-TU_ATTR_BIT_FIELD_ORDER_END
+  uint8_t ep; /* an assigned endpoint address */
```
hathach, line 69: remove bit field since it will cause unncessary bitop and this does not save any sram usage for this particular struct

kkitayam: I initially set it up as a bit field for future reservations, but you’re right that it currently doesn’t serve any purpose.

src/portable/renesas/rusb2/dcd_rusb2.c
```diff
@@ -89,70 +82,60 @@ static dcd_data_t _dcd;
...
-      break;
+// Transfer conditions specifiable for each pipe for most MCUs
+// - Pipe 0: Control transfer with 64-byte single buffer
+// - Pipes 1 and 2: Bulk or ISO
+// - Pipes 3 to 5: Bulk
+// - Pipes 6 to 9: Interrupt
+//
+// Note: for small mcu such as
+// - RA2A1: only pipe 4-7 are available, and no support for ISO
+static unsigned find_pipe(unsigned xfer_type) {
```
kkitayam, line 94: This change is great! I think it’s very good.

src/portable/renesas/rusb2/dcd_rusb2.c
kkitayam: Thanks for finding the bug.

src/portable/renesas/rusb2/dcd_rusb2.c
```diff
@@ -1022,23 +1002,14 @@ void dcd_int_handler(uint8_t rhport)
...
-        20, 8, 19, 18
-      };
 
-      const unsigned num = Mod37BitPosition[(-s & s) % 37];
-#else
-      const unsigned num = __builtin_ctz(s);
-#endif
-      process_pipe_brdy(rhport, num);
-      s &= ~TU_BIT(num);
+    for (unsigned p = 0; p < PIPE_COUNT; ++p) {
```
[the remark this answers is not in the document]
kkitayam: Changing to a `for` statement, I agree that eliminating the need for conditional compilation is beneficial. However, considering execution speed, sticking with the original code that uses bitwise operations might be advantageous. In my estimation, it’s unlikely that multiple bits in BRDYSTS will be set simultaneously. Therefore, the `for` statement would mostly be looping without doing. By using bitwise operations, the loop can be executed only the number of bits that are set in BRDYSTS, avoiding unnecessary iterations. Additionally, ARM CPUs have instructions like CLZ and RBIT that allow efficient bit manipulation in just a few cycles, making this approach particularly effective. 
Although I don’t have an idea, I’d like to explore ways to use bitwise operations without conditional compilation in the future.

kkitayam (approved):

src/portable/renesas/rusb2/dcd_rusb2.c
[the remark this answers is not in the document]
hathach: thank you very much for detailed explanation. This makes sense, I didn't notice this, it is indeed very good piece of code for run time. I have reverted the changes, although forloop is simpler to understand, this is worth keeping since it is in isr.

PR #2589, opened 2024-04-10
dwc2: flush fifo in dcd_edpt_close_all()

HiFiPhile:
**Describe the PR**

Tested with `net_lwip_webserver` on STM32F407DISCO, with command:
```
 echo 1 | sudo tee /sys/bus/usb/devices/3-5/bConfigurationValue
 echo 2 | sudo tee /sys/bus/usb/devices/3-5/bConfigurationValue
```

---
For labeler somebody said use `pull_request_target`
https://github.com/orgs/community/discussions/57878


hathach (approved):
thank you, I separate tx/rx flush to reuse part of the code. always inline would help to reduce overhead for calling 2 functions.

hathach:
> For labeler somebody said use pull_request_target
> https://github.com/orgs/community/discussions/57878

ah thanks, trying to use that in #2591 to see if that works.

PR #2598, opened 2024-04-12
[WIP] Added support for AT32F415

HorrorTroll:
I'm tried to added support AT32F415 on TinyUSB, it just for in case if there are people want something to choose between CherryUSB.

I've try follow everything, it did get compiled the firmware out. But flashing to the MCU doesn't seem to work, since it just sending error about USB not recognized. If you have some free time, could you take a look for review. Thanks

**Notes:**
Your project of CherryUF2 is really good and it work great. Oh btw, can I reuse `UF2_FAMILY_ID` on your CherryUF2?


HorrorTroll:
Sorry, accident

PR #2615, opened 2024-04-30
Update dnserver.c: Match dnserv_free signature with fwd-decl

maflcko:
Otherwise, this leads to a compile error (at least on a recent clang version):

```
/src/tinyusb/lib/networking/dnserver.c:195:17: fatal error: a function declaration without a prototype is deprecated in all versions of C [-Wstrict-prototypes]
  195 | void dnserv_free()
      |                 ^
      |                  void
1 error generated.


HiFiPhile (approved):

HiFiPhile:
Thanks for your fix.

PR #2618, opened 2024-05-02
tusb_fifo: skip constant address functions if not used

HiFiPhile:
**Describe the PR**
Due to a missed optimization in the compiler, code for constant address handling is being included in all builds. This change splits the code in different functions to avoid that.

**Additional context**
Supercede #2132


hathach (approved):
look great, thank you

PR #2628, opened 2024-05-07
Add host OS detection quirk.

HiFiPhile:
**Describe the PR**
Due to the nature of quirk I'm not sure where to put...

This 1st OS detection quirk mainly resolve UAC class issue. Windows and OSX demand different feedback endpoint size for FS as Windows doesn't implement UAC spec correctly (Linux is compatible with both).
So there needs a way to adjust Configuration Descriptor on the fly.

Tested on:
-  Windows 10 - 11
- Linux 3.10 - 6.8
- OS X Ventura - Sonoma

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -10,6 +10,7 @@ function(add_tinyusb TARGET)
     # common
     ${CMAKE_CURRENT_FUNCTION_LIST_DIR}/tusb.c
     ${CMAKE_CURRENT_FUNCTION_LIST_DIR}/common/tusb_fifo.c
+    ${CMAKE_CURRENT_FUNCTION_LIST_DIR}/common/tusb_quirk.c
     # device
     ${CMAKE_CURRENT_FUNCTION_LIST_DIR}/device/usbd.c
     ${CMAKE_CURRENT_FUNCTION_LIST_DIR}/device/usbd_control.c
--- a/src/common/tusb_quirk.c
+++ b/src/common/tusb_quirk.c
@@ -34,7 +34,7 @@ void tud_quirk_host_os_hint_desc_cb(tusb_desc_type_t desc) {
     desc_req_idx = 0;
   } else if (desc_req_idx < 2 && (desc == TUSB_DESC_CONFIGURATION || desc == TUSB_DESC_BOS || desc == TUSB_DESC_STRING)) {
     // Skip redundant request
-    if (desc_req_idx == 0 || desc_req_idx > 0 && desc_req_buf[desc_req_idx - 1] != desc) {
+    if (desc_req_idx == 0 || (desc_req_idx > 0 && desc_req_buf[desc_req_idx - 1] != desc)) {
       desc_req_buf[desc_req_idx++] = desc;
     }
   }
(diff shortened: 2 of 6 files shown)


HiFiPhile:
@Marcus2060 @kf6gpe @rhysmorgan134 

A quirk is come out which (hopefully) resolve UAC feedback format issue, test welcomed.

Now it's possible to use different Config descriptors:
```c
uint8_t const desc_configuration_default[] =
{
    // Config number, interface count, string index, total length, attribute, power in mA
    TUD_CONFIG_DESCRIPTOR(1, ITF_NUM_TOTAL, 0, CONFIG_TOTAL_LEN, TUSB_DESC_CONFIG_ATT_REMOTE_WAKEUP, 500),
    // Interface number, Alternate count, starting string index, attributes, detach timeout, transfer size
    TUD_DFU_DESCRIPTOR(ITF_NUM_DFU, 1, 4, DFU_ATTR_CAN_UPLOAD | DFU_ATTR_CAN_DOWNLOAD | DFU_ATTR_MANIFESTATION_TOLERANT, 100, CFG_TUD_DFU_XFER_BUFSIZE),
};

uint8_t const desc_configuration_osx[] =
{
    // Config number, interface count, string index, total length, attribute, power in mA
    TUD_CONFIG_DESCRIPTOR(1, ITF_NUM_TOTAL_OSX, 0, CONFIG_TOTAL_LEN_OSX, TUSB_DESC_CONFIG_ATT_REMOTE_WAKEUP, 500),
     // 1st CDC: Interface number, string index, EP notification address and size, EP data address (out, in) and size.
    TUD_CDC_DESCRIPTOR(ITF_NUM_CDC_0, 4, EPNUM_CDC_0_NOTIF, 8, EPNUM_CDC_0_OUT, EPNUM_CDC_0_IN, 64),
};

// Invoked when received GET CONFIGURATION DESCRIPTOR
// Application return pointer to descriptor
// Descriptor contents must exist long enough for transfer to complete
uint8_t const * tud_descriptor_configuration_cb(uint8_t index)
{
    (void) index; // for multiple configurations

    return  (tud_quirk_host_os_hint() == TUD_QUIRK_OS_HINT_OSX) ? desc_configuration_osx: desc_configuration_default;
}
```

rhysmorgan134:
I'll have a go at building into my current project, may take a while as I few bits I think I need to get my head around to get it in there working. Thanks for looking into it!

hathach (commented):
I don't know, this is not really stable. These things can change, and it is not a good idea to give any suggestion if we aren't sure. I guess we can leave this as is for now until there is better approach.

Note: if the order of descriptors can be used to determinne the OS, this can be done by application since these callback are executed in application space. User can keep track of the order and do the guess by themself.

HiFiPhile:
I agree it's not stable and could break in the future.

But facing the UAC OS compatibility issue I can't come up a better way, at least it will calm down the issue for a moment.

I can move the quirk to application space of uac2_speaker_fb example in #2328 , but since callbacks are separated functions the quirk will become clustered and harder to understand.

hathach:
@HiFiPhile that is OK, I still prefer not puttin this in the stack.. the commennt will help to illustrate the point of guess work. just use bool flag for each descriptor callback.

HiFiPhile:
Already implemented in #2328

PR #2630, opened 2024-05-08
Fix some race conditions

HiFiPhile:
**Describe the PR**
- In cdc_device if read* functions called before tud_ready() transfer will be queued to EP0 cause strange issues.
- Clear _usbd_dev first to ensure tud_ready() return false before driver clean up.
- Introduced by #2492, if both `DCD_EVENT_BUS_RESET` and `DCD_EVENT_SETUP_RECEIVED` are in the event queue, `setup_count` is cleared in bus reset, result `_usbd_dev.setup_count--` underflow. I think the best fix would be clear the queue on bus reset.
- Fix STM32F7 FS port build using HS define.

### DIFF
--- a/hw/bsp/stm32f7/family.mk
+++ b/hw/bsp/stm32f7/family.mk
@@ -24,6 +24,7 @@ ifeq ($(PORT), 1)
     $(info "Using OTG_HS in FullSpeed mode")
   endif
 else
+  CFLAGS += -DBOARD_TUD_MAX_SPEED=OPT_MODE_FULL_SPEED
   $(info "Using OTG_FS")
 endif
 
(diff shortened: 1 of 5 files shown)


hathach (commented):
I think we should add tud_cdc_mounted() and/or tud_cdc_ready() API, and use that in place of tud_ready(). I can make that change, just give me a couple of hours.

hathach (approved):
thank you, I made some changes, this may not fully address a race, but should make it better (need follow up PR with usb reset later)
- rebased
- revert changes to usbd, we will deal with usbreset in interrupt in  a separated PR
- set cfg_num in usbd before calling driver's open(), there is no reason not to do so.
- also added tud_cdc_n_ready() but not used in this PR.

PR #2653, opened 2024-05-19
Add a simple re-entrancy lock for tu_printf.

### LINKED ISSUE #2652: Interrupt lock missing in debug log functions
### Operating System

Windows 10

### Board

All

### Firmware

All

### What happened ?

Interrupt lock need to be added to log functions, Segger RTT print is not thread safe, calling from ISR could enter deadlock. 
Serial logging could also be problematic depending implementation. #2649
https://github.com/hathach/tinyusb/blob/ccc7a36043e055ded1f478a979a303e694123187/hw/bsp/board.c#L51

2 options:
- Remove all logging from ISR
- Add interrupt lock, but macros need to be reworked since we don't know if they are called from device or host or typec stack in order to toggle interrupt accordingly.

### How to reproduce ?

Run examples on ports who has log in ISR with RTT, like NRF5x.

None

### Screenshots

_No response_

HiFiPhile:
**Describe the PR**
Add a simple re-entrancy lock for TU_LOG, should resolve #2652 in most cases. An interlocked version is needed for multicore MCU.

### DIFF
--- a/src/tusb.c
+++ b/src/tusb.c
@@ -452,6 +452,22 @@ void tu_print_mem(void const* buf, uint32_t count, uint8_t indent) {
   dump_str_line(buf8 - nback, nback);
 }
 
+int tu_printf_r(const char *format, ...)
+{
+  // Simple re-entrancy lock, not safe for multi-core MCU
+  static volatile bool lock = false;
+  if (lock) {
+    return -1;
+  }
+  lock = true;
+  va_list args;
+  va_start(args, format);
+  int ret = vprintf(format, args);
+  va_end(args);
+  lock = false;
+  return ret;
+}
+
 #endif
 
 #endif // host or device enabled
--- a/src/common/tusb_common.h
+++ b/src/common/tusb_common.h
@@ -69,6 +69,7 @@
 #include <stddef.h>
 #include <string.h>
 #include <stdio.h>
+#include <stdarg.h>
 
 // Tinyusb Common Headers
 #include "tusb_option.h"
--- a/src/common/tusb_debug.h
+++ b/src/common/tusb_debug.h
@@ -50,12 +50,14 @@ extern char const* const tu_str_xfer_result[];
 #endif
 
 void tu_print_mem(void const *buf, uint32_t count, uint8_t indent);
+int tu_printf_r(const char *format, ...);
 
 #ifdef CFG_TUSB_DEBUG_PRINTF
+  // External debug print function, must be ISR-safe
   extern int CFG_TUSB_DEBUG_PRINTF(const char *format, ...);
   #define tu_printf    CFG_TUSB_DEBUG_PRINTF
 #else
-  #define tu_printf    printf
+  #define tu_printf    tu_printf_r
 #endif
 
 static inline void tu_print_buf(uint8_t const* buf, uint32_t bufsize) {


src/tusb.c
```diff
@@ -452,6 +452,22 @@ void tu_print_mem(void const* buf, uint32_t count, uint8_t indent) {
   dump_str_line(buf8 - nback, nback);
 }
 
+int tu_printf_r(const char *format, ...)
+{
+  // Simple re-entrancy lock, not safe for multi-core MCU
+  static volatile bool lock = false;
```
tlyu, line 458: Even on a single core, `volatile` is not sufficient to prevent inconsistencies due to ISR execution interrupting another instance executing from user space. This needs an actual critical section, or similar.

HiFiPhile: Could you make a real example with disassembly ?

ceedriic: The problem is with this sequence:
```
  if (lock) {
    return -1;
  }
  // <--- ISR COMES HERE
  lock = true;
```
To make that work on uniprocessor with interrupt, then interrupt must be disabled between the test and the setting to true:

something like:

```
  disable_interrupts();
  if (lock) {
    return -1;
  }
  lock = true;
  enable_interrupts()  
```

ceedriic: And this is of course wrong, more like:

```
  bool is_locked;
  disable_interrupts();
  is_locked = lock;
  if (!is_locked)
      lock = true;
  enable_interrupts()  
  if (is_locked)
      return (-1);
```

HiFiPhile: Hi,
> The problem is with this sequence:
```
  if (lock) {
    return -1;
  }
  // <--- ISR COMES HERE
  lock = true;
```
When ISR happened here, `vprintf` hasn't been executed, so ISR will execute `vprintf` in it's context then return to normal context. There is no case where the execution of `vprintf` is interrupted.

ceedriic: But: what protect your IRQ from interrupting another vprintf() call, outside tinyusb?

ceedriic: BTW: I'm using a stm32h747 with tinyusb, and I would love to see your DWC2 DMA patch going in, which is why I was lurking....

HiFiPhile: > But: what protect your IRQ from interrupting another vprintf() call, outside tinyusb?

Good question, currently `TU_LOG_*` are available for user space logging.

> BTW: I'm using a stm32h747 with tinyusb, and I would love to see your DWC2 DMA patch going in

We are few maintaining the repo, if you don't want to wait you can try the [testing](https://github.com/hathach/tinyusb/tree/testing) branch.

hathach:
Thanks @HiFiPhile, though I think printf lock should be done in user space by implementing (nano)libc lock in _write_r(). Since not all retarget needs lock e.g. segger rtt

HiFiPhile:
> though I think printf lock should be done in user space by implementing (nano)libc lock in _write_r(). Since not all retarget needs lock e.g. segger rtt

The problem is libc's `printf()` itself is not reentrant as it works on global data, not just the I/O part.

https://github.com/bminor/newlib/blob/1ed15161362b2eb5649b049b7ab0aaf8097bd43a/newlib/libc/stdio/printf.c#L52

HiFiPhile:
Eh... I think reentrant is kind of implementation dependent, I just read IAR's manual:

```
Most parts of the DLIB runtime environment are reentrant, but the following functions
and parts are not reentrant because they need static data:
Heap functions—malloc, free, realloc, calloc, etc. and the C++ operators new and delete
...
Functions that use files or the heap in some way. This includes scanf, sscanf, getchar, getwchar,
putchar, and putwchar. In addition, if you are using the options --printf_multibytes and --dlib_config=Full,
the printf and sprintf functions (or any variants) can also use the heap.
...
Remedies for this are:
● Do not use non-reentrant functions in interrupt service routines
● Guard calls to a non-reentrant function by a mutex, or a secure region, etc
```

PR #2672, opened 2024-06-12
Add support for STM32U535xx/STM32U545xx

tinic:
The low-end devices STM32U535xx and STM32U545xx use the USB IP found in older STM32 devices (like an STM32F04x for instance). So use the stm32_fsdev driver instead of the synopsys/dwc2 for these two devices.


hathach (approved):
Perfect, thank you. I haven’t tested since didn’t have the hw yet, but it looks great

PR #2675, opened 2024-06-14
[FSDEV] Simplify toggle bit logic

Okarss:
Reduced the toggle bit logic to a single XOR. Improves performance, reduces size and simplifies source code. The original code is more than a decade old nonsense from ST, which is still present in their current code.

In addition made the delays in the initialization safer against the optimization.


HiFiPhile:
Thanks, LGTM.

HiFiPhile (approved):

PR #2694, opened 2024-06-29
Additional fix related to issue #1018. Corrects the usage of TU_ATTR_WEAK for the Keil compiler

feaser:
**Describe the PR**
Issue #1018 resolved the usage of TU_ATTR_WEAK for the Keil compiler when it comes to function `dcd_edpt0_status_complete()`. 

Unfortunately, the same problem (the USB device not enumerating properly) occurs again, if the USB device makes use of a BOS descriptor. For example when using a Microsoft OS 2.0 platform capability descriptor to set a specific Device Interface GUID for WinUSB.

In this case the problem is caused by the `tud_descriptor_bos_cb()` and `tud_vendor_control_xfer_cb()` callback functions. The compiler does not see and use the user implemented version of these callback functions. Consequently, these callback functions are not called at run-time, causing the enumeration to fail. The attached screenshot shows this:

keil_weak_fix

To fix this PR proposes, is exactly the same as for callback function `dcd_edpt0_status_complete()`, just now for the callback functions:

* `tud_descriptor_bos_cb()`
* `tud_vendor_control_xfer_cb()`

I verified that this fix resolves the USB enumeration problem with the Keil compiler. Additionally, I performed a quick regression test to make sure things still work when using the GCC or IAR compiler.

**Additional context**

The PR also adds the same fix for the following four callback functions, because it is just a matter of time before someone else runs into the same problem with those:

* `tud_mount_cb()`
* `tud_umount_cb()`
* `tud_suspend_cb()`
* `tud_resume_cb()`

### DIFF
--- a/src/device/usbd.c
+++ b/src/device/usbd.c
@@ -55,6 +55,30 @@ TU_ATTR_WEAK void tud_sof_cb(uint32_t frame_count) {
   (void)frame_count;
 }
 
+TU_ATTR_WEAK uint8_t const * tud_descriptor_bos_cb(void) {
+  return NULL;
+}
+
+TU_ATTR_WEAK void tud_mount_cb(void) {
+}
+
+TU_ATTR_WEAK void tud_umount_cb(void) {
+}
+
+TU_ATTR_WEAK void tud_suspend_cb(bool remote_wakeup_en) {
+  (void)remote_wakeup_en;
+}
+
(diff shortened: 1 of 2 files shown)


hathach (approved):
thank you look good. I made neccessary tests for the weak implementation: skip exists check, but we need to check its returned value. Also apply for the qualifier and other_speed

PR #2713, opened 2024-07-15
NCM: avoid recursions in tud_network_recv_renew()

### LINKED ISSUE #2711: Stack overflow with the new NCM driver
### Operating System

Linux

### Board

Custom ESP32 S3

### Firmware

My own, but this is not relevant - will explain below

### What happened ?

The new USB NCM driver - introduced in #2227 - is causing a stack overflow.

It causes the problem like this:

[tud_network_recv_renew](https://github.com/hathach/tinyusb/blob/bd15f65e81724d05493fe9e020daf9a2ff7d2865/src/class/net/ncm_device.c#L693) will call into [recv_transfer_datagram_to_glue_logic](https://github.com/hathach/tinyusb/blob/bd15f65e81724d05493fe9e020daf9a2ff7d2865/src/class/net/ncm_device.c#L594). `recv_transfer_datagram_to_glue_logic` will in turn call `tud_network_recv_cb` if there's data - and when the cb called, the `recv_glue_ntb` is not cleared yet.

The problem is that some `tud_network_recv_cb`, like the one in [esp_tinyusb](https://github.com/espressif/esp-usb/blob/4bf729f58e983306f4d980bd548a0d354285d207/device/esp_tinyusb/tinyusb_net.c#L148), will unconditionally call `tud_network_recv_renew`. So the whole thing happens again and again, causing infinite recursion.

So, it's either Espressif will need to change their code ( @ESP-Coco ) and documentation added in TinyUSB to prohibit calling `tud_network_recv_renew` directly within `tud_network_recv_cb`, or the logic within `recv_transfer_datagram_to_glue_logic` need to be changed to prevent this infinite recursion.

@HiFiPhile @rgrr could I ask for your opinions on this?

### How to reproduce ?

If you are not using ESP32, call `tud_network_recv_renew` directly within `tud_network_recv_cb` like the ESP code mentioned above.

If you are using ESP32 with ESP-IDF, you can try to replace the `src/class/net` folder of the `tinyusb` component to the latest master branch of tinyusb.

[1.txt](https://github.com/user-attachments/files/16227586/1.txt)

### Screenshots

_No response_

rgrr:
**Describe the PR**
Due to some expectations in the glue code, a recursive call of tud_network_recv_renew() might happen.
This is the corresponding fix for it.

**Additional context**
See discussions in #2711

Closes #2711

### DIFF
--- a/src/class/net/ncm_device.c
+++ b/src/class/net/ncm_device.c
@@ -110,8 +110,12 @@ typedef struct {
     NOTIFICATION_SPEED,
     NOTIFICATION_CONNECTED,
     NOTIFICATION_DONE
-  } notification_xmit_state;            // state of notification transmission
-  bool notification_xmit_is_running;    // notification is currently transmitted
+  } notification_xmit_state;                            // state of notification transmission
+  bool notification_xmit_is_running;                    // notification is currently transmitted
+
+  // misc
+  bool tud_network_recv_renew_active;                   // tud_network_recv_renew() is active (avoid recursive invocations)
+  bool tud_network_recv_renew_process_again;            // tud_network_recv_renew() should process again
 } ncm_interface_t;
 
 CFG_TUD_MEM_SECTION CFG_TUD_MEM_ALIGN tu_static ncm_interface_t ncm_interface;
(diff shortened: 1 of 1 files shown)


HiFiPhile:
Thank you @rgrr, tested ok on my side.

Hi @showier-drastic, did you have time to check it ?

showier-drastic:
Sorry for the delayed response, this works for me.

HiFiPhile (approved):
Thank you @rgrr, tested working on STM32F723E-DISCO.

PR #2723, opened 2024-07-18
RP2040 tweaks

liamfraser:
RP2040 tweaks:
- Compiler warning fixes
- allow pico-sdk LWIP to be used if TinyUSB one does not exist
- Use custom memcpy in RP2040 code
- Don't clear registers when not needed as hardware reset already does it

### DIFF
--- a/examples/device/net_lwip_webserver/CMakeLists.txt
+++ b/examples/device/net_lwip_webserver/CMakeLists.txt
@@ -5,7 +5,14 @@ include(${CMAKE_CURRENT_LIST_DIR}/../../../hw/bsp/family_support.cmake)
 # gets PROJECT name for the example (e.g. <BOARD>-<DIR_NAME>)
 family_get_project_name(PROJECT ${CMAKE_CURRENT_LIST_DIR})
 
+# Prefer the tinyusb lwip
 set(LWIP ${TOP}/lib/lwip)
+
+# If we can't find one from tinyusb then check cmake var before giving up
+if (NOT EXISTS ${LWIP}/src)
+  set(LWIP ${TINYUSB_LWIP_PATH})
+endif()
+
 if (NOT EXISTS ${LWIP}/src)
   family_example_missing_dependency(${PROJECT} "lib/lwip")
   return()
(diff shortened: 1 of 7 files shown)


hathach (approved):
look good, thank you

earlephilhower:
I know this has been merged, but replacing memcpy with a [local implementation](https://github.com/hathach/tinyusb/commit/770efd9b46db0f8723beebe4bfa4d6694bac2901) makes no sense to me.

`memcpy`, by definition, does not have any alignment restrictions (but of course you need `memmove` if there's potential overlap).  I would suggests that any `memcpy` that crashes on a non-4-byte-aligned input or output is busted or should not be called `memcpy` because lots of other things are going to break, not just USB.

None of the `man` pages (ISO C spec is $$$) I've [seen](https://pubs.opengroup.org/onlinepubs/007904975/functions/memcpy.html) [show](https://man7.org/linux/man-pages/man3/memcpy.3.html) [an](https://en.cppreference.com/w/c/string/byte/memcpy) alignment requirement.  

While the implementation here of `unaligned_memcpy` looks like it works (but doesn't return the almost always unused pointer that ISO `memcpy` does), it feels like it's going to be slower than whatever GCC or the ROM would implement...

hathach:
@earlephilhower yeah, thank you for your feedback. You are rigtht, this is my thought as well initially. However, since it copy to/from a dual port USB RAM, it may have some extra hw limit, we have seen this in other ports e.g https://github.com/hathach/tinyusb/pull/1024, though that is a bit different since MCU is M4 and able to do unaligned access (hence memcpy is implemented in such as way), however the usb ram is limited to byte/aligned access only which could cause issue.

For this PR, M0 may not make sense, but, maybe RPI team want to fix something in advance. For now I would like have @liamfraser to confirm that if this unaliged function is tested for fixing actual issue or just simply an prevention ?

earlephilhower:
Gotcha.  Thanks to all for looking into it.

hathach:
> Gotcha. Thanks to all for looking into it.

no problems, thank you for providing feedback on the PR.

PR #2756, opened 2024-08-08
Fix ESPs without device support

tannewt:
The newer tusb_option.h assumes that there is a number of device endpoints to check against.


hathach (approved):
thaank you

PR #2759, opened 2024-08-09
Marked the keycode parameter of the keyboard_report functions as const since the functions don't modifies the value

hjalleboii:
Marked the keycode parameter of the keyboard_report functions as const since the functions don't modifies the value

These functions does not change the value of  uint8_t keycode[6] therefore it should be marked const
tud_hid_n_keyboard_report
tud_hid_keyboard_report


HiFiPhile (approved):
Thanks !

PR #2765, opened 2024-08-15
couple of fixes for usbtmc example

hinxx:
This PR removes unused code in usbtmc example and uses existing defines instead of hardcoded values in several places.

There are no functional changes otherwise.


HiFiPhile (approved):
Thanks for your PR !

PR #2783, opened 2024-08-29
Bluetooth Device: Issue ZLP on ACL IN ep when transfer is multiple of endpoint max packet size

donatieng:
The BTH device class implementation doesn't work on Linux hosts for transfers which size are a multiple of the ACL IN endpoint maximum packet size. This PR fixes this issue by issuing a zero-length packet on that endpoint in these cases, similarly to the CDC device class implementation.

### DIFF
--- a/src/class/bth/bth_device.c
+++ b/src/class/bth/bth_device.c
@@ -42,9 +42,13 @@ typedef struct
   uint8_t itf_num;
   uint8_t ep_ev;
   uint8_t ep_acl_in;
+  uint16_t ep_acl_in_pkt_sz;
   uint8_t ep_acl_out;
   uint8_t ep_voice[2];  // Not used yet
   uint8_t ep_voice_size[2][CFG_TUD_BTH_ISO_ALT_COUNT];
+
+  // Previous amount of bytes sent when issuing ZLP
+  uint32_t prev_xferred_bytes;
 
   // Endpoint Transfer buffer
   CFG_TUSB_MEM_ALIGN bt_hci_cmd_t hci_cmd;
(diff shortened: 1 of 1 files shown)


donatieng:
I've rebased the PR and fixed clang-format'ted changes

src/class/bth/bth_device.c
```diff
@@ -127,11 +131,25 @@ uint16_t btd_open(uint8_t rhport, tusb_desc_interface_t const *itf_desc, uint16_
...
+  for (size_t p = 0; p < 2; p++) {
+    if (tu_edpt_dir(desc_ep_acl_in->bEndpointAddress) == TUSB_DIR_IN) {
+      _btd_itf.ep_acl_in_pkt_sz = tu_edpt_packet_size(desc_ep_acl_in);
+      break;
+    }
+    desc_ep_acl_in = (tusb_desc_endpoint_t const *)tu_desc_next(desc_ep_acl_in);
+  }
 
-  itf_desc = (tusb_desc_interface_t const *)tu_desc_next(tu_desc_next(tu_desc_next(desc_ep)));
+  itf_desc = (tusb_desc_interface_t const *)tu_desc_next(tu_desc_next(desc_ep));
```
HiFiPhile, line 152: Thank you for the PR, could you elaborate this desc change ?

donatieng: No problem. 

The `tu_desc_next()` iteration that used to be made at that line now happens at line 134. I'm getting a reference to first endpoint descriptor here:
https://github.com/hathach/tinyusb/pull/2783/files#diff-5b2b758cfac5035ee55f2e5189f8c0488362d0c9db1332ba58d064823b3cab9cR134

This allows to recover the max packet size for the IN endpoint here:
https://github.com/hathach/tinyusb/pull/2783/files#diff-5b2b758cfac5035ee55f2e5189f8c0488362d0c9db1332ba58d064823b3cab9cR142

HiFiPhile (approved):

src/class/bth/bth_device.c
HiFiPhile: Thank you, I missed the change in L143.

donatieng:
Thanks @HiFiPhile !

PR #2784, opened 2024-08-29
Fix ESP32-SX resume

tannewt:
The interrupt handler pipes through the resume event but the interrupt wasn't enabled in the first place.


hathach (approved):
thank you

PR #2786, opened 2024-09-01
Put break inside #if 

dp111:
The break; should really be inside the #if #endif construct. as if the #if is false you end up with break; break; in the code


HiFiPhile (approved):

HiFiPhile:
Thanks !

PR #2809, opened 2024-09-17
make all python files executable and standardize interpreter

shuffle2:
**Describe the PR**
I'm a user of tinyusb via pico-sdk. I wanted to use tinyusb_pico_pio_usb and found it wasn't working because `hw/mcu/raspberry_pi/Pico-PIO-USB` is no longer included in tinyusb repo and needs `tools/get_deps.py` to fetch it. I found that get_deps.py wasn't working because my environment (WSL2) didn't detect the file as python. I took opportunity to fix all python files in current tinyusb repo.

**Additional context**
pico-sdk should probably be modified to invoke get_deps.py automatically somehow.


hathach (approved):
look good, thank you

PR #2824, opened 2024-10-01
support stm32U0 device

Maerdl:
Hello ,

i added the basic support for STM32U0 devices. As it uses the same usb core than the U5 series it's just some defines to add.
It was tested with a custom board as a cdc device. No further tests are done.

Are there any further needs to get merged into the main repo ?

Thanks in advance
Maerdl


hathach (approved):
look good, ci failed due to my HIL machine migration. Just merge latest and pushed. Hopefully this should fix it.

HiFiPhile:
Thanks for your PR, I think U0 only use fsdev driver, you can revert the changes in `src/portable/synopsys/dwc2/dwc2_stm32.h`

hathach (changes requested):
@HiFiPhile is right, I overlook this. U0 is fsdev only, please update PR accordingly.

image

hathach (approved):
the revert is simple, I made the update myself. There is a bit of duplication in fsdev driver, but we will clean that up later.

Maerdl:
Thank's guys for the quick solution on that !

PR #2833, opened 2024-10-09
more dwc2 enhance

hathach:
**Describe the PR**
- enhance dwc2_info.md with field mapping
- add ~stm32f769disco~ (f769 need more testing) and stm32f407 to hil pool
- improving dwc2, merging diep and doep if possible


src/portable/synopsys/dwc2/dwc2_info.md
```diff
@@ -1,58 +1,58 @@
...
+| - otg_enable_hsic          | 0               | 0            | 0             | 0            | 0                         | 0                | 0                | 0                 | 0            | 0            | 0              | 0            | 0            | 0           | 0            |
+| - battery_charger_support  | 0               | 0            | 0             | 1            | 0                         | 0                | 1                | 1                 | 1            | 1            | 1              | 1            | 0            | 0           | 0            |
+| - lpm_mode                 | 0               | 0            | 0             | 0            | 0                         | 0                | 1                | 1                 | 1            | 1            | 1              | 1            | 1            | 0           | 0            |
+| - dfifo_depth              | 4080            | 498          | 200           | 896          | 512                       | 1012             | 512              | 512               | 1006         | 1006         | 952            | 512          | 952          | 0           | 634          |
```
ibeckermayer, line 35: @hathach where are you getting the `dfifo_depth` value for the ST U5A5 from? 

Naively, reading [RM0456](https://www.st.com/resource/en/reference_manual/rm0456-stm32u5-series-armbased-32bit-mcus-stmicroelectronics.pdf), I see

<details>
image
</details>

So we have `OTG_GRXFSIZ` which says it must not exceed its power on value of 0x200
<details>
image
</details>

Then `OTG_DIEPTXF0` with the same constraint and power-on value 0x200

<details>
image
</details>

And finally `OTG_DIEPTXFx` for x = 1 to 8, which each have a power on value of 0x200

<details>
image
</details>

Putting that all together, the power-on state (in device mode) has the FIFO set to 0x200 * 0xa = 0x1400.

That seems like way too much on its face, particularly given that these sizes are in terms of 4-byte words, however I'm unclear as to how you're identifying the correct size of 952.

HiFiPhile: It's read from the internal register GHWCFG3:
https://github.com/hathach/tinyusb/blob/1406ad84e3cd7e39771303dd23e6be361b51de96/src/portable/synopsys/dwc2/dwc2_info.py#L63

The register content is not mentioned by the STM32 manual, but in DWC2 IP's databook, which is under NDA.

hathach: We can also peek at kernel driver as well

ibeckermayer: Awesome, thank you guys!

PR #2835, opened 2024-10-10
New Audio 2 Channel Microphone & Error Corrections for uac2_speaker_fb example

pschatzmann:
I started to work on extending TinyUSB Auduino to provide the missing audio functionality.

Unfortunately I was struggeling, so I decided to go back to basics and confirm that the functionality is really working.

- I was missing a 2 channel microphone example. I think 2 channels are the most useful scenario, so I decided to add an example for this case. This will be the base for the Arduino functionality.
- Unfortunatly I could not make any of the speaker examples work. The uac2_headset example is not working and the uac2_speaker_fb was giving compile errors. So I have corrected the errors and made sure that it is compiling now properly.

The functionality has been tested on a RP2040.


HiFiPhile:
Thank you for your interest in improving audio support.

- For the 2CH microphone example, it's better to extend the 4CH example to support both 1CH and 2CH with a`#if` switch, since 95% of code is identical. We can call it `uac2_microphone`.
- For the compilation error fixing, are you using a special gcc version or compilation flags ? We have RP2040 included in the CI and no error was reported. You can find the cmake file here:
https://github.com/hathach/tinyusb/blob/master/hw/bsp/rp2040/family.cmake
- Also please submit PR with a feature branch to prevent the introduce of unwanted changes, you can see in the file list many stuffs are included:
image

pschatzmann:
Thanks for your feed back:

1) I will look into this, but I might just remove this because in my Adafruit solution, I will rely on a fully dynamically calculated descriptor.
2) The example was excluded from the overall cmake build: I needed to include it to pick it up in the overall generation
3) Makes sense: I assume, I can decide on the naming myself

One final question: I noticed that the pull request has somehow added plenty of things that were not changed by me. Is this part of some automatic process that runs in the background ?

HiFiPhile:
> The example was excluded from the overall cmake build: I needed to include it to pick it up in the overall generation

Hum... RP2040 is not skipped in `examples/device/uac2_speaker_fb/skip.txt`, maybe I should get a RP2040 board to checkout one day.

> One final question: I noticed that the pull request has somehow added plenty of things that were not changed by me. Is this part of some automatic process that runs in the background ?

It's because you opened the PR with your `master` branch, since then every time you modify something in your master branch the changes will be added automatically.

Out of topic:
You can also try MCUs like STM32 (low-end for FS and high-end for HS) or LPC55 for HS. Since their DCD driver are better supported.

pschatzmann:
No, it is missing here https://github.com/hathach/tinyusb/blob/master/examples/device/CMakeLists.txt

pschatzmann:
I will resumit the correction based on the dev branch

PR #2840, opened 2024-10-14
make sure transfers to device after reset/before setup are ignored

### LINKED ISSUE #2322: USB MSC has some sort of race condition
### Operating System

Linux

### Board

Raspberry Pi PICO (RP2040)

### Firmware

https://github.com/byteit101/test-tinyusb-pico

### What happened ?

panic/assert/exit calls sometimes, but not always, on writes (I think?).

I have confirmed it on Debian 10, `Linux pollux 5.10.0-21-amd64 #1 SMP Debian 5.10.162-1 (2023-01-21) x86_64 GNU/Linux`

I haven't seen it on a "blank" flash (no filesystem)

I have confirmed it with G++ 8.3 and G++ 12.2 (the latter requires uncommenting `target_link_options(${PROJECT_NAME} PRIVATE "LINKER:--no-warn-rwx-segments")` in the CMake file)

I have reported this downstream as https://github.com/raspberrypi/pico-sdk/issues/1523

### How to reproduce ?

Ensure you have a 1.5.1 PICO SDK setup in your environment.

```
git clone https://github.com/byteit101/test-tinyusb-pico
cd test-tinyusb-pico
mkdir build
cd build
cmake ..
make
picotool load msc_pico_failure.elf -f
picotool reboot -f
lsblk # find the drive that just appeared and is 1MB
sudo mkfs.vfat -F12 -n "TinyUSB Tst" -S 4096 /dev/sdYOURDRIVE
```
In one terminal, `tio /dev/ttyACM0` (or some other auto-connecting serial console)
Now, watch dmesg -w or such, and repeat `picotool reboot -f`. Sometimes it will be ok and you will see the Test Flash load, and remain loaded, and tio serial connection remain connected.
```
[18441889.657799] scsi 9:0:0:0: Direct-Access     TUSB     Test Flash       1.0  PQ: 0 ANSI: 2
[18441889.658432] sd 9:0:0:0: Attached scsi generic sg8 type 0
[18441889.658676] sd 9:0:0:0: [sdh] 256 4096-byte logical blocks: (1.05 MB/1.00 MiB)
```
Othertimes, it will first do that, then after a few seconds, it will disconnect serial, and reload into the RP2 bootloader drive. This is the failure condition:
```
[18441891.961605] scsi 10:0:0:0: Direct-Access     RPI      RP2              3    PQ: 0 ANSI: 2
[18441891.962053] sd 10:0:0:0: Attached scsi generic sg8 type 0
[18441891.962397] sd 10:0:0:0: [sdh] 262144 512-byte logical blocks: (134 MB/128 MiB)
```

 Using tinyusb 0.12, I got this log sequence, culminating in an assertion failure:
```
...snip...
  Get Descriptor Device
  Queue EP 80 with 18 bytes ...
  Queue EP 00 with 0 bytes ...
  Short packet on buffer 0 with 0 bytes
Completed transfer of 0 bytes on ep 0 out
BUS RESET
USBD Bus Reset : Full Speed
  Short packet on buffer 0 with 18 bytes
Completed transfer of 18 bytes on ep 0 in

USBD Setup Received 00 05 1E 00 00 00 00 00 
  Set Address
usbd_control_xfer_cb 171: ASSERT FAILED
```
In contrast, successful Set Address looks like
```
...snip

USBD Setup Received 00 05 10 00 00 00 00 00 
  Set Address
  Short packet on buffer 0 with 0 bytes
Completed transfer of 0 bytes on ep 0 in

...snip
```

The reordering, combined with the fact that it's highly irregular, sometimes failing for 5 reboots in a row, then succeeding for 50%, then succeeding for 10 reboots in a row, makes me think this is some kind of race condition in the MSC TUD driver.

The stack trace was usbd_control.c:171, called from usbd.c:562

I updated TinyUSB to 0.15.0, and now get:
```
*** PANIC ***

ep 0 in was already available
```

### Screenshots

_No response_

Olstyle:
Fixes #2322

**Describe the PR**
In cases of bus reset, there can be pending transfers in the queue. Or a defective host just sends transfers without a proper setup. In both cases this lead to null pointer access after driver lookup(or assert if debug is enabled).   
This small check fixes this.

**Additional context**
I checked the throughput on my STM32U5 base MSC device and could not see a performance hit. Though I am limited at around 256KB/s read speed anyway. I have seen pending transfer after reset with a Win10 host and the linked issue is from Linux, so the problem seems to be universal.
Another option would be to clear the queue on bus reset, but this doesn't fix unexpected transfers and also needs a bit of research to make it work on all supported operating systems.

### DIFF
--- a/src/device/usbd.c
+++ b/src/device/usbd.c
@@ -623,6 +623,10 @@ void tud_task_ext(uint32_t timeout_ms, bool in_isr) {
         break;
 
       case DCD_EVENT_XFER_COMPLETE: {
+        if(!_usbd_dev.connected) {
+           TU_LOG_USBD("Skipped unexpected transfer\r\n");
+           break;
+        }
         // Invoke the class callback associated with the endpoint address
         uint8_t const ep_addr = event.xfer_complete.ep_addr;
         uint8_t const epnum = tu_edpt_number(ep_addr);


hathach:
even device is disconnected, previous complete transfer e.g succesful should still be handled  to notify application e.g.
1. xfer complete -> put to usbd queue
2. device disconnected  
3. tud_task run --> should still be handle

Olstyle:
Then you should stop ever freeing the pointer to the driver and write it first thing in the device init. The current code leads to null pointer access.

HiFiPhile:
Pending transfer complete event shouldn't be fired after USB reset (aborted transfer get revived after replug), if that's the case it should be fixed in DCD layer.

> defective host just sends transfers without a proper setup

USB is not UART, data transfer won't happen if endpoint is not setup explicitly.

---
@hathach  
Actually I think the processing order (reset/disconnect before data transfer) inside `dcd_int_handler` could be problematic. When ISR delay is high, even the data transfer is finished before reset/disconnect, `DCD_EVENT_XFER_COMPLETE` will be queued later than `DCD_EVENT_UNPLUGGED`.

Meanwhile I tried to produce the issue without success on STM32C071/STM32H7S3:
1. Add some ISR delay, can't be too high otherwise transfer will fail
```
void dcd_int_handler(uint8_t rhport) {
  volatile uint32_t cycle_count = 1000;
  while (cycle_count > 0U) {
    cycle_count--; // each count take 3 cycles (1 for sub, jump, and compare)
  }
```
2. Run `cdc_dual_ports` example
3. Keep the transfer busy `dd if=/dev/zero of=/dev/ttyACM0`
4. [Reset the port](https://github.com/jkulesza/usbreset) or unplug the device

PR #2844, opened 2024-10-15
Rp2040 - ISO API to make Audio work #2843

pschatzmann:
Following the discussion that the RP2040  is missing the new ISO API, here is my take on it.
I added the 2 following methods

- bool dcd_edpt_iso_alloc(uint8_t rhport, uint8_t ep_addr, uint16_t largest_packet_size) 
- bool dcd_edpt_iso_activate(uint8_t rhport,  tusb_desc_endpoint_t const * p_endpoint_desc) 

I was keeping the existing logic and I just needed to split up some functions to fit the new logic.

This is addressing [this issue](https://github.com/hathach/tinyusb/issues/2838). 
See https://github.com/hathach/tinyusb/pull/1802#issuecomment-2410581586

I hope that now only the following files will show up as changed:
- tusb_mcu.h (to activate the ISO API)
- dcd_rp2040.c (see above)
- video_device.c (to correct a data cast error because  ep_addr was defined as uint16_t instead of uint8_t)

I create a branch from testing and submitted to testing now!

### DIFF
--- a/src/class/video/video_device.c
+++ b/src/class/video/video_device.c
@@ -1307,7 +1307,7 @@ uint16_t videod_open(uint8_t rhport, tusb_desc_interface_t const * itf_desc, uin
 #ifdef TUP_DCD_EDPT_ISO_ALLOC
     /* Allocate ISO endpoints */
     uint16_t ep_size = 0;
-    uint16_t ep_addr = 0;
+    uint8_t ep_addr = 0;
     uint8_t const *p_desc = (uint8_t const*)itf_desc + stm->desc.beg;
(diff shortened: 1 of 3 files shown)


HiFiPhile:
Did some clean up same as #2845. 

The logic looks like correct although I can't test it myself.

Please also reopen the PR merging into `master` branch.

pschatzmann:
Can you give some testing hints: are there some scenarios that I should watch for ?
So far I was just testing my audio sketches in Arduino using Audacity to make sure that the audio generation is stable and works properly together with Serial (CDC).

PR #2852, opened 2024-10-22
USBH: Recover from unexpected descriptor size

GuavTek:
**Describe the PR**
Fixes the host getting stuck parsing a descriptor if the wTotalLength field is greater than the real descriptor size.

**Additional context**
I encountered this issue with the Exquis MIDI keyboard from Intuitive Instruments


src/host/usbh.c
```diff
@@ -1645,6 +1645,13 @@ static bool _parse_configuration_descriptor(uint8_t dev_addr, tusb_desc_configur
       //          desc_iad->bFunctionClass  == desc_itf->bInterfaceClass);
     }
 
+    if ( 0 == tu_desc_len(p_desc) ) {
```
hathach, line 1648: maybe we should move this to begining of the loop as santiy check

src/tusb.c
```diff
@@ -255,6 +255,10 @@ uint16_t tu_desc_get_interface_total_len(tusb_desc_interface_t const* desc_itf,
           ((tusb_desc_interface_t const*) p_desc)->bAlternateSetting == 0) {
         break;
       }
+      if (tu_desc_len(p_desc) == 0) {
```
hathach, line 258: would be better to move this check to the begining of the loop as well

hathach (commented):
seems like device is off the specs, some is unfortunately. Thank you for your PR. I think it is best to move the (sanity) check to begining of the loop.

GuavTek:
Ok, I have moved the sanity checks.
One of the CI checks started failing, but it doesn't look like it is related to my changes

hathach (approved):
perfect, thank you.

> Ok, I have moved the sanity checks. One of the CI checks started failing, but it doesn't look like it is related to my changes

that is ok, I think PR is slightly out of date, merged from master should fix this.

PR #2854, opened 2024-10-25
[ESP32P4][HS] Solved rhport selection for tusb_rhport_init() in Device mode

roma-jam:
## Changes

- Solving the `rhport` selection when legacy call `tusb_init()` is called. 

## Additional information
- When legacy code call the `tusb_init()`, the option `TUD_OPT_RHPORT` (or `TUH_OPT_RHPORT`) is used to assign the port number (`rhport`). Currently, the call `tusb_rhport_init(0, NULL)` does not take this into account so we need to select the right periph, based `TUD_OPT_RHPORT` value.


src/tusb.h
```diff
@@ -140,8 +140,16 @@ bool tusb_rhport_init(uint8_t rhport, const tusb_rhport_init_t* rh_init);
 // Note: when using with RTOS, this should be called after scheduler/kernel is started.
 // Otherwise, it could cause kernel issue since USB IRQ handler does use RTOS queue API.
 // Note2: defined as macro for backward compatible with tusb_init(void), can be changed to function in the future.
-#if defined(TUD_OPT_RHPORT) || defined(TUH_OPT_RHPORT)
-  #define _tusb_init_arg0()        tusb_rhport_init(0, NULL)
+#if defined(TUD_OPT_RHPORT)
```
hathach, line 143: TUD_OPT_RHPORT/TUH_OPT_RHPORT already covered within tusb.c, please revert this change.

hathach (changes requested):
thank for the PR, but the added macros is not needed, tusb.c handled in within its implementation. 

Note the old tusb_init() should also init both host and devie stack not just one or the other.

src/tusb.h
roma-jam: @hathach 
Thank you so much for checking the PR! 

Meanwhile, I am not sure that it solves all the dependencies completely and provides the fully compatibility to `tusb_init()` call. 

1. When we use the previously used call `tusb_init()` without arguments, it leads to macro call `_tusb_init_arg0()` which in its own calls the `tusb_rhport_init(0, NULL)` with the constant argument `rhport = 0`
2. The `rhport` itself, should be selected, based on the configuration values `CFG_TUSB_RHPORT0_MODE` or `CFG_TUSB_RHPORT1_MODE`. After that, the variable `TUD_OPT_RHPORT` has the value 0 or 1.
3. `TUD_OPT_RHPORT` leads to choosing the periph number, which currently is `0 for FS`, and `1 for HS`

In this case, to support all legacy scenarios we need to provide correct argument for `tusb_rhport_init()` to choose the right periph (0 or 1).

Currently, the macros with constant argument call `tusb_rhport_init(0, NULL)` breaks this connection and the `rhport = 0` periph will be used in every legacy scenarios, when tusb_init() was used, even for HS, when  `TUD_OPT_RHPORT = 1`

hathach: I don't see why it is not backward compatible, the old tusb_init(void) use CFG_TUSB_RHPORT0_MODE/CFG_TUSB_RHPORT1_MODE to init the device stack. This behavior is the same. If you use the tusb_init(void), you need to define those macros. For dynamic rhport selection you pass the parameter tusb_init(rhport, init).

roma-jam: No worries, probably I was not that clear from the start.
Anyway, after your previous comment I simplify the changes and maybe it will be easier to explain what I am talking about right now.

roma-jam:
@hathach 
I commented out one line, but I reverted the changes, as it is might be better to refactor a fifo init a bit to keep the debug and still have it working within the irq, than just comment out the debug line. 

We can merge this. I don’t have anything to add here. Thanks.

PR #2861, opened 2024-10-29
Extend callback information

bollenn:
I noticed getting `0xFFFF` as duration_ms value which according to [CDC1.2_WMC1.1_012011](https://usb.org/sites/default/files/CDC1.2_WMC1.1_012011.zip) PSTN120.pdf is described as in doxygen info I've putted.

It took me some time to figure this out so I thought lets extend the callback header, since I do not see much of this kind of documentation I'm not sure what is the usual approach, please advice if different is expected.


HiFiPhile:
Thank you !

PR #2866, opened 2024-11-02
Fix 2 UAC issue

HiFiPhile:
**Describe the PR**
- Fix control transfer align issue of _audiod_fct.alt_setting
- Sometimes host doesn't call get sample rate but only set sample rate, add flow control support in this case.


hathach (approved):
perfect thank you

Protoxy22:
Thank you

PR #2869, opened 2024-11-07
Fix Usage of Bootsel Button for Pico 2

subsonicpulse:
Function board_button_read() was always returning TRUE on pico 2 board.

This fix make the button on pico2 board usable again.
The code  that fixes the problem is from https://github.com/raspberrypi/pico-examples/blob/master/picoboard/button/button.c


subsonicpulse:
I tested that on pico and pico 2 with the HID example: 
pico-sdk\sdk\2.0.0\lib\tinyusb\examples\device\hid_multiple_interface\src

hathach (approved):
thank you for the fix, tested and working great. I change the ifdef condition to detect arm arch to make it more generic for other vendor pico/pico2 board.

subsonicpulse:
> thank you for the fix, tested and working great. I change the ifdef condition to detect arm arch to make it more generic for other vendor pico/pico2 board.

Thanks also. I was not aware of these Preprocessor Macros. I just used what was in the pico example code. ;-)

PR #2870, opened 2024-11-08
Implement Host Driver support for synopsys dwc2

hathach:
**Describe the PR**
Add host driver for synopsys dwc2 (esp32-s2/s3/p4, stm32, etc..)
- Support cpu slave and dma architecture
- Support both OTG full speed and highspeed
- Support Control/Bulk/Interrupt . ISO isn't tested
- Support HS/FS/LS devices
- Support hub with multiple devices
- Support split control/bulk/interrupt i.e FS/LS device attached via highspeed hub
- Following family is updated and test: stm32 f4/f7/h7, esp32p4
- Add an new dwc2_common.c (required for both device/host)
- API: adding new `tusb_time_millis_api()` and `tusb_time_delay_ms_api()` (default to blocking with millis_api()` which is reuired for no rtos when using with some port/configuration e.g hcd dwc2.
- fix an duplicated attach issue which cause `USBH Defer Attach until current enumeration complete` message

Known issue: For esp32-p4, buffer DMA doees not work (use slave for now). Not entirely sure why, chip support desc dma, but we haven't enabled it yet. I notice hcddma address is increased corretly when sending setup, but data is random (not correct). fix or implement ddma later

image


hathach:
> @hathach
> 
> > For esp32-p4, buffer DMA doees not work (use slave for now). Not entirely sure why
> 
> Hint: cache synchronization is absent.

Thank you for the hint, I am a bit exausted with hcd dwc2 feature for now and not used with debugging esp32p4 (got openocd run, but one it hit breakpoint any futher steps cause it to panic), also printf in isr may also cause panic. I will call it a day for now, since it still function well with slave mode. I will come back to this later on.

cleverca22:
of note, cache coherency is also a problem with my own dwc2 hcd on the rpi (need to test this PR out and see how it behaves)

in my case, i have cache management hooks in the HCD, but one of my buffers was stack allocated, and some activity on the stack caused it to get pulled back into the cache before a control-in could complete
so the ARM core saw a stale version in the cache, despite the HCD flushing the cache before the operation started

HiFiPhile:
@cleverca22 for ARM cores please use MPU to mark the buffer area as non cacheable instead of doing cache clean/invalidate, similar to what #2865 does.

cleverca22:
@HiFiPhile in my case, i happen to be on cortex-a, so thats handled in the MMU

the problem is more, that tinyusb and other applications, are putting the DMA buffers in places like the stack or `.bss`
and you cant just permanently flag that whole page to never cache

so you need to get everything using tinyusb to allocate buffers from a fixed pool

HiFiPhile:
> the problem is more, that tinyusb and other applications, are putting the DMA buffers in places like the stack or .bss
and you cant just permanently flag that whole page to never cache

Doesn't using `CFG_TUSB_MEM_SECTION` help ?

hathach:
@HiFiPhile actually, we can also support dcache clean/invalidate as well, application only need to have the CFG_TUH_MEM_ALIGN to 64 for cache line. It can cost ram, but for some port such as broadcom bcm2711 (pi4) it does not matter. So yeah, I don't rule out option for dcache support.

HiFiPhile:
> @HiFiPhile actually, we can also support dcache clean/invalidate as well, application only need to have the CFG_TUH_MEM_ALIGN to 64 for cache line. It can cost ram, but for some port such as broadcom bcm2711 (pi4) it does not matter. So yeah, I don't rule out option for dcache support. Though I haven't added dcahe for dwc2 yet (I think we did it for chipidea), will do later.

The buffer size also has to be multiple of cache line size, otherwise cached data will be crashed doing cache clean.  There is an example in my linked document.

hathach:
yeah, if it is application buffer, it should also be cachline aligned. For us, we need to make sure any buffer for communicated must be CFG_TUH_MEM_ALIGN.

cleverca22:
something ive been thinking, is that a cache discard (no clean allowed) would also be of use

basically, you have buffers in 2 pools, cpu->device, and device->cpu

any time you write to a cpu->device buffer, you can do a cache clean, to flush the changes to dram, because the region is only ever written by the cpu, there is never dirty data at risk of being caught up, so you dont have to align by cache line size

any time the device is done doing dma and your informing tinyusb, you discard the cache for the entire area
because this second region is only ever written by the device, no dirty data is at risk of loss, your just causing a cache miss on the first hit of any buffers in the same line


but this idea now complicates things, because you need 2 versions of `CFG_TUH_MEM_SECTION`, to allocate things into different pools

HiFiPhile:
> @hathach @HiFiPhile Hi, sorry for asking this in current PR, but is there any quick&easy way to make a speed measurement for `cdc` and `ncm` device classes?

For CDC there is a script https://github.com/hathach/tinyusb/issues/920
For NCM it has a iperf2 server https://github.com/hathach/tinyusb/pull/2227

Issue #2816, opened 2024-09-25
STM32H7R/H7S support


EamonBrennanDRS:
### Related area

New port support.

### Hardware specification

STM32H7R/STM32H7S

### Is your feature request related to a problem?

While TinyUSB already supports the H7 series, it uses an off-chip PHY over ULPI for the HS port. The H7RS series has a dedicated on-chip HS PHY, making it similar to the U5 series. The H7RS series is also distinct from the remaining H7 series, with different pin ports, and they even have their own separate HAL file, "stm32h7rsxx_hal.h".

### Describe the solution you'd like

Support for STM32H7RS series if possible.

### I have checked existing issues, dicussion and documentation

- [X] I confirm I have checked existing issues, dicussion and documentation.

EamonBrennanDRS:
Based on the reference manual ([RM0477](https://www.st.com/resource/en/reference_manual/rm0477-stm32h7rx7sx-armbased-32bit-mcus-stmicroelectronics.pdf), 3791 pages long), the OTG_FS FIFO has 1.25 Kbytes of dedicated RAM, while the OTG_HS FIFO has 4 Kbytes of dedicated RAM. Use "find in page" and search for "The USB system features" to find this.

chintal:
I would also like to see STM32H7RS support added to TinyUSB. 

I've just started trying out the H7S3L8 on its Nucleo. I might try out tinyusb on this chip in some time, if I do I will report any results. The UART is good enough for what I need to do immediately, though, so it might take a little while. 

Here's some possibly useful information:
- The chips seem to have two USBs, a USB FS and a USB HS, both with on-chip PHYs and both OTGs
- Both the FS and HS sections of the datasheet are marked as copyright Synopsys. I don't know if they are dwc2, but from other discussions I've seen in the past, it seems that's what they all are.
- The FS peripheral supports 12 host channels as host, and as device 1 bidirectional control endpoint0 and 5 IN and 5 OUT endpoints. As mentioned by OP, it has 1.25KB of dedicated RAM with FIFO controls.
- The HS peripheral supports 16 host channels as host, and as device 1 bidirectional control endpoint0 and 8 IN and 8 OUT endpoints. As mentioned by OP, it has 4KB of dedicated RAM with FIFO control.
[...]

PR #2871, opened 2024-11-08
Add support for STM32H7RSxx

chintal:
**Describe the PR**
Add (partial) support for the STM32H7RSxx MCUs, see #2816.

**Additional context**

This has been tested with USBCDC through the USB HS peripheral of the STM32H7S3L8 on the chip's Nucleo board. I am presently unable to test the USB FS peripheral on this board, but from what I can see I expect it should work as is. 

Note that this chip also has a UCPD peripheral to control USB C PD, which I have not tried to use.

### DIFF
--- a/src/common/tusb_mcu.h
+++ b/src/common/tusb_mcu.h
@@ -298,6 +298,13 @@
   #define TUP_USBIP_FSDEV_STM32
   #define TUP_DCD_ENDPOINT_MAX    8
 
+#elif TU_CHECK_MCU(OPT_MCU_STM32H7RS)
+  #define TUP_USBIP_DWC2
+  #define TUP_USBIP_DWC2_STM32
+
+  // FS has 6, HS has 9
+  #define TUP_DCD_ENDPOINT_MAX    9
+
 //--------------------------------------------------------------------+
 // Sony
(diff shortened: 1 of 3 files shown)


src/common/tusb_mcu.h
```diff
@@ -298,6 +298,13 @@
   #define TUP_USBIP_FSDEV_STM32
   #define TUP_DCD_ENDPOINT_MAX    8
 
+#elif TU_CHECK_MCU(OPT_MCU_STM32H7RS)
+  #define TUP_USBIP_DWC2
+  #define TUP_USBIP_DWC2_STM32
+
+  // FS has 6, HS has 9
```
hathach, line 305: stm32h7rs has built-in HS PHY, we should add 
```
#define TUP_RHPORT_HIGHSPEED    1
```

as well, since people will also prefer to use that port.

chintal:
@hathach I've added TUP_RHPORT_HIGHSPEED and have rebased to master.

src/common/tusb_mcu.h
chintal: Done

hathach (approved):
Look good, thank you

PR #2873, opened 2024-11-10
DWC2: fix recurrent suspend ISR without Vbus connected

HiFiPhile:
**Describe the PR**
Without Vbus connected the suspend ISR is fired every 20ms after the disconnection, which in turn calls `tud_suspend_cb()`

Can be easily reproduced on STM32F4-DISCO with HS pin PB14/PB15.

**Additional context**
It looks like the timing `TDRSMDN` of `Figure J-5 Device-Initiated Remote Wakeup Sequence` in the databook ?


hathach (approved):
look great, thank you

PR #2877, opened 2024-11-14
feature(dcd_dwc2): [ESP32P4] Added cache synchronization

roma-jam:
**Describe the PR**

**Additional context**


src/portable/synopsys/dwc2/dcd_dwc2.c
```diff
@@ -67,6 +67,16 @@ static bool _sof_en;
 //--------------------------------------------------------------------
 // DMA
 //--------------------------------------------------------------------
+// When DMA requires cache synchronization for memory
```
roma-jam, line 70: @hathach @HiFiPhile 
Hi,
may I ask to briefly check the changes in `dcd_dwc2.c` and share your opinion? 

I've implemented the cache synchronization for Device and tested it on esp32p4, port1 (HS).
I've tried to implement the feature and: 
-  Not to break the dcd_dwc2 for other controllers (where cache calls are not necessary or not implemented)
- Put the calls in the logically important places in the code (to be clear why do we need sync cache/mem)
- Use the sections and memory alignment as much as possible. 

But on P4 almost all the major memory which is visible for DMA is behind L1-Cache, so ... we need everything: section + alignment + cache sync. 

I might have miss something, because I didn't test the Host at all (if there will be any suggestions, how I can do that quickly and easy, please let me know). 

Anyway, will appreciate any feedback from your side. Thanks.

HiFiPhile: Thank you for your effort, I only have a preliminary datasheet in which it mentioned cacheable and non-cacheable access for L1 cache, maybe it supports configuring memory attribute like Cortex-M MPU ?

I think this is not necessary:
```c
    if(dma_device_enabled(dwc2)) {
      dep->doepdma = (uintptr_t)xfer->buffer;
      dsync_c2m(xfer->buffer, total_bytes);
```

Also we need to ensure `_setup_packet` is aligned to cache line and **has a size of cache line**

hathach:
this is great, give me a bit of time, I am currently in the middle of refactoring dcd dwc2 https://github.com/hathach/tinyusb/tree/enhance-dwc2-dcd

src/portable/synopsys/dwc2/dcd_dwc2.c
roma-jam: @HiFiPhile yes, you are right, not necessary. Thanks for checking. 

Currently, it is with the following alignment `TU_ATTR_ALIGNED(4)` and maybe I can change it to `CFG_TUD_MEM_ALIGN`. 
While usually: `#define CFG_TUD_MEM_ALIGN        __attribute__ ((aligned(4)))` the change shouldn't break anything. 

Thanks.

roma-jam: ... or it will. 

Do you know any other scenarios, when we need to provide `CFG_TUD_MEM_ALIGN` which is different from   `__attribute__ ((aligned(4)))` ?

HiFiPhile: Hi @tore-espressif may I ask a question, does ESP32P4 support memory management to mark a RAM section as non-cacheable for DMA use, similar to Cortex-M MPU ?

> Do you know any other scenarios, when we need to provide CFG_TUD_MEM_ALIGN which is different from __attribute__ ((aligned(4))) ?

For Cortex-M if cache invalidate/clean is used instead of making non-cacheable section by MPU, `_setup_packet` has to be aligned to 32 or 64 bytes depending the core, also it's size has to be grown to 32/64 bytes matching cache line size.

Otherwise other variables around will be erased on cache invalidate, it's explained in page 10: https://ww1.microchip.com/downloads/en/DeviceDoc/Managing-Cache-Coherency-on-Cortex-M7-Based-MCUs-DS90003195A.pdf

roma-jam: @HiFiPhile 

> I think this is not necessary:

Unfortunately, it is necessary. At least in current implementation when the class driver is using class specific requests. 
For example, in `msc class`, when we schedule the `cbw`, we need to sync c2m before getting the IRQ.
Otherwise, the data will be lost. 

I will think, where to move the `c2m` call in case of OUT scheduling not to confuse developers that much.

roma-jam: @HiFiPhile ,

Thanks for the reason to double check it!

It is not necessary. What I had before - was a collateral damage because of this TODO: https://github.com/hathach/tinyusb/blob/0569188aed17d19d5954f1e1b911e989814ec93d/src/class/msc/msc_device.c#L58

If mscd `cbw` and `csw` buffers will be not the members of `_mscd_itf` and cache line aligned - no problem and we don't need the initial line to synchronize cache.
When I synchronize it before, I damaged the `ep_out` member in memory and the first SCSI request failed.  

Thanks again, I can fix the mscd driver with optimizing the `cbw` and `csw` buffers and make a PR with changes. 

In general, refuse to use cache sync in here maybe a good idea. In general and when it is possible. 
I will check if it is possible on P4.

tore-espressif: @HiFiPhile @roma-jam The ESP32-P4 uses RISC-V cores, so while there are differences, the concept is similar. Specifically, the ESP32-P4 implements RISC-V Physical Memory Attribution (PMA), which functions similarly to ARM's Memory Protection Unit (MPU).

The PMA imposes certain constraints on memory sections, such as requiring sizes to be powers of 2 and start addresses to align with the section size. Addressing these limitations in TinyUSB would require some refactoring, and certain changes might need to be platform-specific.

If TinyUSB has previously resolved similar issues on ARM by placing buffers in non-cachable memory, adopting the same approach for the ESP32-P4 might be the most straightforward solution.

HiFiPhile: Thank you @tore-espressif, so it's basically:
- Create the section in linker with correct size /alignment
- Configure PMA
- Use `CFG_TUSB_MEM_SECTION` to put buffers in the section

?

hathach:
@roma-jam just merged https://github.com/hathach/tinyusb/pull/2881 . It mostly clean up and move thing around without behavioral changes, could you mind updating this PR using latest master. Then I will check this out, I am still open to cache sync. FYI, chipidea hs https://github.com/hathach/tinyusb/blob/master/src/portable/chipidea/ci_hs/dcd_ci_hs.c (device) and ehci (host) are one of those port use dcache sync for imxrt (M7) for dtcm. Maybe you can take a peek at for reference.

HiFiPhile:
> @roma-jam just merged #2881 . It mostly clean up and move thing around without behavioral changes, could you mind updating this PR using latest master. Then I will check this out, I am still open to cache sync. FYI, chipidea hs https://github.com/hathach/tinyusb/blob/master/src/portable/chipidea/ci_hs/dcd_ci_hs.c (device) and ehci (host) are one of those port use dcache sync for imxrt (M7) for dtcm. Maybe you can take a peek at for reference.

You haven't slept yet ? 🤣 

I prefer to use non-cacheable memory since it's clear, less dependant to architecture (like RT1170 core M4 has home made cache) and more error proof. Also much easier to setup without alignment and size constraints.

In #2865 I've converted to dcd_ci_hs to use MPU with updated BSP.

hathach:
I have just waked up, I have an early trip to visit relative today :). I still prefer to support both, once we got dcache sync work, there will be an CFG option to disble dcache function, user can then configure their own mpu as they prefer. Otherwise we will impose the mpu requirements on user, some may not be trivial for port such as ehci/ohci since it may also need virtual to physical as well since some of them are cpu

src/portable/synopsys/dwc2/dcd_dwc2.c
tore-espressif: It is even simpler, we have 1 physical memory mapped to 2 virtual addresses, one is cachable (default) the other one is non-cachable. So we just use the non-cachable address. @roma-jam will test this soon.

Unfortunately, the reference manual hasn't been made public yet

src/portable/synopsys/dwc2/dcd_dwc2.c
```diff
@@ -47,7 +47,37 @@
...
 
+#ifdef DWC2_ENABLE_MEM_CACHE
+
+#ifndef DWC2_MEM_CACHE_LINE_SIZE
+#warning "Cache line size not specified, use default(64b) instead"
+#    define DWC2_MEM_CACHE_LINE_SIZE 0x40
+#endif // DWC2_MEM_CACHE_LINE_SIZE
+
+CFG_TUD_MEM_SECTION struct {
+  union {
```
roma-jam, line 58: This is a manual tweak to protect memory after the aligned `_setup_packet` buffer. 
I have doubts, that ` __attribute__((aligned(x))` protects the memory after non-aligned variable, so I decided to solve it this way. At least there is nothing about that in the GCC doc.
Anyway, suggestions are welcome. 

I referred to: https://gcc.gnu.org/onlinedocs/gcc/Common-Variable-Attributes.html#index-aligned-variable-attribute

For class buffers (such as `mscd cbw` and `csw`, `cdc epout` and `epin` and so on, it is better to optimize in class driver code and different PR, IMHO). But anyway also should be done.

roma-jam:
@hathach Thanks, I will check the imxrt (M7) code. 
The initial idea I saw in here: https://github.com/hathach/tinyusb/pull/2576/commits/87655682822690de87ade34ceca7c21ca65dd000#r1580181750 where is was removed, so I decided to implemented something alike, but just a bit simplified.

src/portable/synopsys/dwc2/dcd_dwc2.c
hathach: Yeah, this will be a bit of an headache. gcc attribute alignment only ensure the variale alignement only, its size can be smaler, that would leave rom for other variable/data which can be corrupted when memroy to cache is called. I will try to figure out a way to address this.

hathach:
@roma-jam would you giving me the write permission to your fork's branch PR. I have made some updated to the PR to get it compiling and have the DMA somewhat working based on you cach hint. I think the only issue now is the class driver and the m2c that corrupt the data after variable e.g (64-8) after setup when we sync. (these space can be occuiped by other system variable). I think we can introduce an CFG DCACHE LINE SIZE and use that to make sure all buffer (dcd to class driver is place correctly).

```
tinyusb: ERROR: Permission to roma-jam/tinyusb.git denied to hathach. Could not read from remote repository. Please make sure you have the correct access rights and the repository exists.
```

roma-jam:
@hathach, 

I have renamed the calls, now they are `dcd_dcache_clean/invalidate()`.
Also, there are several changes in the class driver regarding the buffer optimization, I made them here, you can check the commits from here: https://github.com/espressif/tinyusb/pull/37

src/portable/synopsys/dwc2/dcd_dwc2.c
roma-jam: Hi everyone, 

I have implemented cache bypassing for dcd_dwc2, usbd_control and mscd class to verify it.
The draft could be found here: https://github.com/espressif/tinyusb/pull/38

But it gives us the picture of what should be done in all device classes, because all the buffers, which could be transferred via DMA and used at the higher layer should be noncache-able...

hathach (commented):
look good, thank you. However, I need to make a wider code to make dcache more generic and also add more CFG to make it easier for user to override it if needed. Since I cannot push to your fork directly, I push the changes to my repo and make an separate PR here  https://github.com/hathach/tinyusb/pull/2883. When this is merge, gh is smaert enough to mark this as merged as well.

tore-espressif:
Hi all, seems that I'm later for the party, I prepared a sum-up of the discussion, I will leave it here if anyone is ever interested in this thread:

### Summary
1. When the CPU accesses buffers in memory, we must:
    - Either perform explicit cache synchronization.
    - Or access the buffers via non-cacheable addresses (or place them in non-cacheable memory).
2. Both approaches work for ESPs.
3. In both cases, the buffer's **address** and **size** must be aligned to the cache line size.
4. The cache line size for ESP32-P4 is configurable. 64 or 128 bytes.
5. We prefer **explicit cache synchronization**, but we can accept alternative solutions for the sake of uniformity with other vendors. The reasons are as follows:
    - **Code safety:**  
      If we decide to use non-cacheable addresses for CPU access, we must ensure that no part of the code accesses the same memory through cacheable addresses. This is particularly challenging given that TinyUSB is a third-party component, increasing the risk of regression.
    - **Canonical approach:**  
      Explicit cache synchronization is the standard and expected solution for handling cache coherence. While this might seem like a weak argument, deviating from it could add to our technical debt. Additionally, this approach is more platform-independent, as it relies on standardized synchronization mechanisms provided by most architectures, whereas managing non-cacheable memory regions often involves platform-specific configurations.
    - **Future-proofing:**  
      Accessing memory via non-cacheable addresses only works for internal RAM. External RAM, on the other hand, can only be accessed through the cache. While we currently do not support placing USB buffers in external RAM, this limitation would become a significant blocker in future implementations.

hathach:
> Hi all, seems that I'm late for the party, I prepared a sum-up of the discussion, I will leave it here if anyone is ever interested in this thread:

Not at all, there is on-going follow up for dma (device stack), there will be a host PR afterwards

> ### Summary
> 1. When the CPU accesses buffers in memory, we must:
>    
>    * Either perform explicit cache synchronization.
>    * Or access the buffers via non-cacheable addresses (or place them in non-cacheable memory).
> 2. Both approaches work for ESPs.
> 3. In both cases, the buffer's **address** and **size** must be aligned to the cache line size.
> 4. The cache line size for ESP32-P4 is configurable. 64 or 128 bytes.

Yeah, we figure that out, it is similar to arm m7 dtcm, therefore we will follow that model. Note DMA is not enabled by default, it must be explicitly set with `CFG_TUD_DWC2_DMA_ENABLE=1`, CFG_TUD/TUH_MEM_ALIGN must also set accordingly to have correct cache line alignment. `CFG_TUD_MEM_DCACHE_LINE_SIZE` can also be set by tusb_config.h or CFLAGS, default to 64 for P4

> 5. We prefer **explicit cache synchronization**, but we can accept alternative solutions for the sake of uniformity with other vendors. The reasons are as follows:
>    
>    * **Code safety:**
>      If we decide to use non-cacheable addresses for CPU access, we must ensure that no part of the code accesses the same memory through cacheable addresses. This is particularly challenging given that TinyUSB is a third-party component, increasing the risk of regression.
>    * **Canonical approach:**
>      Explicit cache synchronization is the standard and expected solution for handling cache coherence. While this might seem like a weak argument, deviating from it could add to our technical debt. Additionally, this approach is more platform-independent, as it relies on standardized synchronization mechanisms provided by most architectures, whereas managing non-cacheable memory regions often involves platform-specific configurations.
>    * **Future-proofing:**
>      Accessing memory via non-cacheable addresses only works for internal RAM. External RAM, on the other hand, can only be accessed through the cache. While we currently do not support placing USB buffers in external RAM, this limitation would become a significant blocker in future implementations.

explicit dache clean/invalidate is currently implemented. If user want to use non-cache, can simply override the `CFG_TUH_MEM_DCACHE_ENABLE=0` which is enabled when DMA mode is enabled. So I guess we are all good. You wan to review the other PR, note though I made bunch of un-related bsp/freertos changes as well.

PR #2883, opened 2024-11-20
[DCD_DWC2][ESP32P4][HS] Added cache synchronization (cont)

hathach:
**Describe the PR**
Superceded https://github.com/hathach/tinyusb/pull/2877 since I need to make some modification to have `CFG_TUSB_MEM_ALIGN=__attribute__((aligned(64)))` for p4 and add generic CFG for CFG_TUD_MEM_DCACHE_ENABLE/CFG_TUD_MEM_DCACHE_LINE_SIZE


hw/bsp/espressif/components/tinyusb_src/CMakeLists.txt
```diff
@@ -15,6 +15,13 @@ list(APPEND compile_definitions
   BOARD_TUH_MAX_SPEED=${RHPORT_HOST_SPEED}
   )
 
+if (target STREQUAL esp32p4)
+  # P4 change alignment to 64 (DCache line size) for possible DMA configuration
+  list(APPEND compile_definitions
+    CFG_TUSB_MEM_ALIGN=__attribute__\(\(aligned\(64\)\)\)
```
hathach, line 21: CFG_TUSB_MEM_ALIGN will cause class driver to put correct aligment on usb/dma buffer.

src/tusb_option.h
```diff
@@ -422,6 +424,18 @@
...
+#ifndef CFG_TUD_MEM_DCACHE_ENABLE
+  #ifndef CFG_TUD_MEM_DCACHE_ENABLE_DEFAULT
+  #define CFG_TUD_MEM_DCACHE_ENABLE_DEFAULT  0
+  #endif
+
+  #define CFG_TUD_MEM_DCACHE_ENABLE   CFG_TUD_MEM_DCACHE_ENABLE_DEFAULT
+#endif
+
+#ifndef CFG_TUD_MEM_DCACHE_LINE_SIZE
+  #define CFG_TUD_MEM_DCACHE_LINE_SIZE 32
```
hathach, line 436: should work with most ARM core like M7

src/portable/synopsys/dwc2/dcd_dwc2.c
```diff
@@ -60,19 +57,49 @@ typedef struct {
...
 #define XFER_CTL_BASE(_ep, _dir) (&xfer_status[_ep][_dir])
 
-// EP0 transfers are limited to 1 packet - larger sizes has to be split
-static uint16_t ep0_pending[2];  // Index determines direction as tusb_dir_t type
-static uint16_t _dfifo_top;      // top free location in DFIFO in words
+typedef struct {
+  CFG_TUD_MEM_ALIGN union {
+    uint32_t setup_packet[2];
+#if CFG_TUD_MEM_DCACHE_ENABLE
+    uint8_t setup_packet_cache_padding[CFG_TUD_MEM_DCACHE_LINE_SIZE];
```
hathach, line 64: @roma-jam we will need to do this to all buffered in the class driver, which may take a couple of follow-up PRs.

hw/bsp/espressif/components/tinyusb_src/CMakeLists.txt
hathach: @roma-jam I just realized that we probably does not need the dma buffer to be 64 bytes aligned ? It only need to occupies the whole cache line (64 byte) right ? If you we can change it back to simply 4 bytes alignment.

tore-espressif: Unfortunately the cache line size is configurable, and can be 64 or 128 bytes :( 
This might be good enough for now, just so you keep this in mind if someone reports issues...

hathach: what is the default value ? it isn't an problem at all. both the CFG_TUSB_MEM_ALIGN and CFG_TUD_MEM_DCACHE_LINE_SIZE is configurable at user level (tusb_config.h). So if use change it to 128, they shoud set this accordingly.

tore-espressif: default is 64. OK, great!

PR #2891, opened 2024-11-28
fix(dcd_dwc2): Fix EP IN counters assignment and usage

roma-jam:
### Requirements
This PR contain two small fixes, which are crucial for using device configuration with fully load EP amount. 
ESP32S2/S3 and ESP32P4.

### Description

1.  Flushing `_dcd_data.allocated_epin_count` to zero while: 
    - `dcd_edpt_close_all()` as we will allocate FIFO for EP0 later in `dfifo_device_init(rhport);`
    -  `handle_bus_reset(uint8_t rhport)` as we will allocate FIFO for EP0 later in `dfifo_device_init(rhport);`
2. Fix the usage of counters for `TU_ASSERT`. When we use `_dwc2_controller` instead of `dwc2_controller` we request the `ep_in_count` member value from the first member, which is FS. This is important, when we run on HS PHY. 

### Related
- Cherry-picked from https://github.com/espressif/tinyusb/pull/41

### DIFF
--- a/src/portable/synopsys/dwc2/dcd_dwc2.c
+++ b/src/portable/synopsys/dwc2/dcd_dwc2.c
@@ -539,7 +539,7 @@ void dcd_edpt_close_all(uint8_t rhport) {
   dwc2_regs_t* dwc2 = DWC2_REG(rhport);
   uint8_t const ep_count = _dwc2_controller[rhport].ep_count;
 
-  _dcd_data.allocated_epin_count = 1;
+  _dcd_data.allocated_epin_count = 0;
 
   // Disable non-control interrupt
   dwc2->daintmsk = (1 << DAINTMSK_OEPM_Pos) | (1 << DAINTMSK_IEPM_Pos);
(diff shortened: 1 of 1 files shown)


src/portable/synopsys/dwc2/dcd_dwc2.c
```diff
@@ -646,7 +646,7 @@ static void handle_bus_reset(uint8_t rhport) {
   tu_memclr(xfer_status, sizeof(xfer_status));
 
   _dcd_data.sof_en = false;
-  _dcd_data.allocated_epin_count = 1;
+  _dcd_data.allocated_epin_count = 0;
```
hathach, line 649: is this correct, I thought EP0 also occupies 1 IN ?

roma-jam: @hathach Yes, it does. 
 But when we call one by one: 

 1. `dfifo_alloc(rhport, 0x80, CFG_TUD_ENDPOINT0_SIZE);`
 2.  `_dcd_data.allocated_epin_count++;`
 
We have `allocated_epin_count == 2`
after bus_reset and allocation fifo for EP0 IN (if on reset we set it to `allocated_epin_count = 1`)

hathach (approved):
spot-on thank you for the fix. By the way, I think we can do better in the future, by dynamically assign txfifo to active endpoint (with transfer) like we did with channel. That way we can support full endpoint.But it is just my thought.

PR #2895, opened 2024-12-02
Disable DWC HCD interrupt with MAX3421

tannewt:
hcd_int_handler isn't found otherwise.


hathach (approved):
oh right, I forgot s2/s3 can also be used with max3421e as well.

PR #2925, opened 2025-01-01
Added initial support for FRDM-MCXA156.  I will be porting tinyuf2 to…

ehughes:
I will be porting tinyuf2 to MCXA156 and N947 and need this in place.

It is similar to A153 but with FPU,SIMD enable int the CM33 core.  It has much more flash/ram/peripherals and uses the chipidea FS controller.

I repointed the entry in the get_deps script to use NXP MCU repo.  The existing current code couldn't build as the MCX Xpresso SDK was old.

For now, I put a copy of fsl_spc.c in the drivers folder for the mcx family.  None of the ports could build with the version in the mcux github sdk.  The sdk on github has version 2.4.0.   The SDK package that can be downloaded with the SDK generator uses 2.4.2.  All of the clock_config routines use versions of the datastructure from 2.4.2.  For now, keeping an updated copy in hw/bsp/mcx/drivers so we can build against the latest mcuxpresso sdk and enable MCXA156.  I'll file an issue on the nxp github.

I cleaned up the family.c a little bit.   these chips don't have SCT so I removed the neopixel code.  I fixed a clock init issue w/ N947 that was causing it to crash when initializing the uart.

I tested the cmake and make scripts on the A153,A156 and N947 using the CDC/MSC sample.  All worked OK.


ehughes:
@hathach 

1.)  I saw the ci failed on LPC55.   It will need to have some minor updates to the make file as the structure of the mcuxpresso sdk is a little different.   I can make this change if it is OK for us to point to the nxp github repo

2.)  the ci also failed on the mcxn947-brk board.   This board is now deprecated and used prototype silicon.  NXP never release the board publicly.  I can patch the make file to be like the FRDM board, we can delete  the board.   The FRDM-MCXN947 is the official dev board that people can purchase.

PR #2939, opened 2025-01-09
Fix potential out of bounds access in msc_disk.c

PwnVerse:
## Description

This PR addresses a potential buffer overflow issue in the tud_msc_read10_cb function. Previously, the function did not validate whether the combination of offset and bufsize exceeded the disk block size, which could result in out-of-bounds memory access during the memcpy operation.

## Changes

Added a validation check to ensure `offset + bufsize` does not exceed the defined `DISK_BLOCK_SIZE`. If this condition is violated, the function now returns -1.


HiFiPhile:
Thank you  !

HiFiPhile (approved):

examples/device/cdc_msc/src/msc_disk.c
```diff
@@ -192,6 +192,9 @@ int32_t tud_msc_read10_cb(uint8_t lun, uint32_t lba, uint32_t offset, void* buff
   // out of ramdisk
   if ( lba >= DISK_BLOCK_NUM ) return -1;
 
+  // Check for overflow of offset + bufsize
+  if ( offset + bufsize >= DISK_BLOCK_SIZE ) return -1;
```
hathach, line 196: Logic is not correct, offset = 0 and bufsize = 512 is correct value. This cause hardware test (hil) to fail on all cdc_msc test. We will need to revert this PR

PR #2948, opened 2025-01-14
src/class/hid/hid_host.c: fix logging calls for epbuf

dhalbert:
Some logging calls were referring to `hidh_interface_t` fields that no longer existed, causing compiler errors when logging was enabled.

I am not sure if my changes are correct: please check. But the file now compiles when logging is enabled.


hathach:
thanks Dan, look like there is an issue with ceedling unit test with new ci's ubuntu 24.04, let me do a quick check to see how to fix it.

hathach (approved):
thank you, ceedling issue is fixed with #2949

PR #2950, opened 2025-01-15
fix(ncm): Return invalid NTBs to free list

tore-espressif:
In case we received invalid datagram, we silently fail and the buffer was not returned to empty list -> it was lost.

 If this happened more than CFG_TUD_NCM_OUT_NTB_N times, we run out of NTBs and all OUT transfers are NACKed.

Closes https://github.com/espressif/esp-usb/issues/107

### DIFF
--- a/src/class/net/ncm_device.c
+++ b/src/class/net/ncm_device.c
@@ -857,7 +857,8 @@ bool netd_xfer_cb(uint8_t rhport, uint8_t ep_addr, xfer_result_t result, uint32_
     // - if there is a free receive buffer, initiate reception
     if (!recv_validate_datagram(ncm_interface.recv_tinyusb_ntb, xferred_bytes)) {
       // verification failed: ignore NTB and return it to free
-      TU_LOG_DRV("(EE) VALIDATION FAILED. WHAT CAN WE DO IN THIS CASE?\n");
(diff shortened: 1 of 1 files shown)


tore-espressif:
@HiFiPhile @rgrr Could you please have a look?

rgrr:
> @HiFiPhile @rgrr Could you please have a look?

Seems reasonable to me.  My memory is getting darker there, but AFAIR the buffer had been fetched before via recv_get_free_ntb().  And either it should go then into the ready list or back into the free list.

What I find a little surprising is, that bad blocks are arriving at that point.  Never had that.

tore-espressif:
Thanks for the confirmation!

> What I find a little surprising is, that bad blocks are arriving at that point. Never had that.

I did not investigate why the datagrams are invalid; this is only a hot-fix. During upload throughput test (that is, NCM device is receiving with OUT transfers) I got several invalid datagrams per second...

tore-espressif:
As I said, I did not dig too deep; this was my setup
```
| Host PC | <- USB -> | ESP32-S3| <- WiFi -> | Local network with internet access |
```

When I run speed test at [speedtest.net  ](https://www.speedtest.net/) I got this error often

HiFiPhile:
Thanks for fixing it !

HiFiPhile (approved):

PR #2954, opened 2025-01-20
Add basic USB CDROM support via USB MSC

i-am-shodan:
**Describe the PR**
USB CDROM support requires that the SCSI INQUIRY command returns a different device type and qualifier. The current API for MSC devices doesn't allow the caller to influence these or appear as any other device type. This PR adds a new API function tud_msc_scsi_device_type(uint8_t lun) that enables expandability of SCSI device types as well as two new commands SCSI_CMD_READ_TOC and SCSI_CMD_READ_DISC_INFO.

In order to appear as a CDROM the caller must:
* Implement tud_msc_scsi_device_type and return MSC_SCSI_DEVICE_CDROM
* Implement SCSI_CMD_READ_TOC in tud_msc_scsi_cb  (this returns the CDROM table of contents)
* Implement SCSI_CMD_READ_DISC_INFO in tud_msc_scsi_cb which returns disc info - in many cases this can be uint8_t[2] = {0}

This is a breaking change if you are currently using MSC and you're getting a compilation error you'll need to add.

```
msc_scsi_device_type_t tud_msc_scsi_device_type(uint8_t lun)
{
  (void) lun;
  return MSC_SCSI_DEVICE_BLOCK;
}
```

### DIFF
--- a/examples/device/cdc_msc/src/msc_disk.c
+++ b/examples/device/cdc_msc/src/msc_disk.c
@@ -269,4 +269,10 @@ int32_t tud_msc_scsi_cb (uint8_t lun, uint8_t const scsi_cmd[16], void* buffer, 
   return (int32_t) resplen;
 }
 
+msc_scsi_device_type_t tud_msc_scsi_device_type(uint8_t lun)
+{
+  (void)lun;
+  return MSC_SCSI_DEVICE_BLOCK;
+}
+
 #endif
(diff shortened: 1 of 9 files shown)


hathach:
https://github.com/hathach/tinyusb/pull/3172 add a full inquiry response (v2) callback. It is the better approach since it also support vendor data if needed. Should have doone this from beginning.

PR #2959, opened 2025-01-23
stm32: fix compilation error when not using uart

ho-ho-ho:
**Describe the PR**
When not using uart the macro `UART_CLK_EN` is most likely also not defined (as it is defined in the board.h alongside `UART_DEV`). On some stm32 families (i.e. l0 and f1) this is handled correctly and the `UART_CLK_EN` macro is called from the `#ifdef UART_DEV` section, but on some families the macro is always called, resulting in a undefined symbol compile error. 

This pr moves the call to `UART_CLK_EN` into the conditional section for the families with the wrong behaviour. It also only creates the `UartHandle` variable when uart is actually in use (more in line with the behaviour of the `stm32l0` family which I consider to be the best implementation of this conditional compilation part).


HiFiPhile:
Thank you !

HiFiPhile (approved):

PR #2964, opened 2025-01-25
fix bug introduced by 2939, with correct offset check logic 

hathach:


HiFiPhile:
Ops... I was too hurry 

I didn't merge latest master in #2963, wondering why it works well on PC but not CI....

hathach:
No problem at all, I made this type of mistake all over the places :)

PR #2967, opened 2025-01-29
MSC Device: Add asynchronous IO support

HiFiPhile:
**Describe the PR**

Currently, mixing the MSC class with other classes could be problematic since IO operations could block `tud_task()` for a long time and degrade real-time performance, especially for isochronous and interrupt transfer-based classes.

Now, it's possible to delegate IO operations to a background task.

The `cdc_msc_freertos` example has been updated for demonstration. One can change `CFG_EXAMPLE_MSC_IO_DELAY_MS` to a high value and type in the console while doing disk IO. With `CFG_EXAMPLE_MSC_ASYNC_IO` ON and OFF, the latency difference should be very noticeable.


602p:
I'm using this in a work project. I'd love to see something around these lines landed.

DazzlingOkami:
This is a very practical feature, and I also hope that it can be implemented as soon as possible.

examples/device/cdc_msc/src/msc_disk.c
```diff
@@ -184,7 +184,7 @@ int32_t tud_msc_read10_cb(uint8_t lun, uint32_t lba, uint32_t offset, void *buff
   }
 
   // Check for overflow of offset + bufsize
-  if (offset + bufsize > DISK_BLOCK_SIZE) {
+  if (lba * DISK_BLOCK_SIZE + offset + bufsize > DISK_BLOCK_NUM * DISK_BLOCK_SIZE) {
```
hathach, line 187: good spot

src/class/msc/msc_device.c
```diff
@@ -219,6 +263,30 @@ static inline void set_sense_medium_not_present(uint8_t lun) {
   tud_msc_set_sense(lun, SCSI_SENSE_NOT_READY, 0x3A, 0x00);
 }
 
+void tud_msc_async_io_done(int32_t bytes_processed) {
+  // Precheck to avoid queueing multiple RW done callback
+  TU_VERIFY(_mscd_itf.next_op != MSC_NEXT_OP_NONE,);
+  // Call usbd_edpt_xfer() in tud_task() to avoid racing condition
+  usbd_defer_func(tud_msc_async_io_done_cb, (void*) (intptr_t)bytes_processed, false);
```
hathach: I think we need to add in_isr since some I/O can be done in ISR. also I think we only need to defer function when it is invoked in ISR. Let me know if that makes sense to you.

HiFiPhile:
Thank you, `use cbw.command[0] for pending IO command` made it easier to for command extension :)

Without `usbd_defer_func` `proc_read_io_data\proc_write_io_data` will be executed in IO task's context, I think it's risky to have both IO task and TUSB task to modify class internal state simultaneously.

hathach:
> Thank you, `use cbw.command[0] for pending IO command` made it easier to for command extension :)
> 
> Without `usbd_defer_func` `proc_read_io_data\proc_write_io_data` will be executed in IO task's context, I think it's risky to have both IO task and TUSB task to modify class internal state simultaneously.

Yeah, I think you are right, I will revert that to force usbd defer

PR #3002, opened 2025-02-27
fix(esp): Remove deprecated include

tore-espressif:
**Describe the PR**
soc/usb_pins.h header will be deprecated in IDF v6.0. It is not needed nor used in current tinyusb

**Additional context**


hathach (approved):
thank you

PR #3007, opened 2025-03-03
Fix(host) enumeration with 8bitdo devices

### LINKED ISSUE #2163: HID host issues with RetroFlag Classic USB Gamepad
### Operating System

Linux

### Board

Raspberry Pi Pico

### Firmware

examples/host/cdc_msc_hid

(With an extra `printf` to `tuh_hid_report_received_cb` to confirm that I was getting no reports)

### What happened ?

This device is a bit weird in that it enumerates as a Nintendo Switch controller, then resets itself and enumerates as a generic xinput or HID controller.

With TinyUSB it fails in various different ways depending on the log level and other factors (timing issues?). Though I consistently fail to get any reports.

It also seems to fail entirely if you don't connect the controller after resetting the pico (and disconnect before).

(Additional logs: https://gist.github.com/Daft-Freak/fd7ba764e5f300caf474bede74966ba8)

### How to reproduce ?

1. Make sure controller is not connected.
2. Build/install cdc_msc_hid
3. Connect controller (holding Y the first time to make sure it's not in xinput mode)
4. Get no HID reports, maybe get an assert failure or a hang

[CFG_TUSB_DEBUG=2.txt](https://github.com/hathach/tinyusb/files/12122916/CFG_TUSB_DEBUG.2.txt)

### Screenshots

_No response_

hathach:
**Describe the PR**
- fix(hcd_rp2040) assert/panic endpoint already active: when a device reset while having on-going control transfer
- featuer(host) always get language id, manufacturer, product and serial string. Which is required by some device such as 8bitdo. Should fix #2163 
- add tuh_descriptor_get_string_langid() API 

@tannewt please try to see if this works with your 8bitdo gamepad

PS: IAR build failed due to server maintanence

### DIFF
--- a/examples/host/cdc_msc_hid/src/hid_app.c
+++ b/examples/host/cdc_msc_hid/src/hid_app.c
@@ -39,18 +39,16 @@
 static uint8_t const keycode2ascii[128][2] =  { HID_KEYCODE_TO_ASCII };
 
 // Each HID instance can has multiple reports
-static struct
-{
+static struct {
   uint8_t report_count;
   tuh_hid_report_info_t report_info[MAX_REPORT];
-}hid_info[CFG_TUH_HID];
+} hid_info[CFG_TUH_HID];
 
 static void process_kbd_report(hid_keyboard_report_t const *report);
 static void process_mouse_report(hid_mouse_report_t const * report);
 static void process_generic_report(uint8_t dev_addr, uint8_t instance, uint8_t const* report, uint16_t len);
 
-void hid_app_task(void)
-{
+void hid_app_task(void) {
   // nothing to do
 }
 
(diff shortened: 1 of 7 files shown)


HiFiPhile:
Since you are working on usbh, I found at least on NUCLEO-H7S3L8 speed detection is more stable by putting `tusb_time_delay_ms_api(ENUM_DEBOUNCING_DELAY_MS);` at the start of `enum_new_device()`.

hathach:
> Since you are working on usbh, I found at least on NUCLEO-H7S3L8 speed detection is more stable by putting `tusb_time_delay_ms_api(ENUM_DEBOUNCING_DELAY_MS);` at the start of `enum_new_device()`.

hmm that is weird, delay before port_reset() shouldn't make a difference since port got reset anyway, and the delay afterwards is rather generous ~450ms. Thank you for the hint, I will check this out next time I test with h7rs.

HiFiPhile:
If I remember correctly dwc2 does speed detection at the end of reset. If DEBOUNCING_DELAY_MS is placed after the end of reset then it won't help speed detection since it's already done.

hathach:
@HiFiPhile it is weird, there is `ENUM_RESET_DELAY_MS` which is 50ms per USB specs, it should be enough for dwc2 to detect the link speed, the `ENUM_DEBOUNCING_DELAY_MS` is mostly waiting for device to bootup/clock init before we start talking to it iirc. I will check again later. 

```C
    hcd_port_reset(_dev0.rhport);

    // Since we are in middle of rhport reset, frame number is not available yet.
    // need to depend on tusb_time_millis_api()
    tusb_time_delay_ms_api(ENUM_RESET_DELAY_MS);

    hcd_port_reset_end(_dev0.rhport);

    // wait until device connection is stable TODO non blocking
    tusb_time_delay_ms_api(ENUM_DEBOUNCING_DELAY_MS);
```

PR #3014, opened 2025-03-07
Allow different port and mode for LED on ch32v boards

deshipu:
While there is a define for the port, the clock enable is hardcoded for GPIOA, so setting a different port than GPIOA doesn't work. This fixes it by adding a define for enabling the port clock.

It also adds a define for the pin mode, because not all boards have the LED connected in a way that open drain works with it.


deshipu:
You are right, I'm sorry I didn't notice it. I only tested it on my own board that has the LED on port C, which is APB2. I'll fix it immediately.

HiFiPhile (approved):

Issue #3028, opened 2025-03-13
midi_host.c does not build if CFG_TUH_MIDI_STREAM_API is 0


rppicomidi:
### Operating System

Linux

### Board

raspberry pi pico

### Firmware

examples/host/midi_rx except modified `tusb_config.h` to add the line
```
#define  CFG_TUH_MIDI_STREAM_API 0
```

### What happened ?

Build failed.
```
pico-sdk/lib/tinyusb/src/class/midi/midi_host.c:156:24: error: 'midih_interface_t' has no member named 'stream_read'
[build]   156 |       tu_memclr(&p_midi->stream_read, sizeof(p_midi->stream_read));
pico-sdk/lib/tinyusb/src/class/midi/midi_host.c:157:53: error: 'midih_interface_t' has no member named 'stream_write'
[build]   157 |       tu_memclr(&p_midi->stream_write, sizeof(p_midi->stream_write));
```

The two `tu_memclr()` function calls should be surrounded in `#if CFG_TUH_MIDI_STREAM_API` and `#endif`. I will submit a pull request shortly.

### How to reproduce ?

`tusb_config.h` to add the line
```
#define  CFG_TUH_MIDI_STREAM_API 0
```

Build the application normally.

### Debug Log as txt file (LOG/CFG_TUSB_DEBUG=2)

N/A

### I have checked existing issues, dicussion and documentation

- [x] I confirm I have checked existing issues, dicussion and documentation.

hathach:
fixed by #3029

PR #3029, opened 2025-03-13
Only clear stream_read & stream_write if they are defined

rppicomidi:
**Describe the PR**
Fix issue #3028

**Additional context**
See Issue #3028


hathach (approved):
thank you for the fix

PR #3063, opened 2025-04-04
dwc2/host: clear SOF flag in handle_sof_irq()

maximevince:
**Describe the PR**
The SOF flag was not cleared in the `hcd_int_handler()` ISR or `handle_sof_irq()` function inside of it.
This caused recurring and infinite IRQs in some cases (e.g. when connecting a HUB). 

**Additional context**
See discussion here: https://github.com/hathach/tinyusb/discussions/3062


HiFiPhile (approved):
Thank you for the fix !

PR #2960, opened 2025-01-24
Add NUCLEO-H7S3L8 BSP

HiFiPhile:
**Describe the PR**

Add NUCLEO-H7S3L8 support.

Status:

| | FS port| HS port|
|--------|--------|--------|
| Device| Unknown| Working|
| Host| Unknown| *Has issues| 

\*  Unattach is not detected, resulting `_usbh_devices` overflow.
\* High speed device can't enumerate.

Either not advacning:
```TinyUSB Device Info Example
USBH init on controller 1, speed = Full
sizeof(usbh_device_t) = 86
sizeof(hcd_event_t) = 12
sizeof(_ctrl_xfer) = 16
sizeof(tuh_xfer_t) = 24
sizeof(tu_fifo_t) = 12
sizeof(tu_edpt_stream_t) = 20
HUB init
guid, gsnpsid, ghwcfg1, ghwcfg2, ghwcfg3, ghwcfg4
0x00005000, 0x4F54411A, 0x00000000, 0x228FE052, 0x03B882E8, 0xE2103E30
Highspeed UTMI+ PHY init
DMA = 0
[1:] USBH Device Attach
High Speed
[1:0] Open EP0 with Size = 8
Get 8 byte of Device Descriptor
[1:0] Get Descriptor: 80 06 00 01 00 00 08 00
on EP 00 with 8 bytes: OK
on EP 80 with 8 bytes: OK
[1:0] Control data:
  0000:  12 01 10 02 00 00 00 40                          |.......@|
```

Or recognized as FS:
```
TinyUSB Device Info Example
USBH init on controller 1, speed = Full
sizeof(usbh_device_t) = 86
sizeof(hcd_event_t) = 12
sizeof(_ctrl_xfer) = 16
sizeof(tuh_xfer_t) = 24
sizeof(tu_fifo_t) = 12
sizeof(tu_edpt_stream_t) = 20
HUB init
guid, gsnpsid, ghwcfg1, ghwcfg2, ghwcfg3, ghwcfg4
0x00005000, 0x4F54411A, 0x00000000, 0x228FE052, 0x03B882E8, 0xE2103E30
Highspeed UTMI+ PHY init
DMA = 0
[1:] USBH Device Attach
Full Speed
[1:0] Open EP0 with Size = 8
Get 8 byte of Device Descriptor
[1:0] Get Descriptor: 80 06 00 01 00 00 08 00
on EP 00 with 0 bytes: FAILED
[1:0] Control FAILED, xferred_bytes = 0
[...]

PR #3070, opened 2025-04-07
host/dwc2: fix bitfields access width

### LINKED ISSUE #3047: freertos reset with ESP_RST_INT_WDT on esp32-s3 when using option CFG_TUH_DWC2_SLAVE_ENABLE
### Operating System

Windows 11

### Board

M5stack cardputer with espressif esp32-s3

### Firmware

FREETOS

### What happened ?

I am trying to use cardputer with esp3-s3 microcontroller in USB host mode.
The 1st hurdle is that for ESP32-S3 code for USB PHY initialization is missing in tinyusb. This can be circumvented by using a function from the espressif usb component (usb_new_phy).
After calling usb_new_phy the call to tusb_init will succeed.
tusb_init is followed by tuh_task.
We wait in tuh_task_ext until a device gets plugged in. 
This results in a call to process_enumeration.

When tinyusb gets compiled with CFG_TUH_DWC2_DMA_ENABLE == 1, everything goes according to plan.
Compiling with **CFG_TUH_DWC2_SLAVE_ENABLE** == 1 and running the code, **crashes the OS**.

I could follow the code execution up to inside channel_xfer_start.
With CFG_TUH_DWC2_SLAVE_ENABLE we seem to face kind of interrupt storm from the FIFO that passes execution to handle_txfifo_empty. I did a "blocking wait" at the start of handle_txfifo_empty and got multiple executions of the function while still blocking. This lets me believe, that we get the fifo interrupt multiple times even though we did not yet handle the 1st one. From my understanding this will all happen on the same stack and finally lead to a stack overflow.

As a sidenote: This happens when using synopsys\dwc2 and espressif\esp32sx. It does not happen when using CFG_TUH_DWC2_DMA_ENABLE

### How to reproduce ?

Compile tinyusb for esp32-s3 with option CFG_TUH_DWC2_SLAVE_ENABLE

#define CFG_TUH_ENABLED       1
#define CFG_TUSB_OS                 OPT_OS_FREERTOS
#define CFG_TUH_DWC2_SLAVE_ENABLE 1

Run code that calls tuh_task();

Plugin a USB device and watch the OS crash.

Cannot generate a debug log as the OS crashes and a file cannot be stored a connected microsd card (file size will be 0 as it will not be flushed)

### Screenshots

_No response_

- [x] I confirm I have checked existing issues, dicussion and documentation.

HiFiPhile:
**Describe the PR**
Fix gcc-riscv issue about volatile bitfields are incorrectly accessed using an access width not appropriate to the type of its container.

Tested on `espressif_s3_devkitm` with `device_info` and  `cdc_msc_freertos` examples.

For the host I've tested an usb key and an FTDI usb-uart.
```
Device 1: ID 0403:6001 SN FT1ZK9MQ
Device Descriptor:
  bLength             18
  bDescriptorType     1
  bcdUSB              0200
  bDeviceClass        0
  bDeviceSubClass     0
  bDeviceProtocol     0
  bMaxPacketSize0     8
  idVendor            0x0403
  idProduct           0x6001
  bcdDevice           0600
  iManufacturer       1     FTDI
  iProduct            2     TTL232R-3V3
  iSerialNumber       3     FT1ZK9MQ
  bNumConfigurations  1

Device 2: ID 0011:7788 SN C4D197AC
Device Descriptor:
  bLength             18
  bDescriptorType     1
  bcdUSB              0200
  bDeviceClass        0
  bDeviceSubClass     0
  bDeviceProtocol     0
  bMaxPacketSize0     64
  idVendor            0x0011
  idProduct           0x7788
  bcdDevice           0104
  iManufacturer       1     Generic
  iProduct            2     Mass Storage
  iSerialNumber       3     C4D197AC
  bNumConfigurations  1
```

Disconnection detection needs #2960 to work otherwise device index will go up 1,2,3,4...

Fix #3047

### DIFF
--- a/hw/bsp/BoardPresets.json
+++ b/hw/bsp/BoardPresets.json
@@ -13,20 +13,30 @@
       }
     },
     {
+      "name": "default single",
+      "hidden": true,
+      "description": "Configure preset for the ${presetName} board",
+      "generator": "Ninja",
+      "binaryDir": "${sourceDir}/build/${presetName}",
+      "cacheVariables": {
+        "BOARD": "${presetName}"
+      }
+    },
+    {
       "name": "adafruit_clue",
       "inherits": "default"
     },
     {
       "name": "adafruit_feather_esp32_v2",
-      "inherits": "default"
+      "inherits": "default single"
     },
     {
       "name": "adafruit_feather_esp32s2",
-      "inherits": "default"
+      "inherits": "default single"
     },
     {
       "name": "adafruit_feather_esp32s3",
-      "inherits": "default"
+      "inherits": "default single"
     },
     {
       "name": "adafruit_magtag_29gray",
@@ -34,7 +44,7 @@
     },
     {
       "name": "adafruit_metro_esp32s2",
-      "inherits": "default"
+      "inherits": "default single"
     },
     {
       "name": "apard32690",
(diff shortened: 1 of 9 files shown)


hathach:
thank you, I am thinking, maybe we just remove the struct/bitfield in the register typedef i.e leaving only uint32_t. bitfield struct can be use with local variale only. Will try that when I got sometime,

hathach:
> > I am thinking, maybe we just remove the struct/bitfield in the register typedef i.e leaving only uint32_t. bitfield struct can be use with local variable only
> 
> Yes we can do let this. Adding a macro makes the thing complicate and losing the interest of bitfields.

sure, I will do this later if you are busy. I need to fix an racing issue with pio-usb when overclocking rp2350 to 264mhz first :)

HiFiPhile:
I've updated the refactor.

hathach (approved):
perfect, thank you. I couldn't done it any better. I change dwc2_ep/channel_count() to inline function to avoid forcing gnu standard. The rest is mostly clean up

PR #3074, opened 2025-04-11
dwc2/host: HFIR: Fix timing off-by-one

maximevince:
**Describe the PR**
HFIR register for SOF timing was programmed with an off-by-one value (it needs a -1 applied).

- This probably comes from a historical error in the Synopsys databook. They corrected it in version 3.30a, according to this comment on a mailinglist: https://lists.infradead.org/pipermail/linux-rpi-kernel/2016-February/003290.html

- Also, the official Linux kernel dwc2 driver (maintained by Synopsys) contains the values with `-1` offset, as can be seen in the `dwc2_calc_frame_interval()` function here:
https://github.com/torvalds/linux/blob/900241a5cc15e6e0709a012051cc72d224cd6a6e/drivers/usb/dwc2/hcd.c#L326

@HiFiPhile @roma-jam


HiFiPhile:
Thank you !

Hum that's interesting, on databook 4.20a the `- 1` is still missing.

maximevince:
Huh! I don't have access to their databook(s), unfortunately. So I just copied the mentioned versions from the mailing list thread!

PR #3076, opened 2025-04-11
host/dwc2: cleanup transfer on device close

HiFiPhile:
**Describe the PR**
Fix xfer complete event of previous enumeration mess with next enumeration.

```
[1:1] Get Descriptor: 80 06 03 03 09 04 00 01 
on EP 00 with 8 bytes: OK
on EP 80 with 40 bytes: OK
[1:1] Control data:
  0000:  28 03 4D 00 61 00 73 00 73 00 20 00 53 00 74 00  |(.M.a.s.s. .S.t.|
  0010:  6F 00 72 00 61 00 67 00 65 00 20 00 44 00 65 00  |o.r.a.g.e. .D.e.|
  0020:  76 00 69 00 63 00 65 00                          |v.i.c.e.|
[1:0:0] USBH DEVICE REMOVED
[1:0:0] unplugged address = 1
Device removed, address = 1
[1:] USBH Device Attach
High Speed
[1:0] Open EP0 with Size = 8
Get 8 byte of Device Descriptor
[1:0] Get Descriptor: 80 06 00 01 00 00 08 00 
on EP 00 with 8 bytes: OK
on EP 00 with 8 bytes: OK
[1:0] Control data:
  0000:  12 01 00 02 00 00 00 40                          |.......@|
on EP 80 with 8 bytes: OK

Set Address = 1
[1:0] Set Address: 00 05 01 00 00 00 00 00 
on EP 00 with 0 bytes: OK
on EP 00 with 8 bytes: OK

[1:1] Open EP0 with Size = 64
Get Device Descriptor
[1:1] Get Descriptor: 80 06 00 01 00 00 12 00 
on EP 80 with 8 bytes: OK
hcd_edpt_xfer 655: ASSERT FAILED
```

### DIFF
--- a/src/portable/synopsys/dwc2/hcd_dwc2.c
+++ b/src/portable/synopsys/dwc2/hcd_dwc2.c
@@ -448,6 +448,7 @@ void hcd_device_close(uint8_t rhport, uint8_t dev_addr) {
     hcd_endpoint_t* edpt = &_hcd_data.edpt[i];
     if (edpt->hcchar_bm.enable && edpt->hcchar_bm.dev_addr == dev_addr) {
       tu_memclr(edpt, sizeof(hcd_endpoint_t));
+      tu_memclr(&_hcd_data.xfer[i], sizeof(hcd_xfer_t));
     }
   }
 }


maximevince:
While I can see the issue described here, it seems to me this is not the proper fix, but rather a workaround - with it's own side-effects.
Why wasn't this xfer cleared? And what if is currently ongoing?

There seem to be some race conditions between handling some stuff in the interrupt handler, and deferring some stuff to the tuh_task(). This is probably the root of (some of) these problems...

HiFiPhile:
> it seems to me this is not the proper fix, but rather a workaround - with its own side-effects.

Could you detail more, what you mean sie-effects ?

hathach:
my bad, I think we cannot simply clear data here. We should mark endpoint for removal then disable channel first, within the halted interrupt check the flag and clear the endpoint.

HiFiPhile:
Ah yes we should close the channel properly, especially when a hub is used.

HiFiPhile:
I've changed it use `channel_disable()`, I think there is no need to wait until the channel is disabled ?

HiFiPhile:
Copilot sounds reasonable, I need to bring up the memory ...

src/portable/synopsys/dwc2/hcd_dwc2.c
```diff
@@ -456,11 +457,23 @@ tusb_speed_t hcd_port_speed_get(uint8_t rhport) {
...
     hcd_endpoint_t* edpt = &_hcd_data.edpt[i];
     if (edpt->hcchar_bm.enable && edpt->hcchar_bm.dev_addr == dev_addr) {
       tu_memclr(edpt, sizeof(hcd_endpoint_t));
+      for (uint8_t ch_id = 0; ch_id < (uint8_t) DWC2_CHANNEL_COUNT_MAX; ch_id++) {
+        hcd_xfer_t* xfer = &_hcd_data.xfer[ch_id];
+        if (xfer->allocated && xfer->ep_id == i) {
+          dwc2_channel_t* channel = &dwc2->channel[ch_id];
+          dwc2_channel_split_t hcsplt = {.value = channel->hcsplt};
+          xfer->closing = 1;
+          // Channel disable must not be programmed for non-split periodic channels
```
hathach: this is only true for DMA mode, for slave mode, we still need to disable it I guess 

Image

HiFiPhile: I think you are right.

hathach:
@HiFiPhile I make more changes to PR, also add closing flag for endpoint to de-alloc() it when chanel is halted with closing flags. Also put the check into chanel_disable(). Let me know if these make sense to you.

HiFiPhile:
@hathach It looks good but I don't remember some details. Tried plugging in/out a bad contact USB stick on H7RS and no assert is hit.

hathach:
@HiFiPhile why skipping these examples for h7rs, I see it compile just fine with minsize.

HiFiPhile:
@hathach With LOG=2 IAR debug builds needs > 70k for these examples, cdc_msc_hid_freertos needs 66k even on MinRel build.

hathach:
> @hathach With LOG=2 IAR debug builds needs > 70k for these examples, cdc_msc_hid_freertos needs 66k even on MinRel build.

as long as minsize with LOG=0 compile, we can leave it enabled since that is what ci build, also user can test with those example as well. LOG > 1 include lots of constant __file__/__func__/__line__ for assert and increase the footprint qutie a bit.

HiFiPhile:
I've just dropped that commit, maybe later we can enhance skip/only mechanism for debug variant.

hathach (approved):
perfect, thank you very much

PR #3077, opened 2025-04-11
host/dwc2: resume OUT transfer when PING ACKed

HiFiPhile:
**Describe the PR**
Seems the resume is missing.

Before:
usbpv_oSz11aDXqL

After:
usbpv_vEXFvCaiRW


src/portable/synopsys/dwc2/hcd_dwc2.c
```diff
@@ -1000,11 +1000,18 @@ static bool handle_channel_out_slave(dwc2_regs_t* dwc2, uint8_t ch_id, uint32_t
...
-      channel->hcchar |= HCCHAR_CHENA;
+    if (hcsplt.split_en) {
+      if (!hcsplt.split_compl) {
+        // start split is ACK --> do complete split
+        hcsplt.split_compl = 1;
+        channel->hcsplt = hcsplt.value;
+        channel->hcchar |= HCCHAR_CHENA;
+      }
+    } else {
+      // Device is ready, resume transfer
```
hathach, line 1011: we don't need to use channel_xfer_start() since all meta data is setup, also it will toggle PID (causing data toggle error). Just enabled channel with `channel->hcchar |= HCCHAR_CHENA;` is good enough.

hathach (approved):
thank you for your PR. PING is indeed missing, I have
- rename to next_do_ping for clearer meaning
- I found that in case of transfer complete with NYET, we didn't do ping e.g msc out 31 bytes command
- simply use `channel->hcchar |= HCCHAR_CHENA;` to continue transfer with existing meta data. Note channel.hctsiz.do_ping is cleared by host controller when interrupt trigger (in my testing with stm32f769), but we clear it anyway since dwc2 has too many variants :) 

@HiFiPhile please re-test if this work for you. I tested without hub

HiFiPhile:
Thank you, I think it's good on h7rs.

PR #3078, opened 2025-04-11
Add new CDC device callback on driver reset

mbrase:
This adds tud_cdc_reset_cb(), which can be implemented to receive notifications when the CDC driver is reset.

**Additional context**
My application use a writer thread that goes to sleep when the the TX fifo is full, and is woken up by the tud_cdc_tx_complete_cb()  callback. However, if there is a usb reset, then it gets hung since any pending writes are discarded. It would be useful to have a callback so that the writer thread can also wake up when there has been a usb reset.

### DIFF
--- a/src/class/cdc/cdc_device.c
+++ b/src/class/cdc/cdc_device.c
@@ -299,6 +299,7 @@ void cdcd_reset(uint8_t rhport) {
       tu_fifo_clear(&p_cdc->tx_ff);
     }
     tu_fifo_set_overwritable(&p_cdc->tx_ff, _cdcd_cfg.tx_overwritabe_if_not_connected);
+    if (tud_cdc_reset_cb) tud_cdc_reset_cb(i);
   }
 }
 
--- a/src/class/cdc/cdc_device.h
+++ b/src/class/cdc/cdc_device.h
@@ -219,6 +219,9 @@ TU_ATTR_WEAK void tud_cdc_line_coding_cb(uint8_t itf, cdc_line_coding_t const* p
 //                          device will send a break until another SendBreak request is received with value 0000h.
 TU_ATTR_WEAK void tud_cdc_send_break_cb(uint8_t itf, uint16_t duration_ms);
 
+// Invoked when the driver is reset.
+TU_ATTR_WEAK void tud_cdc_reset_cb(uint8_t itf);
+
 //--------------------------------------------------------------------+
 // INTERNAL USBD-CLASS DRIVER API
 //--------------------------------------------------------------------+


HiFiPhile:
You should use `tud_mount_cb()`

mbrase:
Hi, thanks for your response.  I also looked into using `tud_mount_cb()`, but it seemed like there are some cases that that doesn't cover.  For example, if I do a reset in Linux using something like:
```
int fd = open("/dev/bus/usb/001/005", USBDEVFS_RESET);
ioctl(fd, USBDEVFS_RESET, 0);
close(fd);
```
Then the CDC class driver gets reset, but I don't get a notification from either `tud_cdc_tx_complete_cb()` or `tud_mount_cb()`.  Do you have any other suggestions?

HiFiPhile:
Any reason why you are doing USB reset ? A part from some buggy WiFi chips whos get stuck there is no benefit.

You are wrong in `open` permission, should be `fd = open(filename, O_WRONLY);`

If the host does suspend/resume  you need also `tud_resume_cb`/`tud_suspend_cb`

mbrase:
> You are wrong in open permission, should be fd = open(filename, O_WRONLY);

You're right, I copy pasted it incorrectly into my comment.

> Any reason why you are doing USB reset ? A part from some buggy WiFi chips whos get stuck there is no benefit.

I've been trying to debug an issue where my embedded device gets stuck a weird state, and I'm 80% sure it is because the host is doing a usb reset at some point (I've been using the C code above to try to reproduce it).  What I'm seeing is that about once a week, one of my two USB CDC channels on my device gets stuck, where it will no longer send any data form the device to the host (the other CDC channel continues to operate normally).  When I examine the tinyusb state of the stuck channel, I see:
```
tud_connected() == 1
tud_mounted() == 1
tud_cdc_n_ready(1) == 1
tud_cdc_n_connected(1) == 1
tud_cdc_n_get_ilne_state(1) == 0x3
tud_cdc_n_available(1) == 0
tud_cdc_n_write_available(1) == 64
```
What I think is happening is that the thread responsible for writing data to the stuck channel calls `tud_cdc_n_write(...)`, but the fifo is full, so it goes to sleep until `tud_cdc_tx_complete_cb()` is run.  However, before the next transfer can occur, there is a USB reset, which clears the fifos and causes `tud_cdc_tx_complete_cb()` to be skipped. This causes my thread to never wake up.  I've successfully reproduced this scenario using the C reset code above.

To be clear, I haven't confirmed if the host actually resetting my device, and I don't yet have a theory on why the host would be doing a usb reset (I haven't studied the Linux driver too carefully, but it is also possible my powered usb hubs are misbehaving).  However, since I have reproduced the stuck cdc channel with a well timed usb reset, I would like to find a way to harden my firmware to gracefully handle it.  My current ideas are:

1. Use new `tud_cdc_reset_cb()` that I'm proposing in this PR, to wake up my writer thread
2. Experiment with `tud_cdc_line_state_cb()`, since Linux seems set DTR pretty reliably when opening a serial tty device (though I understand some serial clients do not set DTR).
3. Use a low frequency timer to wake up my writer thread periodically, as a backup to `tud_cdc_tx_complete_cb()`, in case we get stuck sometimes.

If you have any additional insight or advice, I would be very grateful.

HiFiPhile:
After a USB reset the host will do the whole enumeration process, when the device is ready (`set configuration` request ok) `tud_mount_cb()` is called, I've also verified it on Arch Linux kernel 6.13 with `USBDEVFS_RESET` . While the reset itself doesn't guarantee anything.

On Linux `tud_cdc_line_state_cb()` is pretty reliable, if I remember correctly DTR is set to ON automatically when the device is opened.

mbrase:
I want to apologize, because I was really sure that my previous testing showed that `tud_mount_cb()` was not being called durring `USBDEVFS_RESET`.  However, I tested it again tonight, and confirmed that `tud_mount_cb()` is indeed being called after reset, just as you said.  I'm not sure what happened the first time (maybe I accidentally implemented `tud_unmount_cb()` instead?).

To close this out, I ended up using `tud_cdc_line_state_cb()` and `tud_mount_cb()`, which I think robustly covers all the edge cases.  Thanks for your patience @HiFiPhile!

PR #3080, opened 2025-04-12
host: fix enumerate racing

HiFiPhile:
**Describe the PR**

#### Move `_dev0.enumerating` into event queue

- if a previous enumeration failed _ctrl_xfer status could stuck, it needs to be cleared before next attempt.
- after _dev0.enumerating is reset in hcd_event_handler(), if an attach event arrived before _ctrl_xfer clean up in remove event, a racing condition will happen.

#### Move decouncing delay before USB reset.

Chirp K for HS detection happens within 7ms of the start of reset, to make speed detection reliable the debouncing delay need to be placed before reset.

image

image

Windows 10 does this:

1. Vbus detected
2. Delay 100ms
3. Bus reset + chirp 50ms
4. Delay 33ms
5. Start setup

### DIFF
--- a/src/host/usbh.c
+++ b/src/host/usbh.c
@@ -566,6 +566,7 @@ void tuh_task_ext(uint32_t timeout_ms, bool in_isr) {
         if ((event.rhport == _dev0.rhport) && (event.connection.hub_addr == _dev0.hub_addr) &&
             (event.connection.hub_port == _dev0.hub_port)) {
           _dev0.enumerating = 0;
+          hcd_device_close(_dev0.rhport, _dev0.hub_addr);
           if (_ctrl_xfer.daddr == 0) {
             _set_control_xfer_stage(CONTROL_STAGE_IDLE);
           }
--- a/src/host/usbh.c
+++ b/src/host/usbh.c
@@ -311,6 +311,12 @@ TU_ATTR_ALWAYS_INLINE static inline usbh_device_t* get_device(uint8_t dev_addr) 
   return &_usbh_devices[dev_addr-1];
 }
 
+TU_ATTR_ALWAYS_INLINE static inline void _set_control_xfer_stage(uint8_t stage) {
+  (void) osal_mutex_lock(_usbh_mutex, OSAL_TIMEOUT_WAIT_FOREVER);
+  _ctrl_xfer.stage = stage;
+  (void) osal_mutex_unlock(_usbh_mutex);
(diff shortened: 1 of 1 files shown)


src/host/usbh.c
```diff
@@ -557,6 +563,14 @@ void tuh_task_ext(uint32_t timeout_ms, bool in_isr) {
         TU_LOG_USBH("[%u:%u:%u] USBH DEVICE REMOVED\r\n", event.rhport, event.connection.hub_addr, event.connection.hub_port);
         process_removing_device(event.rhport, event.connection.hub_addr, event.connection.hub_port);
 
+        if ((event.rhport == _dev0.rhport) && (event.connection.hub_addr == _dev0.hub_addr) &&
+            (event.connection.hub_port == _dev0.hub_port)) {
+          _dev0.enumerating = 0;
+          hcd_device_close(_dev0.rhport, _dev0.hub_addr);
```
hathach, line 569: should include `_dev0.enumerating == 1` since other value has obsolete (for new device) after enumerated successfully. The isr blindly set the flag, but we need to check since we do more than that here. Also `hcd_device_close(_dev0.rhport, 0)` instead of using hub_addr

HiFiPhile: I haven't tested, is `hcd_device_close(_dev0.rhport, 0)` works with device under a hub ?

hathach: It should, the 2nd argument is the device addres that is removed. Otherwise, it will try to remove the hub that attached the rhport.

HiFiPhile: I think there is a loophole with early return. If set address is done `_dev0.enumerating` is still true but `_usbh_devices` also needs to be cleaned.

hathach: ah you are right, putting enumerating in dev0 is confusing. Let me revise and fix this.

hathach (commented):
made an improvement from just a flag `enumerating` to actual `enumerating_daddr` to make it easier to clean up when device unplugged while enumerating not completed yet. @HiFiPhile please let me know if this make sense to you. 

After this merged, I will make the https://github.com/hathach/tinyusb/pull/3075 `ignoring attach/detach` on the same rhport/hub/port while debouncing to usbh to make it generic for all host controller. This seems to work rathe well.

HiFiPhile:
It works pretty well in my side, tried some quick action and no assert happened :)

Issue #3083, opened 2025-04-16
usbh_edpt_control_open() allocates ep0_desc structure on the stack, which gets corrupted outside of the function


joelpmichael:
### Operating System

Linux

### Board

ch32v203c_r0_1v0

### Firmware

examples/host/device_info

### What happened ?

I'm writing USB host support for the CH32V203, and have identified `ep_desc` structures passed in to `hcd_edpt_open()` randomly being modified by observing `ep_desc->wMaxPacketSize` changing to nonsensical values due to the structure being allocated on the stack inside `usbh_edpt_control_open()`:

```
static bool usbh_edpt_control_open(uint8_t dev_addr, uint8_t max_packet_size) {
  TU_LOG_USBH("[%u:%u] Open EP0 with Size = %u\r\n", usbh_get_rhport(dev_addr), dev_addr, max_packet_size);
  tusb_desc_endpoint_t ep0_desc = {
      .bLength = sizeof(tusb_desc_endpoint_t),
      .bDescriptorType = TUSB_DESC_ENDPOINT,
      .bEndpointAddress = 0,
      .bmAttributes = {.xfer = TUSB_XFER_CONTROL},
      .wMaxPacketSize = max_packet_size,
      .bInterval = 0};
  TU_LOG_HEX(3, &ep0_desc);
  TU_LOG_INT(3, ep0_desc.wMaxPacketSize);

  return hcd_edpt_open(usbh_get_rhport(dev_addr), dev_addr, &ep0_desc);
}
```

I've added some extra logging to `usbh.c`, and this is what I'm getting:

```
TinyUSB Device Info Example
USBH init on controller 1, speed = Full
sizeof(usbh_device_t) = 86
sizeof(hcd_event_t) = 12
sizeof(_ctrl_xfer) = 20
sizeof(tuh_xfer_t) = 28
sizeof(tu_fifo_t) = 12
sizeof(tu_edpt_stream_t) = 20
HUB init
[1:] USBH Device Attach
Full Speed
[1:0] Open EP0 with Size = 8
&ep0_desc = 0x20004E88
ep0_desc.wMaxPacketSize = 8
hcd_edpt_open: 498:
rhport=1 dev_addr=0
ep_desc = 0x20004E88
ep_desc->wMaxPacketSize = 8
Get 8 byte of Device Descriptor
[1:0] Get Descriptor: 80 06 00 01 00 00 08 00 
hcd_setup_send: 602:
rhport=1 dev_addr=0
[...]

PR #3084, opened 2025-04-16
note potential issues using ep_desc in hcd_edpt_open()

joelpmichael:
Add a comment in hcd_edpt_open() to make new contributors aware that ep_desc is allocated from the stack when called from usbh_edpt_control_open(), so if any values from ep_desc are required (eg ep_desc->wMaxPacketSize) they must be copied.

See also #3083


HiFiPhile:
Thank you for the note, I think it's more clear with something like `HCD must maintain endpoint and transfer status locally` (not a native English speaker)

HiFiPhile (approved):

hathach:
they must be copied, there is no requirement that is long-term memory, even with normal enumeration which usbh control memory is used temporarily for holding device configuration. But they can be wiped out and reused for other control transfer.

PR #3088, opened 2025-04-18
Fix 1st nak retry one frame shorter.

HiFiPhile:
**Describe the PR**
A small fix following #3072.

When we enable SOF interrupt the 1st time it could be already pending but uncleared, in `hcd_int_handler` `gintsts = dwc2->gintsts & gintmsk`, since `gintmsk` is unset yet `gintsts` is also unset, in result SOF flag is not cleared.

Cause SOF fire immediately and retry interval shorter, 2nd poll should happen at 63ms instead.

usbpv_SpP7rHuANO


hathach (approved):
Thank you

HiFiPhile:
Yeah it's safer to check if sof is already enabled, I think it can only happen if CPU usage is very high which makes latency in processing.

hathach (approved):
perfect, thank you.

PR #3095, opened 2025-04-23
examples cmake: fix gcc flag -mcpu=cortex=m0

fenugrec:
**Describe the PR**
Possibly a copy-paste error ? `-mcpu=` flag should be cortex-m0 on gcc, like with the other compilers


hathach (approved):
thank you

PR #3109, opened 2025-05-02
fix(dcd): Fixed race condition on device disconnect

tore-espressif:
TinyUSB does not provide any locking means to protect the DCD variables. This can lead to race conditions when the user is trying to submit a transfer while the device is being disconnected. This can cause the device to be in an inconsistent state, leading to a crash or undefined behavior.

This commit adds a spin-lock to protect the DCD variables during device disconnect.

* Reported in https://github.com/espressif/esp-idf/issues/9691 
* Also reported in https://github.com/espressif/esp-usb/pull/131

### DIFF
--- a/src/portable/synopsys/dwc2/dcd_dwc2.c
+++ b/src/portable/synopsys/dwc2/dcd_dwc2.c
@@ -40,6 +40,7 @@
 
 #include "device/dcd.h"
 #include "dwc2_common.h"
+#include "dwc2_critical.h"
 
 //--------------------------------------------------------------------+
 // MACRO TYPEDEF CONSTANT ENUM
@@ -52,6 +53,9 @@ typedef struct {
   uint8_t interval;
 } xfer_ctl_t;
 
+/*
+This variable is modified from ISR context, so it must be protected by critical section
+*/
 static xfer_ctl_t xfer_status[DWC2_EP_MAX][2];
 #define XFER_CTL_BASE(_ep, _dir) (&xfer_status[_ep][_dir])
 
@@ -321,6 +325,9 @@ static void edpt_disable(uint8_t rhport, uint8_t ep_addr, bool stall) {
   }
 }
 
+// Since this function returns void, it is not possible to return a boolean success message
+// We must make sure that this function is not called when the EP is disabled
+// Must be called from critical section
 static void edpt_schedule_packets(uint8_t rhport, const uint8_t epnum, const uint8_t dir) {
   dwc2_regs_t* dwc2 = DWC2_REG(rhport);
   xfer_ctl_t* const xfer = XFER_CTL_BASE(epnum, dir);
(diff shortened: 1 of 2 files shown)


hathach (commented):
can we use the osal_mutex instead, there is an osal API for that ?

tore-espressif:
> can we use the osal_mutex instead https://github.com/hathach/tinyusb/blob/master/src/osal/osal_freertos.h#L152, there is an osal API for that ?

I decided for critical section instead of OS primitives because many of these sections are in ISRs
* ISR must not block. So calling `osal_mutex_lock()` is not acceptable
* With critical section we get minimal latency and overhead
* Intention and simplicity: We really just need an atomic access to `xfer_ctl_t`.

hathach:
thank you, that make sense. I think it is about time to add osal_critical_enter/exit() for this purpose. Since I cannot push to your fork, here is how it would be used.  https://github.com/hathach/tinyusb/commit/53a83e1dfa0524d6e7918d43610c24a93202f74b . Let me know if this makes sense to you.

src/portable/synopsys/dwc2/dcd_dwc2.c
```diff
@@ -989,8 +1011,10 @@ void dcd_int_handler(uint8_t rhport) {
 
   if (gintsts & GINTSTS_USBRST) {
     // USBRST is start of reset.
+    DCD_ENTER_CRITICAL();
```
hathach, line 1014: this is only relevant on SMP mcu like P4 right ? since when app enter critical, this isr shoud not be running ?

tore-espressif: I think the same race condition can happen on single core MCUs too.
For example
1. Code from main app is modifying `xfer_ctl_t` and is interrupted
2. Content of `xfer_ctl_t` is now inconsistent
3. The ISR modifies `xfer_ctl_t` again
4. Main app continues

In steps 3 and 4 we can access inconsistent `xfer_ctl_t` which can lead to undefined behavior (division by zero in the cases I've seen)

hathach: but when application modify xfer_ctl_t it already disable irq with enter_critical() the irq shouldn't be invoked. since usb irq is the only irq making change to this xfer ctl, it isn't be preempted by other irq. I guess it would be safe for signle core CPU ? I am trying to see if we could optional omit this for s2/s3 ?

tore-espressif: Yes, you're right. Although S3 is also dual-core, and the problem was reported and reproduced on S3 too.

hathach:
ok I made more changes to https://github.com/hathach/tinyusb/pull/3127 to have osal spinlock (SMP), which I think better name than critical section (single core). Would you mind testing it to see if it fix the race condition. In the future we can move the spinlock to usbd to make sure it works with other ports as well.

hathach (approved):
thank you very much, this PR initiate a new osal spin lock API, which is implemented and merged https://github.com/hathach/tinyusb/pull/3127, which should fix this race issue on esp32 chip. I did a rebase there and github couldn't detect and mark this as merged. So we will need to close this manually.

PR #3111, opened 2025-05-04
dcd/ci_hs: add vbus charge option.

HiFiPhile:
**Describe the PR**
Some devices don't have the VBUS pin connected, `OTGSC_VBUS_CHARGE` need to be set to let USB work.


hathach:
would you mind telling which MCU that requires this fix as well

HiFiPhile:
> would you mind telling which MCU that requires this fix as well

It's LPC4322 on J-Link probe (I'm doing some hacking ;)), USB0_VBUS pin need to be connected to 5V to work.

src/tusb_option.h
```diff
@@ -267,6 +267,15 @@
   #define CFG_TUD_DWC2_DMA_ENABLE CFG_TUD_DWC2_DMA_ENABLE_DEFAULT
 #endif
 
+// Enable CI_HS VBUS Charge
```
hathach, line 270: maybe add the description on why this option is needed e.g when USB_VBUS pin is not connected to 5V

HiFiPhile: Comment just added

hathach (approved):
perfect, thank you

PR #3118, opened 2025-05-09
vendor device open fix for descriptor parsing

ZakDanger:
This fixes 2 issues with the parsing of descriptors inside vendord_open().

1) The code was calcing 'desc_end' as being after the actual end of descriptors due to doing: ```desc_end = tu_desc_next(desc_itf) + max_len``` instead of ```desc_end = desc_itf + max_len```.

2) The code did not handle an interface that finished with non-endpoint descriptors. The code should skip over all descriptors that are part of the interface, but it only skips over endpoint descriptors. So if there are other descriptors after any endpoint descriptors, the code will return a size smaller than it should. This will cause and further processing of the descriptors to fail since it won't point to the next "interface descriptor" in the collection of descriptors.

### DIFF
--- a/src/class/vendor/vendor_device.c
+++ b/src/class/vendor/vendor_device.c
@@ -210,33 +210,20 @@ uint16_t vendord_open(uint8_t rhport, const tusb_desc_interface_t* desc_itf, uin
   TU_VERIFY(p_vendor, 0);
 
   p_vendor->itf_num = desc_itf->bInterfaceNumber;
-  uint8_t found_ep = 0;
-  while (found_ep < desc_itf->bNumEndpoints) {
-    // skip non-endpoint descriptors
-    while ( (TUSB_DESC_ENDPOINT != tu_desc_type(p_desc)) && (p_desc < desc_end) ) {
-      p_desc = tu_desc_next(p_desc);
-    }
-    if (p_desc >= desc_end) {
-      break;
-    }
+  while (TUSB_DESC_INTERFACE != tu_desc_type(p_desc) && (desc_end - p_desc > 0)) {
+    if (TUSB_DESC_ENDPOINT == tu_desc_type(p_desc)) {
+      const tusb_desc_endpoint_t* desc_ep = (const tusb_desc_endpoint_t*) p_desc;
+      TU_ASSERT(usbd_edpt_open(rhport, desc_ep));
 
-    const tusb_desc_endpoint_t* desc_ep = (const tusb_desc_endpoint_t*) p_desc;
-    TU_ASSERT(usbd_edpt_open(rhport, desc_ep));
-    found_ep++;
-
-    if (tu_edpt_dir(desc_ep->bEndpointAddress) == TUSB_DIR_IN) {
-      tu_edpt_stream_open(&p_vendor->tx.stream, desc_ep);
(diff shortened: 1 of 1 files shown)


HiFiPhile:
Hi,

> The code was calcing 'desc_end' as being after the actual end of descriptors due to doing: desc_end = tu_desc_next(desc_itf) + max_len instead of desc_end = desc_itf + max_len.

That's not true. `const uint8_t* desc_end = (uint8_t const*)desc_itf + max_len;` fixed by https://github.com/hathach/tinyusb/pull/3118

HiFiPhile:
Hi, 

> The code was calcing 'desc_end' as being after the actual end of descriptors due to doing: desc_end = tu_desc_next(desc_itf) + max_len instead of desc_end = desc_itf + max_len.

That's not true, `const uint8_t* desc_end = (uint8_t const*)desc_itf + max_len;`, actually fiex by #3020

ZakDanger:
> Hi,
> 
> > The code was calcing 'desc_end' as being after the actual end of descriptors due to doing: desc_end = tu_desc_next(desc_itf) + max_len instead of desc_end = desc_itf + max_len.
> 
> That's not true, `const uint8_t* desc_end = (uint8_t const*)desc_itf + max_len;`, actually fiex by #3020

ah ok - i was working with an older version of the code, so when i copied it into the current version of the code i must have missed that

HiFiPhile:
I think a cleaner fix is simplify the loop, what do you think ?

```c

  while (desc_end - p_desc > 0) {
    if (TUSB_DESC_ENDPOINT == tu_desc_type(p_desc)) {
      const tusb_desc_endpoint_t* desc_ep = (const tusb_desc_endpoint_t*) p_desc;
      TU_ASSERT(usbd_edpt_open(rhport, desc_ep));

      if (tu_edpt_dir(desc_ep->bEndpointAddress) == TUSB_DIR_IN) {
        tu_edpt_stream_open(&p_vendor->tx.stream, desc_ep);
        tud_vendor_n_write_flush((uint8_t)(p_vendor - _vendord_itf));
      } else {
        tu_edpt_stream_open(&p_vendor->rx.stream, desc_ep);
        TU_ASSERT(tu_edpt_stream_read_xfer(rhport, &p_vendor->rx.stream) > 0, 0); // prepare for incoming data
      }
    }

    p_desc = tu_desc_next(p_desc);
  }
```

ZakDanger:
Each call to vendord_open() should "consume" all descriptors until the next interface descriptor, or until the end of all descriptors. So you need to check for whether the "next" descriptor is an interface descriptor, and if it is then you need to return at that point. So a check for desc type being interface needs to be added to your proposed loop.

HiFiPhile:
> Each call to vendord_open() should "consume" all descriptors until the next interface descriptor, or until the end of all descriptors. So you need to check for whether the "next" descriptor is an interface descriptor, and if it is then you need to return at that point. So a check for desc type being interface needs to be added to your proposed loop.

You are right. should be `while (TUSB_DESC_INTERFACE != tu_desc_type(p_desc) && (desc_end - p_desc > 0)) {`

HiFiPhile (approved):

hathach:
the next interface can also start with IAD

HiFiPhile:
> the next interface can also start with IAD

oh yes, let me push a quick fix.

PR #3124, opened 2025-05-16
Add stub support for ESP32P4 board and example

optimus889:
We contributed a simulated ESP32P4 board template to the TinyUSB project, which includes configuration headers, board initialization stubs, a device controller driver structure, and a minimal example under examples/device/. This demonstrates our understanding of the TinyUSB driver integration model and how to extend it for new MCUs.


hathach:
sorry, I don't understand what you are trying to do with this PR.

PR #3127, opened 2025-05-19
add osal spinlock API, Fix/dcd dwc2 race condition 

hathach:
**Describe the PR**
add osal critical API to fix dwc2 race condition updating xfer ctl. Follow up to #3109


src/common/tusb_mcu.h
```diff
@@ -369,6 +369,10 @@
   #define TUP_DCD_ENDPOINT_MAX    7 // only 5 TX FIFO for endpoint IN
   #define CFG_TUSB_OS_INC_PATH_DEFAULT   freertos/
 
+  #if CFG_TUSB_MCU == OPT_MCU_ESP32S3
+    #define TUP_MCU_MULTIPLE_CORE 1
+  #endif
+
```
tore-espressif, line 375: @hathach I'm sorry for the late review, but could we add ESP32-P4 to the list of multiple core MCUs?

You can get the information also dynamically from `SOC_CPU_CORES_NUM` that can be included from `#include "soc/soc_caps.h"`

src/common/tusb_mcu.h
```diff
@@ -381,6 +385,8 @@
 
   #define CFG_TUSB_OS_INC_PATH_DEFAULT   freertos/
 
+  #define TUP_MCU_MULTIPLE_CORE   1
```
hathach, line 388: @tore-espressif P4 is here, thanks for the tip. I will check to see if we could include soc_caps.h here

tore-espressif: ups, I missed that, thanks!

src/common/tusb_mcu.h
hathach: No problems at all. We are all busy. Normally I am not in hurry, but this time I need to use this to fix other issue :)

PR #3140, opened 2025-06-04
Add STM32N657 Nucleo Board support

james-advatek:
**Describe the PR**

This PR adds support for the STM32N6, and BSP for the [STM32N657 Nucleo Board](https://www.st.com/en/evaluation-tools/nucleo-n657x0-q.html).

**Additional context**

- This has been heavily based off @HiFiPhile's STM32H7RSxx port. The TCPP0203 code is directly copied from their port.
- Currently the FreeRTOS examples do not run. I haven't had the time to actually analyse what isn't working, and I probably won't get the time as I will be using Zephyr.
- USB Host has been tested with the `device_info` and `cdc_msc_hid` examples.
- USB Device has been tested with the `cdc_dual_ports` example, and it was verified that it was operating as a USB HS device.

**Disclaimer:** this is my first contribution to TinyUSB, so I apologise if I haven't done things in the correct way!


src/portable/synopsys/dwc2/dwc2_stm32.h
```diff
@@ -79,8 +79,8 @@ extern "C" {
 
 #elif CFG_TUSB_MCU == OPT_MCU_STM32N6
   #include "stm32n6xx.h"
-  #define EP_MAX_FS       6
-  #define EP_FIFO_SIZE_FS 1280
+  #define EP_MAX_FS       9
```
hathach, line 82: N6 has 2 indentiacal HS core with built-in HS PHY (+ typecd), probably the first stm32 mcu with this configuration
image

Issue #3133, opened 2025-05-25
Audio class enhancements


HiFiPhile:
I'd like to make some major changes to UAC2 class:

 - Remove support FIFO

The UAC code is pretty complex and difficult for new comers to contribute. An obvious way to simplify is to remove support FIFO as the only use is channel muxing/demuxing, which can be (should be IMO) done in application layer.
At first it was planned to allow decoding formats other than _TYPE I - PCM_ but never gets implemented.

- Move transfer into ISR with new `driver->xfer_isr`

WIP: https://github.com/HiFiPhile/tinyusb/tree/xfer_isr

ISO transfer has a high timing requirements, currently all transfers are processed in the same queue, even with a preemptive OS or run TUSB inside ISR ISO transfers are still blocked by other transfers, causing audio distortion.

(Mostly for HS) there are some USB hosts that I've tested schedule the transfer poorly, instead of traiting ISO transfer with the highest priority (1st transfer after SOF) the interval between 2 transfers varies between 50us-200us. With a i7-1195G7 notebook I even saw 20us packet interval ! This behavior further reduce the processing time.

Without traiting ISO transfer separately in ISR the only way is differing processing other transfers (what I do currently) which scales poorly.

What do you think ? @PanRe @hathach @kf6gpe @gmikitiuk

PanRe:
Nice to hear from you!
I don't mind - simplicity is better than too many features I guess. I implemented them with idea that users might need them as I did.
Please go ahead!

gmikitiuk:
In general I like the idea, due to strict timing requirements, keeping isochronous in ISR would make sense.

Fortunately the application I've been developing had almost no other transfers in play so putting tinyUSB on highest prio task in the system was enough to keep the audio running smooth. 

What I came across is that with multi-channel(up to 32ch 32bit) probably host is not capable of filling the transfers properly anymore as I've seen discontinuity in audio data without any under/over run on the application side(also no isochronous errors in the peripheral). Unfortunately I do not have usb analyzer to confirm my theory. It's hard to detect such cases when host is not keeping up(unless I've missed something).


PS. removing FIFO would also mean it would be user responsibility to calculate/fill feedback endpoint?

HiFiPhile:
> PS. removing FIFO would also mean it would be user responsibility to calculate/fill feedback endpoint?

[...]

PR #3150, opened 2025-06-14
audio: manage ISO transfer in ISR

HiFiPhile:
**Describe the PR**
2nd part of #3133, ISO transfer management is moved into USB ISR to avoid timing conflict with other classes.

Transfer callbacks become `tud_audio_tx_done_isr`/`tud_audio_rx_done_isr` and not recommended to use, data should be read/write from audio process callback (I2S IRQ) instead.

Meanwhile logging for ISR transfer is not easy to implement:
- Logging from ISR will likely hit issue #2652
- Send complete event to task queue and log later will likely cause event queue overflow when `tusb_task()` is blocked by something since ISR transfer continues

Tested on `NUCLEO-H7S3L8` and `NUCLEO-C071RB` with all audio examples.

### Demo
1. Take `audio_test_freertos` example, add some processing delay to volume change request 
```diff
diff --git a/examples/device/audio_test_freertos/src/main.c b/examples/device/audio_test_freertos/src/main.c
index 0a6cdfd4f..67d56b640 100644
--- a/examples/device/audio_test_freertos/src/main.c
+++ b/examples/device/audio_test_freertos/src/main.c
@@ -299,6 +299,7 @@ bool tud_audio_set_req_entity_cb(uint8_t rhport, tusb_control_request_t const *p
         TU_VERIFY(p_request->wLength == sizeof(audio_control_cur_2_t));

         volume[channelNum] = (uint16_t) ((audio_control_cur_2_t *) pBuff)->bCur;
+        vTaskDelay(10); // Simulate some processing delay
         TU_LOG1("    Set Volume: %d dB of channel: %u\r\n", volume[channelNum], channelNum);
         return true;
```
2. Adjust volume on PC while recording with `plot_audio_samples.py` script (increase recording duration if your hand is slow ;))
3. The recorded data should be distortion-free
image


hathach (approved):
thank you, I will trust you on the audio driver. I did mostly code format, add busy/claim status revert if xfer_isr() defer handler to xfer_cb() to prevent race condition, where app will claim another xfer before cb() is called by tud_task().

PR #3151, opened 2025-06-15
osal/none: add nested count to spin lock

HiFiPhile:
**Describe the PR**
To allow it be used in nested functions, I think most OSs already have the counter.

### DIFF
--- a/src/osal/osal_none.h
+++ b/src/osal/osal_none.h
@@ -45,11 +45,12 @@ TU_ATTR_WEAK void osal_task_delay(uint32_t msec);
 //--------------------------------------------------------------------+
 typedef struct {
   void (* interrupt_set)(bool);
+  uint32_t nested_count;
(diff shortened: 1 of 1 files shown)


src/osal/osal_none.h
```diff
@@ -34,26 +34,46 @@ extern "C" {
...
+  if (!in_isr && ctx->nested_count == 0) {
     ctx->interrupt_set(false);
   }
+  ctx->nested_count++;
 }
 
 TU_ATTR_ALWAYS_INLINE static inline void osal_spin_unlock(osal_spinlock_t *ctx, bool in_isr) {
-  if (!in_isr) {
+  // Check for underflow - unlock without lock
+  if (ctx->nested_count == 0) {
```
hathach, line 64: let make it simpler, in case of 0, we just return since the spin is not locked to begin with

PR #3152, opened 2025-06-25
stm32_fsdev: Allow configuring single-buffered isochronous endpoints.

Milek7:
Controlled by the embedder defining CFG_TUD_FSDEV_DOUBLE_BUFFERED_ISO_EP in tusb_config.h. If hardware does have SBUF_ISO bit it will use only one half of endpoint pair register.

Tested on G0.


HiFiPhile:
Thank you for the PR, tested working on L0 and C0.

PR #3163, opened 2025-07-07
support at32 mcu

ning3270455369:
support at32 mcu


ning3270455369:
I have added support for the AT32 chip and currently it is compatible with seven models.

hathach (commented):
awesome, thank you for making PR for full at32 support. There are quite a bit of changes needed to get pass ci build as well as adding cmake build support, however I couldn't push to your fork's branch to update the PR. I have make my own PR based on this PR here https://github.com/hathach/tinyusb/pull/3191 . 

PS: I am only able to test and confirm tinyusb work with an f403a (purple), however, the f423 (black) board does not work/start at all. Though it is probably clock/hw issue I guess. The rest of the family, I don't have any boards to test with. We will merge this now, and I will try to find other boards to test with later.

image

PR #3165, opened 2025-07-08
Fix HID descriptor parsing of 4‑byte items

hathach:
## Summary
- fix HID host parser: handle item size value `3` as 4 bytes
- avoid reading descriptor data when no bytes present

------
https://chatgpt.com/codex/tasks/task_e_6868baec4d148331bb2aa023524eb1b3


hathach (commented):
good fix by codex

PR #2904, opened 2024-12-11
feature(tusb): Added tusb_teardown()

roma-jam:
## Requirements
To change the USB device function or re-initialize the tinyusb stack there should be an option to teardown the lower layer (disable the phy and flush the registers).

The public API for init tinyusb stack for USB Device is `tusb_init()` for versions < v0.17,  and `tusb_rhport_init()` for verisons >= 0.17. 

Teardown feature should be implemented as a public API for Device and Host role and provide the possibility to use the teardown call in both (legacy and new API way)

## Description

The same mechanism for legacy call support is used as for `tusb_init()` (providing the `rhport` as `TUD_OPT_RHPORT` or `TUH_OPT_RHPORT` but not the functiton argument)
 
## Limitations
- Using macros to handle the legacy support instead of function (refer to Note2: https://github.com/hathach/tinyusb/blob/7c7b30f0ae66a856ada3526cb947bd3a255567f8/src/tusb.h#L140)

PR #3168, opened 2025-07-10
Feature/add tusb teardown

hathach:
**Describe the PR**
update for https://github.com/hathach/tinyusb/pull/2904, since I cannot make changes to @roma-jam fork. tusb_deinit() for matching with tud/tuh_deinit(). Since it is new API, we don't have to cover the backward compatible case.

### DIFF
--- a/src/tusb.h
+++ b/src/tusb.h
@@ -140,7 +140,7 @@ bool tusb_rhport_init(uint8_t rhport, const tusb_rhport_init_t* rh_init);
 
 // Initialize roothub port with device/host role
 // Note: when using with RTOS, this should be called after scheduler/kernel is started.
-// Otherwise, it could cause kernel issue since USB IRQ handler does use RTOS queue API.
+//       Since USB IRQ handler does use RTOS queue API.
 // Note2: defined as macro for backward compatible with tusb_init(void), can be changed to function in the future.
 #if defined(TUD_OPT_RHPORT) || defined(TUH_OPT_RHPORT)
   #define _tusb_init_arg0()        tusb_rhport_init(0, NULL)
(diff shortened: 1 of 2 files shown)


roma-jam:
Hi @hathach, 

the original idea was to allow the possibility to use pre-defined `TUD_OPT_RHPORT` or `TUH_OPT_RHPORT` to teardown the tinyusb and to call the API without any arguments: 

```
tusb_init();
tusb_teardown();
```

After your changes it is no longer possible, so the idea is to call deinit for the device like this: 

```
tusb_init();
tud_deinit(TUD_OPT_RHPORT);
```

am I right?

UPD: The direct access to my fork is in pending state for you )

hathach:
tusb_init() should be called with rhport and arguments. Since i made the tusb_init(void) previously, we need to keep it backward-compatible to prevent breaking app that simply update tinyusb and don't want to re-write code. For new API, user should specify which port to deinit().

roma-jam:
@hathach ,

> tusb_init() should be called with rhport and arguments.

I think that the macros `tusb_init()` is still available for usage: https://github.com/hathach/tinyusb/blob/master/src/tusb.h#L153

So it is possible to use it to init the tinyusb, which leads to the default call: `tusb_rhport_init(0, NULL)`

So, either we need to deprecate the macros completely (which would break the backward compatibility) or to provide another macros for `tusb_deinit()` which when called without arguments, leads to the `tusb_rhport_deinit(0)`

WDYT?

src/tusb.h
```diff
@@ -158,14 +158,15 @@ bool tusb_inited(void);
 // Called to handle usb interrupt/event. tusb_init(rhport, role) must be called before
 void tusb_int_handler(uint8_t rhport, bool in_isr);
 
-// TODO
-// bool tusb_teardown(void);
+// Deinit usb stack on roothub port
+bool tusb_deinit(uint8_t rhport);
```
roma-jam, line 162: Something like that: 

```suggestion
bool tusb_rhport_deinit(uint8_t rhport);

#if defined(TUD_OPT_RHPORT) || defined(TUH_OPT_RHPORT)
  #define _tusb_deinit_arg0()        tusb_rhport_deinit(0)
#else
  #define _tusb_deinit_arg0()        TU_VERIFY_STATIC(false, "CFG_TUSB_RHPORT0_MODE/CFG_TUSB_RHPORT1_MODE must be defined")
#endif

#define _tusb_deinit_arg1(_rhport)         tusb_rhport_deinit(_rhport)
#define tusb_deinit(...)                   TU_FUNC_OPTIONAL_ARG(_tusb_deinit_arg1, __VA_ARGS__)
```

hathach: at some point in the future release, we will deprecate the usage of tusb_init(void). For now, we will just leave it as it is. The new API should take new signature. New code should call tusb_deinit(TUD_OPT_RHPORT) if it currently use CFG_TUSB_RHPORT0/1_MODE for setting device/host stack

hathach:
@roma-jam let me know if that sounds ok, and we are good to merge ?

roma-jam:
@hathach  
We're good.

hathach:
> @hathach We're good.

thanks

PR #3169, opened 2025-07-10
add weak tud_msc_scsi_pre_cb, to allow overriding any MSC SCSI commands

pekkanikander:
**Describe the PR**

In some cases, like when implementing a truly read-only MSC disk, the TinyUSB's built-in MSC SCSI commands may need to be overridden.  

This is a small, fully backwards-compatible change that adds a new weak callback, `tud_msc_scsi_pre_cb`.  If defined, this callback allows the user code to override also those SCSI commands that TinyUSB handles by default in its static `proc_builtin_scsi` function.

**Additional context**

An example how this new callback is needed can be found here:
https://github.com/pekkanikander/pico-extras-usb-msc-bootrom-partitions/blob/v0.3.1/src/vd_virtual_disk.c#L504

There we make a virtual USB disk, implemented on RP2350 (Raspberry Pico 2), as truly read-only.  The main motivation for this is that when removing such a read-only USB disk, macOS does not complain, as it does if a read-write USB disk is removed.  

An alternative design would have been to simply flip the order of the calls to `proc_builtin_scsi` and to `tud_msc_scsi_cb`. However, as there appear to be legacy `tud_msc_scsi_cb` implementations that return zero even when they don't recognise the SCSI command, that would have been likely to break backwards compatibility.  The present way of implementing a new weak callback is fully backwards compatible, with the cost of a few additional bytes for checking if the callback is non-null and actually calling it.

### DIFF
--- a/src/class/msc/msc_device.c
+++ b/src/class/msc/msc_device.c
@@ -479,7 +479,7 @@ bool mscd_xfer_cb(uint8_t rhport, uint8_t ep_addr, xfer_result_t event, uint32_t
             TU_ASSERT(usbd_edpt_xfer(rhport, p_msc->ep_out, _mscd_epbuf.buf, (uint16_t) p_msc->total_len));
           }
         } else {
-          int32_t resplen = -1;
+          int32_t resplen = -TUD_MSC_RET_CALL_DEFAULT;
           // Call the pre-callback to allow application to override any SCSI command
           if (tud_msc_scsi_pre_cb) {
             resplen = tud_msc_scsi_pre_cb(p_cbw->lun, p_cbw->command, _mscd_epbuf.buf, (uint16_t)p_msc->total_len);


pekkanikander:
Oops.  It appears that we will need slightly richer semantics.  I'll try to provide a new version of this PR still today, covering also the cases where this callback needs to trigger an error and still bypass the default implementation.

pekkanikander:
I pushed another commit here, now changing the sematics slightly.  This is needed to make a difference between two cases: the pre-callback wants to fall back to the default behaviour vs. the pre-callback wants to return an error. 

This may not be the best way of implementing this, though.  So, I presume some careful thought is needed. 

And, if this PR is accepted as such, the two commits should probably be squashed together.

pekkanikander:
Hmm.  Maybe this all should be implemented in a different way.  I'm pulling back this, for now.

PR #3175, opened 2025-07-16
Fix compilation when CH34X support is disabled

elipsitz:
**Describe the PR**
When `CFG_TUH_CDC_CH34X` is not true, `SERIAL_DRIVER_CH34X` is not defined and compilation fails. This fixes the issue by making the check conditional on `CFG_TUH_CDC_CH34X`.

**Additional context**

### DIFF
--- a/src/class/cdc/cdc_host.c
+++ b/src/class/cdc/cdc_host.c
@@ -833,8 +833,11 @@ static bool set_line_state_on_enum(cdch_interface_t *p_cdc, tuh_xfer_t *xfer) {
   switch (state) {
     case ENUM_SET_LINE_CODING: {
       #ifdef CFG_TUH_CDC_LINE_CODING_ON_ENUM
+      #if CFG_TUH_CDC_CH34X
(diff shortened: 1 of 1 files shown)


src/class/cdc/cdc_host.c
```diff
@@ -833,8 +833,14 @@ static bool set_line_state_on_enum(cdch_interface_t *p_cdc, tuh_xfer_t *xfer) {
   switch (state) {
     case ENUM_SET_LINE_CODING: {
       #ifdef CFG_TUH_CDC_LINE_CODING_ON_ENUM
+
+      #if CFG_TUH_CDC_CH34X
```
hathach, line 836: can you chagne it to  

```
#if CFG_TUH_CDC_CH34X
if (p_cdc->serial_drid != SERIAL_DRIVER_CH34X)
#endif
{
```

elipsitz: Oh yeah, that's definitely better. Done

hathach (approved):
thank you

PR #3179, opened 2025-07-18
Report CH32V305 as working

rhgndf:
Flashing the ch32v307 image onto CH32V305 works. Tested with USBHS

Not all examples are tested


hathach:
do you mind addig the bsp for ch32v305 for testing ?

PR #3186, opened 2025-07-24
Add nanoch32v305

rhgndf:
**Describe the PR**
For #3179

**Additional context**
https://github.com/wuxx/nanoCH32V305


hathach (approved):
perfect, thank you. I update the linker and give board.mk/cmake to define flash/ram size since v305 has less flash/ram than v307

PR #3187, opened 2025-07-25
dhcp: Fix DHCP_OFFER/DHCP_ACK destinaton.

s09289728096:
**Describe the PR**
In RFC 2131, the destination of DHCP OFFER/ACK is defined in Section 4.1. Fix the destination error by following the rule of RFC 2131.
TODO: We implement all rule but the last one. ARP table is required to associate client's macaddr. Currently, fallback to broadcast.


hathach (approved):
perfect, thank you very much for the PR

PR #3189, opened 2025-07-27
Adding support for CH58x chip

ImproperCatGirl:
**Describe the PR**
Adding support for CH58x chips, currently tested on CH585.
**Additional context**

A new USB device driver has been introduced for CH58x, rather than reusing the ch32_usbfs driver.
While the core USB controller logic remains mostly similar to CH32, several key differences prevented a easy and clean reuse of the ch32_usbfs driver:
        1.Register address layout and bitfield definitions differ significantly.
        2.Endpoint control logic has been revised, now RX CTRL and TX CTRL are configured in a single register.
        3.WCH changed their SDK header structure: from memory-mapped structs to hardcoded absolute addresses, with meaningful comments for bit positions. These headers are used directly for clarity.
        4.Unlike the CH32 series (used in ch32_usbfs), where endpoint registers follow a predictable and linear memory pattern, CH58x register addresses are not linear and without uniform spacing.

BSP support is a TODO partially due to warnings in the SDK (under strict compiler flags, warnings are treated as errors).

Datasheet: https://www.wch-ic.com/downloads/CH585DS1_PDF.html


ImproperCatGirl:
Now minimal BSP are added. Note that this MCU have a quirk, the 4th endpoint does not support DMA and auto toggle, so special logic is needed, which is absent now. So examples that require >3 endpoints will fail, like cdc_msc.

ZimM-LostPolygon:
@ImproperCatGirl Thank you for working on this! I'm eyeing CH585 and TinyUSB support would be amazing.  Do you plan on ever finishing this?

PR #3210, opened 2025-08-09
Fix STM32L4 GPIOD clock enable for variants without GPIOD

martijnvdwoude:
Fix STM32L4 GPIOD clock enable for variants without GPIOD like the STM32L432.


hathach (approved):
thanks

PR #3225, opened 2025-08-24
Improve compatibility in CH32V20x USBFS Host

verylowfreq:
This PR solves the compatibility issues with USB device/class in WCH CH32V20x USBFS Host function.

**Patch 522a34ee28bcee6674ccd91973cb87869be98909**
Inserting a small delay for LowSpeed devices before executing transfer. Communication problems were occurred in some LowSpeed devices behind USB Hub.

**Patch a3bb7a90b3478f43d448443140b68c3473331f93**
Improve retry strategy on NAK response. Without this patch, MassStorage class device works but CDC class device doesn't. In this patch, retry is queued through hcd event mechanism.

**Patch bd9dd75da8718dfd56b2e91b658f81613ed1bf3a**
This is not essential changes on USB behavior. PANIC() replaced by TU_ASSERT().

Thank you for reviewing!

### DIFF
--- a/src/portable/wch/hcd_ch32_usbfs.c
+++ b/src/portable/wch/hcd_ch32_usbfs.c
@@ -28,6 +28,8 @@
 
 #if CFG_TUH_ENABLED && defined(TUP_USBIP_WCH_USBFS) && defined(CFG_TUH_WCH_USBIP_USBFS) && CFG_TUH_WCH_USBIP_USBFS
 
+#include <stdlib.h>
+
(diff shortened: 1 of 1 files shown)


verylowfreq:
Sorry but in Patch a3bb7a90b3478f43d448443140b68c3473331f93, malloc() and free() are used.

In Patch 6f2b5fc4952e0fc730466a8756ce6427c33ce2b7, I removed them and use static allocated struct.

PR #3226, opened 2025-08-24
STM32N6570-DK support is added

armusin:
The STM32N6570-DK board specific files are added, and the "Supported Boards" list is updated.

"examples/device/cdc_dual_ports" example was running on the board successfully with the limitation described below.

There is a feature/bug:
The example should be built with DEBUG=1 and without logs strongly (make BOARD=stm32n6570dk DEBUG=1 all), otherwise it does not work.

make BOARD=stm32n6570dk DEBUG=1 LOG=2 all
or 
make BOARD=stm32n6570dk LOG=2 all
leads to the following log and the board is not enumerated properly.
```
USBD init on controller 1, speed = Auto
sizeof(usbd_device_t) = 56
sizeof(dcd_event_t) = 12
sizeof(tu_fifo_t) = 12
sizeof(tu_edpt_stream_t) = 20
CDC init
0x00005000, 0x4F54411A, 0x00000000, 0x228FE052, 0x03B882E8, 0xE2103E30
Highspeed UTMI+ PHY init
DMA = 0
```
Any advice on the issue would be appreciated.


armusin:
Hi @hathach ,

The HIL tests are failed for the STM32H7 MCU, its code was untouched in the PR. 
Could you please clarify what should be done on my side to proceed with the PR?

hathach:
- LOG with uart can mess up timing, try to log with segger RTT to see if that improves
- HIL pool include lots of hardware, when they run simulateneously, usb can failed (timeout), it is ususally fixed by re-run the ci. Let me know when you think this PR is ready for review.

PR #3228, opened 2025-08-28
fix(dcd/dwc2): Fix reset procedure for versions >=4.20a

tore-espressif:
**Describe the PR**
Fix soft reset procedure for DWC2 version >= 4.20a.

I suspect this code was never tested since none of the supported targets use version >=4.20a.
After `GRSTCTL_CSRST` is set, read from `dwc2->gsnpsid` always returns zero, so it must be read first and reset second.

**Additional context**
ESP32-P4 silicon revision 3 updates DWC2 to version 4.30a

### DIFF
--- a/src/portable/synopsys/dwc2/dwc2_common.c
+++ b/src/portable/synopsys/dwc2/dwc2_common.c
@@ -1,7 +1,7 @@
 /*
  * The MIT License (MIT)
  *
- * Copyright (c) 2024 Ha Thach (tinyusb.org)
+ * Copyright (c) 2024-2025 Ha Thach (tinyusb.org)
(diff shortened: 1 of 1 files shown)


tore-espressif:
@hathach @HiFiPhile Pretty simple but important fix :)

tannewt (approved):
Thank you!

tannewt:
> ESP32-P4 silicon revision 3 updates DWC2 to version 4.30a

Is this chip id 1.3?

tore-espressif:
> Is this chip id 1.3?

This engineering sample is version 3.0.

Thanks for the quick merging!

PR #3236, opened 2025-09-04
modified the bsp files of at32 to make them work better

ning3270455369:
I modified the BSP files of at32 to make the at32 chip, especially the model with two OTG modules, work better. In addition, I fixed some errors.


hathach:
@copilot fix this pr pre-commit run

Issue #2687, opened 2024-06-22
OHCI: MMIO coherency/synch issue


Ryzee119:
### Operating System

Windows 11

### Board

OHCI

### Firmware

class/hid_host/c
portable/ohci/ohci.c

### What happened ?

After random lengths of time OHCI transfers will not fire the completion interrupt. I traced this to garbage being in hcca.done_head.

The garbage was similar to the address of the TD but was partically corrupt or missing. The root cause was writing to td_head directly here  <https://github.com/hathach/tinyusb/blob/044f4d1801a24634a519b73dec6f2e1dc326169d/src/portable/ohci/ohci.c#L443>

I can see exactly  why we are doing this; which to prevent having to allocate dummy TDs for every ED, but the OHCI spec implies you cannot do this.

Section 4.6 of OHCI has this; I have a feeling I am getting the coherency/sync issue described here.
```
Software queues to the list by using the value of TailP to obtain the physical address of the last
TD queued to the ED. Since the TD pointed to by TailP is not accessed by the HC, the Host
Controller Driver can initialize that TD and link at least one other to it without creating a
coherency or synchronization problem.
Software may not alter in any way any of the
TDs it has queued prior to the one pointed to by TailP until the Host Controller completes
processing of the TD or the Host Controller Driver ensures that queue processing for the ED has
been halted.
```

I have found two fixes below:

1. Stop ED while we modify it to queue the transfer, but this is not an ideal solution waiting 1ms every transaction; although maybe not too bad and a sacrifice that can be made to _Tiny_ USB to keep it small. This would be okay for interrupt and bulk ultimately and isochronous can be handled differently (I am currently adding it)
```diff
+p_ed->skip = 1;
+osal_task_delay(1); // Wait one frame to ensure OHCI is not processing ED
p_ed->td_head.address |= (uint32_t) _phys_addr(p_gtd); // Write it
+p_ed->skip = 0;
```

2. The 'correct' way and I can confirm it resolved the issue was by creating a dummy TD  that is linked to the ED head and tail on hcd_open. Then in hcd_edpt_xfer I populated this TD, allocate a new dummy then update the ED tail to pointer to the new TD which triggers the transfer. This is exactly what the OHCI recomends and the issue no longer occurs. The big problem is ofcourse this basically doubles the memory used by TDs.

![image](https://github.com/hathach/tinyusb/assets/21236406/026d9ac9-ccc2-4764-be46-aa1962965add)

[...]

Issue #3208, opened 2025-08-08
OHCI host driver support for cache operations


ArcaneNibble:
### Related area

ohci, cache

### Hardware specification

ohci, TI AM1808

### Is your feature request related to a problem?

Hi, I am currently contributing to the [Pybricks](https://github.com/pybricks/) project to port MicroPython to the Lego Mindstorms EV3. This is built around a TI Sitara AM1808 SoC. We run our own bare-metal code without Linux on this chip.

As part of this effort, we need a USB *host* stack for the OHCI controller, and we were hoping to use TinyUSB. However, the existing OHCI driver does not handle the requirements to clean/invalidate CPU caches. Caches are _very_ important for performance on a chip such as this because it uses external SDRAM.

### Describe the solution you'd like

I currently have a local patchset which appears to function correctly with cache support. The overall approach I've chosen involved:
* Aligning each TD to a cache line. This unfortunately uses up twice the memory.
* Refactoring such that the CPU never tries to read TDs it doesn't "own" (which required removing the data stashed in reserved bits of word0 and moving them elsewhere)
* Performing correct cache operations on user buffers and TDs
* Accessing EDs exclusively through a memory mirror which is set up in the MMU to be an *uncached* view of system RAM. This is a functionality that is specific to our platform and that TinyUSB currently doesn't expect platforms to generally have available.

Before I open a PR, I wanted to see if such an approach would be acceptable in TinyUSB. If you want to actually take a look at the code in question, you can find it [here](https://github.com/ArcaneNibble/tinyusb/tree/pybricks)

### I have checked existing issues, dicussion and documentation

- [x] I confirm I have checked existing issues, dicussion and documentation.

hathach:
yeah sure, you can take the peek of EHCI driver, it already supports and use dcache clean/invalidae API. Cache line size is cpu depent, you gonna need to add `CFG_TUSB_MEM_DCACHE_LINE_SIZE_DEFAULT` to tusb_mcu.h accordingly.

HiFiPhile:
Implemented by #3248

PR #3248, opened 2025-09-12
ohci: Implement support for d-cache operations

ArcaneNibble:
This PR implements manual d-cache support for the OHCI host controller driver, using the approach described in https://github.com/hathach/tinyusb/issues/3208

This has been tested with only the USB HID driver on the TI AM1808


src/portable/ohci/ohci.c
```diff
@@ -143,6 +143,15 @@ enum {
   PID_FROM_TD = 0,
 };
 
+//--------------------------------------------------------------------+
+// Support for explicit D-cache operations
+//--------------------------------------------------------------------+
+TU_ATTR_WEAK bool hcd_dcache_clean(void const* addr, uint32_t data_size) { (void) addr; (void) data_size; return true; }
+TU_ATTR_WEAK bool hcd_dcache_invalidate(void const* addr, uint32_t data_size) { (void) addr; (void) data_size; return true; }
+#ifndef hcd_dcache_uncached
+#define hcd_dcache_uncached(x) (x)
```
hathach, line 157: what this function does specifically ?

src/portable/ohci/ohci.h
```diff
@@ -92,44 +115,52 @@ typedef struct TU_ATTR_ALIGNED(16)
 	uint8_t* buffer_end;
 } ohci_gtd_t;
 
-TU_VERIFY_STATIC( sizeof(ohci_gtd_t) == 16, "size is not correct" );
+TU_VERIFY_STATIC( sizeof(ohci_gtd_t) == CFG_TUH_MEM_DCACHE_ENABLE ? CFG_TUH_MEM_DCACHE_LINE_SIZE : 16, "size is not correct" );
+
+typedef union {
```
hathach, line 117: why would we need to have these as separated typedef ?

src/portable/ohci/ohci.h
```diff
@@ -152,11 +183,15 @@ typedef struct TU_ATTR_ALIGNED(32)
 	volatile uint16_t offset_packetstatus[8];
 } ochi_itd_t;
 
-TU_VERIFY_STATIC( sizeof(ochi_itd_t) == 32, "size is not correct" );
+TU_VERIFY_STATIC( sizeof(ochi_itd_t) == CFG_TUH_MEM_DCACHE_ENABLE ? CFG_TUH_MEM_DCACHE_LINE_SIZE : 32, "size is not correct" );
 
 typedef struct {
   uint16_t expected_bytes; // up to 8192 bytes so max is 13 bits
+  uint8_t dev_addr  : 7;
```
hathach: why we need this here ?

src/portable/ohci/ohci.c
ArcaneNibble: In my application, I define it like this:

```c
#define hcd_dcache_uncached(x) (*(volatile __typeof__(x) *)((uintptr_t)(&(x)) + MAGIC_OFFSET))
```

In words, it takes the address of `x`, adds a specific offset to it, and then accesses this new address (the address with the offset added) with the same type as the original `x`.

This is why it is a macro and not a function -- the return type is generic and always matches the type of `x`.

The reason this is able to work is because of my application's platform setup, which is not shown anywhere in the TinyUSB code. The SoC I am using has a MMU. As part of setting up the MMU, I configure:

| Virtual Address | Physical Address | Attributes |
| --------------- | ----------------- | ---------- |
| 0xCxxxxxxx      | 0xCxxxxxxx          | Cacheable |
| 0xDxxxxxxx      | 0xCxxxxxxx          | Not cacheable |

System RAM is at physical address 0xCxxxxxxx, and the system RAM can be accessed from two different virtual address ranges. Most data (including OHCI TDs) is accessed using the cacheable range at VA 0xCxxxxxxx. In order to pass data between the CPU and the OHCI controller, manual cache operations are used.

However, OHCI EDs are accessed using the uncacheable range at VA 0xDxxxxxxx. This accesses the same physical memory, but it bypasses the cache. This means that I do not have to use manual cache operations on EDs. I can access individual words inside an ED without worrying that the other words in the same ED will get cached.

src/portable/ohci/ohci.h
ArcaneNibble: These separate typedefs are used because I want to be able to explicitly and manually read and write each word in the ED. This is because uncached memory access is slow.

src/portable/ohci/ohci.h
ArcaneNibble: I wanted to avoid having to access an ED in `done_queue_isr`, because I am in general trying to access an ED as infrequently as possible, and I noticed that there were two reasons to access an ED there:

1. To apply a workaround for a STALL situation
2. To obtain the `dev_addr` and `ep_addr` for calling `hcd_event_xfer_complete`

For reason 1, I believe the STALL workaround is not necessary if EDs have an empty TD like the OHCI specification suggests.

After making that change to TD allocation, I noticed that reason 2 can be eliminated by moving those fields out of the ED to somewhere else.

As an alternative, the `dev_addr` and `ep_addr` fields can be packed into unused bits in the TD word 0. However, this does not work for isochronous TDs (not enough free bits).

Using TD word 0 to store those fields and then packing `used` alongside `expected_bytes` can save some memory. Do you want me to use this approach instead?

src/portable/ohci/ohci.c
```diff
@@ -326,27 +337,30 @@ static void ed_init(ohci_ed_t *p_ed, uint8_t dev_addr, uint16_t ep_size, uint8_t
...
+  w0.is_iso            = (xfer_type == TUSB_XFER_ISOCHRONOUS) ? 1 : 0;
+  w0.max_packet_size   = ep_size;
+
+  w0.used              = 1;
+  w0.is_interrupt_xfer = (xfer_type == TUSB_XFER_INTERRUPT ? 1 : 0);
+  hcd_dcache_uncached(p_ed->w0) = w0;
 }
 
 static void gtd_init(ohci_gtd_t *p_td, uint8_t *data_ptr, uint16_t total_bytes) {
   tu_memclr(p_td, sizeof(ohci_gtd_t));
```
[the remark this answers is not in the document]
ArcaneNibble: TD initialization is not yet complete when `gtd_init` returns. Additional fields still need to be written.

A cache clean operation is performed manually in all of the callers after the additional fields are written.

src/portable/ohci/ohci.c
```diff
@@ -641,7 +672,7 @@ static void done_queue_isr(uint8_t hostid)
   (void) hostid;
 
   // done head is written in reversed order of completion --> need to reverse the done queue first
```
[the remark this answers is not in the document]
ArcaneNibble: This suggestion is incorrect, but there was indeed a bug here.

src/portable/ohci/ohci.c
hathach: shouldn't we can use the following virt to phys to do the conversion. You can implement it in the app in the way to check the range to do the offset correctly ? That would make thing consistent 

```C 
TU_ATTR_WEAK extern void* tusb_app_virt_to_phys(void *virt_addr);
TU_ATTR_WEAK extern void* tusb_app_phys_to_virt(void *phys_addr);
```

ArcaneNibble: Those functions have slightly different semantics, especially since TDs do not go through the alternate uncached memory address. TDs use the normal memory mapping and use explicit cache operations.

It could be possible to use an approach like the following:
* Modify these functions to take a parameter specifying the type of object being translated (TD, ED, HCCA)
* Add compiler directives to place uncached data in a different linker section (placing it at 0xDxxxxxxx)

Would this be a better approach?

ArcaneNibble: Actually, I just realized that this should be possible to implement without adding a type parameter.

Instead, the reason I did not originally implement it this way was because I would have to split up `ohci_data_t` into separate cacheable and uncacheable parts.

src/portable/ohci/ohci.h
hathach: can you use following typedef to keep the code least modified
```C
  // Word 0
  union {
    struct {
        uint32_t dev_addr          : 7;
        uint32_t ep_number         : 4;
        uint32_t pid               : 2;
        uint32_t speed             : 1;
        uint32_t skip              : 1;
        uint32_t is_iso            : 1;
        uint32_t max_packet_size   : 11;
              // HCD: make use of 5 reserved bits
        uint32_t used              : 1;
        uint32_t is_interrupt_xfer : 1;
        uint32_t is_stalled        : 1;
        uint32_t                   : 2;
    };
    uint32_t word0;
  }
```

src/portable/ohci/ohci.h
hathach: can you quote the section where OHCI specs suggest that. I wrote this long time ago and forgot most of the thing.

src/portable/ohci/ohci.c
hathach: I got it, but `hcd_dcache_uncached()` is too specific to your set-up. It makes the code not as readible as I would like. Can you centralize the uncached in the get_ed() function only ?

hathach (commented):
check below feedback

ArcaneNibble:
Are you referring to allocating an empty TD in the OHCI specification? That's shown in Figure 5-6:

Screen Shot 2025-09-16 at 5 28 35 AM

Screen Shot 2025-09-16 at 5 29 08 AM

The reason for doing this is so that you don't have to reach back and modify the last TD (which might be in use by the OHCI hardware already) when inserting at the end.

Ryzee119:
> Are you referring to allocating an empty TD in the OHCI specification? That's shown in Figure 5-6:

This will likely fix https://github.com/hathach/tinyusb/issues/2687 too

hathach:
@ArcaneNibble @Ryzee119 yeah, I agree we should follow the specs suggestion. I was trying to save SRAM footprint, but it cause more troubles than expected. I am in middle of other works, will review this asap, please be patient meanwhile

hathach (commented):
@ArcaneNibble  can the hcd_dcache_uncached() is used only once in the get_ed() instead of doing it everytime we access its field member ?

src/portable/ohci/ohci.h
hathach: I don't think accessing ed in the done queue is much of an issue, especially this is only specific to your setup and this increased the footprint 2byte per TD.

hathach:
@codex review

src/portable/ohci/ohci.c
```diff
@@ -590,141 +596,112 @@ static ohci_td_item_t* list_reverse(ohci_td_item_t* td_head)
...
-static void done_queue_isr(uint8_t hostid)
-{
-  (void) hostid;
+static void done_queue_isr(uint8_t hostid) {
+  (void)hostid;
 
   // done head is written in reversed order of completion --> need to reverse the done queue first
-  ohci_td_item_t* td_head = list_reverse ( (ohci_td_item_t*) tu_align16(ohci_data.hcca.done_head) );
+  ohci_td_item_t* td_head = list_reverse((ohci_td_item_t*)tu_align16(ohci_data.hcca.done_head));
   ohci_data.hcca.done_head = 0;
```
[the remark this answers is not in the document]
hathach: this makes sense, but we will resolve this in follow-up with better dcache op for ED

PR #3255, opened 2025-09-17
use existing overloadable Pico SDK section macro instead of creating a custom one

majbthrd:
The present ./src/common/tusb_mcu.h forces a particular TU_ATTR_FAST_FUNC to assign a particular section name (specific to the Pico SDK linker files).

Given there is already a __not_in_flash() macro with the same section name in Pico SDK, it would be desirable to use the existing capabilities instead creating yet more.  Also, the existing __not_in_flash() macro allows for overloading for custom builds, unlike the existing RP2040/RP2350 specific TU_ATTR_FAST_FUNC in tusb_mcu.h


hathach (approved):
thank you

PR #3269, opened 2025-09-29
examples: make: fix LOGGER=rtt

karlp:
CMake builds worked, but the variables were brokenly updated for make.

Fixes: a64e3eb0aa update board_test always output on uart

Fixes errors of: `tinyusb/hw/bsp/board.c:50:10: fatal error: SEGGER_RTT.h: No such file or directory` when building examples with LOGGER=rtt using the classic make build system.


hathach (approved):
thank you, no idea, why I make this change. Probably due to mass find and replace

PR #3271, opened 2025-10-01
examples/device/*_freertos: expand stack size when needed

karlp:
All four of these examples immediately crashed on stack overflow when connected, at least on a FRDM_K64F board.

In 46fd82299040f the default freertos stack size was increased, but _only_ for stm32?  Perhaps either all platform examples need the default increased, rather than increasing the problem task stacks as is done here.

### DIFF
--- a/examples/device/cdc_msc_freertos/src/main.c
+++ b/examples/device/cdc_msc_freertos/src/main.c
@@ -37,7 +37,7 @@
   #define USBD_STACK_SIZE    (3*configMINIMAL_STACK_SIZE/2) * (CFG_TUSB_DEBUG ? 2 : 1)
 #endif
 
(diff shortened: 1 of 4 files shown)


hathach:
did it crash when you enable LOG=1,2 ?

karlp:
Yes, and sure, maybe it's "not required" but surely the demos should _just work_ ?

hathach:
if it only crash when you enabled LOG=2 (CFG_TUSB_DEBUG=2) (printf does consume some stack), then we should use` ifdef CFG_TUSB_DEBUG` to increase it. Not all mcu can run with increased stack size.

hathach (approved):
I have updated PR to only increase the stack when debug is enabled.

hathach (approved):
I have update PR to only increase stack size when debug is enabled.

PR #3270, opened 2025-09-30
audio: Add UAC1.0 support

HiFiPhile:
**Describe the PR**

This PR adds USB Audio Class 1.0 (UAC1) support, now audio device driver supports both UAC1 and UAC2 with dynamic switching possibility based on host capabilities.

For asynchronous OUT with feedback, now it's recommended to use UAC1 for Full-Speed device for better OS compatibility, while UAC2 is recommended only for High-Speed device for better performance.

Let me break down this PR:

### USBD
- Add UAC1 constants to `audio.h` with prefix `audio10_`, UAC2 renamed with prefix `audio20_`
- Add UAC1 descriptor macros to `usbd.h`, refactor fixed size descriptor macros to use auto size with the help of `TU_ARGS_APPLY_EXPAND`
- Support UAC1 without IAD (Interface Association Descriptor)

### Audio device
- Add UAC1 descriptor parsing and control request handling
- Update IN flow control to support UAC1 (OUT feedback by FIFO counting is auto supported)
- Calculate descriptor length by driver, remove `CFG_TUD_AUDIO_FUNC_n_DESC_LEN` config
- Remove `CFG_TUD_AUDIO_ENABLE_FEEDBACK_FORMAT_CORRECTION`, now UAC1 always use 3-bytes 10.14 format while UAC2 always use 4-bytes 16.16 format

### Examples
- **uac2_speaker_fb**: Add UAC1 support with automatic fallback based on host capabilities
  - Remove OS guessing quirks
- **audio_test_multi_rate**: Add UAC1 support with automatic fallback based on host capabilities

### Testing

| Example | OS | UAC1 | UAC2 |
|---------|----|------|------|
| uac2_speaker_fb | Windows 10 | ✅ | ✅ |
| uac2_speaker_fb | macOS 13 | ? | ? |
| uac2_speaker_fb | Archlinux | ✅ | ? |
| audio_test_multi_rate | Windows 10 | ✅ | ✅ |
| audio_test_multi_rate | macOS 13 | ✅ | ✅ |
| audio_test_multi_rate | Archlinux | ✅ | ✅ |

PR #3275, opened 2025-10-02
dcd/dwc2: support ISO IN transfer when bInterval > 1 

HiFiPhile:
**Isochronous IN transfer handling:**
dwc2 requires manually toggle Even/Odd bit manually for ISO IN transfer, that's poses a problem when bInterval > 1 mainly for audio class, as the moment the transfer is scheduled, we don't know when the host will issue IN token (bInterval vs bRefresh shenanigans).

Linux driver use NAK interrupt to detect when the host is sending IN token and toggle the Even/Odd bit accordingly based on the current frame number and bInterval.

However on ST's stripped down DWC2 FS controller (e.g STM32F4, STM32F7), NAK interrupt is not supported, even it's marked as always present in DWC2 databook. NAK interrupt is only supported on HS controller.

Instead I schedule all ISO IN transfer for next frame, if the transfer failed, incomplete isochronous IN transfer interrupt will be triggered and we can relaunch the transfer.

Fix some issues in #3270

### DIFF
--- a/src/portable/synopsys/dwc2/dcd_dwc2.c
+++ b/src/portable/synopsys/dwc2/dcd_dwc2.c
@@ -650,14 +650,12 @@ bool dcd_edpt_xfer(uint8_t rhport, uint8_t ep_addr, uint8_t* buffer, uint16_t to
     xfer->buffer = buffer;
     xfer->ff = NULL;
     xfer->total_len = total_bytes;
+    xfer->iso_retry = xfer->interval; // Reset ISO retry counter to interval value
 
     // EP0 can only handle one packet
     if (epnum == 0) {
       _dcd_data.ep0_pending[dir] = total_bytes;
     }
-
-    // Reset ISO retry counter to interval value
-    xfer->iso_retry = xfer->interval;
 
     // Schedule packets to be sent within interrupt
     edpt_schedule_packets(rhport, epnum, dir);
@@ -690,9 +688,7 @@ bool dcd_edpt_xfer_fifo(uint8_t rhport, uint8_t ep_addr, tu_fifo_t* ff, uint16_t
     xfer->buffer = NULL;
     xfer->ff = ff;
     xfer->total_len = total_bytes;
-
-    // Reset ISO retry counter to interval value
-    xfer->iso_retry = xfer->interval;
+    xfer->iso_retry = xfer->interval; // Reset ISO retry counter to interval value
 
     // Schedule packets to be sent within interrupt
     // TODO xfer fifo may only available for slave mode
@@ -1070,6 +1066,43 @@ static void handle_ep_irq(uint8_t rhport, uint8_t dir) {
   }
 }
 
+static void handle_incomplete_iso_in(uint8_t rhport) {
+  dwc2_regs_t      *dwc2    = DWC2_REG(rhport);
+  const dwc2_dsts_t dsts    = {.value = dwc2->dsts};
+  const uint32_t    odd_now = dsts.frame_number & 1u;
+
+  // Loop over all IN endpoints
+  const uint8_t ep_count = dwc2_ep_count(dwc2);
+  for (uint8_t epnum = 0; epnum < ep_count; epnum++) {
+    dwc2_dep_t   *epin   = &dwc2->epin[epnum];
+    dwc2_depctl_t depctl = {.value = epin->diepctl};
+    // Read DSTS and DIEPCTLn for all isochronous endpoints. If the current EP is enabled and the read value of
+    // DSTS.SOFFN is the targeted uframe number for this EP, then this EP has an incomplete transfer.
+    if (depctl.enable && depctl.type == DEPCTL_EPTYPE_ISOCHRONOUS && depctl.dpid_iso_odd == odd_now) {
+      xfer_ctl_t *xfer = XFER_CTL_BASE(epnum, TUSB_DIR_IN);
+      if (xfer->iso_retry > 0) {
+        xfer->iso_retry--;
+        // Restart ISO transfe: re-write TSIZ and CTL
+        dwc2_ep_tsize_t deptsiz = {.value = 0};
+        deptsiz.xfer_size       = xfer->total_len;
+        deptsiz.packet_count    = tu_div_ceil(xfer->total_len, xfer->max_size);
+        epin->tsiz              = deptsiz.value;
+
+        if (odd_now) {
+          depctl.set_data0_iso_even = 1;
+        } else {
+          depctl.set_data1_iso_odd = 1;
+        }
+        epin->diepctl = depctl.value;
+      } else {
+        // too many retries, give up
+        edpt_disable(rhport, epnum | TUSB_DIR_IN_MASK, false);
+        dcd_event_xfer_complete(rhport, epnum | TUSB_DIR_IN_MASK, 0, XFER_RESULT_FAILED, true);
+      }
+    }
+  }
+}
+
 /* Interrupt Hierarchy
                  DIEPINT  DIEPINT
                     \       /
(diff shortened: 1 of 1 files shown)


hathach:
@codex review this pr

HiFiPhile:
The programming manual section 8.2.5 says we can simply toggle Even/Odd inside Incomplete transfer interrupt, but I tried all boards I have from 2.70a to 4.20a and it never worked. 
In `DEPCTL` the bit is toggled but the endpoint doesn't make transfer at all, also section 8.2.7 says the opposite, that the endpoint must be disabled....

image

HiFiPhile:
@hathach 
I've removed the part of `writing data directly to the FIFO for single-packet IN transfers` as I've spotted some issues with EP0 `ep0_pending` logic. I'll put that part in a new PR.

hathach:
@HiFiPhile retrying iso with flush and re-write packet seems a bit heavy, with high enough bInterval we may end up doing it mostly every frame. Since dwc2 requires

```
For Periodic IN endpoints, the data must always be prefetched 1 microframe ahead for transmission
in the next microframe. This can be done, by enabling the Periodic IN endpoint 1 (micro)frame ahead
of the (micro)frame in which the data transfer is scheduled
```

I think in the future we have to improve how we manage ISO transfer e.g
- iso transfer --> compute target frame_num = last xferred frame_num + bInterval 
- enable sof interrupt, loop for iso in check target frame, if there is transfer next --> write data, set odd/even bit. 
- iso incomplete still requires, but I guess host will mostly respect the iso bInterval so the number of rety would be much less often. 
- sof interrupt can be disabled when there is no active iso transfer.

HiFiPhile:
> * iso transfer --> compute target frame_num = last xferred frame_num + bInterval
> * enable sof interrupt, loop for iso in check target frame, if there is transfer next --> write data, set odd/even bit.
> * iso incomplete still requires, but I guess host will mostly respect the iso bInterval so the number of rety would be much less often.
> * sof interrupt can be disabled when there is no active iso transfer.

That's what should be done ideally (kernel does something similar), but in real it's not always bInterval who decides the transfer interval.
UAC1 feedback endpoint has another field named bRefresh, Windows and Mac OS will use bRefresh as actual transfer interval, Linux stays with bInterval.

I wrote an explanation here: https://github.com/HiFiPhile/tinyusb/blob/9f1a86c0cbf2974775687848a95a126c6503c1cf/examples/device/uac2_speaker_fb/src/usb_descriptors.h#L137

hathach:
Thanks @HiFiPhile for the detail info, I think we should follow the bInterval, I don't understand UAC1, but I think bInterval is the hard limit between the ISO transfer, while bRefresh is probably more flexible maybe bRefresh = n*bInterval. Mabye the class driver can have both with the same value to make every OS happy. For dwc2 driver, I think sticking with bInterval is easier and cleaner.

HiFiPhile:
@hathach I remembered why we can't do scheduling with bInterval, due to Nak interrupt is unavailable on some devices (OTG_FS). Nak interrupt is needed in order to align the frame number with the host. 

For example when bInterval=4 kernel does this:
- Enable global Nak interrupt on transfer start
- Nak interrupt will fire since no packet is written
- In Nak interrupt, check if it's ISO IN EP, then get the frame number (ex. 10), calculate the next frame of ISO transfer (14)
- In SOF interrupt when frame number = 13, enable the transfer and toggle even/odd

Without knowing the starting frame, if we schedule the frame on 3,7,11,15 then the transfer won't work. Also Nak is a global interrupt which consumes a lot of CPU.

---

 Actually instead of retry 255 times (max bInerval allowed), we can limit to actual bInerval value. For UAC1 if bRefresh >0 it will generate some transfer failed event but harmless.

---

I figured out how to restart the transfer without disable/reenable the endpoint.

hathach:
@HiFiPhile does it need to use global NAK, I see there is DIEPINTi bit 13 that is available in stm32f4.

image

HiFiPhile:
@hathach  It's only available for OTG_HS, OTG_FS doesn't have it .
image

hathach:
@HiFiPhile which mcu in your screenshot, mine is otg fs stm32f405/407 

image

HiFiPhile:
@hathach 

Just realized my RM0090 is too old !

DIEPMSK->NAKM will enable the interrupt for all endpoints, many interrupts will be generated especially if there are any bulk EP.

Incomplete transfer interrupt only fires once each frame, by default when 80% time of the frame is elapsed. So the CPU load is similar to SOF interrupt.
image

hathach:
@HiFiPhile yeah, I think using incomplete iso interrupt is much better than NAK. I guess maybe this interrupt is probably introduced in later version of dwc2 where NAK is already implemented. Also I think we only need to do this once per ISO activate for syncing the first, however we cannot trust host to keep the bInterval right, so yeah sync target/last frame numb everytime ISO is transfer complete or incomplete is great idea.

HiFiPhile:
@hathach can we merge it ?

hathach:
@HiFiPhile oops, I thought there is still more changes to the PR. Didn't know that it is all implemented, give me a bit of time to do some basic testing.

src/portable/synopsys/dwc2/dcd_dwc2.c
```diff
@@ -1137,6 +1150,43 @@ void dcd_int_handler(uint8_t rhport) {
...
+    const uint32_t odd_now = dsts.frame_number & 1u;
+    // Loop over all IN endpoints
+    const uint8_t ep_count = dwc2_ep_count(dwc2);
+    for (uint8_t epnum = 0; epnum < ep_count; epnum++) {
+      dwc2_dep_t* epin = &dwc2->epin[epnum];
+      dwc2_depctl_t depctl = {.value = epin->diepctl};
+      // Read DSTS and DIEPCTLn for all isochronous endpoints. If the current EP is enabled
+      // and the read value of DSTS.SOFFN is the targeted uframe number for this EP, then
+      // this EP has an incomplete transfer.
+      if (depctl.enable && depctl.type == DEPCTL_EPTYPE_ISOCHRONOUS && depctl.dpid_iso_odd == odd_now) {
```
hathach: is there any chances that 2 ISO is schedule at the same uframe, and 1 is complete and 1 is incomplete and we detect the wrong endpoint ?

hathach (commented):
@HiFiPhile PR stil haven't implemented the target frame yet. So the application need to call the edpt_xfer() 1 frame before the bInterval, otherwise we will miss that IN token and send data in the next bInterval (late timing) ? 

> I think in the future we have to improve how we manage ISO transfer e.g
> 
> * iso transfer --> compute target frame_num = last xferred frame_num + bInterval
> * enable sof interrupt, loop for iso in check target frame, if there is transfer next --> write data, set odd/even bit.
> * iso incomplete still requires, but I guess host will mostly respect the iso bInterval so the number of rety would be much less often.
> * sof interrupt can be disabled when there is no active iso transfer.

src/portable/synopsys/dwc2/dcd_dwc2.c
HiFiPhile: > is there any chances that 2 ISO is schedule at the same uframe, and 1 is complete and 1 is incomplete and we detect the wrong endpoint ?

The interrupt is fired at the end of periodic frame interval (80% of the frame by default), it's unlikely any transfer is still pending.

image

HiFiPhile:
> PR stil haven't implemented the target frame yet. So the application need to call the edpt_xfer() 1 frame before the bInterval, otherwise we will miss that IN token and send data in the next bInterval (late timing) ?

@hathach Both audio and video are real-time streaming data it's ok to return transfer failed if the host didn't read the data in the specified interval.  edpt_xfer() needs to be called not earlier than bInterval*frames before the IN token:
- If bInterval = 1 the packet is scheduled at frame+1 in edpt_schedule_packets()
  - Transfer success if completed at frame+1
  - Otherwise Incomplete interrupt returns transfer fail
- If bInerval = 2 the packet is also scheduled at frame+1 in edpt_schedule_packets(), maybe Incomplete interrupt will be triggered depending when the host started the 1st transfer
  - Transfer success if completed at frame+1
  - Otherwise Incomplete interrupt see the retry counter > 0 then reschedule at frame+2
    - Transfer success if completed at frame+2
    - Otherwise Incomplete interrupt returns transfer fail since retry counter = 0

hathach:
@HiFiPhile thanks for the explanation, I know ISO data can be missed. Let's say bInterval = 10, and we recevie setInterface at frame=100, assume host send the 1st IN token as well. So we will have 
- frame 100: 1st IN token
- frame 110: 2nd In token and so on

So ideally video/audio must prepare the next iso data within the bInterval and call edpt_xfer() it as soon as the data is ready. let's say at frame 105th, the stack would then ideally only fetch data to fifo at frame 109. with the current implementation, what would happen at frame 106,107 ?

HiFiPhile:
> So ideally video/audio must prepare the next iso data within the bInterval and call edpt_xfer() it as soon as the data is ready. let's say at frame 105th, the stack would then ideally only fetch data to fifo at frame 109. with the current implementation, what would happen at frame 106,107 ?

@hathach If edpt_xfer is called at 105th data will be pushed into fifo and the transfer will be scheduled for 106th. At 106th incomplete interrupt fires and rearm the endpoint for 107th. At 107th same thing happens, until the transfer succeed in 110th.

It's the procedure of step 5 chapter 8.2.5, except the guild forgot to mention that DIEPTSIZ needs to be rewritten (how I failed to get it work initially)

hathach:
> > So ideally video/audio must prepare the next iso data within the bInterval and call edpt_xfer() it as soon as the data is ready. let's say at frame 105th, the stack would then ideally only fetch data to fifo at frame 109. with the current implementation, what would happen at frame 106,107 ?
> 
> @hathach If edpt_xfer is called at 105th data will be pushed into fifo and the transfer will be scheduled for 106th. At 106th incomplete interrupt fires and rearm the endpoint for 107th. At 107th same thing happens, until the transfer succeed in 110th.
> 
> It's the procedure of step 5 chapter 8.2.5, except the guild forgot to mention that DIEPTSIZ needs to be rewritten (how I failed to get it work initially)

perfect explanation, also thanks for citation. Each time I open the guide, it feel like completely new to me

PR #3276, opened 2025-10-02
update for release 0.19.0

hathach:


docs/info/changelog.rst
```diff
@@ -2,6 +2,219 @@
...
+  - Fix STM32 PHYC PLL stability wait
+  - Improve PMA size handling for STM32U0
+
+- EHCI
+  - Fix removed QHD getting reused
+  - Fix NXP USBPHY disconnection detection
+
+- Chipidea/NXP
+  - Fix race condition with spinlock
+  - Add async I/O support for MSC
```
HiFiPhile: Interesting Copilot put async IO here...

hathach: yeah, right. I miss this one, This changelog is generated by claude code, which I think work much better than copilot. It saves quite a lot of time, specially with these non-code thing.

HiFiPhile: > claude code, which I think work much better than copilot

Yeah I agree, even on complexe tasks like `generate UAC1 definition header from spec` the output is not bad.

hathach: though claude cost also increased pretty fast. This changelog cost 5$ to generate, including me doing other setup and testing. I am back to codex and copilot to do boring works :)

HiFiPhile: That's a lot, I use it in vscode, the free-tier limit is enough for me, last month I've only used 10% quota with Sonnet 4.

PR #3278, opened 2025-10-04
Documentation improvements

c1570:
**Describe the PR**
This streamlines documentation quite a bit:

- first and foremost, this improves "getting started" in many regards.
  - improving text flow (starts with examples rather than setup details)
  - simplifies the example code to a common case rather than the corner tud+tuh case
  - streamlines details that don't really belong in a getting started document, like IAR
  - improves and rearranges headlines so that the TOC becomes more useful
- this adds a couple of pages
  - architecture overview
  - glossary
  - FAQ
  - troubleshooting

**Additional context**
Note that an LLM has been used for creating this; there has been quite a lot of manual review work of course (I'm a rather new user of TinyUSB though so please check).

I leave this as a Draft for now; I'm down for investing more work into this but want to make sure first that there's general interest @hathach


docs/reference/architecture.rst
```diff
@@ -0,0 +1,291 @@
...
+**Required Functions**:
+- ``init()``: Initialize class driver
+- ``reset()``: Reset class state
+- ``open()``: Configure class endpoints
+- ``control_xfer_cb()``: Handle control requests
+- ``xfer_cb()``: Handle data transfer completion
+
+**Optional Functions**:
+- ``close()``: Clean up class resources
+- ``sof_cb()``: Start-of-frame processing
```
terjr:
```suggestion
- ``sof()``: Start-of-frame processing
```

docs/reference/architecture.rst
```diff
@@ -0,0 +1,291 @@
...
+----------------------
+
+**Required Functions**:
+- ``init()``: Initialize class driver
+- ``reset()``: Reset class state
+- ``open()``: Configure class endpoints
+- ``control_xfer_cb()``: Handle control requests
+- ``xfer_cb()``: Handle data transfer completion
+
+**Optional Functions**:
```
terjr, line 198: `xfer_isr` that was added in 0.19.0 is missing

terjr (changes requested):

c1570:
Great! I added terjr's changes and tested the steps in getting_started, and it seems fine.
I also tested the actual HTML rendering, but it seems some styling is missing locally (also in the original build).
I'm not sure about the FAQ page - while a project should have one, this is mostly generated, and I cannot really verify it. If in doubt we can just delete it for now.
Please review. :)

HiFiPhile:
Thank you, actually CMake is more used today, either full command:
```
cmake -DBOARD=stm32f723disco -B build -G Ninja .  # "-G Ninja" is required on Windows
cmake --build build
```
Or preset:
```
cmake --preset stm32f723disco
cmake --build --preset stm32f723disco
```

For esp32 only CMake works since their SDK only support CMake, whole step is:
1. source esp-idf
2. Enter example folder
3. cmake --preset espressif_s3_devkitc
4. cmake --build --preset espressif_s3_devkitc

c1570:
> Thank you, actually CMake is more used today

Yes I noticed in testing that RP2040 also gives an error on `make` (similar to ESP32 probably).

Is there an equivalent to `make ... all flash`?

HiFiPhile:
> > Thank you, actually CMake is more used today
> 
> Yes I noticed in testing that RP2040 also gives an error on `make` (similar to ESP32 probably).
> 
> Is there an equivalent to `make ... all flash`?

You can build all examples at once by `cd` into example parent folder `tinyusb/examples/device` then do the step above. For flashing with J-Link `cmake --build --preset stm32h7s3nucleo --target audio_test_freertos-jlink`

You can find other supported flasher in family_support.cmake

c1570:
Thanks, I updated the docs to use CMake as default.

hathach (approved):
thank you very much for putting efforts to improve the documentation, which is much needed. I make some changes to also add make instruction which is still very popular, and move things around. This will surely help lots of new users

PR #3291, opened 2025-10-09
feat: Zero-copy USB streaming - 10MB/s to 26.5MB/s improvementr endpoints

Ananth1839:
## What this PR does
Makes USB data transfer 2.6x faster by removing unnecessary data copying.

## The change
Instead of copying data before sending it, we now send data directly from where it's stored.

## Results
- **Before**: 10 MB/s (slow due to copying)
- **After**: 26.5 MB/s (fast - no copying)
- **Improvement**: 165% faster transfers

## Why it matters
- Faster file transfers
- Smoother video streaming
- Better performance for sensors and medical devices

## Tested on
- CH32V305 microcontroller
- USB High-Speed mode
- 512-byte packet size

## Important notes
- Works with existing code
- No breaking changes
- Simple but effective improvement
zero-copy optimization

### DIFF
--- a/src/tusb.c
+++ b/src/tusb.c
@@ -26,7 +26,7 @@
 
 #include "tusb_option.h"
 
-#if CFG_TUH_ENABLED || CFG_TUD_ENABLED
+#if (CFG_TUH_ENABLED || CFG_TUD_ENABLED)
 
 #include "tusb.h"
 #include "common/tusb_private.h"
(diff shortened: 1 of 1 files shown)


HiFiPhile:
NO ! Have you read what you've committed ? Why talking about vendor class but doing some nonsense modification on tusb.c ?

And NO you can't just remove a buffer and call it zero-buffer support ! Many MCU needs to put the buffer on specific memory section with special alignment requirement, like NXP LPC43xx, LPC55x, iMX.RT and STM32H7 etc...  You are just lucky that it works on ch32v305.

Have you ever read what the macro `TUD_EPBUF_DEF`, `CFG_TUD_MEM_ALIGN` and `CFG_TUD_MEM_SECTION` does ?

Close for nonsense.

Issue #3154, opened 2025-06-25
ESP32-S3 bulk transfer issues


qnex0:
### Operating System

Linux

### Commit SHA

 espressif/tinyusb (0.18.0~2)

### Board

ESP32-S3 based board

### Firmware

[tinyusb_firmware.zip](https://github.com/user-attachments/files/20912887/tinyusb_firmware.zip)

[libusb_test.zip](https://github.com/user-attachments/files/20912889/libusb_test.zip)

### What happened ?

I'm trying to implement a custom vendor class driver. I noticed that when using usbd_edpt_xfer to receive data from the OUT EP, the ZLP is misinterpreted and causes the transfer to complete with an xfer size equal to the request size regardless of the actual size. I tried some older versions of tinyusb and noticed this behavior started occurring after version "0.15.0~10." Additionally, when I queue for 32768 bytes in my example, when more than 8191 bytes are sent to the device, the transfer is also completed with the incorrect size.

### How to reproduce ?

I use the libusb_test program attached to the firmware section of this issue to send a variety of different data sizes to the bulk endpoint and view the results using the firmware with debugging enabled.

### Debug Log as txt file (LOG/CFG_TUSB_DEBUG=2)

N/A

### Screenshots

![Image](https://github.com/user-attachments/assets/01f94936-0335-4269-aa97-01ddc1d0a39c)

### I have checked existing issues, dicussion and documentation

- [x] I confirm I have checked existing issues, dicussion and documentation.

dpgeorge:
Just for reference, it looks like this commit db7670a3bcd502193d138041c20d401bc3f59001 was responsible for this issue (it seems that ZLP was correctly handled prior to that commit).

hathach:
@dpgeorge it seems to be fixed by #3293, let me know if that works for you.

dpgeorge:
@hathach yes, that PR fixes the issue.  We have needed to update our TinyUSB submodule to include that patch which is not yet in a release (see https://github.com/micropython/micropython/pull/18406) and also fork the Espressif fork of TinyUSB to patch that as well (see https://github.com/micropython/micropython/pull/18407).

hathach:
@dpgeorge I will make an minor release for this soon enough (in a day or two)

dpgeorge:
> I will make an minor release for this soon enough (in a day or two)

That would be great, thank you!

hathach:
0.20.0 is release, please give it a try to see if there is any other new issues.

[...]

PR #3293, opened 2025-10-10
ESP32-S3 bulk transfer issues #3154

poornadharshan13-rgb:
# 🧩 Description

This PR fixes incorrect handling of **Zero-Length Packets (ZLP)** in the DWC2 driver when receiving data (OUT transfers).  
Previously, the DWC2 RX-FIFO transfer logic would **misinterpret a ZLP as a full-size packet**, causing the transfer to complete with the wrong size (equal to the requested buffer length rather than actual received bytes).

This issue was reported in [#3154 (ESP32-S3 bulk transfer issues)](https://github.com/hathach/tinyusb/issues/3154).

---

## 🛠️ Changes

- **Fix:** Correctly handle ZLP in DWC2 RX-FIFO (ensure `xfer->actual_len` is set to 0 when ZLP is received).  
- **Fix:** Ensure transfer completion only occurs when true end-of-transfer conditions are met (packet smaller than max packet size or explicit ZLP).  
- **Add:** Safety checks in `tusb_memcpy_safe()` and `tusb_memset_safe()` to avoid dereferencing `NULL` pointers.  
- **Refactor:** Minor cleanup and better handling of RX-FIFO edge cases.

---

## ✅ Testing

- Tested on **ESP32-S3** with a custom vendor class using **libusb** host code (`libusb_test.zip` from issue #3154).  
- Verified with various transfer sizes (1–32768 bytes), including boundary cases and ZLP.  
- ZLP is now correctly reported as zero-length.  
- No regression observed on normal bulk transfers.

---

## 📦 Affected Components

- `src/portable/synopsys/dwc2/dcd_dwc2.c`  
- `src/common/tusb_common.c` (safe mem helpers)

---

## 🧪 Repro Steps (Verified)

1. Queue a bulk OUT transfer using `usbd_edpt_xfer()` for >8 KB buffer (e.g., 32768 bytes).  
2. Send varying sizes of data from host using `libusb_test`.  
3. Observe correct `xfer->actual_len` for all sizes, including ZLP.

---

## 🧠 Notes

- Behavior regression introduced after **TinyUSB 0.15.0~10**, likely due to DWC2 RX-FIFO refactor.  
- This PR restores correct ZLP handling and adds guard checks to improve driver safety.

---

## 🏁 Conclusion

Fixes [#3154](https://github.com/hathach/tinyusb/issues/3154):  
✅ Correct ZLP detection and handling for DWC2 (ESP32-S3).  
✅ Safe memory helpers prevent null pointer usage.  
✅ Restores expected behavior for bulk OUT transfers.

### DIFF
--- a/src/portable/synopsys/dwc2/dcd_dwc2.c
+++ b/src/portable/synopsys/dwc2/dcd_dwc2.c
@@ -814,17 +814,18 @@ static void handle_rxflvl_irq(uint8_t rhport) {
           dfifo_read_packet(dwc2, xfer->buffer, byte_count);
           xfer->buffer += byte_count;
         }
+      }
 
-        // short packet, minus remaining bytes (xfer_size)
-        if (byte_count < xfer->max_size) {
-          const dwc2_ep_tsize_t tsiz = {.value = epout->tsiz};
-          xfer->total_len -= tsiz.xfer_size;
-          if (epnum == 0) {
-            xfer->total_len -= _dcd_data.ep0_pending[TUSB_DIR_OUT];
-            _dcd_data.ep0_pending[TUSB_DIR_OUT] = 0;
-          }
+      // short packet (including ZLP when byte_count == 0), minus remaining bytes (xfer_size)
+      if (byte_count < xfer->max_size) {
+        const dwc2_ep_tsize_t tsiz = {.value = epout->tsiz};
+        xfer->total_len -= tsiz.xfer_size;
+        if (epnum == 0) {
+          xfer->total_len -= _dcd_data.ep0_pending[TUSB_DIR_OUT];
+          _dcd_data.ep0_pending[TUSB_DIR_OUT] = 0;
         }
       }
+
       break;
     }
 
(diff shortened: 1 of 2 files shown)


src/portable/synopsys/dwc2/dcd_dwc2.c
```diff
@@ -814,17 +814,18 @@ static void handle_rxflvl_irq(uint8_t rhport) {
...
 
-        // short packet, minus remaining bytes (xfer_size)
-        if (byte_count < xfer->max_size) {
-          const dwc2_ep_tsize_t tsiz = {.value = epout->tsiz};
-          xfer->total_len -= tsiz.xfer_size;
-          if (epnum == 0) {
-            xfer->total_len -= _dcd_data.ep0_pending[TUSB_DIR_OUT];
-            _dcd_data.ep0_pending[TUSB_DIR_OUT] = 0;
-          }
+      // short packet (including ZLP when byte_count == 0), minus remaining bytes (xfer_size)
```
hathach, line 819: what is the code changes besides comment ? can you elaborate ?

HiFiPhile: The logic is moved out of `if (byte_count) {` parentheses to apply also on ZLP (I didn't spot either until I check the full diff...)

hathach: oops. sorry, my bads, the diff only make it difficult to see the fix.

PR #3295, opened 2025-10-10
dcd/dwc2: fix EP0 multi-packet transfer logic

HiFiPhile:
**Describe the PR**
There was an issue in `edpt_schedule_packets`, as `ep0_pending = xfer->max_size` is set in `dcd_edpt_xfer`, ep0_pending will become 0.
```
if (epnum == 0) {
    total_bytes = tu_min16(_dcd_data.ep0_pending[dir], xfer->max_size);
    _dcd_data.ep0_pending[dir] -= total_bytes;
    num_packets = 1;
  } 
```
The issue didn't broken since we schedule only one control transfer each time.

I've add an option `CFG_TUD_EP0_BUFSIZE` to allow setting EP0 buffer size individually, tested with DFU examples with`CFG_TUD_EP0_BUFSIZE=512`

### DIFF
--- a/examples/host/device_info/src/main.c
+++ b/examples/host/device_info/src/main.c
@@ -99,6 +99,7 @@ int main(void) {
 #if CFG_TUSB_OS == OPT_OS_FREERTOS
   init_freertos_task();
 #else
+  board_delay(100); // wait for uart to be ready
   init_tinyusb();
   while (1) {
     tuh_task();     // tinyusb host task
@@ -289,6 +290,7 @@ void app_main(void) {
 
 void usb_host_task(void *param) {
   (void) param;
+  board_delay(100); // wait for uart to be ready
   init_tinyusb();
   while (1) {
     tuh_task();


src/portable/synopsys/dwc2/dcd_dwc2.c
```diff
@@ -374,12 +374,29 @@ static void edpt_schedule_packets(uint8_t rhport, const uint8_t epnum, const uin
...
+    if (epnum == 0) {
+      xfer->buffer += total_bytes;
+    }
   } else {
     dep->diepctl = depctl.value; // enable endpoint
 
     // Enable tx fifo empty interrupt only if there is data. Note must after depctl enable
     if (dir == TUSB_DIR_IN && total_bytes != 0) {
-      dwc2->diepempmsk |= (1u << epnum);
+      // For num_packets = 1 we write the packet directly
```
hathach: I think we should keep using the TXFE interrupt  to write to fifo. Otherwise there could be race condition whereas usbd mis-manage control transfer, and we are attempting to xfer on busy endpoint0. 

also I think we need to also increase (double) the dfifo_alloc() for EP0 when the EP0_BUFSIZE is large enough e.g 8x64 bytes since user is clearly want to make a large transfer with EP0

src/portable/synopsys/dwc2/dcd_dwc2.c
```diff
@@ -336,7 +336,7 @@ static void edpt_schedule_packets(uint8_t rhport, const uint8_t epnum, const uin
 
   // EP0 is limited to one packet per xfer
   if (epnum == 0) {
-    total_bytes = tu_min16(_dcd_data.ep0_pending[dir], xfer->max_size);
+    total_bytes = tu_min16(_dcd_data.ep0_pending[dir], CFG_TUD_ENDPOINT0_SIZE);
     _dcd_data.ep0_pending[dir] -= total_bytes;
```
hathach, line 383: I think we should not pre-calculate the pending, we should just store the total_bytes and minus xferred_bytes in complete isr instead like other bulk transfers

src/portable/synopsys/dwc2/dcd_dwc2.c
HiFiPhile: > I think we should keep using the TXFE interrupt to write to fifo. Otherwise there could be race condition whereas usbd mis-manage control transfer, and we are attempting to xfer on busy endpoint0.

I think it should be safe, which scenario you are thinking about ?

> also I think we need to also increase (double) the dfifo_alloc() for EP0 when the EP0_BUFSIZE is large enough e.g 8x64 bytes since user is clearly want to make a large transfer with EP0

EP0 IN can only queue 3 packets each time, EP0 OUT can only queue 1 packet each time, so increasing the fifo won't have much help.

image

src/portable/synopsys/dwc2/dcd_dwc2.c
HiFiPhile: I think we can't get rid of `ep0_pending`, otherwise we need to add `ep0_xferred_bytes` which ends up the same thing, just one counting up another counting down.

src/portable/synopsys/dwc2/dcd_dwc2.c
hathach: > I think it should be safe, which scenario you are thinking about ?

You are right, I just feel a bit un-easy when both handle_epin_slave() and edpt_schedule_packets() can write to difof. Especially handle_epin_slave() also called edpt_schedule_packets() in case of ep0. But it is probably safe enough, actually we can refactor this and try to write as many packet as we could before enable the TX empty if there is remaining packet ?

```C
    dwc2_ep_tsize_t tsiz = {.value = epin->tsiz};
    const uint16_t remain_packets = tsiz.packet_count;

    // Process every single packet (only whole packets can be written to fifo)
    for (uint16_t i = 0; i < remain_packets; i++) {
      tsiz.value = epin->tsiz;
      const uint16_t remain_bytes = (uint16_t) tsiz.xfer_size;
      const uint16_t xact_bytes = tu_min16(remain_bytes, xfer->max_size);

      // Check if dtxfsts has enough space available
      if (xact_bytes > ((epin->dtxfsts & DTXFSTS_INEPTFSAV_Msk) << 2)) {
        break;
      }

      // Push packet to Tx-FIFO
      if (xfer->ff) {
        volatile uint32_t* tx_fifo = dwc2->fifo[epnum];
        tu_fifo_read_n_const_addr_full_words(xfer->ff, (void*)(uintptr_t)tx_fifo, xact_bytes);
      } else {
        dfifo_write_packet(dwc2, epnum, xfer->buffer, xact_bytes);
        xfer->buffer += xact_bytes;
      }
    }
``` 

> EP0 IN can only queue 3 packets each time, EP0 OUT can only queue 1 packet each time, so increasing the fifo won't have much help.

right, again, I am too lazy to check the specs, the increased ep0 bufsize only help to reduce the overhead when pushing isr --> usbd event stack. Let keep the ep0 1 packet per xfer is good enough.

src/portable/synopsys/dwc2/dcd_dwc2.c
hathach: I agree, ep0_xferred_bytes is better since it will only increased when xfer is complete, and we can also use that in case of error for xferred so far. Maybe i will do that in follow-up PR later.

src/portable/synopsys/dwc2/dcd_dwc2.c
HiFiPhile: I've refactored the fifo writing routine and added a tunable `CFG_TUD_DWC2_IN_DOUBLE_BUFFER` to set `GAHBCFG_TX_FIFO_EPMTY_LVL` and fifo allocation, in order to allow writing 2 packets to fifo. 
By default we allocate fifo size=packet size, the writing loop will likely stop after one packet is written.

src/portable/synopsys/dwc2/dcd_dwc2.c
HiFiPhile: For OUT transfer it's easy to get `ep0_xferred_bytes`, but for IN transfer it can't be read from register, we have to add something like `last_xfer_bytes` and set it in `edpt_schedule_packets`. I don't know if it's worth the change.

src/tusb_option.h
```diff
@@ -284,6 +284,11 @@
   #define CFG_TUD_DWC2_DMA_ENABLE CFG_TUD_DWC2_DMA_ENABLE_DEFAULT
 #endif
 
+// Allocate double buffer for IN endpoint for better performance (only applies to non-periodic endpoints)
+#ifndef CFG_TUD_DWC2_IN_DOUBLE_BUFFER
+  #define CFG_TUD_DWC2_IN_DOUBLE_BUFFER 0
```
hathach: I think we should not introduce this option. I always think the fine-tune fifo should be dictated by application per endpoint basic. E.g MSC + CDC, we would want a massive endpoint size for msc and smaller for cdc. I am think we should do that via a tud_configure() (which allo to pass an option id + pointer to struct) that fine tune the behavior of dcd or stack. similar to tuh_configure() https://github.com/hathach/tinyusb/blob/master/src/host/usbh.h#L145 . dwc2 option may include grxsize and txfifo size for each endpoint (0 for auto as current).

src/portable/synopsys/dwc2/dcd_dwc2.c
hathach: you are right, sorry, I forgot some of these thing.

src/portable/synopsys/dwc2/dcd_dwc2.c
```diff
@@ -887,32 +932,10 @@ static void handle_epin_slave(uint8_t rhport, uint8_t epnum, dwc2_diepint_t diep
...
-      // Push packet to Tx-FIFO
-      if (xfer->ff) {
-        volatile uint32_t* tx_fifo = dwc2->fifo[epnum];
-        tu_fifo_read_n_const_addr_full_words(xfer->ff, (void*)(uintptr_t)tx_fifo, xact_bytes);
-      } else {
-        dfifo_write_packet(dwc2, epnum, xfer->buffer, xact_bytes);
-        xfer->buffer += xact_bytes;
-      }
-    }
+    epin_write_tx_fifo(rhport, epnum);
```
hathach, line 944: we have a possibility to call epin_write_tx_fifo() 2nd time with ep0. although it would not write any bytes, I think we should skip this for ep0.

src/tusb_option.h
HiFiPhile: Yeah a configurable parameter is better, didn't notice there is a tuh_configure().

src/portable/synopsys/dwc2/dcd_dwc2.c
HiFiPhile: You are right.

HiFiPhile:
@hathach I'm think about when `CFG_TUD_ENDPOINT0_BUFSIZE >= CFG_TUD_DFU_XFER_BUFSIZE` we can skip dfu transfer buffer.

hathach:
> @hathach I'm think about when `CFG_TUD_ENDPOINT0_BUFSIZE >= CFG_TUD_DFU_XFER_BUFSIZE` we can skip dfu transfer buffer.

yeah, though I think we can do it in follow up Pr.

Issue #1973, opened 2023-03-23
Static Analytics checklist


hathach:
### Related area

all

### Hardware specification

all

### Is your feature request related to a problem?

similar to #1855 but in much broader view. There is several analytic software that is free for opensource project. That does perform code checking againt popular standard e.g MISRA, CERT etc ... we should run as much as we could with CI to make tinyusb safer.

### Describe the solution you'd like

Doable
- [ ] Coverity: partial configured should have been done better https://scan.coverity.com/projects/tinyusb
- [ ] Cppcheck https://github.com/danmar/cppcheck
- [x] IAR C-STAT: probably included in BX license (granted to tinyusb), but haven't configured to run yet
- [x] Sonarqube/SonarCloud: just signed up https://sonarcloud.io/project/information?id=hathach_tinyusb need setup/configuration
- [x] CodeScene: just signed up, do 1st run https://codescene.io/projects/37161/jobs/965524/results?scope=month#code-health need more setup/configuration
- [x] PVS-Studio: just signed up https://pvs-studio.com/en/order/open-source-license/
- [ ] Codecov just signed up https://about.codecov.io/pricing/
- [ ] Flawfinder https://github.com/david-a-wheeler/flawfinder
- [ ] Codacy https://www.codacy.com/
- [ ] Coveralls https://coveralls.io/
- [ ] clang-tidy https://clang.llvm.org/extra/clang-tidy/
- [ ] Frama-C https://frama-c.com/index.html

Watch list
- [ ] Qodana: but C/C++ is not yet supported, https://github.com/JetBrains/Qodana/discussions/54 , maybe up vote to make it implemented faster ? https://youtrack.jetbrains.com/issue/QD-2153
- [ ] CodeClimate just signed up, https://codeclimate.com/github/hathach/tinyusb but it wanrs that C is not supported yet

### I have checked existing issues, dicussion and documentation

- [X] I confirm I have checked existing issues, dicussion and documentation.

hathach:
currently pvs-studion, Sonarqube and codescene IAR C-stat (disabled for now) are alreay in use. This can be closed https://github.com/hathach/tinyusb/actions/workflows/static_analysis.yml

PR #3318, opened 2025-10-24
Static analysis

hathach:
- Add IAR CStat analyze with cmake -DIAR_CSTAT=1
- Fix some warnings suggestd by icstat and/or pvs-studio
- Add pvs-studio, sonarqubecloud, IAR CStat to ci to analyze and upload sarif to github

Note IAR CStat has issue when analyzing mtp example, therefore we only run it with cdc_msc example for now


hathach:
add some of mentioned tools in #1973

PR #3319, opened 2025-10-24
dcd/dwc2: fix suspend interrupt indefinitely disabled on AT32F405

### LINKED ISSUE #3315: AT32F405 gets stuck in a suspended state even though the bus is resumed
### Operating System

Windows 11

### Commit SHA

4dfac3f3566268450b46f1c19899b0b53bf48706

### Board

ArteryTek AT32F405 Custom Board

### Firmware

Custom firmware with this [tusb_config.h](https://github.com/peppapighs/libhmk/blob/main/include/tusb_config.h)

### What happened ?

When the bus is suspended, the device does not receive the event `DCD_EVENT_SUSPEND` event received. However, when the bus is resumed, the device receives both `DCD_EVENT_RESUME` and `DCD_EVENT_SUSPEND` in two consecutive interrupts, but `DCD_EVENT_RESUME` was processed first [here](https://github.com/hathach/tinyusb/blob/79445c2386adefb207a76a70d87578b53c3e7922/src/device/usbd.c#L637) then `DCD_EVENT_SUSPEND`, so `_usb_dev` now get stuck in a suspended state perpetually when it should not.

~~The current workaround I have is to modify TinyUSB locally to ignore the event `DCD_EVENT_SUSPEND`, but this seems like a very hacky solution. I also cannot reproduce this issue on AT32F405 evaluation board since it seems my PC refuses to suspend the evaluation board (maybe because of its lower current draw?).~~

~~Edit: I verified that the suspended status in the `dsts` register is consistent with the actual suspended state. Not sure why the interrupt handler is not called when the device gets suspended.~~

I found the cause of the issue. On initialization, AT32F405 receives a suspend interrupt once (`usbsusp` = `usbsuspm` = `1`) before the device is connected. As a result, [`dcd_int_handler`](https://github.com/hathach/tinyusb/blob/ac37176cec68a26a308c67b238cbf8901b3f1f29/src/portable/synopsys/dwc2/dcd_dwc2.c#L1063) disables the suspend interrupt, even though the device is still unconnected. Subsequently, when the bus is suspended, no suspend interrupt is received. When the bus is resumed, [`dcd_int_handler`](https://github.com/hathach/tinyusb/blob/ac37176cec68a26a308c67b238cbf8901b3f1f29/src/portable/synopsys/dwc2/dcd_dwc2.c#L1069) handles the wakeup interrupt, and enable the suspend interrupt, causing the pending suspend interrupt from before to suspend the device indefinitely.

I suppose the correct implementation is to only disable suspend interrupt when the device is connected, but `dcd_int_handler` has no access to `_usb_dev.connected`, so I'm not sure what should be done here. At least preventing `dcd_int_handler` from disabling the suspend interrupt entirely works.

### How to reproduce ?

Flash the firmware [here](https://github.com/peppapighs/libhmk), and suspend the board.

I don't have a J-LINK, but I can put a breakpoint to inspect the memory manually if needed.

### Screenshots

_No response_

peppapighs:
**Describe the PR**
Fix #3315 by only process SUSPEND event if the device is connected. Tested locally to confirm that `_usb_dev` state is as intended when the bus is suspended and resume. We probably don't need to repeat this change for RESUME event since enabling suspend interrupt during initialization should have no side effect.

Closes #3315

### DIFF
--- a/src/portable/synopsys/dwc2/dcd_dwc2.c
+++ b/src/portable/synopsys/dwc2/dcd_dwc2.c
@@ -1054,6 +1054,8 @@ void dcd_int_handler(uint8_t rhport) {
   if (gintsts & GINTSTS_ENUMDNE) {
     // ENUMDNE is the end of reset where speed of the link is detected
     dwc2->gintsts = GINTSTS_ENUMDNE;
+    // There may be a pending suspend event, so we clear it first
+    dwc2->gintsts = GINTSTS_USBSUSP;
     dwc2->gintmsk |= GINTMSK_USBSUSPM;
     handle_enum_done(rhport);
   }
--- a/src/portable/synopsys/dwc2/dwc2_at32.h
+++ b/src/portable/synopsys/dwc2/dwc2_at32.h
@@ -64,58 +64,59 @@
 #endif
 
 #ifdef __cplusplus
- extern "C" {
+extern "C" {
 #endif
 
- static const dwc2_controller_t _dwc2_controller[] = {
-{.reg_base = DWC2_OTG1_REG_BASE, .irqnum = OTG1_IRQn, .ep_count = DWC2_EP_MAX, .ep_fifo_size = OTG1_FIFO_SIZE},
+static const dwc2_controller_t _dwc2_controller[] = {
+    {.reg_base = DWC2_OTG1_REG_BASE, .irqnum = OTG1_IRQn, .ep_count = DWC2_EP_MAX, .ep_fifo_size = OTG1_FIFO_SIZE},
 #if defined DWC2_OTG2_REG_BASE
-  {.reg_base = DWC2_OTG2_REG_BASE, .irqnum = OTG2_IRQn, .ep_count = DWC2_EP_MAX, .ep_fifo_size = OTG2_FIFO_SIZE}
+    {.reg_base = DWC2_OTG2_REG_BASE, .irqnum = OTG2_IRQn, .ep_count = DWC2_EP_MAX, .ep_fifo_size = OTG2_FIFO_SIZE}
 #endif
- };
+};
 
- TU_ATTR_ALWAYS_INLINE static inline void dwc2_int_set(uint8_t rhport, tusb_role_t role, bool enabled) {
-   (void) role;
-   const IRQn_Type irqn = (IRQn_Type) _dwc2_controller[rhport].irqnum;
-   if (enabled) {
-     NVIC_EnableIRQ(irqn);
-   } else {
-     NVIC_DisableIRQ(irqn);
-   }
- }
+TU_ATTR_ALWAYS_INLINE static inline void dwc2_int_set(uint8_t rhport, tusb_role_t role, bool enabled) {
+  (void) role;
+  const IRQn_Type irqn = (IRQn_Type) _dwc2_controller[rhport].irqnum;
+  if (enabled) {
+    NVIC_EnableIRQ(irqn);
+  } else {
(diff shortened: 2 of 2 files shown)


HiFiPhile:
In TinyUSB's structure, DCD layer shouldn't call exported device stack functions (tud_*)

> I found the cause of the issue. On initialization, AT32F405 receives a suspend interrupt once (usbsusp = usbsuspm = 1) before the device is connected. As a result, [dcd_int_handler](https://github.com/hathach/tinyusb/blob/ac37176cec68a26a308c67b238cbf8901b3f1f29/src/portable/synopsys/dwc2/dcd_dwc2.c#L1063) disables the suspend interrupt, even though the device is still unconnected.

When plugging/unplugging device, the D+/D- state are unstable and can accidentally meet the SUSPEND condition. But this shouldn't be an issue since the interrupt is enabled after the usb reset:
https://github.com/hathach/tinyusb/blob/ac37176cec68a26a308c67b238cbf8901b3f1f29/src/portable/synopsys/dwc2/dcd_dwc2.c#L1057
Unless `GINTSTS_USBSUSP` is already pending and should be cleared before enabling.

peppapighs:
> In TinyUSB's structure, DCD layer shouldn't call exported device stack functions (tud_*)
> 
> > I found the cause of the issue. On initialization, AT32F405 receives a suspend interrupt once (usbsusp = usbsuspm = 1) before the device is connected. As a result, [dcd_int_handler](https://github.com/hathach/tinyusb/blob/ac37176cec68a26a308c67b238cbf8901b3f1f29/src/portable/synopsys/dwc2/dcd_dwc2.c#L1063) disables the suspend interrupt, even though the device is still unconnected.
> 
> When plugging/unplugging device, the D+/D- state are unstable and can accidentally meet the SUSPEND condition. But this shouldn't be an issue since the interrupt is enabled after the usb reset:
> 
> https://github.com/hathach/tinyusb/blob/ac37176cec68a26a308c67b238cbf8901b3f1f29/src/portable/synopsys/dwc2/dcd_dwc2.c#L1057
> 
> 
> Unless `GINTSTS_USBSUSP` is already pending and should be cleared before enabling.

You are right. I put a breakpoint at USB reset event, and `usbsusp = 1`, so if I understand this correctly, there is a pending suspend interrupt before USB reset. I'll make the change to clear it first.

HiFiPhile:
I think you can simply write like this, otherwise pending Enum Done interrupt will trigger the IRQ again:
```c
if (gintsts & GINTSTS_ENUMDNE) {
    // ENUMDNE is the end of reset where speed of the link is detected
    dwc2->gintsts = GINTSTS_ENUMDNE;
    // There may be a pending suspend event, so we clear it first
    dwc2->gintsts = GINTSTS_USBSUSP;
    dwc2->gintmsk |= GINTMSK_USBSUSPM;
    handle_enum_done(rhport);
  }
```

peppapighs:
> I think you can simply write like this, otherwise pending Enum Done interrupt will trigger the IRQ again:
> 
> ```c
> if (gintsts & GINTSTS_ENUMDNE) {
>     // ENUMDNE is the end of reset where speed of the link is detected
>     dwc2->gintsts = GINTSTS_ENUMDNE;
>     // There may be a pending suspend event, so we clear it first
>     dwc2->gintsts = GINTSTS_USBSUSP;
>     dwc2->gintmsk |= GINTMSK_USBSUSPM;
>     handle_enum_done(rhport);
>   }
> ```

Right, I wasn't aware you can clear multiple interrupts at once. I've made the change.

PR #3322, opened 2025-10-30
Bufferless (on the fly) video data generation for the video device class

embedded-ideas:
On smaller devices, there isn’t enough RAM to store an entire high-resolution frame. I haven’t found a way to use the video device without such a buffer, so I added a callback mechanism that allows the code to request payload data on the fly. This makes it possible to generate or forward data without storing a whole frame at once.

- The code has been tested on a NanoCH32V305 board.
- I added demo code to the video device example.
- I tried to stay consistent with the existing code and used weak linking for the callback.


HiFiPhile:
Thank you, tested on STM32U5, please fix pre-commit whitespace issue as I can't push to your branch (got permission denied). 
@kkitayam @lijunru-hub what do you think ?

embedded-ideas:
Sorry for the whitespaces, should have looked a bit deeper into the logs. Did some experiments with a new Editor and didn't had my settings right.

HiFiPhile:
Maybe @kkitayam @lijunru-hub are busy, if anything goes wrong we can fix it afterwards.

PR #3326, opened 2025-11-03
Fix code alerts

hathach:
fixed alerts found by pvs-studio

### DIFF
--- a/.PVS-Studio/.pvsconfig
+++ b/.PVS-Studio/.pvsconfig
@@ -1,3 +1,10 @@
-//-V::2506
-//-V::2514
-//-V::2614
+//V_EXCLUDE_PATH */iar/cxarm*
+//V_EXCLUDE_PATH */pico-sdk/*
+
+//-V::2506 MISRA. A function should have a single point of exit at the end.
+//-V::2514 MISRA. Unions should not be used.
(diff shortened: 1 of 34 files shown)


src/portable/synopsys/dwc2/dcd_dwc2.c
```diff
@@ -117,7 +117,7 @@ static void dma_setup_prepare(uint8_t rhport) {
   dwc2_regs_t* dwc2 = DWC2_REG(rhport);
 
   if (dwc2->gsnpsid >= DWC2_CORE_REV_3_00a) {
-    if(dwc2->epout[0].doepctl & DOEPCTL_EPENA) {
+    if(0 != (dwc2->epout[0].doepctl & DOEPCTL_EPENA)) {
```
HiFiPhile: @hathach I don't like adding a redundant `0 !=` into the condition, can we change the static analyzer behavior ?

I think it's better to change `uint32_t board_button_read(void);` to bool.

hathach: yeah, It is rather annoying, maybe we can suppress that rule. let me check

hathach: Some make sense to change to != 0 though like checking on number. For flag with and/or it is indeed unneccessary.

PR #3335, opened 2025-11-07
Prevent tu_edpt_number() from returning an invalid endpoint number

ldube:
**Describe the PR**
The endpoint number is 4 bits wide. The original code extracts 7 bits.
~TUSB_DIR_IN_MASK is 0x7F but we need to use 0x0F.
It is safer not to assume that bits 4..6 of _bEndpointAddress_ are zero.

**Additional context**
See the _bEndpointAddress_ description in the following link.

[EndpointDescriptors](https://beyondlogic.org/usbnutshell/usb5.shtml#EndpointDescriptors)

### DIFF
--- a/src/common/tusb_types.h
+++ b/src/common/tusb_types.h
@@ -103,6 +103,7 @@ typedef enum {
   TUSB_DIR_OUT = 0,
   TUSB_DIR_IN  = 1,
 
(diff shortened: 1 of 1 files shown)


HiFiPhile:
Thank you !

HiFiPhile (approved):

Issue #3281, opened 2025-10-05
index out of bounds


stneng:
### Operating System

Linux

### Commit SHA

6ce46da042eb9ab243e4127ddcac57519c0a226f

### Board

N/A

### Firmware

N/A

### What happened ?

Trace
```
/src/tinyusb/src/device/usbd.c:1294:56: runtime error: index 55 out of bounds for type 'uint8_t[5][2]' (aka 'unsigned char[5][2]')
SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior /src/tinyusb/src/device/usbd.c:1294:56 
```

### How to reproduce ?

Steps to reproduce

- Build oss-fuzz docker
Download files in this folder https://github.com/google/oss-fuzz/tree/master/projects/tinyusb
```
docker build -t cybergym-tinyusb .
docker run -it --rm -e FUZZING_LANGUAGE=c++ -e SANITIZER=undefined cybergym-tinyusb /bin/bash
```

- In docker container
```
compile
cd /out

wget https://github.com/user-attachments/files/22712844/poc.tgz -O poc.bin

./net poc.bin
```

[poc.bin](https://github.com/user-attachments/files/22712844/poc.tgz)

### Debug Log as txt file (LOG/CFG_TUSB_DEBUG=2)

N/A

### I have checked existing issues, discussion and documentation

- [x] I confirm I have checked existing issues, discussion and documentation.

HiFiPhile:
Should be fixed by #3579

PR #3336, opened 2025-11-08
Avoid out-of-bounds errors caused by invalid endpoint number

ldube:
**Describe the PR**
The endpoint number is used to index arrays of size CFG_TUD_ENDPPOINT_MAX. This PR adds a bounds check before the index is used.

**Additional context**
Maybe #3281 will be fixed

### DIFF
--- a/src/device/usbd.c
+++ b/src/device/usbd.c
@@ -143,6 +143,13 @@ typedef struct {
 tu_static usbd_device_t _usbd_dev;
 static volatile uint8_t _usbd_queued_setup;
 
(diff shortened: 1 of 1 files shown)


HiFiPhile:
Thank you but I don't think it's necessary. 

`CFG_TUD_ENDPPOINT_MAX` is already checked in `usbd_edpt_open` and `usbd_edpt_iso_alloc`, #3281 is due to erroneous test adapter sending invalid event which can only happens in stack overflow or RAM corruption.

PR #3351, opened 2025-11-20
Doc: Typo in docs/integration.rst

thamul:
Just a minor typo as I think. Should be port 1 in the second interrupt request handler.


hathach:
thank you

PR #3352, opened 2025-11-20
Use EP0 buffer directly if it's large enough 

HiFiPhile:
This pull request improves how control transfer buffers are managed for both the Audio and DFU device classes. The main enhancement is to allow the use of the endpoint 0 buffer for control transfers when it is large enough, falling back to a dedicated buffer only when necessary.


HiFiPhile:
Or just define a shared EP0 buffer large enough for all classes.

PR #3357, opened 2025-11-21
Fix in multiple example codes, wrong descriptor device qualifier length.

thamul:
I'm not quite sure if it's actually wrong. I found that in five example codes the size of `tusb_desc_device_t` is taken for the `tusb_desc_device_qualifier_t` length field. In multiple other example codes the size of the actual struct `tusb_desc_device_qualifier_t` is taken instead.
But as I said, I'm not sure if it has some other background.


HiFiPhile (approved):
Thank you, device qualifer's length is indeed 10 bytes.  ~~Although this descriptor does nothing actually~~
image

PR #3358, opened 2025-11-21
Add dcd_configure for dwc2 FIFO config

HiFiPhile:
**Describe the PR**
Add dcd_configure to configure dwc2 IN EP double buffer.

Although the gain is minimal as bandwidth limit is close (29MB/s vs 31MB/s). My PC's host controller only does 8 transfer per msof  which yields a maximum throughput of 32MB/s.

### DIFF
--- a/tools/metrics.py
+++ b/tools/metrics.py
@@ -283,11 +283,19 @@ def write_compare_markdown(comparison, path, sort_order='size'):
         else:
             (significant if is_significant(f) else minor).append(f)
 
-    def render_table(title, rows):
-        md_lines.append(f"## {title}")
+    def render_table(title, rows, collapsed=False):
+        if collapsed:
+            md_lines.append(f"<details><summary>{title}</summary>")
+            md_lines.append("")
+        else:
+            md_lines.append(f"## {title}")
+
         if not rows:
             md_lines.append("No entries.")
             md_lines.append("")
+            if collapsed:
+                md_lines.append("</details>")
+                md_lines.append("")
             return
 
         md_lines.append(header)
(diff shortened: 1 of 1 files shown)


src/portable/synopsys/dwc2/dcd_dwc2.c
```diff
@@ -492,9 +507,7 @@ bool dcd_init(uint8_t rhport, const tusb_rhport_init_t* rh_init) {
   // Enable required interrupts
   dwc2->gintmsk |= GINTMSK_OTGINT | GINTMSK_USBRST | GINTMSK_ENUMDNEM | GINTMSK_WUIM;
 
-  // TX FIFO empty level for interrupt is complete empty
   uint32_t gahbcfg = dwc2->gahbcfg;
-  gahbcfg |= GAHBCFG_TX_FIFO_EPMTY_LVL;
```
HiFiPhile, line 497: I think there is no need to set FIFO level to complete empty in any case ?

hathach: I think we use this to write more data to the tx fifo when doing multiple packets ?

hathach: PS: sorry, I forgot, set means complete empty, while default is half-empty. What if we have 2 endpoint 1 is doubles, 1 is single. And half-empty is trigger on the single while we cannot write a full packet to its fifo. I forgot that  if the interrupt trigger again when it is fully empty ? or it keeps interrupting until data is written (until transfer complete)

HiFiPhile: > And half-empty is trigger on the single while we cannot write a full packet to its fifo. 

I think it can't happen. The controller will read one whole packet each time so the fifo level will drop from 64 directly to 0. STM32Cube leave it by default.

hathach: ah, then it make senses. I keep forgetting thing anytime look back at dwc2

PR #3370, opened 2025-11-26
change(usbd): Changed return value to true for weak dcd_deinit() (quick fix)

roma-jam:
## Description

During the recent update (https://github.com/hathach/tinyusb/pull/3326)

`dcd_deinit(rhport);` -> `TU_VERIFY(dcd_deinit(rhport));`

the assertion fails for the ports, where `dcd_deinit()` isn't implemented yet. 

## Changes

This PR introduces a quick fix by changing the default return value from `false` to `true` in the WEAK function, allowing the assert to pass.

The same way, as we have it for: 
```
TU_ATTR_WEAK bool dcd_dcache_clean(const void* addr, uint32_t data_size) {
  (void) addr; (void) data_size;
  return true;
}

TU_ATTR_WEAK bool dcd_dcache_invalidate(const void* addr, uint32_t data_size) {
  (void) addr; (void) data_size;
  return true;
}

TU_ATTR_WEAK bool dcd_dcache_clean_invalidate(const void* addr, uint32_t data_size) {
  (void) addr; (void) data_size;
  return true;
}
```

### Context 

The real implementation may vary,  but for dcd_dwc2 (for example), this might still be not implemented, as we re-configure the controller fully during the following `dcd_init()`.

### DIFF
--- a/src/device/usbd.c
+++ b/src/device/usbd.c
@@ -86,7 +86,7 @@ TU_ATTR_WEAK bool tud_vendor_control_xfer_cb(uint8_t rhport, uint8_t stage, tusb
 
 TU_ATTR_WEAK bool dcd_deinit(uint8_t rhport) {
   (void) rhport;
-  return false;
+  return true;
 }
 
 TU_ATTR_WEAK void dcd_connect(uint8_t rhport) {


HiFiPhile:
I think we can take the chance to add `dcd_deinit` to dwc2, disable interrupt and disconnect may be enough ?

roma-jam:
Hi @HiFiPhile, 

> I think we can take the chance to add dcd_deinit to dwc2

Sure, that is my next step.

This PR is about another thing: the `dcd_deinit()` might be optional, as the dcd_init() could have all the necessary code to start from two states: `was init done` or `wasn't init done`.  
So in this case, if we have an `TU_ASSERT()` for the optional call, we need to make it pass even if the function isn't implemented.

HiFiPhile:
My concern is maybe other port require additional tweak to proper deinit, returning true will hide the issue. I prefer to make `TU_ASSERT(dcd_deinit(rhport));` in `usbd.c` to be safe and in dwc2:
```c
dcd_deinit(uint8_t rhport) {
  (void) rhport;
  // to be filled later
  return true;
}
```
@hathach What do you think ?

We also need to add `tu_varclr(&_usbd_dev);` to `tud_deinit`

roma-jam:
Addressed in https://github.com/hathach/tinyusb/pull/3378

PR #3373, opened 2025-11-27
update Vendor device to omit ep buffer for port with dedicated hwfifo

hathach:
- update vendor device to omit ep buf when dedicated hwfifo is supported
- change tud_vendor_rx_cb() behavior, buffer and bufsize only available when CFG_TUD_VENDOR_RX_BUFSIZE = 0
- add tud_vendor_n_read_discard(), refactor vendor device. Update webusb example for more robust
- add tu_fifo_discard_n() and tu_edpt_stream_discard()

### DIFF
--- a/src/class/vendor/vendor_device.c
+++ b/src/class/vendor/vendor_device.c
@@ -136,6 +136,15 @@ void tud_vendor_n_read_flush(uint8_t idx) {
 }
 #endif
 
+#if CFG_TUD_VENDOR_RX_MANUAL_XFER
+bool tud_vendor_n_read_xfer(uint8_t idx) {
+  TU_VERIFY(idx < CFG_TUD_VENDOR);
+  vendord_interface_t *p_itf = &_vendord_itf[idx];
+  return tu_edpt_stream_read_xfer(p_itf->rhport, &p_itf->stream.rx);
+}
+#endif
+
+
 //--------------------------------------------------------------------+
 // Write API
 //--------------------------------------------------------------------+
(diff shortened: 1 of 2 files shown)


src/class/vendor/vendor_device.c
```diff
@@ -247,38 +287,33 @@ uint16_t vendord_open(uint8_t rhport, const tusb_desc_interface_t* desc_itf, uin
...
+    // invoke callback
+  #if CFG_TUD_VENDOR_RX_BUFSIZE == 0
+    tud_vendor_rx_cb(idx, p_vendor->stream.rx.ep_buf, xferred_bytes);
+  #else
+    tud_vendor_rx_cb(idx, NULL, 0);
+  #endif
 
-    tu_edpt_stream_read_xfer(rhport, &p_vendor->rx.stream);
-  } else if ( ep_addr == p_vendor->tx.stream.ep_addr ) {
+    tu_edpt_stream_read_xfer(rhport, &p_vendor->stream.rx); // prepare next data
```
HiFiPhile, line 317: I think it's useful to add a config whether next transfer is scheduled automatically or manually later in `read_finished()`.

To give processing time for some commands.

hathach: yeah, we can have tud_vendor_configure() for manual read transfer including the intial xfer() when driver is open()

hathach: I add CFG_TUD_VENDOR_RX_MANUAL_XFER (default to 0) instead of dynamic configure, since I think this is only helpful with no-fifo RX.

HiFiPhile: Nice, but I think the logic is inversed in vendord_open and vendord_xfer_cb ?

hathach: ah right, sorry. It is late here already.

PR #3377, opened 2025-11-28
change armgcc setup to manual download due to issue with action

hathach:


hathach:
@copilot fix ci build issue with make-os: matrix: os = "windows-latest", arm-none-abi-gcc is added to GITHUB_PATH but not found when running build.py with make. Possibly the GITHUB_PATH is not the same as cmd.exe

PR #3378, opened 2025-11-28
Add DWC2 deinit support

HiFiPhile:
@roma-jam Hi, I've added deinit support to dwc2.

A dynamic switch example is added and tested on:
- ESP32-S3-DevKitC
- NUCLEO-H7S3L8
- Nucleo-U5A5ZJ-Q
- STM32F407-DISCO
- STM32F723E-DISCO (FS/HS)
- J-Link V9 (STM32F205 FS)


src/tusb.c
```diff
@@ -63,6 +63,16 @@ TU_ATTR_WEAK void* tusb_app_phys_to_virt(void *phys_addr) {
   return phys_addr;
 }
 
+TU_ATTR_WEAK void tusb_pre_init_cb(uint8_t rhport, tusb_role_t role) {
```
hathach: do we really need this, both init() and de-init() is called by application, they should call them before/after those accordingly ?

HiFiPhile: @hathach  It's for host/device specific init, like STM32's GCCFG and GOTGCTL register, which need to be set to different value on host/device. Their meaning is also different across families, I haven't found a easy way to include them in the driver.

https://github.com/hathach/tinyusb/blob/8a466fffa0133783d5e1b97f0b5da57d92d8167a/hw/bsp/stm32h7rs/family.c#L390

hathach: Give me a bit of time, I will try to review and see if we could avoid using these callbacks

HiFiPhile:
@hathach I've rebased on master.

src/common/tusb_fifo.h
```diff
@@ -289,6 +289,12 @@ TU_ATTR_ALWAYS_INLINE static inline bool tu_fifo_empty(const tu_fifo_t *f) {
   return wr_idx == rd_idx;
 }
 
+// Suppress IAR warning
+// Warning[Pa082]: undefined behavior: the order of volatile accesses is undefined in this statement
+#if defined(__ICCARM__)
+#pragma diag_suppress = Pa082
```
hathach, line 295: @HiFiPhile which compile flags to pass to IAR to enable most warning like gcc, we can enable it for tinyusb build as well. 

PS: ah, I just add --warnings-are-errors then try to fix them

src/portable/synopsys/dwc2/dwc2_common.c
```diff
@@ -247,6 +247,19 @@ bool dwc2_core_init(uint8_t rhport, bool is_hs_phy, bool is_dma) {
   return true;
 }
 
+void dwc2_core_deinit(uint8_t rhport) {
+  dwc2_regs_t* dwc2 = DWC2_REG(rhport);
+
+  // Soft disconnect
+  dwc2->dctl |= DCTL_SDIS;
```
hathach: `dwc2->dctl |= DCTL_SDIS` is device specific, it is also called via dcd_disconnect() already ?

HiFiPhile: Yeah it's redundant.

hathach:
@HiFiPhile I push changes that also add mcu specific phy deinit as well (suggested by claude). Please review to see if that makes sense to you. Also I haven't tested with the dynamic_switch example (getting lazy with wiring), please re-check if you already have it setup

PS: I think we can do better with low power/suspend mode by disable clock etc, but that is for later.

src/portable/synopsys/dwc2/dwc2_stm32.h
```diff
@@ -264,6 +264,20 @@ static inline void dwc2_phy_init(dwc2_regs_t* dwc2, uint8_t hs_phy_type) {
   }
 }
 
+// MCU specific PHY deinit, disable PHY power
+static inline void dwc2_phy_deinit(dwc2_regs_t* dwc2) {
+  // Disable on-chip FS PHY
+  dwc2->stm32_gccfg &= ~STM32_GCCFG_PWRDWN;
+
+  // Disable HS PHY if present
+  #ifdef USB_HS_PHYC
```
HiFiPhile: F723 has USB_HS_PHYC but only for HS port.

hathach: I guess we also need the hs_phy_type argument to de-init the right one ?

HiFiPhile: I think yes, also need to take care of when both HS port are used like on N6.

hathach: just update PR

HiFiPhile: I think we need to avoid turning off PHYC on N6 when another port is still active.

hathach: right N6 has 2 instance of USB1/2_HS_PHYC_NS. We can use the dwc2 check to disale the right phy. But we didn't enable it in dwc2_phy_init(). Not sure if N6 is currently working, I don't have the board to test with

HiFiPhile: I think it's okay in fact, didn't notice on N6 it has 2 dedicated PHYC instance.

N6 is working with my N6570-DK.

hathach: hmm, maybe the phy is auto enabled/disabled using dwc2 signal (physel, utmi) with latet stm config. Variant with USB_HS_PHYC seems to use stm32_gccfg to enable/disable phy. I guess that would be ok to merge now ?

HiFiPhile: The PHYC init code is in bsp currently:
https://github.com/hathach/tinyusb/blob/cd7bbba1ddc1ee8f6b4f896c98a5ec452b0953d6/hw/bsp/stm32n6/family.c#L184

It's okay to merge.

hathach: ah thanks, I think we should move it inside our driver that will allow us to disable phy to save power later on. But that can be on its own PR later.

PR #3017, opened 2025-03-08
mentor/dcd_musb - Fix issue reading and writing to USBFIFOn register

fcooperti:
For a single packet read and write operations must pick either 32-bit, 16-bit or 8-bit read or write operations. Mixing and matching can cause corruption of data.

This becomes a bigger issue when using TinyUSB fifo that has to deal with wrapping since read/write operations may be split between multiple transfers of different sizes.

Therefore, create a function that will determine which operation can be used for transfer the entire buffer of data. Then read and write operations can use a parameter to determine what should be used.

fixes hathach/tinyusb#3010

PR #3396, opened 2025-12-08
HWFIFO post fixup

HiFiPhile:
**Describe the PR**

- Add `CFG_TUD_EDPT_DEDICATED_HWFIFO` flag to other supported families.
- There is a weak `dcd_edpt_xfer_fifo()` implementation in usbd.c so no need to guard `usbd_edpt_xfer_fifo()` with `CFG_TUD_EDPT_DEDICATED_HWFIFO` flag, remove the ifdef guard makes the log more verbose when failed (was debugging why audio no longer work)
- In audio_device.c, replace local test by `CFG_TUD_EDPT_DEDICATED_HWFIFO`
- Remove assumption that FS MPS is 64 bytes, which leads to failed transfer if buffer size < 64 bytes

### DIFF
--- a/src/tusb_option.h
+++ b/src/tusb_option.h
@@ -356,12 +356,12 @@
 
 //------------ MUSB --------------//
 #if defined(TUP_USBIP_MUSB)
-  #define CFG_TUD_EDPT_DEDICATED_HWFIFO 1
+  #define CFG_TUD_EDPT_DEDICATED_HWFIFO 0 // need testing to enable
 #endif
 
 //------------ RUSB2 --------------//
 #if defined(TUP_USBIP_RUSB2)
-  #define CFG_TUD_EDPT_DEDICATED_HWFIFO 1
+  #define CFG_TUD_EDPT_DEDICATED_HWFIFO 0 // need testing to enable
 #endif
 
 //--------------------------------------------------------------------


hathach:
Hmm, seems like we have cdc echo issue with musb and rusb2 which just set the CFG_TUD_EDPT_DEDICATED_HWFIFO. @HiFiPhile did you test with these port, I guess we may miss a thing or two for them to work properly ?

HiFiPhile:
> Though I revert the changes with ifdef guard for usbd_xfer_fifo() since it save a bit of flash. Why would it cause the audio issue with logging though, without the CFG_TUD_EDPT_DEDICATED_HWFIFO, the function would return false anyway ?

It's fine to revert now, audio_device used different test condition for FIFO transfer, resulting `usbd_edpt_xfer_fifo` is called when `CFG_TUD_EDPT_DEDICATED_HWFIFO=0`  but failed silently without assert or breakpoint.

> seems like we have cdc echo issue with musb and rusb2

msub needs #3017, maybe rusb2 has the same access limitation ? I don't have board to test with.

hathach:
> It's fine to revert now, audio_device used different test condition for FIFO transfer, resulting `usbd_edpt_xfer_fifo` is called when `CFG_TUD_EDPT_DEDICATED_HWFIFO=0` but failed silently without assert or breakpoint.

ok, I will refactor usbd_xfer and usbd xfer_fifo later. Let revert this for now.

> > seems like we have cdc echo issue with musb and rusb2
> 
> msub needs #3017, maybe rusb2 has the same access limitation ? I don't have board to test with.

right, I guess we need more testing for these. There are quite some ports that support this hwfifo e.g rp2040. Let disable it for now, I will do more test with these ports later on and re-eanble them

PS: I also expect several bug fix regarding the migration as well. So it is better to only do it first with a couple ports :)

PR #3403, opened 2025-12-10
Add tud_vendor_n_write_zlp_if_needed to write zlp when TX FIFO disabled.

gogogoghost:
**Describe the PR**
When `CFG_TUD_VENDOR_TX_BUFSIZE` set to 0. User should write ZLP by themselves. But there is no way to write ZLP. This [check point](https://github.com/hathach/tinyusb/blob/0e48fe862292bb7b3e526a4b42bfa588d97d105e/src/tusb.c#L454) not supports ZLP. I think expose `tu_edpt_stream_write_zlp_if_needed` is a good way.

**Additional context**

### DIFF
--- a/docs/integration.rst
+++ b/docs/integration.rst
@@ -40,7 +40,7 @@ Minimal Example
 
      // init host stack on roothub port 1 for fullspeed host
      tusb_rhport_init_t host_init = {
-       .role  = TUSB_ROLE_DEVICE,
+       .role  = TUSB_ROLE_HOST,
        .speed = TUSB_SPEED_FULL
      };
      tusb_init(1, &host_init);
--- a/src/class/audio/audio_device.c
+++ b/src/class/audio/audio_device.c
@@ -65,40 +65,6 @@
 // MACRO CONSTANT TYPEDEF
 //--------------------------------------------------------------------+
 
-// Use ring buffer if it's available, some MCUs need extra RAM requirements
-// For DWC2 enable ring buffer will disable DMA (if available)
-#ifndef TUD_AUDIO_PREFER_RING_BUFFER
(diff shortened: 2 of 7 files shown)


gogogoghost:
The users can write ZLP like this:
```c
void tud_vendor_tx_cb(uint8_t itf, uint32_t sent_bytes) { 
    // other logic...
	if(sendFinished){
		tud_vendor_write_zlp_if_needed(sent_bytes);
	}
}
```

src/class/vendor/vendor_device.h
```diff
@@ -88,12 +88,17 @@ bool tud_vendor_n_read_xfer(uint8_t idx);
 // Write to TX FIFO. This can be buffered and not sent immediately unless buffered bytes >= USB endpoint size
 uint32_t tud_vendor_n_write(uint8_t idx, const void *buffer, uint32_t bufsize);
 
+
```
HiFiPhile: Thank you, could you remove extra spaces ?

hathach:
maybe we should update the write() to allow zlp instead, let me check that option first.

gogogoghost:
Remove this line should be OK?
https://github.com/hathach/tinyusb/blob/64820791516b4001cb169f095588ccb52eb35ea8/src/tusb.c#L453
Will this affect other logic?

I tried this method. It works fine.
```c
void tud_vendor_tx_cb(uint8_t itf, uint32_t sent_bytes) { 
    // other logic...
	if(sendFinished){
        if(sent_bytes&&(sent_bytes%MAX_PACKET_SIZE)==0){
            tud_vendor_write(NULL,0);
        }
	}
}
```

gogogoghost:
Or we can only check buffer length when FIFO enabled.

HiFiPhile:
Could you check https://github.com/hathach/tinyusb/tree/stream_write

gogogoghost:
Thank you! I will test it on my board on Monday.

PR #3406, opened 2025-12-11
[CT2] [EN-19729] Refactor CDC Write API to return bytes flushed, not buffered

nikhil-nayyar:


nikhil-nayyar:
many apologies, meant to open a PR against our forked repo, not the main repo :disappointed:

nikhil-nayyar:
Although, I've found it useful to have the write method return the number of bytes physically flushed, not just written to the internal CDC buffer. If anyone thinks it's worthwhile, I can try adding this feature as a new method and try opening another PR

hathach:
No I don't think that would be a good idea, application usually check `if (count == write(count)` and if the flushed bytes are less than what is written e.g write(128) and only flush 64, it will cause user confusion thinking that fifo is full.

PR #3411, opened 2025-12-13
tusb: fix stream write logic without fifo

HiFiPhile:
Move ` TU_VERIFY(bufsize > 0);` to fifo mode to allow sending ZLP.

`TU_VERIFY(s->ep_buf != NULL, 0);` is needed like as we don't know if the data is on stack or not.


hathach (approved):
perfect, thank you

PR #3422, opened 2025-12-21
Added tud_vendor_write_clear() which forcefully clears TX buffer

Tails86:
**Describe the PR**
I have run into an edge cast where the host application abruptly stops while there is data still within the vendor output buffer. I need to be able to manually clear this buffer after the application is restarted. This doesn't seem to be a perfect solution as there is still data in the endpoint buffer when the application starts reading. Please let me know if there is any way to fully solve this problem.

### DIFF
--- a/src/class/vendor/vendor_device.c
+++ b/src/class/vendor/vendor_device.c
@@ -161,12 +161,11 @@ uint32_t tud_vendor_n_write_available(uint8_t idx) {
   return tu_edpt_stream_write_available(&p_itf->stream.tx);
 }
 
-uint32_t tud_vendor_n_write_clear(uint8_t idx) {
+bool tud_vendor_n_write_clear(uint8_t idx) {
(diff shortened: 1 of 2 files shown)


HiFiPhile:
Hi, why did you modify tusb_fifo and tusb_private which are unnecessary?

Tails86:
I was unsure if it was possible for the fifo to be changed after the count was retrieved, so I had the count retrieval inside the `ff_lock()` block. I removed that and use `tu_edpt_stream_read_available()` instead.

HiFiPhile:
Could you just replicate cdc ? Since the data will be cleared the count won't worth much ? Anyway there are could be some data already been loaded into endpoint buffer.

https://github.com/hathach/tinyusb/blob/53f9f8618ee19b6c884a25293d860125c8c1a08b/src/class/cdc/cdc_device.c#L240

Tails86:
Done!

PR #3431, opened 2025-12-26
Fix compilation

ceedriic:
**Describe the PR**
A couple of very simple fixes for compilation errors/warnings I got after upgrading my project to tinyusb 0.20.0 release

### DIFF
--- a/src/class/cdc/cdc_host.c
+++ b/src/class/cdc/cdc_host.c
@@ -828,7 +828,9 @@ static bool set_line_state_on_enum(cdch_interface_t *p_cdc, tuh_xfer_t *xfer) {
     ENUM_SET_LINE_CONTROL,
     ENUM_SET_LINE_COMPLETE,
   };
+  #ifdef CFG_TUH_CDC_LINE_CONTROL_ON_ENUM
   const uint8_t idx = get_idx_by_ptr(p_cdc);
+  #endif
   const uintptr_t state = xfer->user_data;
 
   switch (state) {


src/common/tusb_debug.h
```diff
@@ -27,6 +27,8 @@
 #ifndef TUSB_DEBUG_H_
 #define TUSB_DEBUG_H_
 
+#include <stdio.h>
```
HiFiPhile, line 30: It's on purpose that stdio.h is only included when there is no custom printer defined.

src/common/tusb_mcu.h
```diff
@@ -245,13 +245,15 @@
...
 
 #elif TU_CHECK_MCU(OPT_MCU_STM32H7)
-  #include "stm32h7xx.h"
   #define TUP_USBIP_DWC2
   #define TUP_USBIP_DWC2_STM32
 
   #define TUP_DCD_ENDPOINT_MAX 9
 
-  #if __CORTEX_M == 7
+  #if defined(CORE_CM4) == defined(CORE_CM7)
```
HiFiPhile, line 253: Not all H7 devices are dual-core, in which case CORE_CM4/CORE_CM7 are not defined

src/common/tusb_debug.h
ceedriic: Oh I see, the miscompiling code has been removed after v0.20.0 in 7f4a761 - I will drop this commit.

src/common/tusb_mcu.h
ceedriic: Hmm. ok.
So would it be ok to just write:

```
  #ifndef CORE_CM4
    // Enable dcache if DMA is enabled
    #define CFG_TUD_MEM_DCACHE_ENABLE_DEFAULT     CFG_TUD_DWC2_DMA_ENABLE
    #define CFG_TUH_MEM_DCACHE_ENABLE_DEFAULT     CFG_TUH_DWC2_DMA_ENABLE
    #define CFG_TUSB_MEM_DCACHE_LINE_SIZE_DEFAULT 32
  #endif
```

HiFiPhile: I think #ifndef CORE_CM4 is ok.

ceedriic:
I will close this PR and create a new one, I'm no GIT expert...

ceedriic:
Replaced by PR #3433

PR #3432, opened 2025-12-26
Provides new tud_task_ext2() and tuh_task_ext2() entry points

ceedriic:
These new function variants allows the main application loop to limit how much time is spend on these functions, to ensure that other critical code can run fast enough.

**Describe the PR**
An application main loop (without os) that usually tinyusb host and device stack look like:

```
      while(1) { // the mainloop
        application_code();
        tuh_task(); // tinyusb host task
      }
``` 

Unfortunately, without patching, tud_task() and tuh_task() can take arbitrarily long, if the CPU is slow and USB traffic fast.
the other day I got tud_task() running many seconds mounting a FATFS partition using a msc device on slow full speed USB. 
Infinite loop can also happen in weird situations (bugs or stuff like usbh.c line 626).

This PR solves all of these by allowing the caller of these function to limit the number of events processed at once, ensuring the other stack and application_code run quickly. The leftover events will be processed on the next call.

### DIFF
--- a/src/class/cdc/cdc_host.c
+++ b/src/class/cdc/cdc_host.c
@@ -828,7 +828,9 @@ static bool set_line_state_on_enum(cdch_interface_t *p_cdc, tuh_xfer_t *xfer) {
     ENUM_SET_LINE_CONTROL,
     ENUM_SET_LINE_COMPLETE,
   };
+  #ifdef CFG_TUH_CDC_LINE_CODING_ON_ENUM
   const uint8_t idx = get_idx_by_ptr(p_cdc);
+  #endif
   const uintptr_t state = xfer->user_data;
 
   switch (state) {
--- a/src/common/tusb_mcu.h
+++ b/src/common/tusb_mcu.h
@@ -245,13 +245,12 @@
   #define CFG_TUSB_MEM_DCACHE_LINE_SIZE_DEFAULT 32
 
 #elif TU_CHECK_MCU(OPT_MCU_STM32H7)
-  #include "stm32h7xx.h"
   #define TUP_USBIP_DWC2
   #define TUP_USBIP_DWC2_STM32
 
   #define TUP_DCD_ENDPOINT_MAX 9
 
-  #if __CORTEX_M == 7
+  #ifndef CORE_CM4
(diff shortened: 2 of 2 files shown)


HiFiPhile:
I think this is just a poor man's solution, the number of events per cycle is not directly proportional to how long tud_task will run. 

For your MSC case the long latency is due to disk IO in blocked callback functions. The right way is using RTOS + non-blocking asynchronous IO like [cdc_msc_freertos](https://github.com/hathach/tinyusb/tree/master/examples/device/cdc_msc_freertos) example.

Without RTOS you can also use PendSV interrupt (or other free interrupts):
- Put `tud_task()`/`tuh_task()` in PendSV interrupt handler
- `tud_event_hook_cb()` will be called when there is a new event need to be processed, inside the callback you need to trigger PendSV interrupt 
- In `tud_msc_write10_cb()`/`tud_msc_read10_cb()` use a queue to post IO request to your main loop, return TUD_MSC_RET_ASYNC immediately
- In the main loop process the IO request and call `tud_msc_async_io_done()`

ceedriic:
> I think this is just a poor man's solution, the number of events per cycle is not directly proportional to how long tud_task will run.

It might be a poor man's solution, but it still is a solution to a real problem.

> For your MSC case the long latency is due to disk IO in blocked callback functions. The right way is using RTOS + non-blocking asynconous IO like [cdc_msc_freertos](https://github.com/hathach/tinyusb/tree/master/examples/device/cdc_msc_freertos) example.

Well, in my case, disk IO was pretty fast because it is a Ramdisk (I might be a poor man, but I'm not stupid enough to block on callbacks :)

I agree with you that RTOS is better for these, but I'm writing monitoring code on the M4 side (I'm using a rtos on the M7).

In any case TinyUSB claims to be working without RTOS - which is great - so I think this patch adresses a real problem.

ceedriic:
> I think we should have a max event per run such as CFG_TUD_TASK_EVENT_PER_RUN, I guess we don't need to make it dynamic since it is mostly fixed. That way we don't have to add a new tud_task_ext2()

OK, I will update the patch

ceedriic:
Closed in favour of #3478

PR #3439, opened 2025-12-28
device/mtp: queue ZLP when needed

### LINKED ISSUE #3430: MTP does not send ZLPs
### Operating System

Others

### Commit SHA

f6a77b87f04386ea27effb7c1faf8e1a33da0fa1

### Board

RP2040

### Firmware

based on `examples/device/mtp/` , nothing special in fact

### What happened ?

Some two weeks or so ago, I posted #3400 including a fix which was based on my lack of understanding on how USB transfers work. The problem I was observing, that the condition setting `is_complete=true` in `mtpd_xfer_cb()` would never be met a case where I sent 64 bytes of mtp payload, leading to a situation where `tud_mtp_data_complete_cb to be never called could -- seemingly -- be handled by #3401. However, this was only due to the gracefulness of Win 10's MTP client implementation. It failed, when I transmitted files having payload leading to the exact multiple of `bulk_mps` bytes. 

I eventually learned that ZLPs need to be sent always after such a transfer (both incoming and outgoing) has concluded its payload part. In fact, this is currently not accounted for by the current implementation of MTP (in `mtp_device.c`). Therfore, the original approach to identify completion was correct, however the ZLP was never sent. It could theoretically handed over to application's domain, but I would consider that neither correct nor architecturally desirable. I propose the following solution, referring to commit f6a77b87f04386ea27effb7c1faf8e1a33da0fa1:

https://github.com/elektroweber/fork-tinyusb/commit/c41fc220df336cf4d6541790a281d657ab8dd9ab

### How to reproduce ?

up/download files w/ exactly the size of 52+N64 (when using a bulk transfer size of 64 bytes), which gives, together with the header of the GetObject/SendObject Data phase, an integer multiple of 64.

N/A, see screenshot

### Screenshots

Example of transferring a payload with a missing ZLP, using the original implementation
Image

HiFiPhile:
FIx missing ZLP for IN/OUT data transfers.

Data size tested:
52 (zlp), 114, 116 (zlp), 500 (zlp), 578, 1012 (zlp)

Fix #3430

### DIFF
--- a/.github/workflows/ci_set_matrix.py
+++ b/.github/workflows/ci_set_matrix.py
@@ -38,7 +38,8 @@ family_list = {
     "stm32h7": ["arm-gcc", "arm-clang", "arm-iar"],
     "stm32h7rs stm32l0 stm32l4": ["arm-gcc", "arm-clang", "arm-iar"],
     "stm32n6": ["arm-gcc"],
-    "stm32u0 stm32u5 stm32wb stm32wba": ["arm-gcc", "arm-clang", "arm-iar"],
+    "stm32u0 stm32wb stm32wba": ["arm-gcc", "arm-clang", "arm-iar"],
+    "stm32u5": ["arm-gcc", "arm-clang", "arm-iar"],
     "-bespressif_s2_devkitc": ["esp-idf"],
     # S3, P4 will be built by hil test
     # "-bespressif_s3_devkitm": ["esp-idf"],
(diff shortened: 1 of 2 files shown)


elektroweber:
@HiFiPhile please see my comment here: https://github.com/hathach/tinyusb/issues/3430#issuecomment-3695038078

PR #3442, opened 2025-12-31
support for the AT32F45X series

zhiqiang-ch:
Support for the AT32F45X series.


zhiqiang-ch:
> thank you, I have to fixed pre-commit and compile issue since it lack define for SystemCoreClock. But all other change is good


Thank you very much. The definition of SystemCoreClock was missing in the official support package. I consulted them and they will add the definition later. Therefore, I did not specifically modify it. Of course, it works fine as it is now.

PR #3444, opened 2026-01-04
Implement Printer Device Driver

remiberthoz:
**Describe the PR**
Implement a printer device driver for USB Printers. The example illustrates receiving data from a host, the device driver as-is is recognized by the generic printer drivers on linux (and I assume, Windows too).

This allows sending binary data to the device, with no specific drivers, using tools like `lp -o raw`. 

**Additional context**
The example compiles and works on a Pi Pico.

I am not yet sure about integration of the changes in the build toolchain. I have not yet understood if I have to manually edit all the `hw/{bsp,mcu}/` file to include the device driver in cmake source lists.

### DIFF
--- a/examples/device/printer_to_cdc/src/main.c
+++ b/examples/device/printer_to_cdc/src/main.c
@@ -50,12 +50,13 @@
 
 // Forward data from Printer RX to CDC TX
 static void printer_to_cdc_task(void) {
-  if (tud_printer_read_available() == 0 || !tud_cdc_write_available()) {
+  uint32_t avail = tud_printer_read_available();
+  if (avail == 0 || !tud_cdc_write_available()) {
     return;
   }
 
   uint8_t buf[64];
-  uint32_t count = tud_printer_read(buf, sizeof(buf));
+  uint32_t count = tud_printer_read(buf, TU_MIN(sizeof(buf), tud_cdc_write_available()));
   if (count > 0) {
     tud_cdc_write(buf, count);
     tud_cdc_write_flush();
@@ -64,12 +65,13 @@ static void printer_to_cdc_task(void) {
 
 // Forward data from CDC RX to Printer TX
 static void cdc_to_printer_task(void) {
-  if (tud_cdc_available() == 0 || !tud_printer_write_available()) {
+  uint32_t avail = tud_printer_write_available();
+  if (tud_cdc_available() == 0 || avail == 0) {
     return;
   }
 
   uint8_t buf[64];
-  uint32_t count = tud_cdc_read(buf, sizeof(buf));
+  uint32_t count = tud_cdc_read(buf, TU_MIN(sizeof(buf), avail));
   if (count > 0) {
     tud_printer_write(buf, count);
     tud_printer_write_flush();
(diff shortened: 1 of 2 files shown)


remiberthoz:
Well I apparently made (at least) one major mistake when moving the code from my local project to the repository fork. The content of printer.h and printer_device.c is the same, leading to (part of) the compilation errors. I'll work on that and the other problems raised by the CI pipeline.

remiberthoz:
Ok, I'm realizing that I developed this locally against version 0.18.0 but tried to merge it in version 0.20.0. This is not the only problem but I am back on track. I'll see what was changed between the two releases and update the device driver accordingly.

remiberthoz:
In my terminal, compilation and upload of the `printer_to_hid` example now correctly yields :

    $ cat helloworld.ps  # see helloworld.ps content
      %!PS
      8 selectfont
      (Hello world!) show
    
    $ lp -d TinyUSB_Device -o raw ./helloworld.ps  # print helloworld.ps on the device
      request id is TinyUSB_Device-330 (1 file(s))
    
    $   PS 8 selectfont  Hello world   show  # auto-typed by HID

The last line is automatically typed by the HID device. Only alphanumeric characters are handled by the example program, but other characters could be translated to HID keycodes too. The printer endpoint receives UTF8.

----

CircleCI has caught one remaining `[[maybe_unused]]` flag that I'll replace with a `(void)` (in the [`one-random-make-msp430-large-msp430-gcc`](https://circleci.com/gh/hathach/tinyusb/81913) test). I believe the other failing tests are not directly related to the printer device class or its example, but to the build toolchain that is not correctly adjusted yet.

remiberthoz:
Looking forward to your feedback!

I believe I’ve addressed the issues in the sections of code I added. If there’s anything else you'd like me to revisit or any concerns with the implementation, please let me know. I’m happy to make further adjustments.

HiFiPhile:
Hi, thank you for adding printer driver !  I've just tested that it's working on Windows 11 with a manual driver install, I'll leave my reviews below.

remiberthoz:
Thank you, I am looking forward to feedback. The tests I could perform on Windows indeed reveal that detection of the printer is not 100% automatic, and there is a need to "install" the printer (same on linux) from the configuration menu. At least there is no need to use admin privileges, download or execute dependencies, which is sufficient for my use case.

In the interest of TinyUSB as a whole, however, it is interesting to see if that can be improved. I suppose the printer description, which is currently hardcoded here : https://github.com/hathach/tinyusb/pull/3444/files#diff-d649fb44fa8e4a9383e3d84eefb7efd336196be0c8dc66028dd949b45b531437R247 can be adjusted. Switching to a config-defined value is an obvious improvement.

I am also trying to understand what the printer device is supposed to respond to the host on the BULK IN endpoint.

hathach:
thank you very much, would it be better to have printer_to_cdc rather than hid, that way we can test printing huge file. If you are ok, I could make the change to migrate the example.

remiberthoz:
Wow! Thank you for working the extra steps here! That's a lot of improvements. I'm happy this is merged! I'll keep you updated if I find tricks for auto-installation on hosts or other improvements.

PR #3450, opened 2026-01-08
[Do Not Merge] Introduce DWC2_I_WANT_FS_ON_HS_BOARD hack

ceedriic:
**Describe the PR**
This is a proof-of-concept patch to support Full-Speed USB on DWC2 on a board with High-Speed ULPI

**Additional context**
I've a stm32h747 board with a high-speed external PHY.
But for testing/reliability purposes I want to limit the speed to Full-Speed.

I tried with:

`#define BOARD_TUH_MAX_SPEED  OPT_MODE_HIGH_SPEED`
`#define CFG_TUH_MAX_SPEED    OPT_MODE_FULL_SPEED`

In my `tusb_config.h`, but with bad results (hang in trying to setup the low-speed phy)

This patch is just a proof-of-concept that it is not too difficult to make that work, In the hope that someone can clean that up :)

### DIFF
--- a/src/portable/synopsys/dwc2/hcd_dwc2.c
+++ b/src/portable/synopsys/dwc2/hcd_dwc2.c
@@ -425,7 +425,7 @@ bool hcd_init(uint8_t rhport, const tusb_rhport_init_t* rh_init) {
 #endif
   while ((dwc2->gintsts & GINTSTS_CMOD) != GINTSTS_CMODE_HOST) {}
 
-#if DWC2_I_WANT_FS_ON_HS_BOARD
+#ifdef DWC2_I_WANT_FS_ON_HS_BOARD
   // disable high speed mode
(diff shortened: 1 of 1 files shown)


hathach:
why not use the built-in FS PHY instead ? does the mcu only has 1 phy, and that is used by the other usb controller ?

ceedriic:
> why not use the built-in FS PHY instead ? does the mcu only has 1 phy, and that is used by the other usb controller ?

You're right, I could use the built-in FS PHY of the MCU, but the board has been designed with the USB3315C external PHY, to be high-speed capable. So the built-in FS PHY pins are not connected. 

At the moment, for some reasons I'm trying to understand, high-speed devices do not enumerate on that board with the latest TinyUSB, but full-speed devices work well. So with this patch I'm able to run high-speed devices in full speed mode.

ceedriic:
Deleted in favour of #3494

PR #3457, opened 2026-01-12
Fix compiler error with c++ files and -Werror=extra

gabChouin:
**Describe the PR**
When compiling a c++ file that includes `tusb.h`, I get the following error:

`enumerated and non-enumerated type in conditional expression [-Werror=extra]`

Casting the enum values used for the comparison in `tu_edpt_addr` to `uint8_t` fixes the issue.

**Additional context**
We have `-Werror=extra` enabled on our project.


HiFiPhile (approved):
Thank you

PR #3459, opened 2026-01-13
Integrate MemBrowse.

michael-membrowse:
As discussed, this PR adds a MemBrowse integration to TinyUSB.

The integration is wired into the existing CI tests, so it can be used to collect both historical and ongoing memory footprint data.

The configured targets are set in **membrowse-targets.json**. (66 targets)

Run **membrowse-onboard.yml** manually to get historical data.
For an example I ran it from my fork to upload the last 60 commits, which are visible here:
https://membrowse.com/public/membrowse/tinyusb

Ongoing reporting works as follows:
**membrowse-report.yml** runs on every PR update and on pushes to main
It builds the configured targets, analyzes the resulting ELFs, and uploads the results to MemBrowse
A summary of memory changes is posted directly as a PR comment via membrowse-comment.yml
Please let me know if you would like anything adjusted or simplified.

PS: To set the MemBrowse Project Dashboard to be publicly accessible (like the link above), enable it in project settings.

### DIFF
--- a/.github/membrowse-targets.json
+++ b/.github/membrowse-targets.json
@@ -1,662 +1,494 @@
-[
-  {
-    "target_name": "at32f402_405-at_start_f402",
-    "port": "at32f402_405",
-    "board": "at_start_f402",
-    "setup_cmd": "NINJA_URL=https://github.com/ninja-build/ninja/releases/download/v1.13.1/ninja-linux.zip && wget -q $NINJA_URL -O ninja-linux.zip && unzip -q ninja-linux.zip -d $HOME/bin && echo \"$HOME/bin\" >> $GITHUB_PATH && TOOLCHAIN_URL=https://github.com/xpack-dev-tools/arm-none-eabi-gcc-xpack/releases/download/v14.2.1-1.1/xpack-arm-none-eabi-gcc-14.2.1-1.1-linux-x64.tar.gz && mkdir -p $HOME/toolchain && wget -q $TOOLCHAIN_URL -O toolchain.tar.gz && tar -C $HOME/toolchain -xf toolchain.tar.gz && echo \"$HOME/toolchain/xpack-arm-none-eabi-gcc-14.2.1-1.1/bin\" >> $GITHUB_PATH && python3 tools/get_deps.py at32f402_405",
-    "build_cmd": "python3 tools/build.py -s cmake -b at_start_f402",
-    "elf": "cmake-build/cmake-build-at_start_f402/device/cdc_msc/cdc_msc.elf",
-    "ld": "hw/mcu/artery/at32f402_405/libraries/cmsis/cm4/device_support/startup/gcc/linker/AT32F402xC_FLASH.ld",
-    "linker_vars": ""
-  },
-  {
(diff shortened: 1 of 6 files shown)


hathach:
thank you very much for your PR, is there any way we can use/upload the elf file that is built by current build.yml (we can upload the artifact and/or ld file) then download and push to membrowse. That would reduce the number of build job, currently it already takes quite a bit of  time to run ci (10 mins), adding 65 more jobs will hurt since my account can only run like 20 jobs concurrently. 

PS: I think we will need tons more targets since each board can have like 20 examples (elf).

michael-membrowse:
**Done.** MemBrowse now reuses the artifacts from the existing cmake job instead of building
  separately. Added upload-artifacts: true to the cmake job and membrowse downloads and analyzes
  those ELFs.

  **Also fixed a bug I noticed**: the build.yml workflow was running twice on every PR update (both push and
  pull_request events were triggering). Restricted push to only the master branch so PRs trigger once.
  Should save you half of the time.

michael-membrowse:
Just checking in in case this got buried. Happy to make any changes or split things up if that helps.

HiFiPhile:
Hi @membrowse , I think Ha Thach is busy, I'm trying to navigate symbols, is it normal that I got `Error Loading Symbols` on https://membrowse.com/public/membrowse/tinyusb/target/tm4c-ek_tm4c123gxl/symbols?base=71ed9d1498e04f99f69fc8c0336ca41ec7a576af&head=5ad42064ea544080d47c893dafb1b41447ab9f32 ?

michael-membrowse:
Hi @HiFiPhile,
Please try go to the link again. 
The issue was that only one commit was uploaded to the system and it didnt find the parent commit report to compare to.
I will add a better error message to the system to make it clearer. 

I've uploaded more reports using the membrowse-onboard.yml workflow so now the comparison works.

After merging this pull request, i suggest to run the provided membrowse-onboard.yml github workflow as well at least on the last 20 commits (its a parameter of the workflow) so you will have historical information right away.

HiFiPhile:
Thank you, it's a really nice tool. While cdc_msc example has a good coverage, it looks like if we enable analyze for other examples `.github/membrowse-targets.json` will grow pretty large, is it possible to reorganize more efficiently ?

michael-membrowse:
Thanks for your comment. You are right. I've restructured it to be more easily maintainable.

michael-membrowse:
Fixed the flow failures, and verified again that everything is working

hathach:
perfect, thank you @HiFiPhile for reviewing. Specially thanks to @membrowse very much for making the PR. membrowse is super useful for analyzing size, which we are trying to do better with TinyuSB recently.

@dhalbert @tannewt maybe we can integrate this to circuitpython as well.

michael-membrowse:
@hathach 
Thanks! Glad you find it useful.

Ill be happy to help integrating it to CircuitPython.

I already made a PR for Micropython:
https://github.com/micropython/micropython/pull/18558

I also suggest you run the "Onboard to Membrowse" github action once, so you'll have some historical data on MemBrowse dashboard

PR #3461, opened 2026-01-14
remove no-fifo mode in edpt stream API

hathach:
- remove no-fifo mode in edpt stream API
- update vendor_device.c for direct usbd xfer when either RX/TX BUFSIZE is 0 i.e CFG_TUD_VENDOR_TXRX_BUFFERED = 0

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -315,7 +315,7 @@ jobs:
         run: python3 tools/get_deps.py $BUILD_ARGS
 
       - name: Build
-        run: python3 tools/build.py -j 4 --toolchain iar $BUILD_ARGS
+        run: python3 tools/build.py --toolchain iar $BUILD_ARGS
 
       - name: Test on actual hardware (hardware in the loop)
         run: python3 test/hil/hil_test.py hfp.json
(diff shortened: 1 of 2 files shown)


GCRev:
🙌 I was just struggling with this on a project and now I'm unblocked!

hathach:
> 🙌 I was just struggling with this on a project and now I'm unblocked!

glad this help, what is your issue previously.

GCRev:
@hathach I was using no-fifo mode previously in order to get the buffers directly in my callback handlers rather than reading and writing to and from the FIFO. 

IIRC before this change, in that mode there was a logic bug in the vendor_n_write code that was causing the endpoint transfers to be skipped completely, so no messages were ever getting sent out from the device. On the host side, this was causing my "requests" (a write then a read) to the device to time out because nothing reached the host.

https://github.com/hathach/tinyusb/blob/a71cb23647d3eafd3f34b3641fd17ea49b131aec/src/class/vendor/vendor_device.c#L145-L149 This was called when `CFG_TUD_VENDOR_TX_BUFSIZE` was set to 0

Also, when `CFG_TUD_VENDOR_TX_BUFSIZE` was set to 0, https://github.com/hathach/tinyusb/blob/a71cb23647d3eafd3f34b3641fd17ea49b131aec/src/class/vendor/vendor_device.c#L205-L207 this would set the `tx_ff_buf` FIFO to `NULL`

In combination, the `vendor_n_write` -> `tu_edpt_stream_write` caused problems because that FIFO was NULL but the rest of the code didn't provision for it to be NULL.

hathach:
thank you for the detail, yeah, I guess sometimes the unified API call (by integating too many modes) are worth it.

PR #3471, opened 2026-01-26
device/msc: skip command stage if EP out is stalled

HiFiPhile:
**Describe the PR**

Fix `usbd_edpt_xfer` assert when `tud_msc_write10_cb` returned `TUD_MSC_RET_ERROR`.


hathach (approved):
perfect thank you. I move the check to the only place it is viable (after status sent). The other place is open() and recovery should be fine.

PR #3472, opened 2026-01-26
fix stm32u0 data stride

HiFiPhile:
**Describe the PR**
STM32U0 has 1024 byte PMA and 32bit access width.


hathach:
> **Describe the PR** STM32U0 has 1024 byte PMA and 32bit access width.

ah thanks, look like the access scheme is flexible for 1KB ram

Issue #3426, opened 2025-12-24
STM32N6 can't open config "CFG_TUD_DWC2_DMA_ENABLE"


DreamLand7707:
### Operating System

Linux

### Commit SHA

ALL

### Board

STM32N6

### Firmware

tinyusb/src/portable/synopsys/dwc2/dwc2_stm32.h

### What happened ?

When I tried to open DMA, this error occurred. The specific code is:
```c
static mem_region_t uncached_regions[] = {
  // DTCM (although USB DMA can't transfer to/from DTCM)
#if CFG_TUSB_MCU == OPT_MCU_STM32H7
  {.start = 0x20000000, .end = 0x2001FFFF},
#elif CFG_TUSB_MCU == OPT_MCU_STM32H7RS
  // DTCM (although USB DMA can't transfer to/from DTCM)
  {.start = 0x20000000, .end = 0x2002FFFF},
#elif CFG_TUSB_MCU == OPT_MCU_STM32F7
    // DTCM
  {.start = 0x20000000, .end = 0x2000FFFF},
#else
#error "Cache maintenance is not supported yet"
#endif
  CFG_DWC2_MEM_UNCACHED_REGIONS
};
```

However, there are indeed certain addresses in STM32N6 that cannot be accessed by USB DMA. Therefore, no error should be reported here.

### How to reproduce ?

in src/portable/synopsys/dwc2/dwc2_stm32.h

#define CFG_TUD_DWC2_DMA_ENABLE 1
#define CFG_TUSB_MCU OPT_MCU_STM32N6

### Debug Log as txt file (LOG/CFG_TUSB_DEBUG=2)

No debugging logs were found.
Only the aforementioned compilation errors occurred.

### I have checked existing issues, discussion and documentation

- [x] I confirm I have checked existing issues, discussion and documentation.

HiFiPhile:
I don't have stm32n6 board, you are welcomed to open a PR to add support.

HiFiPhile:
Fixed by #3476

PR #3476, opened 2026-01-30
DWC2 Vbus sensing enhancement

### LINKED ISSUE #1476: No DCD_EVENT_UNPLUGGED event for STM32 FSdev
### Operating System

Windows 10

### Board

Custom based on STM32H7B3

### Firmware

portable/st/stm32_fsdev/dcd_stm32_fsdev.c

### What happened ?

Dear,

I have followed the example for CDC/MSC using FreeRTOS and have got pretty much everything working. Expect for the disconnection event: unmount callback (tud_umount_cb) is not being called for this platform and driver. Mount callback (tud_mount_cb) happens as expected.

I am new to this library, so maybe I have messed up something, but as everything else is working, I guess I maybe facing a bug somewhere. Looking into the code, the old synopsys driver evokes the DCD_EVENT_UNPLUGGED event, but the new stm32_fsdev does not.

Any help would be much appreciated.

### How to reproduce ?

Just connect the board using the new stm32_fsdev and look for the mount and unmount callbacks.

### Debug Log as txt file

_No response_

### Screenshots

_No response_

HiFiPhile:
**Describe the PR**
- Add VBus sensing option to `tud_configure()`, for STM32 `GCCFG` register config is now included in the stack, which reduces BSP compicity.
- Also turn off device mode Vbus output in F4/F7's BSP to fix disconnection detection.
- Add Vbus detection support to Nucleo-U5A5AZJ which uses ADC to read Vbus voltage.
- Also fix N6 BSP to make both ports work

Other device families should not be infected as `GOTGCFG` default value stays the same.

Device and host modes are tested and functional:
| Board | Type | Disconnection detection |
|--------|--------|--------|
| F407-DISCO | Internal FS | OK |
| F723E-DISCO | Internal FS / Internal ULPI HS | Both ports OK [1](#1) |
| H747I-DISCO | External ULPI HS | OK |
| Nucleo-H7S3L8 | Internal UTMI HS | NO Vbus connection |
| Nucleo-U5A5AZJ | Internal UTMI HS | OK | 
| N6570DK | Internal UTMI HS | NO Vbus connection |

1: Needs 104cf33239eda282b34c1f7971fbfc8e310eabf8

Should fix #1476, #3426

**Additional context**
No more callback is needed for #3378

### DIFF
--- a/hw/bsp/stm32f2/family.c
+++ b/hw/bsp/stm32f2/family.c
@@ -106,10 +106,8 @@ void board_init(void) {
 
 #if CFG_TUD_ENABLED
   // Enable VBUS sense (B device) via pin PA9
-  tud_configure_dwc2_t cfg = {
-    .bm_double_buffered = 0,
-    .vbus_sensing = true
-  };
+  tud_configure_dwc2_t cfg = CFG_TUD_CONFIGURE_DWC2_DEFAULT;
+  cfg.vbus_sensing = true;
   tud_configure(0, TUD_CFGID_DWC2, &cfg);
 #endif
 }
--- a/hw/bsp/stm32f4/family.c
+++ b/hw/bsp/stm32f4/family.c
@@ -180,10 +180,8 @@ void board_init(void) {
 #endif
 
 #if CFG_TUD_ENABLED
-  tud_configure_dwc2_t cfg = {
-    .bm_double_buffered = 0,
-    .vbus_sensing = VBUS_SENSE_EN
-  };
+  tud_configure_dwc2_t cfg = CFG_TUD_CONFIGURE_DWC2_DEFAULT;
+  cfg.vbus_sensing = VBUS_SENSE_EN;
   tud_configure(BOARD_TUD_RHPORT, TUD_CFGID_DWC2, &cfg);
   board_vbus_set(BOARD_TUD_RHPORT, false);
 #endif
--- a/hw/bsp/stm32f7/family.c
+++ b/hw/bsp/stm32f7/family.c
@@ -157,10 +157,8 @@ void board_init(void) {
 #endif // vbus sense
 
 #if CFG_TUD_ENABLED && BOARD_TUD_RHPORT == 0
-  tud_configure_dwc2_t cfg = {
-    .bm_double_buffered = 0,
-    .vbus_sensing = OTG_FS_VBUS_SENSE
-  };
+  tud_configure_dwc2_t cfg = CFG_TUD_CONFIGURE_DWC2_DEFAULT;
+  cfg.vbus_sensing = OTG_FS_VBUS_SENSE;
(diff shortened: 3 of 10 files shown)


HiFiPhile:
@brentmellor  Could you check if Vbus detection works on Nucleo-U575 ?

brentmellor:
@HiFiPhile 
I'll try to check this code sometime this week. I modified the development board trying to get VBUS detection working natively and so I'll need to put things back to original state.

src/portable/synopsys/dwc2/dwc2_stm32.h
```diff
@@ -300,6 +303,136 @@ static inline void dwc2_phy_update(dwc2_regs_t* dwc2, uint8_t hs_phy_type) {
...
+static inline void dwc2_stm32_gccfg_cfg(dwc2_regs_t* dwc2, bool vbus_sensing, bool is_host) {
+  if (is_host) {
+    vbus_sensing = false;
+  }
+#if CFG_TUSB_MCU == OPT_MCU_STM32F1
+  // F1: Basic FS-only core, no VBUS sensing support
+  // Only PWRDWN bit is used (set in dwc2_phy_init)
+  (void) vbus_sensing;
+
+#elif CFG_TUSB_MCU == OPT_MCU_STM32F2 || CFG_TUSB_MCU == OPT_MCU_STM32F4
```
hathach: f412 (3.20a) has VBDEN, I think we should use the dwc2 core version instead. I am making the change locally, trying to wrap this up before TET.

HiFiPhile: I've thought about that, but I don't have all boards to ensure the correspondence between core version and ST's customization.

hathach: I have changed to use the GUID instead, which is ST internal dwc2 configuration I guess. I don't have all the boards as well, we will rely on the dwc2_info.py/md. And implement the best we could. Anything else would require PRs.

src/portable/synopsys/dwc2/dwc2_stm32.h
```diff
@@ -300,6 +303,66 @@ static inline void dwc2_phy_update(dwc2_regs_t* dwc2, uint8_t hs_phy_type) {
   }
 }
 
+//------------- GCCFG configuration -------------//
+static inline void dwc2_stm32_gccfg_cfg(dwc2_regs_t* dwc2, bool vbus_sensing, bool is_host) {
```
hathach, line 307: @HiFiPhile I change the function to work based on the GUID. You could check the dwc2_info.md for reference. We can update the logic if needed.

HiFiPhile: Looks good to me !

PR #3478, opened 2026-02-01
Limit events processed by tud_task_ext() / tuh_task_ext()

ceedriic:
**Describe the PR**
Add CFG_TUD_TASK_EVENTS_PER_RUN and CFG_TUH_TASK_EVENTS_PER_RUN options to limit the number of events processed in a single call to tud_task_ext() or tuh_task_ext().

This is a simplified replacement for pull request #3432 after [hathach](https://github.com/hathach) last comment.

**Additional context**

An application main loop (without os) with tinyusb host and device stack look like:

```
      while(1) { // the mainloop
        application_code();
        tuh_task(); // tinyusb host task
      }
```

Unfortunately, without patching, tud_task() and tuh_task() can take arbitrarily long, if the CPU is slow and USB traffic fast.
the other day I got tud_task() running many seconds mounting a FATFS partition using a msc device on slow full speed USB.
Infinite loop can also happen in weird situations (bugs or stuff like usbh.c line 626).

This PR solves all of these by adding two new simple configuration options:

```
#define CFG_TUD_TASK_EVENTS_PER_RUN  16
#define CFG_TUH_TASK_EVENTS_PER_RUN  16
```

Once this limit is reach, tud_task() or tuh_task() exit, ensuring the other stack and application_code run quickly. The leftover events will be processed on the next call.

The default value of 16 is high enough that it should not be reached in healthy situations. A value of 0 disable the feature and go back to infinite number of events processed by call.

### DIFF
--- a/src/device/usbd.c
+++ b/src/device/usbd.c
@@ -671,6 +671,7 @@ void tud_task_ext(uint32_t timeout_ms, bool in_isr) {
 #if CFG_TUD_TASK_EVENTS_PER_RUN > 0
     if (epr >= CFG_TUD_TASK_EVENTS_PER_RUN) {
       TU_LOG_USBD("USBD event limit (" TU_XSTRING(CFG_TUD_TASK_EVENTS_PER_RUN) ") reached\r\n");
+      break;
     }
 #endif
     dcd_event_t event;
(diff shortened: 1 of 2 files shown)


ceedriic:
Please wait, I'm retesting everything with the missing breaks...

PR #3482, opened 2026-02-04
Fully asynchrounous host code

ceedriic:
**Describe the PR**
This remove all calls to tusb_time_delay_ms_api() from inside tuh_task_ext().

**Additional context**
Use of the host API was not really possible with OPT_OS_NONE due to blocking calls in tuh_task_ext().

With an OS, it will also improve the situation because:

1. The timeout_ms parameter of tuh_task_ext() is now respected.
2. The traffic with device 1 will not stop when device 2 is plugged.

I've split this in 3 main patches (the last 6 are fixes for warnings discovered after generating the PR)

1. Long boring preparation patch to split the functions calling tusb_time_delay_ms_api().
3. Executing the continuation functions asynchronously.
4. cleaning up the handling of deferred attachments with a separate queue (note: I've not really tested that last patch because with my hub and 2 devices for some reason I never got into this situation)

### DIFF
--- a/src/tusb_option.h
+++ b/src/tusb_option.h
@@ -692,8 +692,7 @@
 // use tusb_time_millis_api() instead of tusb_time_delay_ms_api() in tuh_task()
 // tuh_task_ext() will be asynchronous and never sleep in tusb_time_delay_ms_api()
 #ifndef CFG_TUH_TASK_USE_TIME_MILLIS_API
-  #if CFG_TUSB_OS == OPT_OS_RTX4 || CFG_TUSB_OS == OPT_OS_PICO || defined(ESP_PLATFORM) || \
-     (CFG_TUSB_OS == OPT_OS_NONE && (TUSB_MCU_VENDOR_ESPRESSIF || CFG_TUSB_MCU == OPT_MCU_RP2040))
+  #if CFG_TUSB_OS == OPT_OS_RTX4 || CFG_TUSB_OS == OPT_OS_PICO || defined(ESP_PLATFORM)
     // these boards/os do not implements the required tusb_time_millis_api()
     #define CFG_TUH_TASK_USE_TIME_MILLIS_API  0
   #else


ceedriic:
How to handle the few platforms missing tusb_time_millis_api()? 
Is there a define that indicates that this function is available?
Is there really platforms that wants the HOST API but cannot implement tusb_time_millis_api()?

HiFiPhile:
> How to handle the few platforms missing tusb_time_millis_api()?

I see you can't wait longer :)

I've asked Claude if OSAL ports has millis API:

✅ Already supported:


- OPT_OS_FREERTOS: Uses xTaskGetTickCount()
- OPT_OS_MYNEWT: Uses os_time_ticks_to_ms32(os_time_get()
- OPT_OS_PICO: Uses to_ms_since_boot(get_absolute_time()
- OPT_OS_RTTHREAD: Uses rt_tick_get() converted to milliseconds
- OPT_OS_ZEPHYR: Uses k_uptime_get_32()
- OPT_OS_CUSTOM: Requires user implementation

❌ Not yet supported in board_api.h:
- OPT_OS_RTX4: Would need to add time API (RTL.h doesn't expose millisecond counter directly)
Recommendation

We can add a `osal_millis_get` for those who support.

---

Another challenge is control transfer and buffer management.  To save ressource USBH only manage one control transfer  (composed by setup+data+status packets) each time. 
https://github.com/hathach/tinyusb/blob/707fb5244103c5d28f50f26e779c456675c6ee35/src/host/usbh.c#L729

With the state managed here:
https://github.com/hathach/tinyusb/blob/707fb5244103c5d28f50f26e779c456675c6ee35/src/host/usbh.c#L191

And the buffer, which is also exposed to class driver as temporary buffer by `usbh_get_enum_buf()`
https://github.com/hathach/tinyusb/blob/707fb5244103c5d28f50f26e779c456675c6ee35/src/host/usbh.c#L202

ceedriic:
Ok, now all compilation checks pass.

> > How to handle the few platforms missing tusb_time_millis_api()?
> 
> I see you can't wait longer :)

Hey, I've customers waiting :)

Regarding tusb_time_millis_api(), I've added a configuration option, CFG_TUH_TASK_USE_TIME_MILLIS_API, which is enabled by default except for OPT_OS_RTX4 and OPT_OS_PICO. I've also disabled the feature for expressif.

For OPT_OS_PICO this is weird because in ./hw/bsp/board_api.h, board_millis() is defined, but I think ./hw/bsp/board.c, which implements tusb_time_millis_api() by calling board_millis(), is simply not compiled for some platforms (rp2040, expressif in particular).

I'll look at the control transfer issue after lunch.

ceedriic:
> Another challenge is control transfer and buffer management. To save ressource USBH only manage one control transfer (composed by setup+data+status packets) each time.

Ok, I've looked at tuh_control_xfer() and I think I see some of the issues. here are a few of my observations:

1) tuh_control_xfer() without a callback can recurse into tuh_task(). I was scared when I saw that but fortunately tuh_control_xfer() is never called without callback in the enumeration code or anywhere in my app. However, I believe I will need to update the patch to make sur tuh_task() does not wake up timers inside such a recursive path.

2) As you mentionned, ctrl_xfer_info can be owned by class driver code, which means the enumeration will fail if we're unlucky and if device 2 start an enumeration when device 1 set the DTR or something. The whole enumeration will forever fail, because:
      2.1) The return value of tuh_control_xfer() is not handled by enumaration code (ignored or ASSERTed).
      2.2) tuh_control_xfer() does not call the callback if it cannot acquire the buffer, therefore we do not get to the 3x retry code.
      2.3) there is no timeouts anywhere in the enumeration code.

3) In the same way, application code for Device 1 could fail unexpectedly if an enumeration from Device 2 use the buffer needed by, say, acm_set_line_coding().

4) Worse, in the case of acm_set_line_coding(), the failed call will also corrupt the enumeration buffer, and likely trash the enumeration process, because it writes to enum_buf outside of the protection of the ctrl_info->stage state check and the mutex.

5) All of this means that the current code is fine with one device (which is my immediate need, luckily) but need some overhaul / refactoring / improvements to work well with multiple devices.

6) None of the above is related to this PR (except the first point which I will fix) and can be handled later by further improvements / PRs. I'd be happy to help for the parts that I understand.

Am I correct? do we agree on all that?

ceedriic:
My goodness, is there a way to silence this "advanced" security bot?

hathach:
> My goodness, is there a way to silence this "advanced" security bot?

that is OK, just ignore it if it does not make sense. It is static analyzer warning from PVS-studio and/or other tool, but I currently configure some of the tusb config (this thing take lots of time to tweak and tune-up) therefore code may be mistaken as unreachable. I got annoyed by it as well, but we do review with a pair of human eye.

HiFiPhile:
> Ok, I've looked at tuh_control_xfer() and I think I see some of the issues. here are a few of my observations:

1 . Yes
2.1/2.2 Yes, currently there is no retry if tuh_control_xfer failed early. 
2.3 To support timeout we need to modify HCD drivers to manage devices who sends NAK but no transfer error. (Windows use 20s timeout)
3. Yes
4. Oh, I saw it....
5. I tried to do something last week but realized that to have full concurrency we may need to add transfer status and buffer to each devices, it's a hard decision to make for smaller MCUs ...
6. I think you are right

@hathach what do you think ?

ceedriic:
> 5. I tried to do something last week but realized that to have full concurrency we may need to add transfer status and buffer to each devices, it's a hard decision to make for smaller MCUs ...

Reading the code, adding a transfer status to each device for use after enumeration was also my first idea (but keeping a global transfer status and buffer global for enumeration).

For the buffer, acm_set_line_coding() and friends could maybe just put the small buffers they need right into the CDC private structure, or worst case use malloc/free instead of the shared buffer when the system is configured with > 1 device.

Honestly, I can see Logitech selling a mouse with TinyUSB and a 8051... but I don't think many people would want to use TinyUSB as host with a "smaller MCU", especially with multiple devices....

ceedriic:
Deleted in favour of #3490 (same final code)

PR #3483, opened 2026-02-06
Skip Zephyr and Membrowse comment temporarily

hathach:
Summary
- Deleted .github/pull_request_template.md.
- Updated .github/workflows/build.yml:
  - Simplified workflows by removing check-paths job dependencies and code path validation.
  - Adjusted conditions for jobs like hil-build, hil-tinyusb, and hil-hfp.
  - Marked specific workflows (e.g., zephyr) to be skipped explicitly using if: false or similar.
 - Updated family_support.cmake:
  - Added FOLDER properties for various post-build and flashing targets, improving organization and project structure.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -4,9 +4,33 @@ on:
   workflow_dispatch:
   push:
     branches: [master]
+    paths:
+      - 'src/**'
+      - 'examples/**'
+      - 'lib/**'
+      - 'hw/**'
+      - 'tools/build.py'
+      - 'tools/get_deps.py'
+      - '.github/actions/**'
+      - '.github/workflows/build.yml'
+      - '.github/workflows/build_util.yml'
+      - '.github/workflows/ci_set_matrix.py'
   pull_request:
+    paths:
+      - 'src/**'
+      - 'examples/**'
+      - 'lib/**'
+      - 'hw/**'
+      - 'test/hil/**'
+      - 'tools/build.py'
+      - 'tools/get_deps.py'
+      - '.github/actions/**'
+      - '.github/workflows/build.yml'
+      - '.github/workflows/build_util.yml'
+      - '.github/workflows/ci_set_matrix.py'
   release:
     types: [ published ]
+
 concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}
   cancel-in-progress: true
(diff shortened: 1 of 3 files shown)


.github/workflows/build.yml
```diff
@@ -4,9 +4,33 @@ on:
...
+      - 'src/**'
+      - 'examples/**'
+      - 'lib/**'
+      - 'hw/**'
+      - 'test/hil/**'
+      - 'tools/build.py'
+      - 'tools/get_deps.py'
+      - '.github/actions/**'
+      - '.github/workflows/build.yml'
+      - '.github/workflows/build_util.yml'
```
michael-membrowse, line 29: The flow is entirely skipped if the changes are not affecting the build.
This will cause MemBrowse not to register those commits and break the commit chain.
(The history view will not be correct).
Before this change, the build itself was skipped, but the flow still sent the data to MemBrowse so the commit chain would stay intact.

.github/workflows/build.yml
```diff
@@ -347,28 +325,19 @@ jobs:
...
     secrets: inherit
 
   membrowse-comment:
-    needs: [check-paths, membrowse]
-    if: >
-      always() &&
-      github.event_name == 'pull_request' &&
-      needs.check-paths.outputs.code_changed == 'true'
+    needs: membrowse
+    # skip membrowse comment since it is too verbal
```
michael-membrowse, line 339: We can improve that. Maybe show just a subset of targets? Or mention just the affected targets?

hathach: yeah, I am not entirely sure, I am only interested in the file changes in tinyusb src, instead of the whole target which also pull in other code such vendor hal, mcu and 3rd party lib. 

Also I break it out into > 1%, < 1% and not changed (which is hidden by default). However, my needs is a bit specific. I think your current implementation work for most end-user exe target. I am still not familliar with membrowse yet, but if it is posible to have pr comment from a (md) file, it would be great for user to extract/customize the pr report. 

https://github.com/hathach/tinyusb/pull/3451#issuecomment-3728594295

Note: my home-build metrics does size average for all targets (same object file) so the number is not exact but work best since it cover large configuration options

michael-membrowse: Ok, thanks for the feedback, I think I can make it work. I'll open a PR.

PR #3487, opened 2026-02-08
fix transfer length when high speed capable device/host working at full speed

HiFiPhile:
When high speed capable device/host working at full speed, transfer length is not updated accordingly, leading to ZLP management issue. 
Use speed specific transfer length instead of compile time macro.

### DIFF
--- a/src/class/vendor/vendor_device.h
+++ b/src/class/vendor/vendor_device.h
@@ -54,12 +54,12 @@ extern "C" {
 
 // RX FIFO can be disabled by setting this value to 0
 #ifndef CFG_TUD_VENDOR_RX_BUFSIZE
-  #define CFG_TUD_VENDOR_RX_BUFSIZE 64
+  #define CFG_TUD_VENDOR_RX_BUFSIZE TUD_EPSIZE_BULK_MAX
 #endif
 
 // TX FIFO can be disabled by setting this value to 0
 #ifndef CFG_TUD_VENDOR_TX_BUFSIZE
-  #define CFG_TUD_VENDOR_TX_BUFSIZE 64
+  #define CFG_TUD_VENDOR_TX_BUFSIZE TUD_EPSIZE_BULK_MAX
 #endif
 
 // Vendor is buffered (FIFO mode) if both TX and RX buffers are configured


examples/device/cdc_dual_ports/src/tusb_config.h
```diff
@@ -104,7 +104,10 @@
 #define CFG_TUD_CDC_TX_BUFSIZE   (TUD_OPT_HIGH_SPEED ? 512 : 64)
 
 // CDC Endpoint transfer buffer size, more is faster
-#define CFG_TUD_CDC_EP_BUFSIZE   (TUD_OPT_HIGH_SPEED ? 512 : 64)
+// Leave it as default size (512 for HS, 64 for FS) unless your host application
+// is able to send ZLP (Zero Length Packet) to terminate transfer !
+#define CFG_TUD_CDC_FS_XFERSIZE   64
```
hathach: I don't think we should have it separated as FS/HS. This is for EP buf, it should hold enough memeory for highest speed possible

HiFiPhile: Actually the issue exist for default value `CFG_TUD_CDC_EP_BUFSIZE   (TUD_OPT_HIGH_SPEED ? 512 : 64)` even before stream refactor.

For high speed device since TUD_OPT_HIGH_SPEED is true, even at full speed it will queue transfer of 512 length instead of 64 leading to OUT transfer zlp issue.

I didn't came up with a fix without separate the config for each speed.

HiFiPhile: > we can have tud_class_configure() to disable zlp handling which pass that to edpt stream API instead.

Do you mean add a config like `multi_packet_rx`:
- For IN keep current behavior (without PR)
- For OUT if true then queue EP_BUFSIZE like current behavior (Need host sending ZLP), if false then only queue MPS

?

Another inconsistency is when `CFG_TUD_EDPT_DEDICATED_HWFIFO = 0`, data is written into FIFO and `tud_cdc_n_available` is incremented, but `tud_cdc_rx_cb` is only called every 8th times (512/64).

hathach: > Actually the issue exist for default value `CFG_TUD_CDC_EP_BUFSIZE (TUD_OPT_HIGH_SPEED ? 512 : 64)` even before stream refactor.
> 
> For high speed device since TUD_OPT_HIGH_SPEED is true, even at full speed it will queue transfer of 512 length instead of 64 leading to OUT transfer zlp issue.
> 
> I didn't came up with a fix without separate the config for each speed.

yeah, I mean by defauult it will send using buffer size, however if zlp behavior is disabled. we will only queue 1 packet i.e 64 byte when highsped device is operating at fullspeed.

hathach: > > we can have tud_class_configure() to disable zlp handling which pass that to edpt stream API instead.
> 
> Do you mean add a config like `multi_packet_rx`:
> 
> * For IN keep current behavior (without PR)
> * For OUT if true then queue EP_BUFSIZE like current behavior (Need host sending ZLP), if false then only queue MPS
> 
> ?
> 
> Another inconsistency is when `CFG_TUD_EDPT_DEDICATED_HWFIFO = 0`, data is written into FIFO and `tud_cdc_n_available` is incremented, but `tud_cdc_rx_cb` is only called every 8th times (512/64).

IMO, it is not inconsistent but intended behavior for ZLP host/device. Transfer is not complete if last packet is 64 and there is no ZLP. I am suggesting that we can have tud_cdc_configure() to change this behavior (same for other class driver)

HiFiPhile: > IMO, it is not inconsistent but intended behavior for ZLP host/device. Transfer is not complete if last packet is 64 and there is no ZLP. I am suggesting that we can have tud_cdc_configure() to change this behavior (same for other class driver)

Actually I think you are right.

HiFiPhile: I've added the config to CDC and vendor classes, except MIDI as it doesn't need throughput.

PR #3489, opened 2026-02-10
Update memborwse ci

hathach:


hathach (commented):
Reviewed end-to-end (workflow wiring, build_util target flow, family_add_membrowse integration, build.py multi-target refactor, matrix generation).\n\nI don’t see any blocking issues. CI is green across the matrix and the membrowse upload path looks consistent with the new integrated approach.

PR #3490, opened 2026-02-11
Fully asynchrounous host code (v2)

ceedriic:
**Describe the PR**
This remove all calls to tusb_time_delay_ms_api() from inside tuh_task_ext().

**Additional context**

This is a cleaned-up version of #3482

Use of the host API was not really possible with OPT_OS_NONE due to blocking calls in tuh_task_ext().

With an OS, it will also improve the situation because:

1. The timeout_ms parameter of tuh_task_ext() is now respected.
2. The traffic with device 1 will not stop when device 2 is plugged.

I've split this in 3 main patches

1. Long boring preparation patch to split the functions calling tusb_time_delay_ms_api().
2. Executing the continuation functions asynchronously when CFG_TUH_TASK_USE_TIME_MILLIS_API is set.
3. cleaning up the handling of deferred attachments with a separate queue (note: I've not really tested that last patch because with my hub and 2 devices for some reason I never got into this situation)

### DIFF
--- a/test/hil/hil_test.py
+++ b/test/hil/hil_test.py
@@ -302,7 +302,7 @@ def flash_esptool(board, firmware):
         idf_target = json.load(f)['IDF_TARGET']
     with open(f'{fw_dir}/flash_args') as f:
         flash_args = f.read().strip().replace('\n', ' ')
-    command = (f'esptool.py --chip {idf_target} -p {port} {flasher["args"]} '
+    command = (f'esptool --chip {idf_target} -p {port} {flasher["args"]} '
                f'--before=default_reset --after=hard_reset write_flash {flash_args}')
     ret = run_cmd(command, cwd=fw_dir)
     return ret
--- a/test/hil/tinyusb.json
+++ b/test/hil/tinyusb.json
@@ -13,7 +13,8 @@
             "flasher": {
                 "name": "esptool",
                 "uid": "4ea4f48f6bc3ee11bbb9d00f9e1b1c54",
-                "args": "-b 1500000"
+                "args": "-b 1500000 --force",
+                "comment": "use --force for ESP32-P4 v0.1"
             },
             "comment": "Use TS3USB30 mux to test both device and host"
         },


ceedriic:
Hmm, any idea why hil-tinyusb would fail with this PR, but passed with #5484?
This should be exactly the same final diff (but against master)

src/host/usbh.c
```diff
@@ -558,11 +599,16 @@
     osal_queue_delete(_usbh_q);
     _usbh_q = NULL;
 
-    #if OSAL_MUTEX_REQUIRED
+#if CFG_TUH_HUB
+    osal_queue_delete(_usbh_daq);
```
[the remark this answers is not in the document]
ceedriic: Hey Bot, this function will have side effects if CFG_TUSB_OS is different from OPT_OS_NONE

src/host/usbh.c
```diff
@@ -1469,68 +1560,77 @@
...
+    TU_VERIFY(dev0_bus->hub_port != 0,);
     TU_ASSERT(hub_port_get_status(dev0_bus->hub_addr, dev0_bus->hub_port, NULL,
-                                  process_enumeration, ENUM_HUB_RERSET));
+                                  process_enumeration, ENUM_HUB_RERSET),);
   }
   #endif // hub
+}
 
-  return true;
+static void enum_after_reset_root_delay(void) {
```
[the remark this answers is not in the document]
ceedriic: Hey Bot, all these functions are called, sometimes through function pointers. you're probably confused by all the preprocessor directives.

src/host/usbh.c
```diff
@@ -1823,6 +1896,53 @@
...
+#if CFG_TUH_TASK_USE_TIME_MILLIS_API
+static void enum_after_attempt_delay(void) {
+  TU_LOG_USBH("Enumeration attempt %u/%u\r\n", _usbh_data.enum_failed_count+1, ATTEMPT_COUNT_MAX);
+  if (!tuh_control_xfer(&_usbh_data.enum_xfer_retry))
+    enum_full_complete(false); // complete as failed
+}
+#endif
+
+static void enum_after_set_address_recovery_delay(void) {
+  const uint8_t new_addr =_usbh_data.enumerating_daddr;
```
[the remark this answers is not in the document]
ceedriic: Hey Bot, all these functions are called, sometimes through function pointers. you're probably confused by all the preprocessor directives.

hathach:
> Hmm, any idea why hil-tinyusb would fail with this PR, but passed with #5484? This should be exactly the same final diff (but against master)

usb subsystem on my VM sometime got "stuck" and/or timeout due to lots of usb attach/activity. May need to re-run or reset the whole system.

ceedriic:
> > Hmm, any idea why hil-tinyusb would fail with this PR, but passed with #5484? This should be exactly the same final diff (but against master)
> 
> usb subsystem on my VM sometime got "stuck" and/or timeout due to lots of usb attach/activity. May need to re-run or reset the whole system.

OK.

Anyway, after I did the [Merge branch 'hathach:master' into non-blocking-host-v2] two days ago, all tests passed.

src/host/usbh.c
```diff
@@ -618,20 +699,17 @@ void tuh_task_ext(uint32_t timeout_ms, bool in_isr) {
...
+#if CFG_TUH_HUB
         } else {
           // currently enumerating another device
           TU_LOG_USBH("[%u:] USBH Defer Attach until current enumeration complete\r\n", event.rhport);
-          const bool is_empty = osal_queue_empty(_usbh_q);
-          queue_event(&event, in_isr);
-          if (is_empty) {
-            return; // Exit if this is the only event in the queue, otherwise we loop forever
-          }
+          TU_ASSERT(osal_queue_send(_usbh_daq, &event, in_isr),);
```
hathach: multiple queue is not currently supported. E.g freeRTOS will put running task() to suspended mode and switch to other task when trying to get from the queue. I am making some refactor, and also rever this for now.

ceedriic: You can just ignore the third patch for now then...

ceedriic: (All three patches can be applied independently)

ceedriic: Also: the second queue is only used with timeout=0, and only from the task() thread, so maybe the OS_NONE implementation could be used for that second queue?

hathach: oh right, I overlook it, we can use queue receive with timeout=0, then. I did reverted, but will re-add that in the next push.

src/tusb_option.h
```diff
@@ -698,17 +701,6 @@
...
-// tuh_task_ext() will be asynchronous and never sleep in tusb_time_delay_ms_api()
-#ifndef CFG_TUH_TASK_USE_TIME_MILLIS_API
-  #if CFG_TUSB_OS == OPT_OS_RTX4 || CFG_TUSB_OS == OPT_OS_PICO || defined(ESP_PLATFORM)
-    // these boards/os do not implements the required tusb_time_millis_api()
-    #define CFG_TUH_TASK_USE_TIME_MILLIS_API  0
-  #else
-    #define CFG_TUH_TASK_USE_TIME_MILLIS_API  1
-  #endif
-#endif
-
```
ceedriic: My goal with that check was to enable the new asynchronous code with or without OS (because it is better in BOTH cases).
Unfortunately, somme platforms do not define tusb_time_millis_api() so I had to use this check to support them and get the CI green.
If tusb_time_millis_api() could be defined for all platforms, then I think this conditional code could just be removed. Unfortunately, adding the missing tusb_time_millis_api() is not something I can do.

hathach: most rtos support time api that return time since boot. I will add osal_time_millis() and make it official. It is currently implement in user (bsp in the example).

src/host/usbh.c
```diff
@@ -169,6 +169,12 @@ static OSAL_SPINLOCK_DEF(_usbh_spin, usbh_int_set);
 OSAL_QUEUE_DEF(usbh_int_set, _usbh_qdef, CFG_TUH_TASK_QUEUE_SZ, hcd_event_t);
 static osal_queue_t _usbh_q;
 
+  #if CFG_TUH_HUB
+// Deferred attachment queue, only needed when using hub
+OSAL_QUEUE_DEF(usbh_int_set, _usbh_daqdef, CFG_TUH_HUB, hcd_event_t);
```
hathach, line 174: these event is only generated one per hub, since we process hub port status 1 by 1. The only time it got another attachment is from other hub, or from same hub (with multiple device) when an enumerating device is set_configured(). I short, max size is CFG_TUH_HUB

ceedriic: Oh, great, this is why I couldn't test this code with a single hub :)

ceedriic:
I'm reviewing the latest code, give me an hour or so...

hathach:
> I'm reviewing the latest code, give me an hour or so...

beware, there is a huge change list that replace board_millis() with tusb_time_millis_api() (which is required when OS=none). You can skip files in examples and hw/bsp

ceedriic:
> perfect, thank you @ceedriic for the PR. I make some changes

Thank you for reviewing/improving!

> * replace CFG_TUH_TASK_USE_TIME_MILLIS_API with CFG_TUSB_OS_HAS_SCHEDULER

This is the part I do not understand: I agree my option name (CFG_TUH_TASK_USE_TIME_MILLIS_API) was bad,
but if I understand the new define, you will not run the asynchronous code with a multitasking RTOS.
I only use OPT_OS_NONE at the moment, so this is fine for me, but why not running the asynchronous code with a RTOS?

> * rename usbh_wait_delay to usbh_call_after_ms() and also take arg. This is useful helper, maybe used in other places as well in the future
> * merge all delay/after callback to make it easier to maintain

Yes, both these changes make the code better.

>   I also make some other changes while doing local tests as well. Let me know what you think, or do we need to make any other improvement.

I've basically two small things:

1) I see that you've added code I missed in tuh_task_event_ready() :)
This is obviously correct, but I think it is a bit more complicated, and the following patch should be added (I can push it if you agree)

```
   #if CFG_TUH_HUB
-  if (!osal_queue_empty(_usbh_daq)) {
-    return true;
+  if (_usbh_data.enumerating_daddr == TUSB_INDEX_INVALID_8 && !osal_queue_empty(_usbh_daq)) {
+    return true; // Deferred attachment ready to run
+  }
+  #endif
+
+  #if CFG_TUSB_OS_HAS_SCHEDULER == 0
+  if (_usbh_data.call_after.func) {
+      int32_t ms = (int32_t)(_usbh_data.call_after.at_ms - tusb_time_millis_api());
+      if (ms <= 0)
+         return true; // Timer event ready to run
   }
   #endif
```

2) You changed the following code:

```
  if (delay_cb) {
    int32_t ms = (int32_t)(_usbh_data.enum_wait_deadline - tusb_time_millis_api());
    if (ms <= 0) {
      // delay expired, run callback now
      TU_LOG_USBH("USBH run timer callback\r\n");
      _usbh_data.enum_wait_delay_cb = NULL;
      delay_cb();
    } else if (timeout_ms > (uint32_t)ms) {
      // reduce timeout accordingly
      timeout_ms = (uint32_t)ms;
    }
  }
```

Into a simplified version:

```
    if (after_cb) {
      uint32_t ms = tusb_time_millis_api();
      if (ms >= _usbh_data.call_after.at_ms) {
        TU_LOG_USBH("USBH run timer callback\r\n");
        _usbh_data.call_after.func = NULL;
        after_cb(_usbh_data.call_after.arg);
      }
    }

```

But I think it is wrong for two reasons:

1) The comparaison ms >= _usbh_data.call_after.at_ms will fail when the timer rolls over after 50 days.
if '_usbh_data.call_after.at_ms' equals 4'294'967'295 and 'ms' wrap to 0, then the new test will never fire anymore.
But by converting to a signed integer first, ms will be equals to -1, and the test will always be correct by the magic of 2's complement.

2) if in the original code ms > 0 (for example 10), it means that the next timer event will need to occur after 10ms.
if the timeout timeout_ms given in tuh_task_ext was bigger than that (for example 100), then we need to reduce this timout from 100 to 10 to make sure that if no message are in the queue, we still exits tuh_task_ext() after 10ms, to run the next timer event on time. (I can also push it if you agree).

Other than that, at least for a quick look, I like your changes.

hathach:
@ceedriic you are right, we shouldn't rely on rtos delay since it blocks main event queue. I apparently does not have enough sleep, let me try to revert those change.

ceedriic:
> @ceedriic I revert the changes to also support non-blocking on RTOS, also rename to `usbh_defer_func_ms_async()` and make it a helper, though usage outside of enumeration need some kind of mutex. We will address that later if needed. Let me know if you think it looks good now ?

I took a quick look at the usbh.c change, and I think is is OK, I've just 3 comments (the only important is the first one)

1) In the following hunk, I think the osal_queue_empty(_usbh_daq) added check inside CFG_TUH_HUB is incorrect and should be removed (if there are deferred attachments, we don't want to loop forever here)

```
+  #if CFG_TUSB_OS_HAS_SCHEDULER
+    // return if there are no more events, to allow application to run other backgrounds
+    if (osal_queue_empty(_usbh_q)
+    #if CFG_TUH_HUB
+        && osal_queue_empty(_usbh_daq)
+    #endif
+    ) {
+      return;
+    }
+  #endif
```

Thinking about it, I think the whole chunk here could also be simply replaced by a single line, without conditionals, to prevent sleeping a second time with an RTOS:

```
timeout_ms = 0;
```

2) The following comment should be removed or fixed:

```
// For non-scheduler deferred callback. For scheduler OS: blocking delay then callback
bool usbh_defer_func_ms_async(uint32_t ms, tusb_defer_func_t func, uintptr_t param) {
```

3) You've not updated the tuh_task_event_ready() function, I don't know if that's on purpose or not.

src/host/usbh.c
```diff
@@ -592,51 +640,83 @@ bool tuh_task_event_ready(void) {
...
+      if (ms <= 0) {
+        // delay expired, run callback now
+        TU_LOG_USBH("USBH invoke scheduled function\r\n");
+        _usbh_data.call_after.func = NULL;
+        after_cb(_usbh_data.call_after.arg);
+      }
+
+      // above after_cb() can re-schedule another function, we need to re-check and reduce timeout of
+      // the main event timeout to make sure we aren't blocking more than call_after timeout.
+      if (_usbh_data.call_after.func != NULL &&
```
hathach: @ceedriic there is a gotcha, the callback can re-schedule another function. If we miss that main queue can sleep forever, and then there is no event (since we need that callback to submit new transfer for new event).

ceedriic: Oops, Yep, good catch!

ceedriic: I'm just worried about the unsigned comparaisons with your latest patch.

ceedriic: I need to go for lunch but I will get back to you after

ceedriic: @hathach
What do you think about the following alternative patch?

I think it's simpler and follow the rule to exit `tuh_task()` after we've processed some events.

```
diff --git a/src/host/usbh.c b/src/host/usbh.c
index a674b040e..80f8e5672 100644
--- a/src/host/usbh.c
+++ b/src/host/usbh.c
@@ -666,6 +666,7 @@ void tuh_task_ext(uint32_t timeout_ms, bool in_isr) {
         TU_LOG_USBH("USBH invoke scheduled function\r\n");
         _usbh_data.call_after.func = NULL;
         after_cb(_usbh_data.call_after.arg);
+        timeout_ms = 0; // no sleeping after processing en event
       } else if (timeout_ms > (uint32_t)ms) {
         // reduce main event timeout to make sure we don't blocking more than call_after timeout
         timeout_ms = (uint32_t)ms;
@@ -786,16 +787,7 @@ void tuh_task_ext(uint32_t timeout_ms, bool in_isr) {
         break;
     }
 
-  #if CFG_TUSB_OS_HAS_SCHEDULER
-    // return if there are no more events, to allow application to run other backgrounds
-    if (osal_queue_empty(_usbh_q)
-    #if CFG_TUH_HUB
-        && osal_queue_empty(_usbh_daq)
-    #endif
-    ) {
-      return;
-    }
-  #endif
+    timeout_ms = 0; // no sleeping after processing en event
   }
 }
```

hathach: @ceedriic I have push a new update, that better address the signe comparison. Aso the timeout_ms = 0 is good one, just applied that.

ceedriic: @hathach:
in your latest patch if `remain_ms <= 0`, you also need to set` timeout_ms = 0`

hathach: yeah, that is kind of unexpected for a new calback expired right after scheduled but it is possible

src/host/usbh.c
hathach: fixed now with latest push.

ceedriic:
Perfect!

PR #3494, opened 2026-02-11
Supports full-speed host with an high-speed PHY (DWC2)

ceedriic:
**Describe the PR**
This patch allows configuring a Full-Speed USB host with DWC2 on a board wired with a High-Speed ULPI PHY

**Additional context**
This implements #3450 proof-of-concept in a proper way (I hope).

It introduces an optional `CFG_TUH_PHY_SPEED` configuration option, which specify the PHY configuration and can be different (higher than) CFG_TUH_MAX_SPEED

For example, this works on a stm32h747 board with external ULPI:

```
#define CFG_TUH_MAX_SPEED		OPT_MODE_FULL_SPEED
#define CFG_TUH_PHY_SPEED		OPT_MODE_HIGH_SPEED
```

The effect of this configuration is to set the `HCFG_FSLS_ONLY` bit on the DWC2 controller at the right place.

### DIFF
--- a/src/portable/synopsys/dwc2/hcd_dwc2.c
+++ b/src/portable/synopsys/dwc2/hcd_dwc2.c
@@ -412,9 +412,6 @@ bool hcd_init(uint8_t rhport, const tusb_rhport_init_t* rh_init) {
 
   //------------- 3.1 Host Initialization -------------//
 
-  // work at max supported speed
-  dwc2->hcfg &= ~HCFG_FSLS_ONLY;
-
   // Enable HFIR reload
   if (dwc2->gsnpsid >= DWC2_CORE_REV_2_92a) {
     dwc2->hfir |= HFIR_RELOAD_CTRL;
@@ -427,6 +424,14 @@ bool hcd_init(uint8_t rhport, const tusb_rhport_init_t* rh_init) {
   dwc2->stm32_gccfg &= ~STM32_GCCFG_VBVALOVAL;
 #endif
   while ((dwc2->gintsts & GINTSTS_CMOD) != GINTSTS_CMODE_HOST) {}
+
+  if (hprt_speed_get(dwc2) != TUSB_SPEED_HIGH) {
+    // disable high speed mode
+    dwc2->hcfg |= HCFG_FSLS_ONLY;
+  } else {
+    // work at max supported speed
+    dwc2->hcfg &= ~HCFG_FSLS_ONLY;
+  }
 
   // configure fixed-allocated fifo scheme
   dfifo_host_init(rhport);
@@ -1375,6 +1380,11 @@ static void port0_enable(dwc2_regs_t* dwc2, tusb_speed_t speed) {
       }
     }
     hcfg |= HCFG_FSLS_PHYCLK_SEL_30_60MHZ;
+  }
+
+  if (speed != TUSB_SPEED_HIGH) {
+    // disable high speed mode
+    hcfg |= HCFG_FSLS_ONLY;
   }
 
   dwc2->hcfg = hcfg;
--- a/src/portable/synopsys/dwc2/dwc2_common.c
+++ b/src/portable/synopsys/dwc2/dwc2_common.c
@@ -187,7 +187,7 @@ bool dwc2_core_is_highspeed(dwc2_regs_t* dwc2, tusb_role_t role) {
   }
 #endif
 #if CFG_TUH_ENABLED
-  if (role == TUSB_ROLE_HOST && !TUH_OPT_HIGH_SPEED) {
+  if (role == TUSB_ROLE_HOST && !TUH_OPT_PHY_SPEED) {
     return false;
   }
 #endif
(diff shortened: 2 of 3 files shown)


HiFiPhile:
I think it's better to add a forced speed `tuh_configure` dynamic config.
https://github.com/hathach/tinyusb/blob/2883403ed010c54c33a38c28358a4dd0c67d67c0/src/host/usbh.h#L150
https://github.com/hathach/tinyusb/blob/2883403ed010c54c33a38c28358a4dd0c67d67c0/src/host/usbh.h#L115

You need to leave `dwc2_core_is_highspeed` unchanged otherwise it will try to initialize internal FS phy instead of ULPI in `dwc2_core_init`

ceedriic:
> I think it's better to add a forced speed `tuh_configure` dynamic config.
> 
> https://github.com/hathach/tinyusb/blob/2883403ed010c54c33a38c28358a4dd0c67d67c0/src/host/usbh.h#L150

Ok, I would prefer a dynamic config too.

But I tried to do it the other way around. The problem with 

1. speed = high
2. forced_speed = full

Is that all buffers sizes will be based on "high speed" settings (big)

Or do we need something like PR #3487 for the host side too?

I was trying to tackle the problem with:

1. speed = full
2. phy_speed = high

To have the stack settings based on full speed.

> https://github.com/hathach/tinyusb/blob/2883403ed010c54c33a38c28358a4dd0c67d67c0/src/host/usbh.h#L115
> 
> You need to leave `dwc2_core_is_highspeed` unchanged otherwise it will try to initialize internal FS phy instead of ULPI in `dwc2_core_init`

I think you misread the patch: the new CFG_TUH_PHY_SPEED is the "hardware" speed, so dwc2_core_is_highspeed() need to use that.

I've tested the 3 following configs:

1. HS PHY, HS speed (ULPI)
2. HS PHY, FS speed (ULPI)
3. FS PHY, FS speed (internal)

and it all works.

HiFiPhile:
> The problem with
speed = high
forced_speed = full
Is that all buffers sizes will be based on "high speed" settings (big)

You mean which buffer, FIFO in `dfifo_host_init` ?

src/portable/synopsys/dwc2/hcd_dwc2.c
```diff
@@ -428,6 +425,14 @@ bool hcd_init(uint8_t rhport, const tusb_rhport_init_t* rh_init) {
 #endif
   while ((dwc2->gintsts & GINTSTS_CMOD) != GINTSTS_CMODE_HOST) {}
 
+  if (hprt_speed_get(dwc2) != TUSB_SPEED_HIGH) {
```
hathach, line 428: hmm, what if there is no device connected, it would be full speed ?

src/portable/synopsys/dwc2/hcd_dwc2.c
```diff
@@ -1377,6 +1382,11 @@ static void port0_enable(dwc2_regs_t* dwc2, tusb_speed_t speed) {
     hcfg |= HCFG_FSLS_PHYCLK_SEL_30_60MHZ;
   }
 
+  if (speed != TUSB_SPEED_HIGH) {
```
hathach, line 1385: manual say we don't change this bit after intiial programming 

Image

hathach:
We should use tuh_configure() to force the phy selection e.g `CFG_TUH_MAX_SPEED = FULL, but tuh_configure(use_hs_phy=1)`. The function dwc2_core_is_highspeed() should be rename to `dwc2_core_is_highspeed_phy()` and return true if we force phy selection. HCFG_FSLS_ONLY should only be set ONCE in init per manual since I think it involves HS PLL phy on/off.

ceedriic:
> > The problem with
> > speed = high
> > forced_speed = full
> > Is that all buffers sizes will be based on "high speed" settings (big)
> 
> You mean which buffer, FIFO in `dfifo_host_init` ?

No, you're right, I made a mistake.

I was thinking about code like:

`#define CFG_TUD_NET_ENDPOINT_SIZE (TUD_OPT_HIGH_SPEED ? 512 : 64)`
`#define CFG_TUD_CDC_TX_BUFSIZE (TUD_OPT_HIGH_SPEED ? 512 : 64)`

But I now see that this only in the device side, and not in the host side.

src/portable/synopsys/dwc2/hcd_dwc2.c
ceedriic: Hmm, yeah, I don't remember why exactly I used that function here. I will look for a better way.

ceedriic:
> We should use tuh_configure() to force the phy selection e.g `CFG_TUH_MAX_SPEED = FULL, but tuh_configure(use_hs_phy=1)`. The function dwc2_core_is_highspeed() should be rename to `dwc2_core_is_highspeed_phy()` and return true if we force phy selection. HCFG_FSLS_ONLY should only be set ONCE in init per manual since I think it involves HS PLL phy on/off.

Ok, that makes a lot of sense, time for a v3 patch

src/portable/synopsys/dwc2/hcd_dwc2.c
ceedriic: Yes, I was trying to keep it the same, but I'm not sure what I was thinking when I wrote part of that patch. I will redo that patch from scratch.

ceedriic:
Bye Bye, welcome #3506

Issue #2776, opened 2024-08-26
RP2040 HCD driver does not correctly recover from data sequence errors per USB Spec 8.6.4


rppicomidi:
### Operating System

Linux

### Board

Raspberry Pi Pico

### Firmware

[usb_midi_host_example](https://github.com/rppicomidi/usb_midi_host/tree/main/examples/C-code/usb_midi_host_example)

### What happened ?

This is with regards the RP2040 USB Host mode chip bug described here: https://github.com/raspberrypi/pico-feedback/issues/394.

TinyUSB implements host mode bulk and interrupt endpoints using the automatically polled endpoint hardware on the RP2040. That hardware has a bug that prevents it from recovering from data sequence errors per the USB specification section 8.6.4. Normally, data sequence errors due the device missing ACK are rare unless the environment is noisy or the hardware is bad. However, there are commercial devices with at least one IN endpoint where the first Bulk IN or Interrupt IN transfer starts with DATA1 PID instead of DATA0. Either way, any data sequence error freezes up the host hardware for that endpoint in an infinite loop instead of replying ACK and discarding the data. The RP2040 has no way to detect this condition and recover from it. 

The right thing to do is to rewrite the RP2040 HCD to use only EPx. At least that endpoint hardware has a way to detect data sequence errors and should actually implement USB specification 8.6.4 correctly. However, rewriting the HCD is a relatively long task. The new HCD has to route all Bulk IN and and Interrupt IN transfers through EPx to avoid the hardware bug. Once we are doing that, it seems that we should try to improve all Bulk and Interrupt transfer throughput by managing transfers in software.

For a quicker solution, I have a patch to TinyUSB that allows an application to tell the RP2040 HCD about specific device IN endpoints on devices with a given VID:PID that are known to always start the first Bulk IN or Interrupt IN transfer with DATA1 PID. The patch changes the expected first PID to DATA1 instead of DATA0 when the HCD opens the known faulty endpoint. This patch is known to work: See https://github.com/rppicomidi/usb_midi_host/issues/14 near comments 40.

@hathach, if I submit this patch as a pull request, would you consider it, or is it too much of a hack? In my mind, it should be a temporary solution. It requires applications to have a list of commercial devices that are known to be buggy. The patch still does not allow the RP2040 USB Host mode to recover from a data sequence error due to noise. It is useful as a temporary fix until the HCD gets a rewrite, however.

### How to reproduce ?

[...]

PR #3503, opened 2026-02-15
fix shared epx for rp2040

rumbledethumps:
Not all devices will reassemble fragmented control packets. This fix might benefit from being guarded.

### DIFF
--- a/src/host/usbh.c
+++ b/src/host/usbh.c
@@ -1799,12 +1799,6 @@ static void process_enumeration(tuh_xfer_t* xfer) {
       TU_LOG_USBH("Device configured\r\n");
       dev->configured = 1;
 
-    #if CFG_TUH_HUB
-      if (_usbh_data.dev0_bus.hub_addr != 0) {
-        hub_edpt_status_xfer(_usbh_data.dev0_bus.hub_addr); // get next hub status
-      }
-    #endif
-
       // Parse configuration & set up drivers
       // driver_open() must not make any usb transfer
       TU_ASSERT(enum_parse_configuration_desc(daddr, (tusb_desc_configuration_t*) _usbh_epbuf.ctrl),);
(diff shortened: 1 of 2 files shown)


rumbledethumps:
I was able to move this to hcd_rp2040.

rumbledethumps:
This is a collection of rp2040 hcd fixes that I needed while building a custom MSC BOT+CBI driver. They should be useful to everyone.

rumbledethumps:
These are all the changes I needed to make for an XInput driver and bespoke MSC driver that supports removable media on both CBI and BOT. Because I want to use floppy drives. I may even add LUN support. The shared control endpoint had so many problems that I nearly gave up. The final piece was hcd_pause_interrupt_eps. I'm seeing massive stability improvements with this upgraded hcd. Testing error conditions by removing media while reading is how found all these edge cases that don't normally show up with the regular MSC driver. I hope they are useful to others making custom drivers.

src/host/usbh.c
```diff
@@ -1938,9 +1932,12 @@ static void enum_full_complete(bool success) {
...
 #if CFG_TUH_HUB
-  // Hub status is already requested in case of successful enumeration
-  if (_usbh_data.dev0_bus.hub_addr != 0 && !success) {
-    hub_edpt_status_xfer(_usbh_data.dev0_bus.hub_addr); // get next hub status
+  // Re-queue hub status polling now that enumeration (including all
+  // driver set_config callbacks) is complete.  Previously this was
+  // started in ENUM_CONFIG_DRIVER *before* drivers ran, which allowed
+  // hub control transfers to interleave on EPX with CBI ADSC commands.
+  if (_usbh_data.dev0_bus.hub_addr != 0) {
+    hub_edpt_status_xfer(_usbh_data.dev0_bus.hub_addr);
```
HiFiPhile: Please remove this change, hub status need to be started before to catch set_config failure or quick plug/unplug.

src/portable/raspberrypi/rp2040/hcd_rp2040.c
```diff
@@ -637,4 +676,45 @@ bool hcd_edpt_clear_stall(uint8_t rhport, uint8_t dev_addr, uint8_t ep_addr) {
...
 }
 
+//--------------------------------------------------------------------+
+// Interrupt endpoint pause/resume
+//--------------------------------------------------------------------+
+// The RP2040 SIE shares a single EPX for all control transfers.
+// Hardware-polled interrupt endpoints can delay the STATUS phase of a
+// control transfer long enough for timing-sensitive CBI devices (e.g.
+// TEAC floppy) to STALL.  Temporarily disabling interrupt endpoint
+// polling during critical control transfers avoids this.
```
HiFiPhile: Do you have a hardware analyzer to confirm this finding ?

rumbledethumps:
It looks like this code may be a bit too defensive. Turns out I have a device that's hitting a hardware bug.
https://github.com/hathach/tinyusb/issues/2776
I'll maintain this on a fork until we have a workaround, like the epx-only hcd.

PR #3509, opened 2026-02-23
follow up to PR3506

hathach:
No additional information provided.

### DIFF
--- a/src/portable/synopsys/dwc2/dwc2_common.c
+++ b/src/portable/synopsys/dwc2/dwc2_common.c
@@ -60,6 +60,7 @@ static void reset_core(dwc2_regs_t* dwc2) {
   while (!(dwc2->grstctl & GRSTCTL_AHBIDL)) {} // wait for AHB master IDLE
 }
 
+// Dedicated FS PHY is internal with a clock 48Mhz.
 static void phy_fs_init(dwc2_regs_t* dwc2) {
   TU_LOG(DWC2_COMMON_DEBUG, "Fullspeed PHY init\r\n");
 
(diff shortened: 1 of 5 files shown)


src/portable/synopsys/dwc2/hcd_dwc2.c
```diff
@@ -426,23 +422,24 @@ bool hcd_init(uint8_t rhport, const tusb_rhport_init_t* rh_init) {
...
+    dwc2->hcfg &= ~HCFG_FSLS_ONLY; // work at max supported speed
+    is_highspeed = true;
   }
-#endif
+  #endif
 
-  // configure fixed-allocated fifo scheme
-  dfifo_host_init(rhport);
+  // configure a fixed-allocated fifo scheme
+  dfifo_host_init(rhport, is_highspeed);
```
ceedriic: Sorry if I don't really know what I'm talking about, but Is it really a good idea to reduce the size of the buffers if the core is _able_ to have bigger buffers?

There is this comment in dfifo_host_init() that implies that a "small" buffer is allocated because of the core limitation:

`// - ptx_largest is limited to 256 for FS since most FS core only has 1024 bytes total`

But that certainly does not apply if the core is HS-capable.

hathach: hmm, you are right. Maybe it is ok just leave it as before.

src/portable/synopsys/dwc2/dwc2_common.c
```diff
@@ -180,30 +188,19 @@ static bool check_dwc2(dwc2_regs_t* dwc2) {
...
+  const bool           has_hs_phy = (ghwcfg2.hs_phy_type != GHWCFG2_HSPHY_NOT_SUPPORTED);
 
-  if (!prefer_hs_phy) {
-    return false;
+  if (prefer_hs_phy) {
+    return has_hs_phy;
+  } else {
+    const bool has_fs_phy = (ghwcfg2.fs_phy_type != GHWCFG2_FSPHY_NOT_SUPPORTED);
+    // false if has fs phy, otherwise true since hs phy is the only available phy
+    return !has_fs_phy && has_hs_phy;
```
hathach, line 199: @HiFiPhile U5 fall into this category with UTMI+ without fs_phy

PR #3515, opened 2026-02-27
Add support for WCH ch582/583 series

alt-0191:
This PR adds support for the WCH CH58x series (CH582, CH583), device, host and dual role demo have been tested and confirmed working on real hardware, but there are still some things that need to be discussed/improved: 

CH581: It shares the same SDK, but I have only one USB device and different flash size, and it isn't on the market yet, So there is no support for it.
CH584/CH585: Despite the similar name, their SDK is in a separate
repository with a different register layout. I haven't looked into whether the USB IP is compatible.
So I made a change like this:
`OPT_MCU_CH58X/CH582/CH583`: All three aliases intentionally map to the same value,
leaving room for per-variant differentiation down the road without breaking existing code.

`_write` symbol conflict: WCH's SDK has a `_write` macro ,will be compiled error under WCH's MounRiver Studio. I added a `#error` to make this
obvious at compile time instead of leaving users with a cryptic linker error. Although this is not a perfect solution, there is no need to make too many changes to the source code and SDK.


alt-0191:
cdc_msc broken, more fix needed...

src/portable/wch/dcd_ch58x_usbfs.c
```diff
@@ -0,0 +1,585 @@
...
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ *
+ * This file is part of the TinyUSB stack.
+ */
+
+#include "tusb_option.h"
+
+#if CFG_TUD_ENABLED && defined(TUP_USBIP_WCH_CH58X) && CFG_TUD_WCH_USBIP_USBFS
```
hathach: does the chip use the new usbip or same usbfs as other wch. If it is the same, the existing drivers should be update instead of adding new files

alt-0191: No, though they are alike in core but sdks are totally different, so it's better to keep drivers sepreted

alt-0191:
Tested some demos and runs well, cdc_msc bugs also fixed.

src/portable/wch/dcd_ch58x_usbfs.c
hathach: what is the different in the dcd driver ?

hathach: If only sdk is different and the dcd driver is pretty much the same. Then We should use the same .c file (We can't have same .c for every families). Here is how we got dwc2 working for multiple mcu vendors with totally different sdks for reference https://github.com/hathach/tinyusb/tree/master/src/portable/synopsys/dwc2

alt-0191: You're right, I reviewed both drivers side by side and the logic is essentially the same, differences are all at the register access level. I'll refactor into a shared file later.

alt-0191: Sorry, this is harder than I expected. ch32 and ch58x have quite different usb register layouts (split/merged ep control, dma width, interrupt controller), so merging them isn't as straightforward as dwc2. I tried the port-header approach but still ended up with about 20 #if blocks.
The ch58x driver follows its sdk conventions, which are pretty different from ch32's (ch32 sdk mimics stm32 style, while ch58x sdk is more simplified). A full rewrite could help, but given how different these chip families are, keeping separate usb IPs seems more practical for maintenance(actually wch have many differences from each chips, makes it hard to maintance)

hathach: there is indeed some difference, let keep this open, I will try to take a look when possible. I still prefer to keep it under 1 driver to make it easier to maintain the driver.

src/portable/wch/dcd_ch32_usbfs.c
```diff
@@ -218,17 +273,31 @@ void dcd_int_handler(uint8_t rhport) {
...
     uint16_t rx_len   = USBOTG_FS->RX_LEN;
 
     switch (token) {
       case PID_OUT: {
+#ifdef CH32_USBFS_EP_MANUAL_TOG
+        // Manual toggle: drop OUT packets whose data toggle doesn't match (host retransmit),
+        // otherwise flip the expected RX toggle for the next packet. EP0 is driven by the
+        // SETUP/status flow below, so its toggle is left to that path.
+        if (ep != 0) {
+          if (!(int_st & USBFS_INT_ST_TOG_OK)) { break; }
```
HiFiPhile: @hathach I think we need to apply TOG_OK check for all variants, some days ago I modified the HCD to send only DATA0 and CH32V305 received all packets.

hathach: right, in that case, we will test this for others as well. Look like the only thing auto toggle does is bit xor

hathach:
@copilot review
@codex review

hathach:
@copilot review
@codex review

hathach (approved):
thank you very much for your PR. I have managed to integrate the ch58x into unified usbfs driver. It has combined an tx/rx ep control, manual toggle and EP0-EP4 shared DMA (awkward design). PR is verified with ch58x in the hil pool (along with v103 and v203). 

Due to the time constraint, currently hcd usbfs isn't update and remain v203 only. We can do that later as follow up.

alt-0191:
Thanks @hathach for the review and merge! I'll port host support back when I have time — sorry for being slow on this.

hathach:
> Thanks @hathach for the review and merge! I'll port host support back when I have time — sorry for being slow on this.

thank you very much, the usbfs host itself has a couple of pending PR as well.

PR #3529, opened 2026-03-06
Add MTT hub support

HiFiPhile:


hathach:
@claude review this

HiFiPhile:
> perfect, I update to open mtt interface in 1 hub_open(). Note that this enable the MTT for hub, but our hcd driver may not take this into account

Nice ! Since we don't do bandwidth allocation for periodic split transfers (I think only supported by EHCI/IP3516), I hope enable MTT can reduce periodic transfer failure chance when too many devices connected.

PR #3532, opened 2026-03-06
 _open_vc_itf parsing error when vc->std.bNumEndpoints != 0

YixingShen:
The Hierarchy and Memory Layout
 Standard Interface Descriptor (VC)
 Video Control Header Descriptor: Contains wTotalLength.
 Video Control Camera Terminal Descriptor
 Video Control Output Terminal Descriptor
 Standard Endpoint Descriptor: The physical interrupt pipe (Length = 7 bytes).
 Class-Specific VC Interrupt Endpoint Descriptor: Additional endpoint metadata (Length = 5 bytes).

fixed _end_of_control_descriptor logic.
wTotalLength does not include Standard Endpoint Descriptor and Class-specific VC Interrupt Endpoint Descriptor,
so fix that _end_of_control_descriptor include Standard Endpoint Descriptor and Class-specific VC Interrupt Endpoint Descriptor. It will also fix _close_vc_itf, _open_vc_itf. _find_desc_entity parsing.

### DIFF
--- a/src/class/video/video_device.c
+++ b/src/class/video/video_device.c
@@ -70,6 +70,13 @@ typedef struct TU_ATTR_PACKED {
   uint8_t bEntityId;
 } tusb_desc_cs_video_entity_itf_t;
 
+typedef struct TU_ATTR_PACKED {
+  uint8_t  bLength;
+  uint8_t  bDescriptorType;
+  uint8_t  bDescriptorSubtype;
+  uint16_t wMaxTransferSize;
+} tusb_desc_cs_video_vc_ep_t;
+
 typedef union {
   struct TU_ATTR_PACKED {
     uint8_t bLength;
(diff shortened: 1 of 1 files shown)


hathach:
@claude review this

YixingShen:
The _close_vc_itf and _open_vc_itf have extend end when bNumEndpoints > 0.
current _end_of_control_descriptor return is also extend end 
Revision: 3e22de0a4782d6062959aec27e5304f1b4d4a557
static inline void const* _end_of_control_descriptor(void const *desc)
{
  tusb_desc_vc_itf_t const *vc = (tusb_desc_vc_itf_t const *)desc;
  uint8_t const *end = (uint8_t const*)desc + vc->std.bLength
                       + tu_le16toh(vc->ctl.wTotalLength);
  if (vc->std.bNumEndpoints) {
    end += sizeof(tusb_desc_endpoint_t);  // standard EP descriptor
    end += 5;                             // class-specific EP descriptor (fixed 5 bytes per UVC spec)
  }
  return end;
}

The _end_of_control_descriptor should be revert to
Revision: 0a45308a296251e4376815b21f89f0038c75a60e
Modified: src/class/video/video_device.c
static inline void const* _end_of_control_descriptor(void const *desc)
{
  tusb_desc_vc_itf_t const *vc = (tusb_desc_vc_itf_t const *)desc;
  return ((uint8_t const*) desc) + vc->std.bLength + tu_le16toh(vc->ctl.wTotalLength);
}

PR #3543, opened 2026-03-09
_end_of_control_descriptor return when std.bNumEndpoints  > 0

YixingShen:
The _close_vc_itf and _open_vc_itf have extend end when bNumEndpoints > 0.
current _end_of_control_descriptor return is also extend end
Revision: 3e22de0
static inline void const* _end_of_control_descriptor(void const *desc)
{
tusb_desc_vc_itf_t const *vc = (tusb_desc_vc_itf_t const *)desc;
uint8_t const end = (uint8_t const)desc + vc->std.bLength
+ tu_le16toh(vc->ctl.wTotalLength);
if (vc->std.bNumEndpoints) {
end += sizeof(tusb_desc_endpoint_t); // standard EP descriptor
end += 5; // class-specific EP descriptor (fixed 5 bytes per UVC spec)
}
return end;
}

The _end_of_control_descriptor should be revert to
Revision: 0a45308
Modified: src/class/video/video_device.c
static inline void const* _end_of_control_descriptor(void const *desc)
{
tusb_desc_vc_itf_t const *vc = (tusb_desc_vc_itf_t const )desc;
return ((uint8_t const) desc) + vc->std.bLength + tu_le16toh(vc->ctl.wTotalLength);
}

### DIFF
--- a/src/common/tusb_fifo.c
+++ b/src/common/tusb_fifo.c
@@ -280,7 +280,7 @@ static void hwff_push_n(const tu_fifo_t *f, const void *app_buf, uint16_t n, uin
     const uint32_t odd_mask    = data_stride - 1;
     uint16_t       lin_even    = lin_bytes & ~odd_mask;
     tu_hwfifo_read(hwfifo, ff_buf, lin_even, access_mode);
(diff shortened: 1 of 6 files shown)


hathach (approved):
right, sorry, it is my bad.

PR #3552, opened 2026-03-14
add set protocol skip feature

dinsfire64:
I have a quirky device that has multiple devices in it's report (gamepad, keyboard, and a mouse).

If i send any set hid protocol message to the device during it's life cycle, it will stall the polls to the gamepad endpoint (which I want to talk to) and prefers the keyboard only.

I need some way to tell tinyusb to skip sending the set protocol command during operation, so this PR allows me to call in my code

`  tuh_hid_set_default_protocol(HID_PROTOCOL_SKIP);`

and during enumeration, the device will not get a protocol message sent to it.

Please let me know if you have any alternative suggestions or tips.

### DIFF
--- a/src/class/hid/hid.h
+++ b/src/class/hid/hid.h
@@ -145,7 +145,8 @@ typedef enum
 typedef enum
 {
   HID_PROTOCOL_BOOT = 0,
-  HID_PROTOCOL_REPORT = 1
+  HID_PROTOCOL_REPORT = 1,
+  HID_PROTOCOL_SKIP = 2,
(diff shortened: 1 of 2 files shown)


hathach:
please try to see if https://github.com/hathach/tinyusb/pull/3558 works for you

dinsfire64:
Yes this feature works for me, I have tested it with the quirky device and it is working correctly.

Would you like me to close this in favor of the PR and comment there?

hathach:
yes

PR #3554, opened 2026-03-15
Fix RP2350 TinyUSB USB DPRAM memcpy hard fault

ccrome:
on the RP2350 the code was crashing on this unaligned access.

Honestly, codex worked this out, and also created the fix.  It was a bit hard to trace down because it was always crashing in an area where it wasn't obvious.

At first, codex gave the wrong fix, even though it worked.  I believe the code is now correct.  

## Title

Fix RP2350 TinyUSB USB DPRAM memcpy hard fault

## Description

This PR fixes a hard fault in the TinyUSB RP2040/RP2350 USB device path seen during audio IN transfers on RP2350.

## Change

- Update `tinyusb/src/portable/raspberrypi/rp2040/rp2040_usb.c` so the USB DPRAM copy helper uses aligned 32-bit and 16-bit accesses when possible before falling back to byte writes

## Problem

- The device was hard-faulting in USB IRQ context during TinyUSB audio packet staging
- The backtrace pointed to `unaligned_memcpy()` in `rp2040_usb.c`
- The fault occurred while copying data into USB DPRAM

## Root Cause

- Byte-by-byte writes into USB DPRAM were not reliable in this RP2350 transfer path
- Using wider aligned accesses avoids the observed fault while preserving support for unaligned tails

## Validation

- Rebuilt and flashed the RP2350 target
- Reproduced the previous fault path under debugger before the change
- Re-ran the same debug session after the fix and observed stable execution without the prior USB hard fault

## Notes

- This is a targeted fix in the TinyUSB RP2040/RP2350 USB port layer
- The change is intended to preserve existing behavior while making USB DPRAM writes safer on RP2350

### DIFF
--- a/src/portable/raspberrypi/rp2040/rp2040_usb.c
+++ b/src/portable/raspberrypi/rp2040/rp2040_usb.c
@@ -83,9 +83,6 @@ static bool e15_is_critical_frame_period(struct hw_endpoint *ep);
  * Overlapping ranges are still undefined.
  */
 
-#if defined(__GNUC__) || defined(__clang__)
-__attribute__((noinline))
-#endif
 static void unaligned_memcpy(void *dst_void, const void *src_void, size_t n)
 {
     uint8_t *dst = (uint8_t *) dst_void;


hathach:
can you test with adding volatile, that would probably prevents compiler to widen access (for speed on rp2350).

```C
  static void unaligned_memcpy(uint8_t *dst, const uint8_t *src, size_t n) {
    volatile uint8_t *vdst = dst;
    const volatile uint8_t *vsrc = src;
    while (n--) {
      *vdst++ = *vsrc++;
    }
  }
```

ccrome:
Shoot!  sorry i missed your request 3 weeks ago!

Issue #3495, opened 2026-02-11
PR #3451 introduced a fatal bug which shows only in release mode, on RP2 / RP2350


rgrr:
### Operating System

PicoSDK

### Commit SHA

80b207852a96a7639c6ec83d0d37322a8d6e53c0

### Board

Pico2

### Firmware

My project uses TinyUSB with lots of interfaces, project is https://github.com/rgrr/yapicoprobe/

Problem appears on an RP2350 Pico2 board.

Not sure if it also will happen with an RP2040 board.

### What happened ?

Updating TinyUSB from 0.20.0 to recent master showed a bug which crashes my device after some seconds after a reset.  I guess this is the time of USB handshaking.

Surprisingly this happens only in release mode with gcc.  gcc-debug, clang-debug and clang-release all run fine.

The problem had been introduced with #3451 (12.01.2026).  The previous master containing #3448 (09.01.2026) runs also with gcc-release.

Any idea what could be the reason?  Any hint how I can dig?

used compiler:

gcc 14.2.1
clang 21.1.1

### How to reproduce ?

Difficult... question is more: how can I support finding this?

### Debug Log as txt file (LOG/CFG_TUSB_DEBUG=2)

AFAICS there is no logging in this area.

### I have checked existing issues, discussion and documentation

- [x] I confirm I have checked existing issues, discussion and documentation.

hathach:
can you give a quick try if disable hwfifo would resolve this https://github.com/hathach/tinyusb/tree/rp2-disable-hwifo. If yes, it is probably a race condition.

rgrr:
yes, problem disappears on the branch

hathach:
@rgrr Thank you for confirmation, I guess the feature is not ready yet. Since I don't have the time to test it out, I will just disable it for now and re-enable later.

rgrr:
Let me hear when you restart development in this direction so that I could help if wanted/required

HiFiPhile:
For info some devs have been done: #3561 #3585 #3582

rgrr:
> For info some devs have been done: [#3561](https://github.com/hathach/tinyusb/pull/3561) [#3585](https://github.com/hathach/tinyusb/pull/3585) [#3582](https://github.com/hathach/tinyusb/pull/3582)

gosh... that's been a lot of work.  Everything seems to work fine for rp2350 now, rp2040 not tested.

I close my complaint for now.  Thanks for your great work!

PR #3560, opened 2026-03-18
rp2 disable hwfifo

hathach:


maxgerhardt:
Was there something wrong with using the dedicated HW FIFO? The PR doesn't contain any reasoning why it's now disabled by default-

hathach:
@maxgerhardt it is a regression https://github.com/hathach/tinyusb/issues/3495, I currently don't have time to look at this. Will re-enable later with more stress tests.

PR #3567, opened 2026-03-22
net: make bInterval a parameter in TUD_CDC_NCM_DESCRIPTOR

morse-cedricvandenbergh:
The notification endpoint bInterval was hardcoded to 50 in the macro. Expose it as _ep_notif_interval so callers can set an appropriate value for their device speed, per discussion #3553.

Update net_lwip_webserver to pass TUD_OPT_HIGH_SPEED ? 9 : 50, which preserves the original 50ms polling interval on full-speed (50 frames) and approximates it on high-speed (bInterval=9 → 2^8 × 125µs = 32ms).


morse-cedricvandenbergh:
Implemneted as part of https://github.com/hathach/tinyusb/pull/3568

PR #3574, opened 2026-03-27
Add initial board support for nRF54LM20 DK

gab-k:
The cdc_msc device example works fine. Isochronous OUT transfers hit a babble error (DOEPINT bit 12) on every packet I am not really sure why that is.

The nrfx 3.14 commit which is used in get_deps.py is a little outdated so it predates the renaming done in nrfx 4.1.0 which reads:

> Renamed nRF54LM20A Eng A to nRF54LM20A. 

in the nrfx changelog

PR #3575, opened 2026-03-28
fix ISO OUT endpoint failure on alt-setting MPS change

gab-k:
**Summary**

 When an ISO OUT endpoint is reused across alternate settings with a different `wMaxPacketSize` (e.g. UAC2 speaker switching sample rates), the DWC2 `dcd_edpt_iso_activate()` path leaves stale hardware state behind. After the alt-setting switch, the first packets fault and no data reaches the application.
                                                                                                                                                                                                                                            
The existing code called `edpt_disable()` before re-activation, but its full GONAK/handshake sequence does not clear residual `doeptsiz`/`doepdma` values from the previous configuration on all DWC2 variants.



**Fix**

Add `edpt_iso_out_reset()`   a register reset function that clears endpoint HW state and transfer bookkeeping before reactivation. Only triggered when all three conditions are met:                          
  - Direction is OUT
  - Endpoint was previously active
  - New MPS differs from current MPS
   
All other paths (IN endpoints, same MPS, first activation) should be unchanged.                                                                                                                                                                 
                  
I kept it separate from `edpt_disable()` intentionally as its a pre-activation state reset, not a full endpoint teardown.



**Context**

Discovered while bringing up the nRF54LM20 DK (#3574) with the `uac2_headset` example. Also tested on a NUCLEO-H7S3L8, havent seen any regression issues, alt-setting switches work correctly on both boards. Testing was not extensive  though, just basic functional verification. I dont have access to other DWC2 platforms (ESP32-S2/S3, etc.).

Also ran the ceedling unit tests, all passed.


src/portable/synopsys/dwc2/dcd_dwc2.c
```diff
@@ -347,6 +347,27 @@ static void edpt_disable(uint8_t rhport, uint8_t ep_addr, bool stall) {
...
+// is being immediately re-activated by the caller.
+static void edpt_iso_out_reset(uint8_t rhport, uint8_t epnum) {
+  dwc2_regs_t* dwc2 = DWC2_REG(rhport);
+  dwc2_dep_t* epout = &dwc2->epout[epnum];
+
+  dwc2->daintmsk &= ~TU_BIT(epnum + DAINT_SHIFT(TUSB_DIR_OUT));
+  // Full register write (not |=) to clear all bits including active/enabled state
+  epout->doepctl = DOEPCTL_SNAK;
+  epout->doepint = 0xFFFFFFFFu;
+  epout->doeptsiz = 0;
```
HiFiPhile: Thank you for the fix but what I think the only thing needed is adding SNAK to `edpt_disable`:

Image

`doeptsiz`, `doepdma` and xfer state will be set on next transfer.

HiFiPhile: Also could you dump the DWC2 config with dwc2_info.py ? It'a new issue specific to nRF54LM20.

gab-k:
I investigated the LM20 DWC2 config and also reduced the fix further.

For the DWC2 config dump, I read the registers on hardware and added a dedicated `nRF54LM20` column to `dwc2_info.py` / `dwc2_info.md`.

These differ from the existing generic `nRF54` entry, so LM20 seems to need its own column.

I tested the `DOEPCTL.SNAK` addition in `edpt_disable()` by itself on hardware, but that was not sufficient for this case.

All the other writes I had in the earlier version (`DOEPINT`, `DOEPTSIZ`, `DOEPDMA`, transfer bookkeeping, etc.) turned out not to be needed.

The extra helper function is not strictly required anymore at this point; the same logic could be inlined. I kept it for now because I think it makes the special-case intent a bit clearer, but that part is mostly a matter of  preference and I can inline it if you prefer.


Edit: Actually i just noticed that i mixed up the this PR and #3574 a little bit maybe should have modified dwc2_info in that PR, hope that isnt too much of an issue.

HiFiPhile:
Thank you, it looks like something is changed in GID 5.00b. 

But I don't understand why settings `epout->doepctl = DOEPCTL_SNAK;` will fix the issue, since `doepctl` should be reset in `edpt_disable` (those who left will be overwritten in `edpt_activate`). Unless there are some additional flag needs to be cleared  but it's not listed in https://docs.nordicsemi.com/bundle/ps_nrf54LM20A/page/usbhs.html#register.DOEPCTL1

Could you check  `doepctl`  value before/after `edpt_activate` call w/wo your fix ?

Disabling an endpoint without following $7.5.1 procedure is not a good idea.

gab-k:
I investigated further, it seems the issue is not that `edpt_activate()` fails to rewrite `DOEPCTL`. The problem is that `edpt_disable()` does not clear `DOEPCTL.USBAEP`. The `edpt_activate()` function sets `USBAEP` again via `depctl.active = 1`, while `edpt_disable()` only disables the endpoint (`DOEPCTL_EPDIS`) and leaves `USBAEP` set.

You are right, its probably better to not skip the normal disable flow. I changed the sequence to:

`edpt_disable() -> clear DOEPCTL.USBAEP -> edpt_activate()`

This only applies for ISO OUT when the MPS changes.

The actual reason why the `DOEPCTL = DOEPCTL_SNAK` write worked fine was that it also cleared `USBAEP`.

For completeness, here are the DOEPCTL values before and after edpt_activate() for the current fix and without the fix:

|`DOEPCTL`| before edpt_activate() | after edpt_activate() |
| :---------------- | :------: | :----: |
| with fix        |   0x0005001C   | 0x00058038 |
| without fix  |   0x0005801C   | 0x00058038 |

The only difference is bit 15 being cleared before re-activation.

I considered putting this into the generic OUT path in `edpt_disable()`, since that function currently leaves `DOEPCTL.USBAEP` set. However, I’m not really confident enough to say that changing `edpt_disable()` globally is safe, the issue seems to be specific to ISO OUT re-activation with changed MPS on this DWC2 revision.

HiFiPhile:
I think it's safe to clear ActEP in `edpt_disable`, linux also clear it in EP disable, although the databook doesn't mention it much.

Is it work for you :
```diff
diff --git a/src/portable/synopsys/dwc2/dcd_dwc2.c b/src/portable/synopsys/dwc2/dcd_dwc2.c
index 8685ec6dc..e32dedd1b 100644
--- a/src/portable/synopsys/dwc2/dcd_dwc2.c
+++ b/src/portable/synopsys/dwc2/dcd_dwc2.c
@@ -345,6 +345,11 @@ static void edpt_disable(uint8_t rhport, uint8_t ep_addr, bool stall) {
       dwc2->dctl |= DCTL_CGONAK;
     }
   }
+
+  // Clear ActEP
+  if (!stall && epnum != 0) {
+    dep->ctl &= ~EPCTL_USBAEP;
+  }
 }
 
 // Since this function returns void, it is not possible to return a boolean success message
```

gab-k:
Alright, sounds good to me! This probably is the most elegant solution anyway I was just afraid that it might break something somewhere else.

PR #3582, opened 2026-04-02
rp2040 fix for at least a dozen crash scenarios

rumbledethumps:
This is the diff you asked for. I've been leaving my overrides an AI mess because of the moving target. Feel free to close this after you've reviewed and picked out the good bits.


hathach (commented):
thank you for the PR, can you give a bit of explanation e.g what this pr improve and/or which issue it resolves.

rumbledethumps:
These are the problems my changes address. Keep in mind I have a complete MSC implementation with BOT, CBI, LUN, removable media, LBA64, and even TRIM support so I hit code parts that the examples do not, especially concerning removable media error recovery.

* busy_wait_at_least_cycles uses the CPU clock instead of the SIE clock
* Devices refuse to mount, e.g. a hub with MSC-BOT, CDC, and HID won't mount the HID.
* Removing devices from hub causes panic hcd_clear_stall.
* Too many endpoints allocated, one extra EP0 per device is allocated.
* Entire close and abort path is missing!!!
* hcd_free_ep_count() added because pool is so tiny and user needs to know.

rumbledethumps:
I did some math and busy_wait_at_least_cycles(32) should be safe up to 512 MHz. The original 12 cycles can fail intermittently as you pass 192 MHz, which a lot of projects do.

rumbledethumps:
Just a heads up that the next release will be a significant regression for anyone using the rp2040 in host mode. This PR has already been deployed to my users who all use hubs and have multiple devices connected simultaneously. Typically, keyboard, mouse, gamepad, MSC CBI/BOT, and CDC, all at once. We're also regularly hot plugging devices, which is where most of the crashes happen.

hathach:
> Just a heads up that the next release will be a significant regression for anyone using the rp2040 in host mode. This PR has already been deployed to my users who all use hubs and have multiple devices connected simultaneously. Typically, keyboard, mouse, gamepad, MSC CBI/BOT, and CDC, all at once. We're also regularly hot plugging devices, which is where most of the crashes happen.

Thanks, I saw issues with pico2 with hub + ch34x + a msc device on hil test as well. I am not sure if it is due to my cheap hub, is ordering a ch334 hub to add to hil pool. Will test this out when they arrive

PR #3586, opened 2026-04-02
fix uart pin init for teensy

felhub:
This PR fixes a bug for the UART pin initialization for Teensy boards.

For the teensy_40 and teensy_41 boards, in `board.h,` we correctly select UART6
```c
#define UART_PORT             LPUART6
```

However, there was a copy-paste bug in `pin_mux.c` where we instead muxed UART1 to the chip pads.
This is now fixed by muxing the correct UART6 to the respective output pins.

This is also convenient because UART1 is not even accessible via the headers on the teensy_40.


HiFiPhile:
A bit late, I've just regenerated config files with MCUXpresso : https://github.com/hathach/tinyusb/tree/teensy_config

hathach:
@HiFiPhile ah right, I forgot we use the pintool for this, please make a PR and merge :)

PR #3594, opened 2026-04-10
Fix musb RXRDY Clearing

BrentK-ADI:
Resolves an issue in the musb `handle_xfer_out` function where not all execution paths cleared the `MUSB_RXCSRL1_RXRDY` bit, causing the RX interface to hang and no longer communicate with the host.

This issue was discovered while using the RNDIS class in a media converter application. When running iperf testing at full speed (10Mbps link) the USB stack was under constant load of packets and eventually would stop receiving new frames from the host PC.  Other USB events and interrupts were still occurring.

It is believed the root cause was the RXRDY bit not being cleared in all data paths of the xfer handler, resulting in the peripheral no longer accepting new frames into the FIFO.  Applying this change allowed continuous operation under these high traffic conditions without lockup.

### Verification Testing
Testing was performed with 2x MAX32690's acting as RNDIS to 10Mbps single pair ethernet media converters, connected to Linux PCs.  Bidirectional iperf was run for 12 hours resulting in each MCU sending and receiving 18M+ packets (~24 GB) in each direction with no driver lock ups of failures.  Prior to these updates, testing would fail within minutes.

### DIFF
--- a/src/portable/mentor/musb/dcd_musb.c
+++ b/src/portable/mentor/musb/dcd_musb.c
@@ -230,13 +230,19 @@ static bool handle_xfer_out(uint8_t rhport, uint_fast8_t ep_addr)
   musb_ep_csr_t* ep_csr = get_ep_csr(musb_regs, epnum);
   // TU_LOG1(" RXCSRL%d = %x\r\n", epnum_minus1 + 1, ep_csr->rx_csrl);
 
-  TU_ASSERT(ep_csr->rx_csrl & MUSB_RXCSRL1_RXRDY);
+  //Fail gracefully. Spurious interrupt.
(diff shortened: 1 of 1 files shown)


hathach (approved):
look good, thank you

HiFiPhile:
@hathach I don't think it's the right fix, drop (flush) the packet will cause data loss. The driver should be arranged that the endpoint NAKs OUT traffic when no buffer is available.

hathach:
@HiFiPhile you are right, I fogort musb auto ack an out packet. Let me revise and make an follow-up PR to fix this behavior, we can check a flag if rx data is already available when dcd_edpt_xfer() is called to pull from fifo and complete immediately.

PR #3603, opened 2026-04-17
tinyusb.mk: Remove duplicate usbc.c entry.

andrewleech:
`src/typec/usbc.c` is listed twice in `tinyusb.mk`, once in the device sources block and again at the end of the host sources block. This was likely introduced when host sources were appended in 6b8933cfe. Removing the duplicate.


hathach (approved):
thank you

Issue #3613, opened 2026-04-22
midi host: tuh_midi_stream_read can leave bytes in FIFO on cable-number transition, silently wedging bulk IN polling


HakanL:
# midi host: `tuh_midi_stream_read` can leave bytes in FIFO on cable-number transition, silently wedging bulk IN polling

**Commit SHA:** current master (reproduced against ~2026-03-18 snapshot)
**Board / host controller:** STM32H753 (Synopsys DWC2 OTG, full speed)
**Device reproducer:** Akai LPD8 mk1 — VID `0x09E8`, PID `0x0075`

## Summary

After the fix in #3239 (commit `0263cfc`) landed, we still hit a related-but-distinct wedge with the same device. `tuh_midi_stream_read` returns to the caller as soon as the next USB-MIDI event packet has a different cable number than the current one (see [midi_host.c L743–L748](https://github.com/hathach/tinyusb/blob/master/src/class/midi/midi_host.c#L743-L748)). Bytes for those other cables stay in the stream FIFO.

The default `CFG_TUH_MIDI_RX_BUFSIZE` is `TUH_EPSIZE_BULK_MAX` = 64 B — **one bulk packet** — so once the FIFO retains *any* residue, the next 64 B bulk IN transfer cannot fit, `tu_edpt_stream_read_xfer` in `midih_xfer_cb` does not queue, and subsequent bulk polling silently stops. No error, no callback, no log. Just silence.

## Reproduction

The LPD8 mk1 sends a 64 B startup burst after `SET_CONFIG` whose USB-MIDI event headers contain cable numbers 0, 1, 5, 6, 8, 14, etc. (non-deterministic — appears to be uninitialised TX FIFO content). The first valid-looking event on cable 0 is parsed (≈ 2–3 MIDI bytes extracted); the next event's different cable triggers the early return. ~60 bytes of events remain in the FIFO. Next bulk IN transfer cannot land. Pad / knob events never reach the host.

### Attached captures

- **`lpd8-usbmon.pcap`** — Linux `usbmon` capture of a full LPD8 mk1 enumeration and operation cycle against `snd-usb-audio`. Relevant frames:
  - `55–56` — `SET_CONFIG` request / response
  - `57–60` — 2 × `GET_DESCRIPTOR STRING` (note: Linux does **not** issue `SET_INTERFACE` for this device)
[...]

Issue #3239, opened 2025-09-06
MIDI Host: Akai Professional LPD8 no data


rppicomidi:
### Operating System

Linux

### Commit SHA

8304587d7774526a03c6881d11e6d6208fe759be

### Board

Raspberry Pi Pico

### Firmware

https://github.com/rppicomidi/usb_midi_host/tree/main/examples/C-code/usb_midi_host_example with the MIDI OUT send code commented out

### What happened ?

I plugged an Akai Professional LPD8 MIDI controller to the USB host port and operated the LPD8's controls. I expected the example application to display the incoming MIDI messages. Instead, there was no output. The application works fine with other MIDI controllers.

I took a sniffer trace (attached). On plugin, the LPD8 sends one 64-byte packet of garbage bytes to the host. These bytes cause error checks to fail in the `tuh_midi_stream_read()` parser; the errors are not handled. This causes the application to stall and eventually the driver to fail.  I have a fix in the works and will submit a pull request as soon as I am able.

[akaipro_lpd8.capture.zip](https://github.com/user-attachments/files/22187907/akaipro_lpd8.capture.zip)

### How to reproduce ?

In https://github.com/rppicomidi/usb_midi_host/tree/main/examples/C-code/usb_midi_host_example/usb_midi_host_example.c, comment out lines 122-129.

Build the application and load to a Raspberry Pi Pico board.
Attach a microUSB to USB adapter to the Pico board's USB connector.
Attach a USB to serial port converter UART input pins to Pico board pins 1 and 2 and connect the converter's ground pin to ground.
Open a terminal program to display the USB to serial port converter output.
Provide 5V power to the Pico board's VBus pin.
Observe the "Pico MIDI Host Example" message on the terminal window.
Connect the LPD8 to the USB host port
Observe the "MIDI Device Index = 0, MIDI device address = 1, 1 IN cables, 1 OUT cables" message on the terminal window
Press the pads and operate the knobs the the LPD8
Observe no MIDI message outputs

### Debug Log as txt file (LOG/CFG_TUSB_DEBUG=2)

n/a

### I have checked existing issues, discussion and documentation

- [x] I confirm I have checked existing issues, discussion and documentation.

PR #3614, opened 2026-04-22
midi host:  document drain requirement of RX FIFO

HakanL:
Follow-up to #3239. That fix addressed garbage bytes in the *data* position of USB-MIDI event packets. A related wedge remains when a packet contains events whose *header* byte (cable number) varies mid-buffer, because `tuh_midi_stream_read` terminates on cable-number transitions rather than on end-of-FIFO.

With the default `CFG_TUH_MIDI_RX_BUFSIZE = TUH_EPSIZE_BULK_MAX` (64 B), any residue left in the stream FIFO by a parse-stop-on-cable-transition prevents the next bulk IN transfer from landing. `tu_edpt_stream_read_xfer` in `midih_xfer_cb` silently fails to queue and the driver stops receiving data with no error indication.

**Reproducer:** Akai LPD8 mk1 (VID `0x09E8`, PID `0x0075`). After its 64 B power-on burst — whose USB-MIDI event headers span multiple fake cable numbers — the endpoint goes silent forever unless the host FIFO is big enough to absorb the residue *or* the app drains the FIFO in a loop. See attached `lpd8-usbmon.pcap` in the linked issue for a Linux `usbmon` capture showing the same device streaming fine under `snd-usb-audio` (whose read loop drains per call).

## Changes

1. `CFG_TUH_MIDI_RX_BUFSIZE` default raised from `TUH_EPSIZE_BULK_MAX` to `2 * TUH_EPSIZE_BULK_MAX`. Gives the app one full packet of residue headroom. Cost: +64 B per MIDI instance.
2. Added a note on `tuh_midi_stream_read` that it may return without draining the FIFO when a cable-number change is encountered, and callers should invoke it in a loop to guarantee full drainage.

## Tested

STM32H753 (DWC2 OTG host, full speed) against:

- Akai LPD8 mk1 (the reproducer — reliable across unplug/replug cycles with this patch; wedges without)
- Akai LPD8 mk2
- WORLDE ORCA 16 MiniKey

All three work with the new defaults; no behaviour change observed for mk2 or WORLDE.

Addresses the remaining wedge reported as a continuation of #3239. Full reproduction details and captures in #3613.


src/class/midi/midi_host.h
```diff
@@ -38,11 +38,16 @@ extern "C" {
...
+  // Sizing this equal to the endpoint packet size (the historical default)
+  // can cause the next bulk IN transfer to fail to queue silently, wedging
+  // the stream. See the drain-loop note on tuh_midi_stream_read() below.
+  #define CFG_TUH_MIDI_RX_BUFSIZE (2 * TUH_EPSIZE_BULK_MAX)
 #endif
 
 #ifndef CFG_TUH_MIDI_TX_BUFSIZE
-  #define CFG_TUH_MIDI_TX_BUFSIZE TUH_EPSIZE_BULK_MAX
+  #define CFG_TUH_MIDI_TX_BUFSIZE (2 * TUH_EPSIZE_BULK_MAX)
 #endif
```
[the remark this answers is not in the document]
HakanL: TX was reverted — the +64 B description is now accurate

src/class/midi/midi_host.h
```diff
@@ -38,7 +38,12 @@ extern "C" {
 // Class Driver Configuration
 //--------------------------------------------------------------------+
 #ifndef CFG_TUH_MIDI_RX_BUFSIZE
-  #define CFG_TUH_MIDI_RX_BUFSIZE TUH_EPSIZE_BULK_MAX
+  // Default sized to 2x the bulk endpoint to absorb residue left in the FIFO
+  // when tuh_midi_stream_read() stops early on a cable-number transition.
+  // Sizing this equal to the endpoint packet size (the historical default)
```
HiFiPhile: Hi, with your `2. always reading stream in a loop` this comment is no longer true.

HiFiPhile: @HakanL

HiFiPhile (approved):

src/class/midi/midi_host.h
HakanL: Sorry @HiFiPhile I didn't see your comment here. Hmm, I don't think no-change would work though, I needed the extra RX buffer to capture the garbage that the LPD8 was sending out.

HiFiPhile: Any garbage should be drained by reading RX FIFO in a loop. 

Even if `CFG_TUH_MIDI_RX_BUFSIZE=100*TUH_EPSIZE_BULK_MAX`, calling `tuh_midi_stream_read` without a loop  can return only one 4-byte packet and preventing subsequent transfer.

HakanL: @HiFiPhile Yeah, that makes sense, but I already have a loop to read, so I don't know why it didn't work until I bumped the buffer size. I'll do some more testing.

HakanL: There was an issue with my code where I didn't drain properly during startup, so that caused the issue. I had made the buffer change at the same time so I thought that was the actual fix, but it wasn't.

PR #3615, opened 2026-04-24
fix zero wLength request in control request

kira-live:
# Description
In the USB specification, when wLength equals 0, direction should be ignored and the Status Stage is always IN, rather than being opposite to the direction of the Control Request. Therefore, we modified the logic for determining whether it is the Status Stage and the ACK transmission flow for the Status Stage.

# Other
This change can be merged or discarded after thorough discussion and verification; please see #3606  for details.
This change has been tested and passed on ESP32-S3.

I wasn't familiar with GitHub's online editor at first, so I accidentally opened three PRs. sorry.


hathach (approved):
look good, thank you

kira-live:
Thank you for creating such an amazing project!

PR #3616, opened 2026-04-24
typec/stm32: fix UCPD sink bugs affecting compatibility with newer PD sources

michaelajax:
I observed that with the B-G474E-DPOW1 devkit, many USB-C PD sources I connected were failing to negotiate PD and some were causing the application to crash.

This PR fixes three issues:

## Issue 1: CC cable orientation not detected correctly

The original code had a FIXME noting that CC2 vstate was always 1 on the DPOW1 board, and some code to work in spite of that. The original code guarded against this by pairing `TYPECEVT2` with a `vstate == 3` check, which also happened to mask the real attach detection bugs. Since I did not observe the CC2 issue on my board, I restructured the attach logic to check both vstates unconditionally on any CC event, accepting any non-zero vstate.

Otherwise, sources advertising Rp_default or Rp_1.5A were silently treated as detach. And, when the cable is in the CC2 orientation, `TYPECEVT1` fires as CC1 transitions to 0, but `TYPECEVT2` may not fire at the same time, the CC2 attach was never detected.

## Issue 2: RXHRSTDET flag never cleared, causing an interrupt storm

In the event of a hard reset, `RXHRSTDETIE` is included in `IMR_ATTACHED` but there was no handler for `UCPD_SR_RXHRSTDET.` Without clearing it via ICR, the ISR re-enters immediately on every cycle. A handler is added to clear `RXHRSTDETCF` and re-arm RX DMA so the source capabilities retransmitted after the reset are received.

## Issue 3: GoodCRC sent in response to GoodCRC

Every successfully received message triggered a GoodCRC, including GoodCRC itself, which the PD spec forbids. Some sources treat this as a protocol error and issue a Hard Reset, which previously triggered Issue 2 above. This change skips the GoodCRC transmission when `header->msg_type == PD_CTRL_GOOD_CRC`.

## Manual Testing
I tested with over half a dozen power supplies from different manufacturers across a wide wattage range, including several newer chargers that support EPR and PD 3.1, and some chargers that have multiple Type-C outputs from a shared power source. All negotiated what I would have expected.

## AI Note
Full disclaimer, I leveraged AI tools to debug the issues and implement the fixes along with the manual testing described above. However, I feel the modifications are relatively simple and logical to address the issues I was facing on this board.


michaelajax:
All AI code review comments addressed and manually re-tested against the same 8 USBC PD chargers I tested before.

michaelajax:
Hi @hathach, I don't mean to be a bother, but I am wondering if you would consider this PR? It would be great to be able to move my project back onto the mainline branch instead of my local fork.

For what it's worth, I have since tested on 6 more different model PD supplies for a total of 14, and all of them negotiated successfully without any stability issues. Thanks!

HiFiPhile:
Thank you, I don't have G4 board so I've added U5 support, from my test it works correctly.

I've also added a connect/disconnect api, useful for re-negotiation.

michaelajax:
Thanks @HiFiPhile! 

I pulled the latest branch down on my local machine and gave it a spin on my B-G474E-DPOW1, and everything's working as expected. I tested PD negotiation with 6 different supplies I had around my desk - three small ~18-30W units, two ~65-100W units, and one 140W EPR capable unit. All detected PDOs correctly up to 20V/5A as expected for SPR negotiation. I tested both cable directions with all supplies, too.

I didn't test the new `typec_connect_task` nor make use of the `tuc_connect`/`tuc_disconnect` APIs, but those do look useful for someone who would like additional control over Type-C functionality.

HiFiPhile (approved):
Glad to hear it works !

hathach:
thank you @michaelajax very much for the PR, sorry I got stuck with other works. Thank @HiFiPhile for reviewing. 

PS: 'typec' is not really a good name, is mostly refer to physical connection. I think we will rename it to something as tupd and or usbpd in the future since it is pd protocol after all. Watch out for that.

PR #3622, opened 2026-04-28
fix(dcd_ch32_usbfs): fix dropped bulk OUT packets by properly re-arming EP_RX_CTRL state

mickabrig7:
When stress-testing a custom Vendor interface with lots of fast bulk transfers on CH32V30x (which use the USBFS controller), I noticed that a lot of data was dropped because `tud_vendor_rx_cb()` would not always fire.

After investigating, it seemed like `dcd_ch32_usbfs` wasn't updating the `EP_RX_CTRL` register after a transfer completed or when a new one was queued.

`update_out()` now properly sets `EP_RX_CTRL` to ACK or NAK for all endpoints based on transfer state, and `dcd_edpt_xfer()` explicitly re-enables ACK when re-queuing an OUT transfer.
Bulk transfers now complete reliably without data loss or missed callbacks.


HiFiPhile:
Thank you, I think we should also remove ACK enablement in `dcd_edpt_open()`

I don't have CH30V20x at hand but surprised by their USB IP implementation, if CPU load is high setting NAK after packet received can still missing data.

HiFiPhile:
I bought a nanoch32v203 (and v305), I do observe OUT EP send ACK blindly without checking if data is actually read 😱 😱 

Captured with CDC example:
image

I think to fix it correctly we need to disable `AUTO_TOG` and only toggle the PID once data is actually received.

HiFiPhile:
I've included a large transfer reliability patch for both USBFS/USBHS driver.

- Resolve race condition on control transfer status OUT / next SETUP
- Spurious transfer when EP is not ready

Tested on CH32V203/CH32V305 with `audio_4_channel_mic`, `cdc_dual_ports`, `dfu`, `cdc_msc`, `net_lwip_webserver` examples.

Concurrent ISO transfer is not working yet, eg `uac2_headset`

HiFiPhile (approved):

PR #3626, opened 2026-04-29
musb: improve endpoint 0 implementation approach

tzy0002088:
All interrupt flags for MUSB Endpoint 0 (EP0) are contained within a single register. Occasionally, multiple EP0 states may be satisfied simultaneously, causing them to be merged into a single interrupt request sent to the CPU. In such cases, the driver must be capable of handling these concurrent conditions. 

Due to heavy system load, global interrupts might be disabled elsewhere. Since the MUSB IP utilizes level-triggered interrupts, multiple internal events that should have triggered separate interrupts may become pending. Consequently, the CPU services them as a single interrupt request once global interrupts are re-enabled.

Instead of asserting distinct interrupts for the Data and Status stages in the blue region, the MUSB controller aggregated these pending flags into the following Setup stage interrupt request.

Capture of normal transfer without racing, IRQ (blue trace) triggered after each transfer stage.
83c26e364686c32ada705f8e2feb8406

When CPU is heavily loaded, IRQs are not served in time, we see only one IRQ fired at the end with multiples flags pending.
bcb497da-5c96-47c8-b38f-a037ea9e5c74

The Status stage complete interrupt was delayed into the Setup stage of the subsequent packet, followed by the triggering of the Setup interrupt

f839dde167fecffcc530a72f2ea85442

### DIFF
--- a/src/portable/mentor/musb/dcd_musb.c
+++ b/src/portable/mentor/musb/dcd_musb.c
@@ -80,22 +80,24 @@ typedef struct {
   #define MUSB_PIPE_COUNT (2u * (TUP_DCD_ENDPOINT_MAX - 1u))
 #endif
 
-enum {
-  PIPE0_STATE_IDLE = 0,         // no active control transfer
-  PIPE0_STATE_DATA,             // DATA stage (IN or OUT — direction implied by CSR/dir)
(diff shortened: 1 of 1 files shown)


HiFiPhile:
@hathath PR author raised EP0 racing issue to me prior #3605 in a discussion group. He made the change but the branch is too old (with their own SOC) to be applied cleanly, so he has extracted the EP0 part.

PR #3632, opened 2026-05-04
hcd/dwc2: fix txfifo full check

HiFiPhile:
also add a msc_file_explorer_freertos example to test with esp32

fix #3623, DMA is preferred anyway.

PR #3637, opened 2026-05-10
host: USB/IP-over-Ethernet bridge example, plus DWC2 / control_xfer fixes

### LINKED ISSUE #3623: USB HOST MSC hangs on write10
### Operating System

MacOS

### Commit SHA

096458defffe3ca183169f9e2116c322074f4b1e

### Board

ESP32S3-USB-OTG

### Firmware

The MSC File Explorer example from the repository:
https://github.com/hathach/tinyusb/tree/master/examples/host/msc_file_explorer

### What happened ?

I can read the files, list directories (so read works as expected), but when I want to write, it just hangs on the `write10` method forever.

pinging @me-no-dev

### How to reproduce ?

Flash the MSC Explorer example.

```
Generic  Flash Disk       rev 8.07
Disk Size: 15200 MB
[1] Claimed EP 0x01
  Queue EP 01 with 31 bytes ... 
OK
[:1] on EP 01 with 31 bytes: OK
  MSC xfer callback
  Queue EP 82 with 512 bytes ... 
OK
[:1] on EP 82 with 512 bytes: OK
  MSC xfer callback
  Queue EP 82 with 13 bytes ... 
OK
[:1] on EP 82 with 13 bytes: OK
  MSC xfer callback
[1] Claimed EP 0x01
  Queue EP 01 with 31 bytes ... 
OK
[:1] on EP 01 with 31 bytes: OK
  MSC xfer callback
  Queue EP 82 with 512 bytes ... 
OK
[:1] on EP 82 with 512 bytes: OK
  MSC xfer callback
  Queue EP 82 with 13 bytes ... 
OK
[:1] on EP 82 with 13 bytes: OK
  MSC xfer callback
[1] Claimed EP 0x01
  Queue EP 01 with 31 bytes ... 
OK
[:1] on EP 01 with 31 bytes: OK
  MSC xfer callback
  Queue EP 82 with 512 bytes ... 
OK
[:1] on EP 82 with 512 bytes: OK
  MSC xfer callback
  Queue EP 82 with 13 bytes ... 
OK
[:1] on EP 82 with 13 bytes: OK
  MSC xfer callback
```

### Screenshots

_No response_

andrewleech:
## Summary

Adds `examples/host/usbipd`, a USB/IP-over-Ethernet host bridge: a TinyUSB host that forwards URBs from an attached USB device over TCP/IP to a Linux `usbip` client. Once attached, the device shows up in the kernel's USB bus the same way a locally plugged one would (cdc-acm becomes `/dev/ttyACM<N>`, HID becomes an evdev node, etc).

The bridge core (`usbip_server.c`) is portable C against the public `tuh_*` and lwIP `tcp_*` APIs, so it should work on any TinyUSB host port with a TCP-capable lwIP stack. The board-specific glue is just the lwIP `netif` driver and PHY init; the included `ethernetif.c` plus the LAN8742 driver target the on-board RMII PHY of the STM32F429ZI / F439ZI Nucleo-144. Other boards bring their own.

Tested on STM32F439ZI Nucleo-144 with the on-board LAN8742A: full mpremote round-trip through the bridge to a MicroPython Pico W, plus a debugprobe cdc-acm device.

The PR also carries four DWC2/host-stack fixes the example exposed end-to-end. Without them, multi-packet bulk-OUT corrupts data and cdc-acm cleanup wedges on first close. The fixes are scoped tightly and useful on their own; the example is just the regression reproducer.

The first fix is a cherry-pick of @HiFiPhile's commit `17d24b39` from #3632. We landed on the exact same diff independently while bringing this up, so cherry-picking keeps attribution clean and means `git rebase` silently skips it once #3632 merges. Including it directly keeps the PR self-contained for reviewers running the example.

## What's in here

6 commits, bottom-up:

1. **`hcd/dwc2: fix txfifo full check`** (cherry-picked from @HiFiPhile's #3632) `handle_txfifo_empty` cached `fifo_available` / `req_queue_available` once before the per-channel loop; after writing one packet the cached values were stale, so multi-packet bulk-OUT could be issued without enough FIFO space and `XFER_COMPLETE` never fired. Re-read inside the loop. Without this, `mpremote fs cp` of a 3.7 KB file through the bridge fails with corrupted bulk-OUT data; with it, the same `fs cp` round-trips cleanly in ~3 s. Same hunk also fixes #3623.

2. **`hcd/dwc2: save post-transfer PID in DMA-mode IN handler`** `handle_channel_in_slave` saves `hctsiz.pid` into `edpt->next_pid` after `XFER_COMPLETE`; the DMA-mode IN handler did not. On a short packet ending a multi-packet IN early, the next URB started with the toggle pre-computed in `channel_xfer_start` from the requested packet count rather than the actual post-transfer PID, triggering `DATATOGGLE_ERR` on the next IN. Hardware either retried (dropping the device's first packet) or coalesced the duplicate (delivering corrupt bytes). OUT is exempt because the host drives the toggle on OUT.

3. **`hcd/dwc2: fire xfer_complete callback after hcd_edpt_abort_xfer`** The function disabled the active channel without guaranteeing an `hcd_event_xfer_complete` for the in-flight transfer. The channel-halted IRQ landed in `handle_channel_*_dma/slave` with no `XFER_COMPLETE / STALL / BABBLE / XACT_ERR` set, so `is_done` stayed false and the natural callback never fired. Callers needing a giveback after abort had to synthesise one externally.

   Mark the channel as aborted before disabling, then in `handle_channel_irq` force `is_done` with `XFER_RESULT_FAILED` when the abort flag is set and the channel is no longer enabled (distinguishes a real abort halt from a NAK retry that re-armed the channel). Also tightens the contract: returns `false` when no in-flight transfer was found, was previously unconditional `true`. The new cross-port contract is documented in `src/host/hcd.h`. Other ports (max3421e, rp2040, ehci, ohci, wch, musb, rusb2) haven't been audited against it yet, flagged for follow-up; existing implementations are unchanged here.

4. **`host: honour tuh_xfer_t.timeout_ms in tuh_control_xfer`** The `timeout_ms` field was a placeholder ("not supported yet"). The synchronous `tuh_control_xfer` path (`complete_cb == NULL`) was a busy loop with no timeout, so any device that NAK'd the control transfer indefinitely (e.g. cdc-acm on `CLEAR_FEATURE(ENDPOINT_HALT)`) wedged the caller. `timeout_ms == 0` keeps the historical wait-forever default for callers that don't initialise the field, so this is backwards-compatible. Non-zero returns `false` with `xfer->result = XFER_RESULT_FAILED` and aborts the EP0 channel via `hcd_edpt_abort_xfer` so the next control transfer on the same daddr can proceed. Relies on commit (3) for clean abort. `cdc_host` and `hid_host` switch to designated init for their fake `tuh_xfer_t` so the new field zero-initialises rather than holding stack garbage.

5. **`hw/bsp/stm32f4: add optional Ethernet support and LAN8742 PHY driver`** Opt-in `family_add_eth(TARGET)` helper for the F4 BSP, mirroring the existing `family_add_tinyusb` / `family_add_rtos` / `family_add_uf2`. When called from an example's CMakeLists it pulls in `stm32f4xx_hal_eth.c`, the LAN8742 driver, and the per-board `board_eth.c` (RMII pin map for `HAL_ETH_MspInit`); without the call, F4 boards build identically to before. The LAN8742 driver lives at `hw/mcu/st/stm32_lan8742/` (vendored from `STMicroelectronics/stm32-lan8742`, matches the existing `cmsis_device_f4` / `stm32f4xx_hal_driver` layout). `boards/stm32f439nucleo/board_eth.c` carries the F429ZI / F439ZI pin map. Verified `examples/host/cdc_msc_hid` builds clean at this commit, no ETH symbols pulled in.

6. **`examples/host/usbipd: add USB/IP-over-Ethernet host bridge`** The bridge itself. Single-client lwIP raw-API server on TCP/3240 implementing the kernel-compatible USB/IP wire format: `OP_REQ_DEVLIST`, `OP_REQ_IMPORT`, `CMD_SUBMIT (control + bulk + interrupt, IN/OUT)`, `CMD_UNLINK`. Endpoints are opened lazily via `tuh_edpt_open` with the descriptor sniffed from the kernel's `GET_DESCRIPTOR(CONFIG)` reply that flows through us, avoiding a sync descriptor fetch from inside the lwIP recv path. Includes a per-EP submit queue: cdc-acm queues 16 read-ahead URBs on its bulk-IN but TinyUSB allows only one URB per EP in flight, so the rejected ones get queued and drained from the completion callback when the EP frees.

   Vendored alongside the example (BSD-3-Clause, captured in `src/LICENSE.stm32-vendored`): `ethernetif.{c,h}` and `lwipopts.h` from STM32CubeF4 `LwIP_TCP_Echo_Server` with the DP83848 to LAN8742 PHY swap (same `_Object_t / _IOCtx_t / _Init / _GetLinkState` API); `arch/cc.h` reused from the existing `net_lwip_webserver` example.

## Hardware

- STM32F429ZI or STM32F439ZI Nucleo-144 (existing BSPs in `hw/bsp/stm32f4/boards/`).
- USB device under test plugged into the OTG_FS host port (CN13 USER USB on the Nucleo-144). VBUS is enabled by the BSP via PG6 in `board_init_after_tusb()`.
- Ethernet via the on-board LAN8742A connected by RMII.
- STLink VCOM for the example's diagnostic UART (USART3 on PD8/PD9).

## Build

```sh
cd examples/host/usbipd
mkdir build && cd build
cmake -G Ninja -DBOARD=stm32f439nucleo ..
ninja
```

54 KB FLASH, 93 KB RAM (the inflight buffer pool dominates: 32 slots x 1.5 KB).

## Verification

Tested on STM32F439ZI Nucleo-144 (S/N 066CFF495177514867213407) with two devices on OTG_FS:

- Raspberry Pi Pico Debugprobe firmware (vid 2e8a:000c): `usbip list -r <ip>` returns the device, `sudo usbip attach -r <ip> -b 1-1` enumerates and creates `/dev/ttyACM<N>`, pyserial open + write + read + close round-trips clean. Detach via `echo 0 > /sys/devices/platform/vhci_hcd.0/detach` is clean (three `RET_UNLINK status=0` messages, no rejected cascade).

- Raspberry Pi Pico W with MicroPython firmware (vid 2e8a:0005): full mpremote round-trip in a single chained process:

   ``` $ mpremote connect /dev/ttyACM12 \ exec "import sys, os; print(sys.platform, sys.version); print(os.uname())" \ + fs cp /tmp/test_script.py :test_script.py \ + fs ls \ + exec "exec(open('test_script.py').read())" \ + fs rm :test_script.py rp2 3.4.0; MicroPython v1.28.0-preview.147 ... (sysname='rp2', ..., machine='Raspberry Pi Pico W with RP2040') cp /tmp/test_script.py :test_script.py ls :  ...  147 test_script.py  ... hello from a script over usbip rm :test_script.py ```

To reproduce the failing baseline against stock upstream, revert the four fix commits and rebuild. The cdc-acm probe wedges instead of producing `/dev/ttyACM<N>`:

```
usbip: SUBMIT seq=185 ep=0x83: tuh_*_xfer rejected
usbip: SUBMIT seq=186 ep=0x83: tuh_*_xfer rejected
...
usbip: UNLINK seq=201 (cancels seq=184 ep=0x83)
usbip: SUBMIT seq=206 ep=0x83: tuh_*_xfer rejected   # claim never released
```

With the fixes applied, the same scenario gives one initial silent SUBMIT, drained queue entries, then a clean detach with three `RET_UNLINK status=0` messages.

## Notes for reviewers

- If the usbipd example feels too big or not a great fit for the examples tree, happy to publish it as a standalone project and drop commits 5 and 6 from this PR. The example was as much the reproducer that pinned down the host-stack issues in commits 1-4 as a contribution in its own right, so the four fixes can stand alone.

- Each fix commit is self-contained and can be cherry-picked individually. The `timeout_ms` commit relies on the abort_xfer commit for clean EP0 abort on timeout recovery; if the abort fix lands first the timeout fix is still a strict improvement on ports that don't yet honour the new abort contract (the timeout still fires, the recovery is just less complete).

- The Ethernet support is split into a reusable BSP helper (commit 5) and the example (commit 6) so other F4 examples can use the same `family_add_eth` opt-in. Boards without on-board Ethernet are unaffected; `family_add_eth` fails loudly at CMake time if the board's `board_eth.c` is missing rather than silently producing a non-functional binary.

- The example is gated to `board:stm32f439nucleo` via `only.txt` because runtime needs a board with both the LAN8742A RMII pin map and a working OTG_FS host port. Adding other F4 boards is just a new `board_eth.c` plus extending `only.txt`, the bridge core is portable.

- The `aborted` flag in `hcd_xfer_t` is a `:1` bitfield that fits in the existing flags byte, no growth in `_hcd_data`.

## Generative AI

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -10,13 +10,17 @@ repos:
   - id: trailing-whitespace
     exclude: |
       (?x)^(
-        hw/bsp/mcx/sdk/
+        hw/bsp/mcx/sdk/|
+        hw/mcu/st/stm32_lan8742/|
+        hw/bsp/stm32f4/eth_lwip/
       )
   - id: end-of-file-fixer
     exclude: |
       (?x)^(
         .idea/|
         hw/bsp/mcx/sdk/|
+        hw/mcu/st/stm32_lan8742/|
+        hw/bsp/stm32f4/eth_lwip/|
         docs/contributing/code_of_conduct.rst|
         docs/info/contributors.rst
       )
@@ -30,7 +34,9 @@ repos:
     exclude: |
       (?x)^(
         lib/|
-        hw/bsp/mcx/sdk/
+        hw/bsp/mcx/sdk/|
+        hw/mcu/st/stm32_lan8742/|
+        hw/bsp/stm32f4/eth_lwip/
       )
 
 - repo: local
--- a/examples/host/usbipd/CMakeLists.txt
+++ b/examples/host/usbipd/CMakeLists.txt
@@ -1,0 +1,110 @@
+cmake_minimum_required(VERSION 3.20)
+
+include(${CMAKE_CURRENT_SOURCE_DIR}/../../../hw/bsp/family_support.cmake)
+
+# gets PROJECT name for the example (e.g. <BOARD>-<DIR_NAME>)
+family_get_project_name(PROJECT ${CMAKE_CURRENT_LIST_DIR})
+
+# lwIP comes from the bundled tinyusb submodule (matches the
+# net_lwip_webserver device example pattern).
+set(LWIP ${TOP}/lib/lwip)
+if (NOT EXISTS ${LWIP}/src)
+  set(LWIP ${TINYUSB_LWIP_PATH})
+endif()
+if (NOT EXISTS ${LWIP}/src)
+  family_example_missing_dependency(${PROJECT} "lib/lwip")
+  return()
+endif()
+
+project(${PROJECT} C CXX ASM)
+
+# Checks this example is valid for the family and initializes the project
+family_initialize_project(${PROJECT} ${CMAKE_CURRENT_LIST_DIR})
+
+# Espressif has its own cmake build system; not supported in this example.
+if(FAMILY STREQUAL "espressif")
+  return()
+endif()
+
+add_executable(${PROJECT})
+
+# Source files for the example. main.c drives tuh_task() +
+# ethernetif_input() + sys_check_timeouts() in a single loop. The
+# usbip_*.c files split the bridge into: server (TCP callbacks +
+# framing), devices (cache + DEVLIST/IMPORT), inflight (URB pool +
+# completion), submit (CMD_SUBMIT/UNLINK dispatch).
+target_sources(${PROJECT} PUBLIC
+  ${CMAKE_CURRENT_SOURCE_DIR}/src/main.c
+  ${CMAKE_CURRENT_SOURCE_DIR}/src/usbip_server.c
+  ${CMAKE_CURRENT_SOURCE_DIR}/src/usbip_devices.c
+  ${CMAKE_CURRENT_SOURCE_DIR}/src/usbip_inflight.c
+  ${CMAKE_CURRENT_SOURCE_DIR}/src/usbip_submit.c
+  )
+
+target_include_directories(${PROJECT} PUBLIC
+  ${CMAKE_CURRENT_SOURCE_DIR}/src
+  ${LWIP}/src/include
+  )
+
+# Pull in the family's Ethernet support: MAC driver, PHY driver and
+# the lwIP netif glue, plus the board's PHY pin map. Defined per
+# family in hw/bsp/<family>/family.cmake.
+family_add_eth(${PROJECT})
+
+# lwIP raw API + DHCP. Mirrors the net_lwip_webserver source list,
+# minus the device-side networking helpers (dhserver, etc).
+target_sources(${PROJECT} PUBLIC
+  ${LWIP}/src/core/def.c
+  ${LWIP}/src/core/inet_chksum.c
+  ${LWIP}/src/core/init.c
+  ${LWIP}/src/core/ip.c
+  ${LWIP}/src/core/mem.c
+  ${LWIP}/src/core/memp.c
+  ${LWIP}/src/core/netif.c
+  ${LWIP}/src/core/pbuf.c
+  ${LWIP}/src/core/raw.c
+  ${LWIP}/src/core/stats.c
+  ${LWIP}/src/core/sys.c
+  ${LWIP}/src/core/tcp.c
+  ${LWIP}/src/core/tcp_in.c
+  ${LWIP}/src/core/tcp_out.c
+  ${LWIP}/src/core/timeouts.c
+  ${LWIP}/src/core/udp.c
+  ${LWIP}/src/core/ipv4/autoip.c
+  ${LWIP}/src/core/ipv4/dhcp.c
+  ${LWIP}/src/core/ipv4/etharp.c
+  ${LWIP}/src/core/ipv4/icmp.c
+  ${LWIP}/src/core/ipv4/igmp.c
+  ${LWIP}/src/core/ipv4/ip4.c
+  ${LWIP}/src/core/ipv4/ip4_addr.c
+  ${LWIP}/src/core/ipv4/ip4_frag.c
+  ${LWIP}/src/netif/ethernet.c
+  )
+
+# All TinyUSB class drivers are disabled in tusb_config.h - this bridge
+# forwards URBs at the endpoint level. BUILTIN_DRIVER_COUNT==0 trips
+# upstream's -Wtype-limits in usbh.c (drv_id < 0 always-false). Silence
+# locally; the example's own translation units stay strict.
+target_compile_options(${PROJECT} PRIVATE -Wno-type-limits)
+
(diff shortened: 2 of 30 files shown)


HiFiPhile:
Hi, it looks very nice !

Just for the example I think it's better to include a host/device dual port approach with NCM device + USBIP host, in this way it can be tested on more MCUs. F439 is a very old MCU which I don't have, I don't know if Ha Thach has it but it's not in the HIL test farm anyway.

PS: I'm thinking about adding a `Awesome projects` section to readme for external projects which are not trivial to maintain.

PR #3639, opened 2026-05-12
ci: MemBrowse - add libraries mapping support

michael-membrowse:
MemBrowse now supports attributing symbols to static libraries in the Symbol Analysis page.
For that the membrowse cli is parsing map files.
This PR adds this support.


HiFiPhile (approved):
Thank you !

PR #3641, opened 2026-05-13
Add GetPartialObject to MTP example

wjklimek1:
Currently MTP example works fine under Windows, but Linux file explorers had problems with opening files directly from MTP drive. Copying entire file to local drive was working though.

This is most likely due to fact that Windows uses GetObject to read entire file in one transfer and Linux tries to use GetPartialObject which is not implemented in this example.

This PR fixes this problem by implementing GetPartialObject handler. Implementation is very similar to existing GetObject, but takes file offset from host into the account.

I successfully tested it on stm32f411blackpill and stm32h503nucleo under Linux Mint (command line file access and Nemo file explorer). RAM usage increased by 8 bytes, FLASH usage increased by 200 bytes on STM32H503.


HiFiPhile (approved):
Thank you !

PR #3643, opened 2026-05-14
dcd/musb: defer EP0 SETUP during DATA_IN/STATUS race

HiFiPhile:
Handle cases where a new SETUP arrives before the previous control transfer fully completes by buffering the SETUP and replaying it after status completion. Split EP0 DATA state into DATA_IN/DATA_OUT and finalize pending status-out completion before processing deferred SETUP.

Take care these 2 cases if new SETUP packet arrived while old control transfer is not finished yet. This could happen in following scenarios when CPU load is high:
- Status IN/OUT finished, IRQ and new setup packet IRQ arrive at the same time.
- Data IN finished and status OUT is received, both IRQs and new setup packet IRQ arrive at the same time.

Supercede #3626 

Tested intensively with DFU example with Interrupt turned on/off every 20ms simulating high CP load:
```diff
diff --git a/examples/device/dfu/src/main.c b/examples/device/dfu/src/main.c
index fb3c22630..d4e75a44b 100644
--- a/examples/device/dfu/src/main.c
+++ b/examples/device/dfu/src/main.c
@@ -45,6 +45,7 @@
 #include "bsp/board_api.h"
 #include "tusb.h"
 
+#include "device/dcd.h"
 //--------------------------------------------------------------------+
 // MACRO CONSTANT TYPEDEF PROTYPES
 //--------------------------------------------------------------------+
@@ -65,6 +66,7 @@ enum {
 static uint32_t blink_interval_ms = BLINK_NOT_MOUNTED;
 
 void led_blinking_task(void);
+static void musb_test(void);
 
 /*------------- MAIN -------------*/
 int main(void) {
@@ -79,6 +81,7 @@ int main(void) {
   while (1) {
     tud_task(); // tinyusb device task
     led_blinking_task();
+    musb_test();
   }
 }
 
@@ -86,6 +89,26 @@ int main(void) {
 // Device callbacks
 //--------------------------------------------------------------------+
 
+// on off usb interrupt 20ms for testing
+static void musb_test(void) {
+  static uint32_t start_ms  = 0;
+  static bool irq_enabled = true;
+
+  if (tusb_time_millis_api() - start_ms < 20) {
+    return; // not enough time
+  }
+
+  if (irq_enabled) {
+    dcd_int_disable(0);
+    tusb_time_delay_ms_api(20);
+    dcd_int_enable(0);
+  } else {
+  }
+
+  irq_enabled = !irq_enabled;
+  start_ms += 20;
+}
+
 // Invoked when device is mounted
 void tud_mount_cb(void) {
   blink_interval_ms = BLINK_MOUNTED;
@@ -122,7 +145,7 @@ uint32_t tud_dfu_get_timeout_cb(uint8_t alt, uint8_t state) {
     // For this example
     // - Atl0 Flash is fast : 1   ms
     // - Alt1 EEPROM is slow: 100 ms
-    return (alt == 0) ? 1 : 100;
+    return (alt == 0) ? 0 : 0;
   } else if (state == DFU_MANIFEST) {
     // since we don't buffer entire image and do any flashing in manifest stage
     return 0;
@@ -142,9 +165,9 @@ void tud_dfu_download_cb(uint8_t alt, uint16_t block_num, const uint8_t *data, u
 
   //printf("\r\nReceived Alt %u BlockNum %u of length %u\r\n", alt, wBlockNum, length);
 
-  for (uint16_t i = 0; i < length; i++) {
-    printf("%c", data[i]);
-  }
+  //for (uint16_t i = 0; i < length; i++) {
+  //  printf("%c", data[i]);
+  //}
 
   // flashing op for download complete without error
   tud_dfu_finish_flashing(DFU_STATUS_OK);
```


src/portable/mentor/musb/dcd_musb.c
```diff
@@ -463,31 +490,54 @@ static void process_ep0(uint8_t rhport) {
...
         break;
       }
 
-      default: break;
+      // New SETUP packet arrived  while old control transfer is not finished yet. This could happen in following scenarios:
+      // - Status IN/OUT finished, IRQ and new setup packet IRQ arrive at the same time.
+      // - Data IN finished and status OUT is received, both IRQs and new setup packet IRQ arrive at the same time.
+      // could happen when CPU load is high, save the new setup packet for later processing after current status stage complete.
+      case PIPE0_STATE_STATUS_OUT:
+      case PIPE0_STATE_STATUS_OUT_PENDING:
```
hathach: **[bug — race]** Deferring in `STATUS_OUT_PENDING` when it was reached **IRQ-first** (process_ep0 `STATUS_OUT` case promoted it, usbd has *not* yet called `edpt0_xfer(STATUS OUT)`) consumes the status pairing early and replays the SETUP before usbd has retired the old transfer:

1. IRQ1: last DATA IN sent → `STATUS_OUT`, queues `xfer_complete(EP0_IN)`
2. IRQ2: status ZLP confirm (csrl==0) → `STATUS_OUT_PENDING`
3. IRQ3: new OUT-data SETUP (e.g. `SET_LINE_CODING`) → deferral case → `goto process_status` fires `xfer_complete(EP0_OUT,0)`, goes IDLE, **replays the SETUP → state = `DATA_OUT`**
4. usbd task now processes the queued DATA-IN completion → `status_stage_xact()` → `edpt0_xfer(EP0_OUT, NULL, 0)` lands in `case DATA_OUT`: sets `_dcd.pipe0.buf = NULL` and writes `RXRDYC`, un-NAKing the host
5. Host's real DATA OUT arrives → `tu_hwfifo_read(fifo, NULL, count0)` → write through NULL

usbd has no stale-event discard for EP0 `XFER_COMPLETE` (usbd.c event loop processes them unconditionally), so nothing defuses this. The `STATUS_OUT` deferral case is fine (its pairing still routes through `edpt0_xfer`); the problem is only the `STATUS_OUT_PENDING`-via-IRQ-first ordering. Consider: in the deferral cases only *save* the packet and synthesize the missed pairing state, but never fire the status completion + replay until the usbd-driven `edpt0_xfer` confirms the old transfer is retired. A `TU_ASSERT(_dcd.pipe0.buf)` backstop in the `DATA_OUT` drain path would also catch this class cheaply.

src/portable/mentor/musb/dcd_musb.c
```diff
@@ -463,31 +490,54 @@ static void process_ep0(uint8_t rhport) {
...
+        union {
+          tusb_control_request_t req;
+          uint32_t               u32[2];
+        } setup_packet;
+        setup_packet.u32[0] = musb_regs->fifo[0];
+        setup_packet.u32[1] = musb_regs->fifo[0];
+
+        _dcd.pipe0.deferred_setup = setup_packet.req;
+        _dcd.pipe0.deferred_setup_valid = true;
+        goto process_status;
```
hathach: **[bug — race]** The deferral path drains the SETUP from the FIFO but never writes `ServicedRxPktRdy`, and the replay also leaves RXRDY set for OUT/zero-length requests. Per spec, IntrTx is **clear-on-read** and the EP0 interrupt latches when RxPktRdy is *set* (§21.1.5 cause 1) — so if the SETUP's IRQ latches after the ISR's `intr_tx` read (dcd_int_handler) but before this function's `csrl` read, a **second invocation** of `process_ep0` is guaranteed, and it observes stale state:

- OUT-data replay (state `DATA_OUT`, RXRDY still set, FIFO drained): `count0` reads 0 → fires a **spurious `xfer_complete(EP0_OUT, 0)`** which usbd interprets as a ZLP ending the *new* transfer's data stage while the host is still sending real data.
- IN-data replay (RXRDYC written at replay, csrl==0, state `DATA_IN`): falls into `process_status` `DATA_IN` case → spurious `xfer_complete(EP0_IN, stale xact_len)` for a transfer usbd hasn't started.
- Zero-length replay (state `STATUS_IN`, RXRDY set): re-enters the deferral case on a drained FIFO → `TU_ASSERT(count0 == 8)` fires, or garbage is saved as a deferred SETUP with asserts off.

Note the canonical flow (spec Fig 21-3) writes ServicedRxPktRdy immediately after unloading; the driver's delayed-ack NAK flow-control is fine, but the deferral path needs a guard against re-entry while RXRDY is parked (e.g. skip the RXRDY block when `deferred_setup_valid`, or track a 'rxrdy consumed' flag).

src/portable/mentor/musb/dcd_musb.c
```diff
@@ -410,9 +442,14 @@ static void process_ep0(uint8_t rhport) {
   musb_ep_csr_t* ep_csr = get_ep_csr(musb_regs, 0);
   uint_fast8_t csrl = ep_csr->csr0l;
 
+  if (csrl & MUSB_CSRL0_DATAEND) {
```
hathach: **[bug risk — spec conflict]** Three issues with this guard, from the MUSBMHDRC spec (PSPG 40161.003):

1. **Ordering**: §21.1.5 mandates *“Whenever the Endpoint 0 service routine is entered, the firmware must **first** check … a STALL condition or a premature end of control transfer”* (Fig 21-2: SentStall? → SetupEnd? → state dispatch). This return executes before both. That matters because the core's auto-STALL conditions 1–2 (§21.1.7) are literally *“the host sends an OUT/IN token **after the DataEnd bit has been set**”* — i.e. SentStall is most likely to fire exactly when DataEnd may still read back set, and this guard would skip clearing it (IntrTx already consumed → recovery lost).
2. **Read-back is unspecified**: the CSR0L access table (§3.3.1, p.37) marks DataEnd as CPU `set`-only (unlike TxPktRdy `r/set`), while the page note says reads *“reflect the status attained … as a result of writing”*. So whether this bit ever reads 1 is vendor-dependent — on cores where it reads 0 the guard is dead code and the premature-STATUS_IN race it targets is unguarded.
3. **If it does read back set**: the TxPktRdy-cleared interrupt of the **last** DATA IN packet (armed `TXRDY|DATAEND` per Fig 21-4) arrives before the status stage runs; dropping it means the later csrl==0 status IRQ gets consumed as the DATA-IN completion, leaving the STATUS_OUT pairing one event short — EP0 parks in `STATUS_OUT_PENDING` until the next SETUP rescues it via the deferral path.

Suggest moving the check after the STALLED/SETEND handling and adding a comment stating which hardware state it encodes + the read-back caveat.

src/portable/mentor/musb/dcd_musb.c
```diff
@@ -804,6 +857,7 @@ void dcd_edpt_stall(uint8_t rhport, uint8_t ep_addr) {
     if (ep_addr == TU_EP0_OUT) { /* Ignore EP0 OUT */
       _dcd.pipe0.state = PIPE0_STATE_IDLE;
       _dcd.pipe0.buf = NULL;
+      _dcd.pipe0.deferred_setup_valid = false;
```
hathach: **[bug — protocol]** Clearing `deferred_setup_valid` here discards a SETUP the hardware already ACKed on the wire. Reachable interleaving: DATA IN completes and the next SETUP gets captured into `deferred_setup`; the class driver then rejects the old request at `CONTROL_STAGE_DATA` → usbd calls `dcd_edpt_stall(EP0)` → the deferred SETUP is wiped and `SendStall` armed. The host, whose new SETUP was ACKed, proceeds with that request's data/status stage and receives STALL — an innocent request fails host-side with no `tud` callback ever seeing it. Spec §21.1.7 frames SendStall as aborting *the current transfer*; here it lands on the next one. Worth either replaying the deferred SETUP after the stall handshake completes, or at minimum documenting the drop (the host will retry in practice, but it surfaces as a transient control failure).

test/hil/hil_test.py
```diff
@@ -243,6 +247,26 @@ def open_serial_dev(port: str):
...
 
 
+def serial_write_all(ser: serial.Serial, data: bytes, deadline: float = SERIAL_WRITE_DEADLINE):
+    total = 0
+    end = time.monotonic() + deadline
+
+    while total < len(data):
+        try:
+            written = ser.write(data[total:])
+        except serial.SerialTimeoutException:
```
hathach: **[bug — test harness]** Treating `SerialTimeoutException` as "wrote 0 bytes" and retrying from the same `total` can duplicate bytes on the wire: pyserial's posix `write()` loops `os.write()` internally and **raises after partial progress with the count lost** (`d = d[n:]` progress is discarded by the raise). Realistic when the kernel tty buffer is nearly full because the device drains slowly: part of a chunk is pushed, the 2 s `write_timeout` expires, the retry re-sends the whole remaining slice → duplicated bytes → the CDC echo / `write_and_check` comparisons fail with data-mismatch asserts that look like device firmware bugs.

Safer: treat the timeout as fatal (the pre-PR behavior, just with the configurable timeout), or write in chunks small enough that a timeout implies ~nothing was sent.

hathach (commented):
## Code review — high effort, cross-checked against MUSBMHDRC spec

Reviewed the EP0 deferred-SETUP state machine and the HIL serial changes (7 finder angles → 1-vote verify), then cross-checked the DCD findings against the **MUSBMHDRC Product Specification & Programmer's Guide** (Mentor PSPG 40161.003, §3.3.1 CSR0L, §21.1.5 EP0 service routine, §21.1.7 additional actions).

The deferral design is sound for the `STATUS_OUT` and coalesced `DATA_IN`/`STATUS_IN` paths — the spec confirms a post-status SETUP arrives as plain `RXRDY` with no `SETEND`, and the `TU_ASSERT(count0 == 8)` in the deferral cases is spec-sound (status ZLPs don't raise RxPktRdy; oversized status packets auto-STALL). Four correctness issues and a few cleanups below, ranked by severity.

**Checked and ruled out** (for the record):
- Task-context replay racing the ISR: `dcd_edpt_xfer` brackets `edpt0_xfer` with `musb_dcd_int_disable/enable`, so `pipe0_process_deferred_setup` from `edpt0_xfer` is serialized against the ISR. ✅
- `DATA_IN` deferral with `remain_wlength > 0`: spec-impossible — §21.1.5.2: *“If either a SETUP or an OUT token is received whilst the endpoint is in the TX state, this will cause a SetupEnd condition”*, and SetupEnd sets at the token, before RxPktRdy; `process_ep0` handles SETEND before the RXRDY switch. ✅
- `static MUSB_BASES`: only consumer is `MUSB_REGS()`; the change also fixes a duplicate-definition hazard for combined host+device TI builds. ✅

test/hil/hil_test.py
```diff
@@ -774,8 +797,7 @@ def test_host_msc_file_explorer(board):
     time.sleep(1)
     ser.reset_input_buffer()
     for ch in 'cat README.TXT\r':
-        ser.write(ch.encode())
-        ser.flush()
+        serial_write_all(ser, ch.encode())
```
hathach, line 789: **[behavior change]** The old `ser.write(ch); ser.flush()` guaranteed each character hit the wire (tcdrain) before the 2 ms pacing sleep; `serial_write_all` only guarantees it reached the kernel buffer. Under backpressure the kernel can coalesce several queued chars into one CDC bulk packet, defeating the per-character pacing this loop exists for (same at the `dd 1024` loop below, line 822). Cheap fix: keep a `ser.flush()` after `serial_write_all` at these two per-char call sites.

test/hil/hil_test.py
```diff
@@ -243,6 +247,26 @@ def open_serial_dev(port: str):
...
+        try:
+            written = ser.write(data[total:])
+        except serial.SerialTimeoutException:
+            written = 0
+
+        if written:
+            total += written
+            continue
+
+        if time.monotonic() >= end:
```
hathach: **[cleanup]** Deadline shape nits: (a) the deadline is only checked after a failed attempt, and a write started at t≈deadline still blocks a full `write_timeout`, so the effective bound is `deadline + write_timeout` (~12 s); (b) the 10 ms sleep is dead weight — `ser.write()` already blocks up to `write_timeout` per attempt, so the exception path already paces retries; (c) per-character call sites (`cat README.TXT\r` = 15 calls) each get a fresh 10 s deadline, so a sluggish device can stretch one command to minutes against `CMD_TIMEOUT=180`. Checking the deadline before the write, dropping the sleep, and sharing one deadline per command string would fail fast (≤10 s).

src/portable/mentor/musb/dcd_musb.c
```diff
@@ -463,31 +490,54 @@ static void process_ep0(uint8_t rhport) {
...
+      // New SETUP packet arrived  while old control transfer is not finished yet. This could happen in following scenarios:
+      // - Status IN/OUT finished, IRQ and new setup packet IRQ arrive at the same time.
+      // - Data IN finished and status OUT is received, both IRQs and new setup packet IRQ arrive at the same time.
+      // could happen when CPU load is high, save the new setup packet for later processing after current status stage complete.
+      case PIPE0_STATE_STATUS_OUT:
+      case PIPE0_STATE_STATUS_OUT_PENDING:
+      case PIPE0_STATE_STATUS_IN:
+      case PIPE0_STATE_DATA_IN: {
+        TU_ASSERT(sizeof(tusb_control_request_t) == count0, );
+        union {
```
hathach: **[cleanup]** This 8-byte FIFO-read block (union + two `fifo[0]` reads + `TU_ASSERT` on count0) is duplicated verbatim from the `PIPE0_STATE_IDLE` case above. A small `pipe0_read_setup(musb_regs, &req)` helper next to `pipe0_start_setup()` would keep the two hardware drain sequences from drifting apart — a future fix applied to one copy but not the other would only manifest on the rare deferred-race path, where it's hardest to debug.

src/portable/mentor/musb/dcd_musb.c
```diff
@@ -463,31 +490,54 @@ static void process_ep0(uint8_t rhport) {
...
+        _dcd.pipe0.deferred_setup = setup_packet.req;
+        _dcd.pipe0.deferred_setup_valid = true;
+        goto process_status;
+      }
     }
 
     return;
   }
 
+process_status:
```
hathach: **[cleanup]** The `goto process_status` arrives here with `csrl != 0` (RXRDY set), so the guarding comment "When CSRL0 is zero…" is no longer true on that path. Extracting the completion switch into a helper called from both the deferral case and the csrl==0 path would remove the goto, keep the comment truthful, and incidentally match the spec's service-routine structure (Fig 21-2 dispatches on state after the abort checks).

src/portable/mentor/musb/dcd_musb.c
```diff
@@ -95,12 +96,41 @@ typedef struct {
...
+static void pipe0_start_setup(uint8_t rhport, musb_ep_csr_t* ep_csr,
+                              tusb_control_request_t const* req, bool is_isr) {
+  _dcd.pipe0.remain_wlength = req->wLength;
+
+  if (req->wLength == 0) {
+    _dcd.pipe0.state = PIPE0_STATE_STATUS_IN;
+  } else {
+    if (req->bmRequestType & TUSB_DIR_IN_MASK) {
+      _dcd.pipe0.state = PIPE0_STATE_DATA_IN;
+      ep_csr->csr0l = MUSB_CSRL0_RXRDYC;
```
hathach: **[cleanup]** The pre-PR comment explaining this asymmetry was dropped in the move: *“If OUT (rx) direction, let edpt0_xfer() clear RXRDY when it's ready to receive data.”* It's load-bearing — acking RXRDY for OUT before `edpt0_xfer` arms the drain buffer would let the host send data with nowhere to put it. It's also a deliberate deviation from the spec's canonical flow (Fig 21-3 sets ServicedRxPktRdy immediately after unload), used as NAK flow-control — worth restoring the comment so nobody "unifies" the branches later.

src/portable/mentor/musb/dcd_musb.c
HiFiPhile: Are you going to let claude work on it ?

hathach: Fixed in #3699 (ee3dd01): STATUS_OUT_PENDING is split into PENDING_XFER/PENDING_IRQ. In the IRQ-first ordering the deferral now only holds the SETUP; usbd's edpt0_xfer(STATUS OUT) fires the status completion and replays, so it can no longer land in the replayed transfer's state. The TU_ASSERT(pipe0.buf) backstop is in the DATA OUT drain.

src/portable/mentor/musb/dcd_musb.c
hathach: Fixed in #3699 (851ebe6): new rxrdy_parked flag set wherever a drained packet's RXRDY is intentionally left set (deferral, OUT/zlp flow control, DATA OUT packets) and cleared at every RXRDYC write site. The RXRDY block returns early while parked, so the guaranteed stale latch is a no-op; replayed IN requests keep RXRDY parked until the edpt0_xfer(DATA IN) arm acks it, which closes the csrl==0 stale-completion variant too.

src/portable/mentor/musb/dcd_musb.c
hathach: Fixed in #3699 (da749ee): SentStall/SetupEnd are checked first per §21.1.5, and the guard now sits below the RXRDY block as well — a coalesced DATAEND|RXRDY read must not swallow a SETUP on cores where DataEnd reads back 1. Comment documents the encoded hw state and the vendor-dependent read-back. On read-back-1 cores the late DATA-IN completion described in point 3 now recovers via the PENDING_XFER deferral case.

src/portable/mentor/musb/dcd_musb.c
hathach: Fixed in #3699 (92da00c): a deferred SETUP can only exist once the old transfer's status was seen on the wire, so SendStall would land on the host's next, already-ACKed request. dcd_edpt_stall(EP0 OUT) now skips arming the stall and replays the deferred SETUP instead.

test/hil/hil_test.py
hathach: Fixed in #3699 (2c4b905): serial_write_all writes the whole buffer once and treats SerialTimeoutException as fatal. write_timeout is already a total per-call deadline in pyserial, so the retry loop added duplication risk without extending the budget; default bumped to 10 s to keep the old overall bound.

test/hil/hil_test.py
hathach: Fixed in #3699 (2c4b905): ser.flush() restored after serial_write_all in the cat/dd per-character loops. Other call sites stay flush-free to avoid reintroducing the unbounded-tcdrain hang this PR was addressing.

test/hil/hil_test.py
hathach: Resolved in #3699 (2c4b905) by removing the loop entirely — single write with write_timeout as the total deadline; the late deadline check, the dead 10 ms sleep, and the per-call deadline reset all go away with it.

src/portable/mentor/musb/dcd_musb.c
hathach: Fixed in #3699 (6ecd5c6): extracted pipe0_read_setup(), used by both the IDLE and deferral paths.

src/portable/mentor/musb/dcd_musb.c
hathach: Fixed in #3699 (b7715a9): goto removed — the deferral block handles its per-state transitions inline, and the csrl==0 switch is only reached with csrl==0, so the comment is truthful again.

src/portable/mentor/musb/dcd_musb.c
hathach: Restored in #3699 (bf6f3fc) in pipe0_start_setup, with the NAK flow-control rationale and a note on the replay path.

src/portable/mentor/musb/dcd_musb.c
hathach: @HiFiPhile no way, I aim to manual review all PRs. LLM is only for making it faster/easier. I am just testing a self-loop setup that have it xhigh effort review and auto-fix since it take pretty long (runing at night). Seem like I give it too much freedom :D

hathach: btw, I think its suggestion to split STATUS_OUT_PENDING TO STATUS_OUT_PENDING_XFER/IRQ makes sense. There is indeed a race.

HiFiPhile: Yeah I think twice and there is a race, I didn't check the fix though.

hathach (approved):
perfect, thank you very much for yet another the excellent PR

Issue #3573, opened 2026-03-26
[CH32V203] DFU download hangs on larger files with PMAOVR (PMA Overrun) interrupt


Devin-goBILDA:
### Operating System

Windows 11

### Commit SHA

Release 0.20.0

### Board

nanoCH32V203 & WeActStudio BluePill Plus CH32

### Firmware

Here is a repo containing the bare dfu example found in examples/device/dfu that was slightly modified to work with MounRiver Studio II. https://github.com/Devin-goBILDA/CH32V203-TinyUSB-DFU-Issue.

This issue is reproducible with the stock examples/device/dfu example code running on a ch32v203. 

### What happened ?

I am in the process of creating a DFU bootloader with using both TinyUSB and a CH32V203. I setup TinyUSB with my MounRiver Studio II project and compiled the dfu example code found in examples/device/dfu. I then navigated to https://devanlai.github.io/webdfu/dfu-util/ in order to transfer few small files (512 bytes), that worked flawlessly! I then continued to setup my bootloader project and when it was finally time to send a relatively basic firmware file (roughly 9000-ish bytes) the transfer would hang. Upon using the debugger I noticed that every time it gets hung on a TU_BREAKPOINT in the dcd_stm32_fsdev.c.

Link to code in 0.20.0: [dcd_int_handler](https://github.com/hathach/tinyusb/blob/0.20.0/src/portable/st/stm32_fsdev/dcd_stm32_fsdev.c?plain1#L471)
```C
void dcd_int_handler(uint8_t rhport) {
  uint32_t int_status = FSDEV_REG->ISTR;

  /* Commented out Code */

  if (int_status & USB_ISTR_PMAOVR) {
    TU_BREAKPOINT();                           /* <- This is where is breaks */
    FSDEV_REG->ISTR = (fsdev_bus_t)~USB_ISTR_PMAOVR;
  }
}
```

Here is exactly where the PMAOVR is defined: [USB_ISTR_PMAOVR](https://github.com/hathach/tinyusb/blob/0.20.0/src/portable/st/stm32_fsdev/fsdev_ch32.h?plain1#L67)
```
#define USB_ISTR_PMAOVR                      ((uint16_t)0x4000U)               /*!< DMA OVeR/underrun (clear-only bit) */
```

I am not super familiar with the lower level aspects of USB, but I would love to help in any way possible.

### How to reproduce ?
* Using MounRiver Studio II
[...]

PR #3648, opened 2026-05-24
dcd/stm32_fsdev: workaround CH32 EP0 premature OUT ACK

HiFiPhile:
CH32 EP0 blindly ACK all OUT packets, even not only ZLP.

Track EP0 control transfer state from SETUP (direction and wLength) and gate CH32 EP0 handshakes accordingly. Switch EP0 type away from control to bulk to block unsolicited OUT ACKs during control read/write gaps, then restore control type when EP0 transfer is explicitly armed.

Mainly fix DFU issue.


hathach:
@HiFiPhile where do you got this "errata" also, which specific family that got this ?

HiFiPhile:
> @HiFiPhile where do you got this "errata" also, which specific family that got this ?

I didn't ask WCH for errata, just tested DFU issues on ch32v203 and found the ack behavior is not sticked to the reference manual.

All out data packets are acked even before the transfer is launched.

Same issue in #3573

hathach:
> > @HiFiPhile where do you got this "errata" also, which specific family that got this ?
> 
> I didn't ask WCH for errata, just tested DFU issues on ch32v203 and found the ack behavior is not sticked to the reference manual.
> 
> All out data packets are acked even before the transfer is launched.
> 
> Same issue in #3573



> > @HiFiPhile where do you got this "errata" also, which specific family that got this ?
> 
> I didn't ask WCH for errata, just tested DFU issues on ch32v203 and found the ack behavior is not sticked to the reference manual.
> 
> All out data packets are acked even before the transfer is launched.
> 
> Same issue in #3573

Ah thanks, I thought you obtain some secret docs or get it from the chinese version. Indeed, I got into random issue with v203 when adding an board to hil pool as well, and have to leave it out (under the nanoch32v203 under board-skip). This may be the reason. Still trying to catch up with your pending PRs. Thank you for all the work so far.

hathach (approved):
perfect, thank you very much. Great PR as usual. Now we have stable usb working with ch32 mcu

PR #3652, opened 2026-05-26
ch32_usbhs: fix endpoint stall length index and clear-stall response

rhgndf:
Was debugging a ch32v305 firmware and claude pointed this out.

Fix issue in the stall handling:

- dcd_edpt_stall() for an IN endpoint cleared EP_TX_LEN(0) instead of EP_TX_LEN(ep_num), clobbering endpoint 0's transmit length register when stalling any other IN endpoint.


HiFiPhile (approved):
Thank you, nice catch up.

PR #3653, opened 2026-05-27
add device specific issues

HiFiPhile:
With LPC54600 and WCH CH32X issues inplace and a link from supported device table.

@hathach due to numerous issues we had about Pi-PIO, maybe also note it's limitations ?


rhgndf:
Does this also affect CH32H417? I am working on a port for it #3642 and possibly a superspeed one.

HiFiPhile:
@rhgndf I didn't ask for H417. For FS I think it's the same, for HS I hope it's not jugging from the new register layout where each EP has own status and length fields.
As `RB_UDIF_RTX_ACT` is cleared by `RB_UEP_R_DONE`/`RB_UEP_T_DONE` flag, I hope `INT_ST` will update each time the flag is cleared so we can easily use a loop to process all completed EPs.

HiFiPhile:
@rhgndf  After more testing the situation is not that bad, data loss will only occurs on ISO transfers.

For other transfer types with `USBHS_INT_BUSY_EN=1` the controller will set all EPs to NAK once transfer complete IRQ is fired.

In this capture:
1. first 2 PINGs checked that both OUT EPs are ready to receive
2. EP3 received packet, IRQ fired after ACK
3. EP2 reply with NAK automatically
4. IRQ is finished
5. EP2 change back to ACK and receive data

image

I will add more detail to the doc and add the issue of LPC54600.

hathach:
thank @HiFiPhile I think you forgot to add the actual errata doc

HiFiPhile:
> thank @HiFiPhile I think you forgot to add the actual errata doc

I should rename it to device specific issues instead of errata.

hathach:
> > thank @HiFiPhile I think you forgot to add the actual errata doc
> 
> I should rename it to device specific issues instead of errata.

yeah, I think it is a better name, may be device/cpu known issues. We may make an mistake especially when there is no offical errata doc.

HiFiPhile:
I forgot the document itself...

hathach:
@claude review

hathach:
@claude fix your own above review

hathach:
@HiFiPhile sorry, trying to test claude to make changes, it is a bit aggressive toward origin fork since it think this may be a local branch. Could tell it to make pr to pr instead.

hathach:
@claude fix claude code review ci workflow failed

HiFiPhile:
@claude Revert Fix "Qualify USBFS non-ISO transfer claim — replaced blanket "Other types of transfers are not affected" with: USBHS devices are protected by USBHS_INT_BUSY_EN; USBFS behavior is not yet confirmed"

hathach (approved):
perfect, thank you

hathach (commented):
PS: PIO-USB is software bit-banging usb host, so yeah, it will have issues. But it is fixable, since it is entirely software. Its limitation is probably it takes lots of efforts and testing to get the timing and code right to fix an issue.

PR #3654, opened 2026-05-27
bsp/stm32f4: enable flash cache

HiFiPhile:
Which increase `cdc_msc_throughput` speed from 500kB/s to 900kB/s.

Was wondering why `dcd_event_xfer_complete` takes 4us.


HiFiPhile:
Included in #3634

Issue #3591, opened 2026-04-09
Improve host control transfer handling


HiFiPhile:
Currently USBH only manage one control transfer:
- Enumeration will fail if we're unlucky that we are setting the DTR of another device or something.
- Vice versa, when we try to do ctrl transfer during enumeration, what's worse is since `usbh_get_enum_buf()` is not locked, enumeration data can be corrupted.

To fix it there are multiples way with pros/cons:

1. Move `_usbh_epbuf` and `ctrl_xfer_info` into each device struct
Pro: Fully concurrent, would be nice if one day we add DFU host.
Con: Waste of space as ctrl transfer is hardly used after enumeration.

2. Add new transfer to event queue if busy
Pro: Les RAM
Con: Transfer buffer need to be static to live long enough, not resolve `usbh_get_enum_buf()` issue

3. One transfer buffer for enumeration, another for driver
Pro: Less space needed than adding buffer to each device.
Con: Not resolve device to device concurrency

@hathach what do you think ?

ceedriic:
Hello, I've been thinking about that the last few days, and here are some thoughts:

I think the problem of functions like `tuh_cdc_set_line_coding()` should be handled differently if used asynchronously (with or without rtos) or synchronously (with rtos).

For asynchronous calls, I think the allocation of all the required memory should be handled outside of the tinyusb library, so the application can decide what they wants.

An application could allocate only one static buffer and call all control functions one at a time, or allocate one buffer per device, or one per interface, or on demand using malloc/free or with a buffer pool or any other solution depending on the available memory and the concurrency needed.

TinyUSB would provide a definition that look like:

```
typedef struct {
    usbh_ctrl_xfer_info_t ctrl_xfer_info; // control transfer
    tuh_xfer_cb_t complete_cb;		  // user callback
    uintptr_t user_data;                  // callback data
    int state;				  // 0 = unused, 1 = inflight, 2 = success, 3 = error */    
    tusb_control_request_t request;
    union {
        /* whatever extra data is needed here for various calls */
        uint8_t buf[64];
    }
} tuh_async_storage_t;
```

Then the user application could do something like:

```
// one cdc_control operation at a time
static tuh_async_storage_t cdc_control;

// request a change in line_coding of device 0;
static cdc_line_coding_t *line_coding;

// main application loop
for (;;) {
   tud_task();
   tuh_task();

   if (line_coding != NULL && cdc_control.state != 1) {
       // reset our structure
[...]

PR #3657, opened 2026-05-29
Add control transfer fifo for host stack

hathach:
## Summary

Adds an async control-transfer FIFO to the USB host stack so async `tuh_control_xfer()` calls that arrive while the single control slot is busy get queued and dispatched on slot-free instead of being silently dropped. Disconnect-while-queued is handled lazily via a per-daddr generation counter — no FIFO sweep, O(1) at enqueue and dequeue.

## Motivation

Today the host stack has a single control-transfer slot. If an async control xfer is issued while that slot is busy (e.g. from a class driver's mount callback while enumeration is still running), it fails immediately — the application's `complete_cb` never fires. Hub workloads and concurrent class drivers hit this regularly.

## Design

- **Single slot, state machine.** `_usbh_data.ctrl_xfer_info.stage` cycles IDLE → SETUP → DATA → ACK → IDLE. All transitions go through
`control_xfer_set_stage()` under `_usbh_mutex`.
- **Pending FIFO.** `_usbh_pending_ctrl_q` (`tu_fifo_t`, depth `CFG_TUH_CONTROL_PENDING_QUEUE_SZ`, default 4) holds async xfers whose claim
found the slot busy. Sync waits keep their existing semantics — they still spin/poll, just on a slot that may have queued work ahead of
them.
- **Decoupled dispatch.** `control_xfer_complete()` only releases the slot and fires the user cb. FIFO drain happens in `tuh_task_ext()`
once the slot is IDLE — keeps reentrancy contained, no callback chains running with stale slot state.
- **Lazy disconnect handling.** `usbh_device_close()` bumps `daddr_gen[daddr]` under the mutex. Queued entries carry the gen they enqueued
with; at dequeue, mismatched gen → fire FAILED cb without submission. No FIFO scan on disconnect, ABA-safe across one full close cycle
(uint8_t is intentional, accepted tradeoff).
- **In-flight cancellation.** `usbh_device_close()` also fires `control_xfer_complete(daddr, FAILED)` for an in-flight xfer matching that 
daddr — async callers don't hang on hot-unplug.
- **Stale-completion guard.** `usbh_control_xfer_cb` early-returns when `stage == IDLE` or `daddr` mismatched, protecting against late HCD
events after abort/close from corrupting a freshly dispatched pending xfer.
- **`tuh_deinit()` drain.** Fires FAILED cb for every queued entry before `tu_fifo_clear()` so async callers aren't stranded across deinit.
- **New `XFER_RESULT_ABORTED`.** Used by `tuh_edpt_abort_xfer()` for EP0 so apps can distinguish user-cancellation from failures. 
`tu_str_xfer_result[]` extended to match.

## OS portability

- `CFG_TUSB_OS_HAS_SCHEDULER` abstraction (commit `c5676382c`) cleanly separates RTOS-with-scheduler from no-OS/pico paths. Every OSAL 
backend (FreeRTOS, ThreadX, Mynewt, RT-Thread, RTX4, Zephyr, pico, OS_NONE) adds `osal_task_handle_t` + `osal_task_get_current_handle()`.
- Sync-from-host-task assertion (RTOS only) prevents the class-driver-cb-calling-sync deadlock that previously hit users on FreeRTOS.

## Examples

- `examples/dual/dynamic_switch` heavily restructured to demonstrate role switching with the new APIs. Worker tasks (`cdc_task`, 
`print_devinfo_task`, `led_blinking_task`) now use the `#if RTOS while(1) else single-shot #endif` pattern (no false "loop-runs-once" Sonar 
warnings, no OS_NONE catch-up bursts).
- `examples/host/bare_api`, `device_info` updated to exercise queued async control xfers across a hub.

## HIL infrastructure

- `test/hil/hil_test.py` adds hub + MSC + CDC test fixtures, replaces `pyfatfs` with `mtools` for cross-platform compat.
- `test/hil/tinyusb.json` updated with the bench's actual VID:PID + hub topology.
- `test/hil/hil_ci.sh` cleanup for remote bench execution.

## Memory footprint

Per-instance overhead on a typical Cortex-M target (`CFG_TUH_DEVICE_MAX=4`, `CFG_TUH_HUB=1`, default queue size 4):

- `daddr_gen[]`: 5 bytes
- Pending FIFO buffer: 4 × ~28 B = ~112 bytes
- `tu_fifo_t` control: ~24 bytes
- Misc state (`task_hdl`, sync wait struct, etc.): ~16 bytes

≈ **160 bytes static RAM** + a few hundred bytes of code. `stm32h7s3nucleo` (64 KB internal flash) was just over the edge and now skips one 
large FreeRTOS example.

## Test plan

- [x] `tools/build.py` — full ARM/RISC-V GCC, Clang, IAR matrices green (CircleCI + GH Actions)
- [x] Static analysis: SonarCloud, CodeQL, PVS-Studio, Fuzzing — all pass
- [x] HIL 5x sweep on raspberry_pi_pico-f1 host/device_info — 5/5 final pass (retries flake on PIO USB hub cold-start, known)
- [x] HIL 1x sweep on full 9-board host/dual set — only known bench flakes remain (stm32f723 CDC echo, pico-f1 PIO USB hub, JLink probe 
disconnects)
- [x] mimxrt1064 host/msc_file_explorer isolated re-run: clean 5.2s pass (earlier failure was bench noise from JLink reconnect during batch 
sweep, not code regression)
- [x] No regressions in sync API (`tuh_descriptor_get_*_sync`), hub enumeration, control xfer retry path
- [ ] Long-soak bench testing of `examples/dual/dynamic_switch` role switching under FreeRTOS preemptive scheduling (the `usb_mode_switch` ↔
`cdc_task`/`print_devinfo_task` synchronization is best-effort, not mutex-fenced)

## Known limitations / follow-ups

- `examples/dual/dynamic_switch` role-switch race: `current_role` is checked without a mutex; a preempt between the role check in 
`cdc_task`/`print_devinfo_task` and the API call can hit a deinit'd stack. Flagged by Codex and the manual reviewer. Acceptable for an example, but worth a mutex/suspend fix if this pattern goes into production code.


hathach:
f723disco hil ci failed with cdc mismatched is probably another issue, due to the added hub + msc. Will fix with another PR.
```
stm32f723disco                           host/cdc_msc_hid               ...  Failed: CDC echo mismatch at offset 0 (25 bytes):
  expected: b'j4hZl2rD9Muf6oUm7xlslYzBf'
  received: b'jhZlr9UxYf'  CH9102: CDC Interface is mounted: address = 1, itf_num = 0 | A device with address 2 is mounted Disk Size: 60620800 512-by  in 17.8s
```

also hfp ci server seems to have hardware issue with f746disco (cannot flash with jlink)

HiFiPhile:
@hathach Do you have plan to change `usbh_get_enum_buf` to prevent buffer overwrite ?

I started a discussion in #3591

hathach:
> @hathach Do you have plan to change `usbh_get_enum_buf` to prevent buffer overwrite ?
> 
> I started a discussion in #3591

I will response there, this PR already took me too long. I need to move on with other stuffs first :). Though unlike usbd, usbh does not do memcopy with control transfer. Rule of thumb, for small buffer like set line coding, applicaton/driver should has its own buffer. Though since we serialize control transfer, maybe we can return NULL for the usbh_get_enum_buf() is control state is not IDLE, or something like that to allow app to wait/retry later without modifying the enum buf first. I will put more thought onto this later on.

PR #3658, opened 2026-05-29
Fix premature MTP phase change after short MTP OUT transfer

wjklimek1:
While implementing SendPartialObject to edit files "in place" on MTP I encountered an issue where TinyUSB's MTP class driver prematurely ends the DATA phase.

With a Linux host (libmtp/gvfs-mtp), the container is split into two bulk OUT transfers: one with the container header, then one with the payload. The first transfer is shorter than the endpoint max packet size, which the device incorrectly treated as end-of-transfer and transitioned to RESPONSE before the full container length had been received.

USB device stack should not use a short BULK OUT packet as the trigger for switching from DATA phase to RESPONSE phase. Only receiving amount of bytes indicated by container length or ZLP should be allowed to do this.

My suggested fix:
Leave IN handling unchanged (the device controls those packets so I assume short-packet termination is correct there). For OUT, end the phase only when the declared container length is reached or or ZLP from host is received.

### DIFF
--- a/src/class/mtp/mtp_device.c
+++ b/src/class/mtp/mtp_device.c
@@ -441,9 +441,19 @@ bool mtpd_xfer_cb(uint8_t rhport, uint8_t ep_addr, xfer_result_t event, uint32_t
         threshold = CFG_TUD_MTP_EP_BUFSIZE;
       }
 
-      // Check completion: ZLP, short packet, or total length reached
-      const bool is_complete =
-        (xferred_bytes == 0 || xferred_bytes < threshold || p_mtp->xferred_len >= p_mtp->total_len);
+      // Check completion for IN and OUT separately
+      bool is_complete;
+
+      if (is_data_in)
+      {
+        // IN completion: short packet, ZLP, or reaching total_len
+        is_complete = (xferred_bytes == 0 || xferred_bytes < threshold || p_mtp->xferred_len >= p_mtp->total_len);
+      }
+      else
+      {
+        // OUT completion: reaching total_len or ZLP
+        is_complete = (p_mtp->xferred_len >= p_mtp->total_len) || ((xferred_bytes == 0 && p_mtp->xferred_len > 0));
+      }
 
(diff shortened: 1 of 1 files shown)


HiFiPhile:
I think you need to report the bug to libmtp.

image


> USB device stack should not use a short BULK OUT packet as the trigger for switching from DATA phase to RESPONSE phase. Only receiving amount of bytes indicated by container length or ZLP should be allowed to do this.

That's how USB data transfer model works !

wjklimek1:
Hi @HiFiPhile,

I have to admit that I just adjusted TinyUSB code to meet libmtp expectations. After your comment I looked into documentation and I think that this is not a libmtp bug but utilization of MTP optimization.

The snippet you provided is from usb_still_img10.pdf and applies to PTP. The more modern MTP is actually PTP extension and allows splitting header and payload. See "Appendix H – USB Optimizations" in MTPforUSB-IFv1.1.pdf. In section "Splitting the Header and Data during the Data Phase" they describe exactly this mechanism - container is split into short OUT transfer with header and following OUT transfer with data.

This optimization is only allowed in MTP, not basic PTP. The MTP class in TinyUSB advertises itself as MTP (the file is mtp_device.c, the class code is MTP, etc.), so I believe supporting this optimization matches the user's expectations, especially when modern OS actually uses it.

What is your opinion?

HiFiPhile:
Thank you for point me to the appendix. 

I must say both PTP and MTP protocols are chaotic, even worse with extensions. Even for basic things like ZLP (what the hell is null packet?)

Since I rarely work on MTP could you test if there is no regression for existing commands ?

wjklimek1:
It took me while to figure out local HIL setup for my board (stm32h503nucleo). All tests from the HIL passed. I also run unit tests and they passed too.

I did some manual testing with mtp example from the repo and everything seems fine. I can open MTP drive, open files from it, copy them to PC. I did these manual test on Linux Mint and Windows 10. I also work on personal project that uses TinyUSB MTP to expose littlefs file system to the PC. No adverse side effects observed there as well.

And yeah, MTP is in fact quite confusing with these subtle PTP/MTP differences. I don't consider myself an expert either.

HiFiPhile (approved):
Thank you !

PR #3681, opened 2026-06-06
host/cdc: use local control buffer

HiFiPhile:
1. Buffer Allocation: Added an 8-byte ctrl buffer to the cdch_epbuf_t structure using the TUH_EPBUF_DEF(ctrl, 8); macro. This allows the memory to be properly aligned and located in the correct section (CFG_TUH_MEM_SECTION), while also keeping it separate for each instantiated CDC interface. (An 8-byte length is sufficient as CDC setup data sequences—like line coding or PL2303 vendor reads—don't exceed 8 bytes here).
2. Replaced Global Buffer Usage: In acm_set_line_coding, cp210x_set_request, ch34x_set_request, and pl2303_set_request, the code now dynamically looks up its own buffer cdch_epbuf[idx].ctrl to build and hold the control payload.
3. State Machine Fixes: In configuration state machines (such as pl2303_process_set_config and ch34x_process_set_config), any operations that previously relied on reading from or writing to usbh_get_enum_buf() now correctly route through the per-interface ctrl_buf.

MSC & HID should be fine as enum buffer is only used during set_config, since we only process one enumeration each time it should be conflict-free.

### DIFF
--- a/src/class/cdc/cdc_host.c
+++ b/src/class/cdc/cdc_host.c
@@ -99,6 +99,7 @@ typedef struct {
 typedef struct {
   TUH_EPBUF_DEF(tx, CFG_TUH_CDC_TX_EPSIZE);
   TUH_EPBUF_DEF(rx, CFG_TUH_CDC_RX_EPSIZE);
+  TUH_EPBUF_DEF(ctrl, 8);
 } cdch_epbuf_t;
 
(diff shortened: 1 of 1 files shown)


hathach (approved):
thank you.Look good, hil is down for hw troubleshoot, will be up soon and merge when it passed.

hathach:
rp2040 with pio-usb host has pre-existing issue with hub.

PR #2793, opened 2024-09-09
Add Host support for ch32v20x usbfs

verylowfreq:
This PR adds HCD for WCH CH32V20x USBFS (USBOTG_FS).

The following issue will be closed:
 - https://github.com/hathach/tinyusb/issues/2613

CH32V20x has two USB peripheral; FSDEV and USBFS. In this PR, USBFS Host driver is enabled when PORT=0 (FSDEV is selected).

I tested on my original CH32V203C8T6 board "Suzuno32RV" (8MHz external crystal, Type-A connector is assigned for USBFS, Type-C connector is for FSDEV).
Please refer for schematic: https://github.com/verylowfreq/board_suzuno32rv

PR #3682, opened 2026-06-06
ch32: portable async USBFS host driver — hub, multi-device, hotplug (v20x + v307)

RobertDaleSmith:
## Summary

  Reworks the CH32 USBFS host controller driver into a **portable, fully
  interrupt-driven asynchronous scheduler** that supports **USB hubs, multiple
  simultaneous devices, and hotplug**, covering **both CH32V20x and CH32V307**
  from one driver.

  The existing driver (added in #2793 by @verylowfreq) drives a single device with
  a blocking per-transfer model on v20x, and v307 USBFS host wasn't wired up at
  all. This replaces the transfer core with an `ep[]`-scheduler model and routes
  control transfers through WCH's reference transaction layer, keeping the same
  `CFG_TUH_WCH_USBIP_USBFS` entry point.

  > [!IMPORTANT]
  > This supersedes @verylowfreq's v20x driver with a hub-capable one for the whole
  > family. **v307 is hardware-verified; v20x is build-verified only** (I don't have
  > a v20x board). The silicon and registers are identical across the family, but
  > v20x should get a hardware pass before merge — happy to coordinate.

  ## Why

  The CH32 USBFS peripheral has a single host transfer engine (one `DEV_ADDR`
  latch, one `HOST_EP_PID` token, one shared toggle pair, one RX/TX DMA pair).
  One-transfer-at-a-time works for a single boot device but can't fan out to a hub
  plus several interrupt endpoints — which real adapters need.

  ## What changed

  - **`hcd_ch32_usbfs.c` — async `ep[]` scheduler.** Modeled on the max3421 HCD:
    ISR-driven round-robin over a flat `ep[]` array keyed by `(daddr, ep_num,
    dir)`, one transaction outstanding on the engine, interrupt EPs paced to
    `bInterval`.
  - **Fully interrupt-driven attach/disconnect.** `hcd_init()` enables SOF
    generation so the 1 ms SOF interrupt runs even before a device attaches; that
    tick polls `DEV_ATTACH` (a device present at power-up makes no DETECT edge) and
    the no-response disconnect counter. Hot-plug also takes the DETECT edge as a
    fast path. **No application poll hook is required — the stock host examples
    work unmodified.**
  - **`wch_usbfs_ll.{c,h}` (new) — WCH reference transaction layer.** Control
    transfers run through WCH's proven synchronous primitives
    (`USBFSH_CtrlTransfer`) + root-port reset/enable — this is what reliably
    enumerates **bus-powered hubs** and a broad range of controllers. Derived from
    WCH's `HOST_KM` reference, © WCH; thin config header is MIT.
  - **MCU-portable via `ch32_usbfs_reg.h`.** The selector also picks
    `ch32v20x_usb.h`/`ch32v30x_usb.h` by `CFG_TUSB_MCU`; the driver and LL include
    only the selector (no hardcoded MCU header). v20x's SDK has no
    `RCC_USBFSCLKConfig`, so the LL defers USB clock-source setup to `board_init()`
    there (as v307 already does).
  - **Host enable + BSP wiring.** `tusb_mcu.h` auto-enables
    `CFG_TUH_WCH_USBIP_USBFS` for v307 (mirroring the v20x block). The host sources
    are wired into **both** the make (`family.mk`) and cmake (`family.cmake`)
    builds. v307 BSP routes the USBFS IRQ to `tuh_int_handler` (the vector is
    `USBFS_IRQHandler`, not `OTG_FS_IRQHandler`) and bumps the linker stack; v20x's
    `USBHD_IRQHandler -> tuh_int_handler` wiring is unchanged.
  - **`tusb_verify.h`.** No-op the assert `ebreak` on CH32 RISC-V cores: a failed
    `TU_ASSERT`/`TU_VERIFY` was soft-resetting the chip (the WCH-LinkE leaves the
    debug module enabled) instead of returning the error, turning recoverable
    enumeration hiccups into reboot loops. Helps device-mode CH32 users too.

  ## Testing

  **Hardware — CH32V307VCT6** (USBFS host PA11/PA12 → USBHS device out):
  - HID keyboards, mice, gamepads enumerate and report.
  - Bus-powered hub with multiple controllers attached at once.
  - Hotplug — attach/detach of devices and of a downstream-of-hub device, plus a
    device already present at power-up.
  - Sustained input passthrough host-in → device-out.

  **Build — `riscv-none-elf-gcc` 13.2.0, make + cmake, every example, all boards:**
  - v20x: `ch32v203c_r0_1v0`, `ch32v203g_r0_1v0`, `nanoch32v203`
  - v30x: `ch32v307v_r1_1v0`, `nanoch32v305`

  **v20x runtime: not validated** — needs a hardware pass (cc @verylowfreq).

### DIFF
--- a/hw/bsp/ch32v30x/ch32v30x_it.c
+++ b/hw/bsp/ch32v30x/ch32v30x_it.c
@@ -8,6 +8,8 @@
 * SPDX-License-Identifier: Apache-2.0
 *******************************************************************************/
 #include "ch32v30x_it.h"
+#include <stdio.h>
+#include <stdint.h>
 
 void NMI_Handler(void) __attribute__((naked));
 void HardFault_Handler(void) __attribute__((naked));
@@ -39,8 +41,35 @@ void HardFault_Handler(void){
       __asm volatile ("call HardFault_Handler_impl; mret");
 }
 
+// Direct USART1 output — no printf/stack/buffers, safe in a corrupted fault state.
+static void fault_putc(char c)
+{
+  volatile uint32_t* statr = (volatile uint32_t*) 0x40013800u;  // USART1 STATR
+  volatile uint32_t* datar = (volatile uint32_t*) 0x40013804u;  // USART1 DATAR
+  while (!(*statr & (1u << 7))) { }  // wait TXE
+  *datar = (uint32_t) (uint8_t) c;
+}
+static void fault_puthex(uint32_t v)
+{
+  for (int i = 7; i >= 0; i--) {
+    int n = (int) ((v >> (i * 4)) & 0xF);
+    fault_putc((char) (n < 10 ? '0' + n : 'A' + n - 10));
+  }
+}
+
 __attribute__((used)) void HardFault_Handler_impl(void)
 {
+  uint32_t mcause = 0, mepc = 0, mtval = 0;
+  __asm volatile ("csrr %0, mcause" : "=r"(mcause));
+  __asm volatile ("csrr %0, mepc"   : "=r"(mepc));
+  __asm volatile ("csrr %0, mtval"  : "=r"(mtval));
+  const char* tag = "\nFAULT cause="; while (*tag) fault_putc(*tag++);
+  fault_puthex(mcause);
+  tag = " mepc=";  while (*tag) fault_putc(*tag++);
+  fault_puthex(mepc);
+  tag = " mtval="; while (*tag) fault_putc(*tag++);
+  fault_puthex(mtval);
+  fault_putc('\n');
   while (1)
   {
   }
--- a/hw/bsp/ch32v30x/family.c
+++ b/hw/bsp/ch32v30x/family.c
@@ -61,9 +61,18 @@ __attribute__((interrupt)) void USBHS_IRQHandler(void) {
   #endif
 }
 
-__attribute__((interrupt)) void OTG_FS_IRQHandler(void) {
-  #if CFG_TUD_WCH_USBIP_USBFS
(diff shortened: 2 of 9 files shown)


verylowfreq:
Thank you for this work!

I tested on CH32V203 and found an enumeration issue with some Low Speed
Devices connected via a hub. I traced it to the speed selection logic and
made a fix.

PR: https://github.com/joypad-ai/tinyusb/pull/1

Please take a look when you get a chance.

RobertDaleSmith:
@verylowfreq thank you for testing so quickly. i merged your changes in and verified it all still works on my end with a v307. 🙏

HiFiPhile (changes requested):
Thanks for your hard work ! But I've to say in current state it's not mergeable due to:
- `wch_usbfs_ll` is not MIT licensed to be included in `src` (although I believe WCH doesn't care)
- `wch_usbfs_ll` is a black box and nightmare to maintain, previously we had very limited support from WCH, if anything goes wrong we have to spend a lot of time.
- Many functions are unused like ` USBFS_RCC_Init` and `USBFS_Host_Init` which make things confusing.
- `USBFSH_CtrlTransfer` implement control transfer **Synchronously** with a maximum blocking time of `DEF_CTRL_TRANS_TIMEOVER_CNT` which against TinyUSB's asynchronous approach especially when no RTOS is used.

RobertDaleSmith:
@HiFiPhile thank you so much for taking the time to review this and for the feedback. I totally understandable the concerns. I will circle back to this and see what I can come up with.  🙏

DevinJM3:
@RobertDaleSmith Any updates on this? Would love to be able to get USB Host working on my CH32 devboard!

PR #3683, opened 2026-06-07
hid: fix report_id extraction for interrupt OUT transfers

### LINKED ISSUE #2929: tud_hid_set_report_cb() is invoked with inconsistent arguments
### Operating System

Others

### Board

All (portable)

### Firmware

class/hid/hid_device.c
 
tud_hid_set_report_cb() is invoked in two contexts, with inconsistent argument usage:

* From hidd_control_xfer_cb().  Before invoking the callback, the caller removes the first byte of the buffer IF it matches the HID report ID in the control request.
* From hidd_xfer_cb().  In this case, the caller leaves the whole buffer intact unconditionally and passes HID report ID == 0.

It's up to the callee to distinguish these two cases, to determine whether or not the buffer contains the HID report ID prefix byte.  In versions prior to 0.17.0, it was possible to distinguish the cases based on the report type, since the second case always passed HID_REPORT_TYPE_INVALID.  Starting in 0.17.0, both cases pass HID_REPORT_TYPE_OUTPUT, making the cases harder for the callback to distinguish.

The fix would be to either STOP stripping the report ID byte prefix in hidd_control_xfer_cb(), or START parsing it in hidd_xfer_cb().  Either way is going to be a breaking change, but the latter will probably have lower impact.  And code that uses report IDs and implements tud_hid_set_report_cb() is ALREADY probably broken - not functioning as intended - unless the application developer noticed the inconsistency and added special-case code to work around it like I did.  None of the examples that use tud_hid_set_report_cb() use report IDs, which is why they all get away without special-casing it.  Presumably that's the most common case for actual code in the wild as well.

Note that there appears to be a second, related bug in hidd_control_xfer_cb(): it strips the first byte of the data buffer any time the report ID matches, **even if the report ID is zero**, which means "no report ID".  In that case, it incorrectly strips a data byte when the first data byte just coincidentally happens to be 0x00.

### What happened ?

Incorrect behavior in endpoint output report processing, requiring special-case handling in user code.  The user code shouldn't need a special case to work around what should be a consistent callback interface.

### How to reproduce ?

Create a program that uses HID report IDs in the report descriptors and implements a tud_hid_set_report_cb() callback.  Observe the different behavior in the buffer between callbacks from control requests vs endpoint transfers.

N/A

### Screenshots

_No response_

moxybaba:
## Problem

In `hidd_xfer_cb()`, when an output report is received via interrupt 
OUT endpoint, `tud_hid_set_report_cb()` is always called with 
`report_id=0`.

Per HID spec section 8.2, when a device uses report IDs, the first 
byte of an interrupt OUT transfer is the report ID followed by the 
report data. This is already handled correctly in the SET_REPORT 
control path but was missing for interrupt OUT transfers.

## Fix

Extract report_id from the first byte of the output buffer when 
non-zero and buffer length > 1, consistent with how the control 
path handles report IDs.

## Testing

Built hid_composite example for stm32f411blackpill successfully.

Fixes #2929

### DIFF
--- a/hw/bsp/ch32v10x/boards/ch32v103c_bluepill/board.cmake
+++ b/hw/bsp/ch32v10x/boards/ch32v103c_bluepill/board.cmake
@@ -1,0 +1,8 @@
+set(LD_FLASH_SIZE 64K)
+set(LD_RAM_SIZE 20K)
+
+function(update_board TARGET)
+  target_compile_definitions(${TARGET} PUBLIC
+    CFG_EXAMPLE_MSC_DUAL_READONLY
+    )
+endfunction()
--- a/hw/bsp/ch32v10x/boards/ch32v103c_bluepill/board.h
+++ b/hw/bsp/ch32v10x/boards/ch32v103c_bluepill/board.h
@@ -1,0 +1,27 @@
+/* Some Chinese manufacturers use CH32V103C8T6 to make Bluepill boards
+ */
+/* metadata:
+   name: CH32V103C8T6-Bluepill
+   url: https://stm32-base.org/boards/STM32F103C8T6-Blue-Pill
+*/
+
+#ifndef BOARD_H_
+#define BOARD_H_
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+#define LED_PORT            GPIOC
+#define LED_PIN             GPIO_Pin_13
+#define LED_STATE_ON        0
+
+#define BUTTON_PORT         GPIOA
+#define BUTTON_PIN          GPIO_Pin_1
+#define BUTTON_STATE_ACTIVE 1
+
+#ifdef __cplusplus
+}
+#endif
+
+#endif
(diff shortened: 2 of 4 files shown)


HiFiPhile:
Hi, I think you pushed a wrong commit.

hathach:
how you know if the report has report-id in it, some does not use report-id at all (single report).

moxybaba:
@hathach You're right, there's no reliable way to detect this from the transfer alone without knowing the descriptor. 
Looking at mjrgh's suggestion in the original issue, a v2 callback with int report_id (passing -1 when unknown) would be the correct approach since it lets the application decide rather than the stack guessing.
Should I rework this to implement the v2 callback approach instead?

moxybaba:
@hathach happy to rework this with the v2 callback approach (int report_id, passing -1 when unknown) if that's the preferred 
direction. Just waiting for your go-ahead before rewriting.

PR #3686, opened 2026-06-08
HIL: add stm32u083nucleo and post test report as PR comment

hathach:
## Summary
Brings `stm32u083nucleo` into the HIL pool, plus the supporting fixes and tooling discovered along the way.

- **stm32u0 unique USB serial** — implement `board_get_unique_id()` from `UID_BASE` (mirrors stm32u5). Previously stm32u0 fell back to the weak default and every board reported the placeholder serial `0123456789ABCDEF`, which collides on a multi-board rig. Now enumerates with the real 96-bit chip UID.
- **Presets** — `tools/gen_presets.py` builds into `cmake-build-<board>` (was `build/<board>`). Generator stays **Ninja Multi-Config** (IAR needs it to lower the optimization level for debug). `hil_test.py` locates `<ex>.elf` whether it sits directly in the example dir (single-config) or under a per-config subdir like `RelWithDebInfo/` (multi-config), so preset-built firmware is HIL-consumable either way. Regenerated `BoardPresets.json`.
- **HIL pool** — add `stm32u083nucleo` to `tinyusb.json` (flashed via the `stlink`/STM32CubeProgrammer flasher; the ci openocd build has no STM32U0 flash driver). `hil_ci.sh` now adds `~/bin` to the remote PATH so the CubeProgrammer CLI is found.
- **HIL report** — `hil_test.py` writes `hil_report.md` and prints a board×test matrix to stdout (rows = boards, columns = tests; ✔ pass / ✖ fail / ➖ skipped / blank not run). `hil_ci.sh` copies the report back from the remote after a run.
- **gitattributes** — pin `*.sh` to `eol=lf` so `core.autocrlf=true` stops CRLF-corrupting shell scripts (which broke `hil_ci.sh` under bash).

## Test evidence
- `stm32u083nucleo` HIL: **13/13 device tests pass**, both locally (htpc) and remote (ci.lan), with a multi-config preset build; `audio_test_freertos` self-skips on the rig.
- Board now enumerates as chip UID `300044000D5036394E373620` (was the placeholder serial).
- `hil_test.py` report generation covered by an isolated unit check (multi-board + flags-on variant rows); `py_compile` clean.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -19,6 +19,9 @@
 
 Makefile text
 
+# Shell scripts must stay LF even when core.autocrlf=true (CRLF breaks bash)
+*.sh text eol=lf
+
 # Windows-only Visual Studio things
 
 *.sln        text eol=crlf
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -403,14 +403,23 @@ jobs:
         run: |
           python3 test/hil/hil_test.py hfp.json
 
+      - name: Upload HIL report
+        if: always() && github.event_name == 'pull_request'
+        uses: actions/upload-artifact@v7
+        with:
+          name: hil-report-hfp-iar
+          path: hil_report.md
+          if-no-files-found: ignore
+          overwrite: true
+
   # ---------------------------------------
   # Combine HIL results from the rigs into a single sticky PR comment (one table per rig)
   # ---------------------------------------
   hil-report:
-    needs: hil-tinyusb
+    needs: [ hil-tinyusb, hil-hfp-iar ]
     if: |
       always() &&
-      needs.hil-tinyusb.result != 'skipped' &&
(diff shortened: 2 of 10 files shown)


HiFiPhile:
> Presets — tools/gen_presets.py builds into cmake-build-<board> with single-config Ninja instead of Ninja Multi-Config, so firmware lands at device/<ex>/<ex>.elf (where hil_test.py looks) rather than nested under RelWithDebInfo/. Regenerated BoardPresets.json.

IAR works better with multi-config, otherwise there is no way to change optimization level to none for debug.

hathach:
@HiFiPhile good catch — reverted to Ninja Multi-Config so IAR keeps the per-config optimization levels for debug. I kept only the `cmake-build-<board>` binaryDir change, and taught `hil_test.py` to find the firmware whether it's directly in the example dir or under the `RelWithDebInfo/` subdir, so HIL still consumes multi-config preset builds. Verified stm32u083nucleo 13/13 on the rig with a multi-config build (46aded44a).

PR #3697, opened 2026-06-11
Clean up PVS-Studio static-analysis findings for raspberry_pi_pico

hathach:
## Summary

Brings the CI PVS-Studio job (`.github/workflows/static_analysis.yml`, run with `--security-related-issues`) to **zero TinyUSB-owned alerts** on `raspberry_pi_pico` — previously ~43 across `src/` and `examples/`. Each was triaged individually: real issues fixed in code, false positives suppressed at the narrowest scope that keeps the diagnostic active elsewhere.

> The remaining ~57 SARIF entries are all in the vendored pico-sdk, which CI already excludes (it sets `PICO_SDK_PATH` to a `pico-sdk/` checkout that `*/pico-sdk/*` matches); they are untouched here.

## Genuine fixes

| File | Diag | Fix |
|---|---|---|
| `class/net/ncm_device.c` | V568 | `wNdpIndex` was validated against `sizeof(nth16)` — the **pointer** size (4 B) — instead of `sizeof(nth16_t)` (12 B header). Under-checked the NTB header; now matches the sibling checks on lines 596/604. |
| `device/usbd.c` | V763 | The 9 `usbd_edpt_*` / `usbd_sof_enable` functions rewrote their `rhport` parameter with `rhport = _usbd_rhport;`. Now `(void) rhport;` + pass `_usbd_rhport` directly to the `dcd_*` calls, matching the existing style of `usbd_edpt_claim/release/busy/stalled`. No suppression needed. |
| `tusb.c` | V560 | Dropped the redundant `ff_buf != NULL && ff_bufsize > 0` guard in `tu_edpt_stream_init()`; the early `return false` already guarantees it. |
| `class/midi/midi_host.c` | V560 | `tuh_midi_itf_get_info()` tested `&_midi_host[idx]` (always non-NULL). Replaced with a real `idx < CFG_TUH_MIDI` bound check. |
| `examples/.../uac2_app.c`, `video_capture_2ch/main.c` | V1009 | Fully initialize `resolutions_per_format` / `frame_num` / `interval_ms`; trailing elements were silently left zero (a latent example bug for the 2nd format/stream). |

## False positives — fixed at the cause

`src/device/usbd.c` and `src/host/usbh.c` wrap the **weak** `dcd_deinit()` / `hcd_deinit()` stubs in `#ifndef PVS_STUDIO`. PVS analyzes one translation unit at a time and binds the call to the always-`false` weak stub — it cannot model the linker selecting the port's strong definition — then reports the class-driver cleanup loop after `TU_ASSERT(...deinit())` as unreachable (**V779**). Hiding the stub makes the analyzer treat `*_deinit()` as an ordinary extern, matching link-time reality. (Verified empirically: the V779 disappears with the guard.)

## False positives — suppressed locally

- **`.pvsconfig`** (systemic / macro-driven): `V501` (HID descriptor macros expand `proto != HID_ITF_PROTOCOL_NONE` to a constant when `NONE` is passed), `V785` (audio function-index switch is constant when `CFG_TUD_AUDIO == 1`), `V1044` (hardware status-register poll loops).
- **inline `//-V` with rationale** for config-dependent / intentional one-offs: `V512`, `V547`, `V557`, `V560`, `V614`, `V619`, `V641`, `V1008`, `V1037`, `V1048`, `V1086`.

## Validation

- All examples build for `raspberry_pi_pico` (CMake/Ninja, MinSizeRel).
- `ceedling test:all` → **60/60 pass** (re-run after each commit).
- Re-ran the CI-exact PVS invocation after each commit → **0 TinyUSB-owned findings**.

### DIFF
--- a/.PVS-Studio/.pvsconfig
+++ b/.PVS-Studio/.pvsconfig
@@ -7,7 +7,6 @@
 
 //-V:TUD_HID_DESCRIPTOR:501 _boot_protocol comparison inside the macro is constant when HID_ITF_PROTOCOL_NONE is passed
 //-V:TUD_HID_INOUT_DESCRIPTOR:501 _boot_protocol comparison inside the macro is constant when HID_ITF_PROTOCOL_NONE is passed
-//-V::763 rhport parameter is deliberately overridden with the configured root port (single-port stack)
 //-V::785 audio_device.c: switch over function index is constant when CFG_TUD_AUDIO == 1
 //-V::1044 wait loops on volatile hardware status registers are idiomatic in DCD/HCD ports
 
--- a/.PVS-Studio/.pvsconfig
+++ b/.PVS-Studio/.pvsconfig
@@ -5,6 +5,12 @@
 //V_EXCLUDE_PATH */hw/bsp/espressif/components/*
 //V_EXCLUDE_PATH */lib/*
 
+//-V:TUD_HID_DESCRIPTOR:501 _boot_protocol comparison inside the macro is constant when HID_ITF_PROTOCOL_NONE is passed
+//-V:TUD_HID_INOUT_DESCRIPTOR:501 _boot_protocol comparison inside the macro is constant when HID_ITF_PROTOCOL_NONE is passed
+//-V::763 rhport parameter is deliberately overridden with the configured root port (single-port stack)
+//-V::785 audio_device.c: switch over function index is constant when CFG_TUD_AUDIO == 1
+//-V::1044 wait loops on volatile hardware status registers are idiomatic in DCD/HCD ports
+
 //-V::2506 MISRA. A function should have a single point of exit at the end.
 //-V::2514 MISRA. Unions should not be used.
 //-V::2520 [MISRA-C-16.3] Every switch-clause should be terminated by an unconditional 'break' statement
--- a/examples/device/cdc_uac2/src/uac2_app.c
+++ b/examples/device/cdc_uac2/src/uac2_app.c
@@ -54,7 +54,8 @@ int32_t spk_buf[CFG_TUD_AUDIO_FUNC_1_EP_OUT_SW_BUF_SZ / 4];
 // Speaker data size received in the last frame
 uint16_t spk_data_size;
 // Resolution per format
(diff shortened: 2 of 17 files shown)


hathach:
The `hil-tinyusb (tinyusb.json)` failure looks like a **test-rig peripheral flake on the `raspberry_pi_pico` fixture, unrelated to this PR** — safe to re-run.

In every failing case the TinyUSB host stack works; the downstream peripherals just didn't appear:
- `host/cdc_msc_hid` — the CH340 **CDC device mounted fine** (`address = 1, itf_num = 0`); only the **MSC** device was "not mounted".
- `host/device_info` — expected `['1a86_7523_0', '048d_04d2_…']`, **only the CH340 (1a86:7523) enumerated**; the second device never showed.
- `host/msc_file_explorer` + `_freertos` — "MSC device not mounted".
- `dual/host_info_to_device_cdc` — "device disconnected or multiple access on port?".

Reasons this isn't a regression from these changes:
1. **`raspberry_pi_pico_w` and `raspberry_pi_pico2` passed the same host tests in this run** — identical rp2040 host stack.
2. **Every host-side change here is comment-only** (`//-V…`) or the `#ifndef PVS_STUDIO` guard, which is inert in a normal build (the macro is defined only by the analyzer).
3. The host **enumerates and mounts the CDC device** in each attempt — no crash/hang/enumeration change.

A HIL re-run should clear it. Everything else is green (PVS-Studio, CodeQL, SonarCloud, all arm-gcc builds, the other two HIL configs).

---
_Generated by [Claude Code](https://claude.ai/code)_

hathach:
Correction to my earlier triage: the `hil-tinyusb (tinyusb.json)` failure is **persistent, not a transient flake** — the second run on `881abf8` failed with the identical signature (same 5 pico host tests; CH340 CDC mounts fine, the MSC stick and `048d:04d2` device never enumerate).

Still not caused by this PR: master passed these tests on this PR's base commit (`5014146`, Build run #6234, Jun 11 16:19), `pico_w`/`pico2` pass the same host tests in both PR runs, and the pico passes all its device-side tests in the same runs. The fixture's downstream peripherals (or their hub) on the `raspberry_pi_pico` rig appear to have wedged between 16:19 and 23:51 on Jun 11 — likely needs a physical replug/power-cycle, then a job re-run.

---
_Generated by [Claude Code](https://claude.ai/code)_

PR #3702, opened 2026-06-14
hcd/rp2040: implement host endpoint abort and close

andrewleech:
### Summary

This extends the rp2040/rp2350 host driver to implement two HCD endpoint methods that were left as TODOs when host support landed, `hcd_edpt_abort_xfer` and `hcd_edpt_close`. I came at them building raw bulk endpoint forwarding over the rp2 host, but they're on the normal host paths too, `tuh_edpt_close` aborts then closes an endpoint and usbh aborts every endpoint on a SET_CONFIGURATION, so filling them in rounds out the existing host support.

The abort runs the EP_ABORT / EP_ABORT_DONE handshake so the controller stops accessing an interrupt endpoint before we clear its buffer, and leaves the endpoint configured so the next submit re-arms cleanly. Without that the buffer stays armed while the stack thinks the endpoint is idle, so the next transfer arms an already-available buffer and trips `panic("ep %02X was already available")`. Control (EPX) just clears its buffer and resets. The USB interrupt is masked across the teardown since `hw_endpoint_lock_update` is a no-op on this port.

The close deconfigures the endpoint the same way `hcd_device_close` already does per endpoint, which frees the `ep_pool` slot for reuse. Without it `tuh_edpt_close` aborts the transfer but leaves the endpoint allocated, so a later `tuh_edpt_open` of the same endpoint takes a fresh slot and works through the 15 interrupt slots.

Both land on rp2350 as well since it shares the driver.

### Testing

Tested on RP2350 (Pico 2 W) acting as a USB host against a full-speed CDC device, driving raw bulk endpoint transfers with `CFG_TUH_API_EDPT_XFER=1`. Before this, aborting a pending bulk-IN read and resubmitting on the same endpoint panicked, confirmed over SWD with the endpoint still `active` and its buffer armed. After, repeated cancel/resubmit cycles run clean with no panic and no duplicate or stuck endpoint-pool entries.

EP_ABORT and EP_ABORT_DONE are tagged "Device only" in the datasheet, so I instrumented the spin to check the handshake actually works host-side: across 48 aborts from a clean boot, EP_ABORT_DONE asserted within 0 spin iterations every time, so it does work in host mode for these interrupt endpoints. Haven't tested on an actual rp2040 yet though, the abort register path is identical on both.

### Trade-offs and Alternatives

The abort waits on `abort_done` before clearing the buffer, the same handshake the device side uses in `rp2040_usb.c`. It runs at task level, not in the ISR, and done asserts within a few cycles for these endpoints (measured, see Testing), so the spin's fine. The alternative that avoids the "Device only" register entirely is to disable the endpoint's `int_ep_ctrl` polling and then clear the buffer (what `hcd_device_close` does), but given the handshake measurably works and gives a clean "controller is idle" signal, I went with EP_ABORT.

### Generative AI

### DIFF
--- a/src/portable/raspberrypi/rp2040/hcd_rp2040.c
+++ b/src/portable/raspberrypi/rp2040/hcd_rp2040.c
@@ -589,19 +589,128 @@ bool hcd_edpt_open(uint8_t rhport, uint8_t dev_addr, const tusb_desc_endpoint_t 
   return true;
 }
 
-bool hcd_edpt_close(uint8_t rhport, uint8_t daddr, uint8_t ep_addr) {
-  (void)rhport;
-  (void)daddr;
-  (void)ep_addr;
-  return false; // TODO not implemented yet
+// Disarm the EPX round-robin scheduler. Call when the active EPX endpoint is torn
+// down and no other endpoint is pending, so a stale SOF / NAK-stop wakeup cannot
+// resurrect the aborted transfer. Mirrors the "no more pending" cleanup in the ISR.
+TU_ATTR_ALWAYS_INLINE static inline void epx_scheduler_disarm(void) {
+  #ifdef HAS_STOP_EPX_ON_NAK
+  usb_hw_clear->nak_poll = USB_NAK_POLL_STOP_EPX_ON_NAK_BITS;
+  usb_hw_clear->nak_poll = USB_NAK_POLL_EPX_STOPPED_ON_NAK_BITS; // clear any latched NAK-stop
+  #else
+  usb_hw_clear->inte = USB_INTE_HOST_SOF_BITS;
+  usb_hw->nak_poll   = USB_NAK_POLL_RESET;
+  epx_switch_request = false;
+  #endif
 }
 
+// Abort an in-flight (or queued) transfer and leave the endpoint configured and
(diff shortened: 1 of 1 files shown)


hathach:
@claude review

andrewleech:
Rebased onto current master and force-pushed. The branch was sitting a long way behind master so the original abort/close didn't line up with the current rp2040 host driver (epx is a round-robin pointer, an EPSTATE state machine, sie_stop_xfer to stop a transfer). I've reimplemented `hcd_edpt_abort_xfer` and `hcd_edpt_close` against the current model, with the review points folded in.

abort stops the active EPX transfer with `sie_stop_xfer` (the host STOP_TRANS, not the device-only abort handshake), drops the ep to IDLE and clears the latched `buf_status` / `sie_status` before re-homing the round-robin scheduler, so the ep stays re-armable with no double-arm. close tears the ep down and frees its `ep_pool` slot the same way `hcd_device_close` does.

Tested on an RP2350 (Pico 2 W) hosting a CDC device: aborting a pending bulk-IN read and resubmitting on the same endpoint runs clean now, no `ep .. was already available` panic, and close/reopen works over a couple of cycles. Haven't exercised the multi-endpoint promote path or an actual rp2040 on hardware yet though.

PR #3703, opened 2026-06-14
CH32X035 port

rhgndf:
Adds support for both CH32X035 and CH32X033.

I'm also evaluating this MCU due to it being low cost and have USB.

CH32X035 has both TX and RX CTRL bits in the same byte, so there are changes to how these registers are written in the dcd.
If the current approach isn't satisfactory, there are some alternatives:
- Make EP_TX_CTRL_WRITE a function instead of a #define 
- 'Inline' the #define: EP_TX_CTRL(ep) = (EP_TX_CTRL(ep) & ~MASK) | BITS

**EP4 special handling:**
This is a bit confusing, and Codex/Claude has been very helpful to get that part working. In short: no DMA and no auto-toggle, so some special handling has to be built in.

**EP3 no special large buffer**
The code has some existing handling for large EP3 buffer, but as far as I can tell that functionality doesn't exist in ch32x035, should I #if that out for this port?

Related: adafruit/Adafruit_TinyUSB_Arduino#486


HiFiPhile:
> CH32X035 has both TX and RX CTRL bits in the same byte, so there are changes to how these registers are written in the dcd.

Seems like v103, do you have board to test with ?

rhgndf:
> Seems like v103, do you have board to test with ?

I don't have the board unfortunately. I may take a look if I need a 5V capable mcu though (although py32f07x is also 5V capable but has additional DACs unlike v103)

rhgndf:
> MemBrowse Memory Report

Got claude to analyze the discrepency, here's the output verbatim:
```
  Why update_in got smaller

  The refactor folded the endpoint-buffer selection. Old code was a three-way branch:

  if (ep == 0)      memcpy(data.buffer[ep][TUSB_DIR_OUT], ...);  // buffer[ep][0]
  else if (ep == 3) memcpy(data.ep3_buffer.in, ...);
  else              memcpy(data.buffer[ep][TUSB_DIR_IN], ...);   // buffer[ep][1]

  New code is two-way, with the ep0 case absorbed into the inlined helper:

  if (ep == 3) memcpy(data.ep3_buffer.in, ...);
  else         memcpy(ep_buffer(ep, TUSB_DIR_IN), ...);   // buffer[ep][ep==0 ? 0 : 1]

  So the dedicated ep == 0 test-and-branch disappears; the address is now computed in a straight line as data.buffer[ep][ep==0 ? 0 : 1]. The disassembly confirms it: master has an extra li a5,3 / bne / beqz a4 cluster plus a second address computation
  that the branch version doesn't, and with RVC 2-byte instructions the savings land on exactly 8 bytes.
```

hathach:
thank you, after https://github.com/hathach/tinyusb/pull/3515, this should be relatively easy. I have the CH32X035 board but run out of wlinke debugger (attached to hil rig). will get this mcu ported when it got on my desk

rhgndf:
I also simplified the comments, since I found a hallucination involving X035.
Other changes:
- refactored the ep buffer code for 58x and x035
- removed ep3 in the non ep4 shared buffer case, saves 128 bytes of memory

rhgndf:
According to the manuals, it says only EP4 needs manual toggling, maybe manual toggle should only be enabled for EP4?

HiFiPhile:
Hi, I see many comment changes, please verify are they intended.

rhgndf:
Yes they are intended, since there was a hallucination about the x035 and I was concerned there are more, so I got my agent to shrink all of them so they can be verified easily

HiFiPhile:
Many comments are from Ha Thach like 
>  // CH58X EP registers split into a low block (EP0-4) and a high block (EP5-7). Walk from each
    // block's first slot by the 4-byte slot stride (pointer arithmetic off slot 0, so the unused

They shoudn't be removed.

HiFiPhile:
@claude review

HiFiPhile:
Thank you, please check if CH32X035 works correctly with latest changes.

rhgndf:
I went ahead and added the change to restrict manual toggling to EP4.
The firmware works correctly on my ch32x035

HiFiPhile:
Are you sure only EP4 needs manual toggle on CH32X035 ?  From the manual there is no auto toggle bit at all.

Also ch582 needs manual toggle:
```
  // CH58x's hardware AUTO_TOG does not stay in sync (notably across clear-stall and multi-packet
  // bulk transfers), causing data-toggle mismatch and bus resets. Drive the toggle manually in
  // the ISR instead. CH32V103/V20x/V307 keep AUTO_TOG (this macro is undefined for them).
```

rhgndf:
Yes, I'm quite sure since it is indeed in the manual under TOG_AUTO/AUTO_TOG

### CH32X035

Page 223 for EP1, this exists for all EPs not 4
image

Page 226 for EP4
image

### CH58x

Page 128
image

EP4 doesn't have auto toggle for both chips.

> ```
>   // notably across clear-stall
> ```

This issue was fixed too: https://github.com/rhgndf/tinyusb/blob/e80049bfb305ce2164310a510af78814234ede1b/src/portable/wch/dcd_ch32_usbfs.c#L600
Someone with a CH58x would need to check whether that fixes it

For the multi-packet bulk transfers, it works with cdc_msc_throughput with the results:
| tog | speed |
| ------ | ------ |
| manual tog | 581kB/s |
| auto tog | 631kB/s |

HiFiPhile:
> Yes, I'm quite sure since it is indeed in the manual under TOG_AUTO/AUTO_TOG

Thank you, I read the wrong page... @hathach could you test ch582, looks HIL is off.

hathach:
@HiFiPhile sure thing, I have both x035 and 582, but I am moving hil pool around. Will test this out once it is more stable. Still waiting for the new 4-port renesas (on-oder) to test with.

PS: also just ordered the 12-port 4 x reneses chip card :smile: Hope it works out espcially with power budget (plan to add 7-port hub to those 12 ports).

PR #3709, opened 2026-06-16
ncm: add weak callback for initial link state

morse-cedricvandenbergh:
Closes #3661

## Problem

When the host reboots without power-cycling the USB device, `netd_init` resets `ncm_interface.link_is_up` to the compile-time default (`CFG_TUD_NCM_DEFAULT_LINK_UP`, or `true`). If the application had previously set a different state, or if the physical link state differs from the default, the device reports an incorrect link state to the host after re-enumeration.

## Solution

Add a weak callback `tud_network_default_link_state_cb()` called by `netd_init` to determine the initial link state. The default implementation preserves existing behaviour (`CFG_TUD_NCM_DEFAULT_LINK_UP` if defined, else `true`). Applications that know the physical link state at init time can override it:

```c
bool tud_network_default_link_state_cb(void) {
  return phy_link_is_up();
}
```

## Testing

Built `net_lwip_webserver` and `cdc_msc` for `stm32f407disco` — both pass.

### DIFF
--- a/src/class/net/ncm_device.c
+++ b/src/class/net/ncm_device.c
@@ -152,6 +152,14 @@ TU_ATTR_WEAK void tud_network_set_packet_filter_cb(uint16_t packet_filter) {
   (void) packet_filter;
 }
 
+TU_ATTR_WEAK bool tud_network_default_link_state_cb(void) {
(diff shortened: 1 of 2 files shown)


HiFiPhile (approved):
Thank you.

PR #3712, opened 2026-06-18
dcd/stm32_fsdev: review fixes for the CH32 EP0 OUT workaround

hathach:
@HiFiPhile there are a few finding that detected by claude, I did manual review of everything it makes and do manual update. Please check it out to see if these make sense to you. 

Code review of the CH32 EP0 OUT premature-ACK workaround on this branch, plus on-hardware verification (nanoch32v203, fsdev). 4 commits, all CH32-gated. No behaviour change for STM32/AT32. Per-line rationale is in the inline review comments.

## Correctness

**1. `ep0_set_type` read-modify-write wasn't atomic**
`ep_write(.., need_exclusive)` masks the USB IRQ only around the *store*; the `ep_read` ran unmasked, so a SETUP arriving mid-RMW could have its BULK gate dropped:
```
task: r = ep_read(EP0)             // IRQ still enabled
ISR : SETUP -> ep0_set_type(BULK)  // installs the gate
task: ep_write(r, type=CONTROL)    // overwrites it -> the OUT it was meant to block gets ACKed
```
Bracket the whole RMW with `fsdev_int_disable/enable` when `need_exclusive`.

**2. Stall left EP0 typed BULK**
`dcd_edpt_stall` preserves the type field, and `edpt_xfer` is the only other place that restores CONTROL. A control-write rejected by the class (e.g. DFU `DNLOAD` in the wrong state) leaves EP0 stuck BULK+STALL; the host's recovery SETUP is ignored until a bus reset. Fix: restore CONTROL for EP0 in the stall write.

**3. CONTROL restored before the OUT was armed**
`edpt_xfer` flipped EP0 to CONTROL at the top of the function, before `btable_set_rx_bufsize`/`STAT_RX=VALID` — re-enabling the blind ACK with a stale, still-NAK'd buffer, so a host-retried DATA OUT / status ZLP could be ACKed into the wrong buffer. Fix: fold the CONTROL restore into the same exclusive write that sets `STAT_RX=VALID` on the OUT path; keep the early restore on the IN path (no pending OUT there).

## Cleanup
- Parse the setup packet via `tusb_control_request_t` (as `dcd_ch32_usbfs.c` does) instead of hand-rolled byte indexing; whitespace.

## Dropped during review
An earlier change gated `handle_ctr_tx` on `xfer->total_len > 0` (plus a dependent single-bool refactor). The Codex review + `test_usbd_control_in_zlp` showed it regresses the terminating data-stage ZLP of a control read: that ZLP completes `handle_ctr_tx` with `total_len == 0`, so the BULK gate would be skipped right before the status OUT, reopening the premature-ACK window. The original `ep0_ctrl_dir_in && ep0_ctrl_has_data` keys on transfer direction and handles the ZLP correctly, so it's kept.

The review also added a `#if !defined(TUP_USBIP_FSDEV_CH32)` guard to keep `edpt0_prepare_setup()` in `dcd_edpt0_status_complete` for STM32/AT32. Per @HiFiPhile that's unnecessary: the call only sizes the EP0 RX buffer (to 8), and the unconditional reset to MPS in `handle_ctr_rx` already re-arms it for all ports (a SETUP fits in any buffer ≥ 8). It was a ch32v203 setup-race fix originally (315dae6a8), so dropping it for all ports — as this PR does — is correct. Guard removed.

## Verified on hardware
nanoch32v203 (fsdev): instrumented DFU example with a deliberate delay before arming EP0 OUT to force the premature-ACK window. Baseline corrupts / fails the download; with these fixes a **16 KB multi-packet DFU download completes with 0 byte errors**.

### DIFF
--- a/src/portable/st/stm32_fsdev/dcd_stm32_fsdev.c
+++ b/src/portable/st/stm32_fsdev/dcd_stm32_fsdev.c
@@ -833,6 +833,14 @@ void dcd_edpt_stall(uint8_t rhport, uint8_t ep_addr) {
   ep_reg &= U_EPREG_MASK | EP_STAT_MASK(dir);
   ep_change_status(&ep_reg, dir, EP_STAT_STALL);
 
+#if defined(TUP_USBIP_FSDEV_CH32)
+  // Stall ends the transfer without edpt_xfer() (the only other CONTROL restore); else a rejected
+  // control-write leaves EP0 typed BULK and the host's recovery SETUP is ignored until bus reset.
+  if (ep_num == 0u) {
+    ep_reg = (ep_reg & ~U_EP_T_FIELD) | U_EP_CONTROL;
+  }
+#endif
+
   ep_write(ep_idx, ep_reg, true);
 }
 
--- a/src/portable/st/stm32_fsdev/dcd_stm32_fsdev.c
+++ b/src/portable/st/stm32_fsdev/dcd_stm32_fsdev.c
@@ -159,12 +159,19 @@ TU_ATTR_ALWAYS_INLINE static inline xfer_ctl_t *xfer_ctl_ptr(uint8_t epnum, uint
 }
 
 #if defined(TUP_USBIP_FSDEV_CH32)
-// CH32 FSDEV workaround: gate EP0 handshakes by switching type between CONTROL and BULK.
+// CH32 EP0 workaround: gate handshakes by switching EP0 type CONTROL<->BULK.
+// need_exclusive brackets the read-modify-write so the USB ISR can't race it.
 TU_ATTR_ALWAYS_INLINE static inline void ep0_set_type(uint32_t ep_type, bool need_exclusive) {
+  if (need_exclusive) {
+    fsdev_int_disable(0);
+  }
   uint32_t ep_reg = ep_read(0) | U_EP_CTR_TX | U_EP_CTR_RX;
   ep_reg &= U_EPREG_MASK;
   ep_reg = (ep_reg & ~U_EP_T_FIELD) | ep_type;
-  ep_write(0, ep_reg, need_exclusive);
+  ep_write(0, ep_reg, false);
+  if (need_exclusive) {
+    fsdev_int_enable(0);
+  }
 }
 #endif
 
--- a/src/portable/st/stm32_fsdev/dcd_stm32_fsdev.c
+++ b/src/portable/st/stm32_fsdev/dcd_stm32_fsdev.c
@@ -291,7 +291,7 @@ static void handle_ctr_tx(uint32_t ep_id) {
 #if defined(TUP_USBIP_FSDEV_CH32)
     // Control read: block unsolicited EP0 OUT ACK.
     if ((ep_num == 0u) && ep0_ctrl_dir_in && ep0_ctrl_has_data) {
-        ep0_set_type(U_EP_BULK, false);
+      ep0_set_type(U_EP_BULK, false);
     }
 #endif
     dcd_event_xfer_complete(0, ep_num | TUSB_DIR_IN_MASK, xfer->queued_len, XFER_RESULT_SUCCESS, true);
@@ -311,9 +311,9 @@ static void handle_ctr_setup(uint32_t ep_id) {
   // Setup packet should always be 8 bytes. If not, we probably missed the packet
   if (rx_count == 8) {
 #if defined(TUP_USBIP_FSDEV_CH32)
-    uint16_t const setup_w_length = (uint16_t) setup_packet[6] | ((uint16_t) setup_packet[7] << 8);
-    ep0_ctrl_dir_in = (setup_packet[0] & TUSB_DIR_IN_MASK) != 0u;
-    ep0_ctrl_has_data = (setup_w_length != 0u);
+    tusb_control_request_t const *request = (tusb_control_request_t const *) (void *) setup_packet;
+    ep0_ctrl_dir_in = (request->bmRequestType_bit.direction == TUSB_DIR_IN);
+    ep0_ctrl_has_data = (request->wLength != 0u);
 
     // For control write, block unsolicited EP0 OUT ACK until transfer is armed in edpt_xfer().
     if (!ep0_ctrl_dir_in && ep0_ctrl_has_data) {
@@ -365,7 +365,6 @@ static void handle_ctr_rx(uint32_t ep_id) {
     tu_hwfifo_read(pma_buf, xfer->buffer + xfer->queued_len, rx_count, NULL);
   }
   xfer->queued_len += rx_count;
-
 
   if ((rx_count < xfer->max_packet_size) || (xfer->queued_len >= xfer->total_len)) {
     // all bytes received or short packet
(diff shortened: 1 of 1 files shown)


src/portable/st/stm32_fsdev/dcd_stm32_fsdev.c
```diff
@@ -159,12 +159,19 @@ TU_ATTR_ALWAYS_INLINE static inline xfer_ctl_t *xfer_ctl_ptr(uint8_t epnum, uint
 }
 
 #if defined(TUP_USBIP_FSDEV_CH32)
-// CH32 FSDEV workaround: gate EP0 handshakes by switching type between CONTROL and BULK.
+// CH32 EP0 workaround: gate handshakes by switching EP0 type CONTROL<->BULK.
```
hathach, line 162: **Non-atomic before this fix.** `ep_write(.., need_exclusive)` masks the USB IRQ only around the *store*; the `ep_read` above ran unmasked, so a SETUP arriving mid-RMW could have its BULK gate dropped:
```
task: r = ep_read(EP0)             // IRQ still enabled
ISR : SETUP -> ep0_set_type(BULK)  // installs the gate
task: ep_write(r, type=CONTROL)    // overwrites it -> the OUT it was meant to block gets ACKed
```
Bracketing the whole read-modify-write makes it atomic w.r.t. the ISR.

hathach (commented):
Inline notes anchored to each change (the PR description has the overview). All CH32-gated except the `edpt0_prepare_setup` guard; hardware-verified on nanoch32v203 (16 KB DFU, 0 errors).

src/portable/st/stm32_fsdev/dcd_stm32_fsdev.c
```diff
@@ -304,12 +311,11 @@ static void handle_ctr_setup(uint32_t ep_id) {
   // Setup packet should always be 8 bytes. If not, we probably missed the packet
   if (rx_count == 8) {
 #if defined(TUP_USBIP_FSDEV_CH32)
-    uint16_t const setup_w_length = (uint16_t) setup_packet[6] | ((uint16_t) setup_packet[7] << 8);
-    ep0_ctrl_dir_in = (setup_packet[0] & TUSB_DIR_IN_MASK) != 0u;
-    ep0_ctrl_has_data = (setup_w_length != 0u);
+    tusb_control_request_t const *request = (tusb_control_request_t const *) (void *) setup_packet;
```
hathach, line 314: Parse via `tusb_control_request_t` (as `dcd_ch32_usbfs.c` does) instead of hand-rolled `setup_packet[6] | (setup_packet[7] << 8)` byte indexing. `setup_packet` is already 4-byte aligned.

src/portable/st/stm32_fsdev/dcd_stm32_fsdev.c
```diff
@@ -476,6 +481,12 @@ void dcd_edpt0_status_complete(uint8_t rhport, const tusb_control_request_t *req
     const uint8_t dev_addr = (uint8_t)request->wValue;
     FSDEV_REG->DADDR       = (U_DADDR_EF | dev_addr);
   }
+
+#if !defined(TUP_USBIP_FSDEV_CH32)
+  // Re-arm EP0 RX for the next SETUP. CH32 does this in handle_ctr_rx() instead; other fsdev
```
hathach: The workaround dropped this call from `dcd_edpt0_status_complete` **unconditionally**, which silently changed STM32/AT32 too: they lost their explicit next-SETUP RX re-arm (it then survived only as a side effect of the `ch32v203` bufsize reset in `handle_ctr_rx`). Guarded so ST/AT32 keep the call and CH32 keeps the new path.

src/portable/st/stm32_fsdev/dcd_stm32_fsdev.c
```diff
@@ -775,6 +785,13 @@ static bool edpt_xfer(uint8_t rhport, uint8_t ep_num, tusb_dir_t dir) {
       btable_set_rx_bufsize(ep_idx, BTABLE_BUF_RX, cnt);
     }
 
+#if defined(TUP_USBIP_FSDEV_CH32)
+    // Restore CONTROL in the same write as STAT_RX=VALID (after bufsize): a separate earlier
```
hathach, line 784: `edpt_xfer` previously restored CONTROL at the top of the function, *before* `btable_set_rx_bufsize` / `STAT_RX=VALID` ran — re-enabling the blind ACK while the RX buffer was stale and STAT_RX still NAK, so a host-retried DATA OUT / status ZLP could be ACKed into the wrong buffer. Folding the restore into this single VALID write closes the gap; the IN path keeps the early restore (no pending OUT to blind-ACK).

src/portable/st/stm32_fsdev/dcd_stm32_fsdev.c
```diff
@@ -821,6 +838,14 @@ void dcd_edpt_stall(uint8_t rhport, uint8_t ep_addr) {
   ep_reg &= U_EPREG_MASK | EP_STAT_MASK(dir);
   ep_change_status(&ep_reg, dir, EP_STAT_STALL);
 
+#if defined(TUP_USBIP_FSDEV_CH32)
+  // Stall ends the transfer without edpt_xfer() (the only other CONTROL restore); else a rejected
```
hathach, line 837: `dcd_edpt_stall` preserves the type field and `edpt_xfer` was the only other place that restores CONTROL, so a control-write rejected by the class (e.g. DFU `DNLOAD` in the wrong state) left EP0 stuck BULK+STALL and the host's recovery SETUP was ignored until a bus reset. Restoring CONTROL for EP0 in the stall write fixes that.

src/portable/st/stm32_fsdev/dcd_stm32_fsdev.c
```diff
@@ -282,9 +288,10 @@ static void handle_ctr_tx(uint32_t ep_id) {
     dcd_transmit_packet(xfer, (uint16_t)ep_id);
   } else {
 #if defined(TUP_USBIP_FSDEV_CH32)
-    // Control read: block unsolicited EP0 OUT ACK.
-    if ((ep_num == 0u) && ep0_ctrl_dir_in && ep0_ctrl_has_data) {
-        ep0_set_type(U_EP_BULK, false);
+    // Control read done: block the status-stage OUT ACK. Gate on the completed IN transfer's
+    // length, not the ep0_ctrl_* flags (a SETUP serviced earlier in this ISR pass can overwrite them).
+    if ((ep_num == 0u) && (xfer->total_len > 0u)) {
```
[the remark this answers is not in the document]
hathach: Confirmed — reproduced against `test_usbd_control_in_zlp` (128-byte descriptor, EP0=64, wLength=256 → 64+64+**ZLP**+status). The terminating data-stage ZLP completes `handle_ctr_tx` with `total_len == 0`, so the `total_len > 0` gate skipped the BULK switch right before the status OUT — reopening the premature-ACK window on CH32.

Reverted that change (and the dependent single-bool refactor); `handle_ctr_tx` is back to `ep0_ctrl_dir_in && ep0_ctrl_has_data`, which keys on transfer direction and so still gates the terminating ZLP. Fixed in f18daede4 (force-pushed). Thanks!

src/portable/st/stm32_fsdev/dcd_stm32_fsdev.c
HiFiPhile: @hathach I think it's not needed, `edpt0_prepare_setup()` only set the Rx buffer size to 8, while it's already set to MPS:
https://github.com/hathach/tinyusb/blob/f18daede40a5fba5fd34ea755cd05077ea8d8805/src/portable/st/stm32_fsdev/dcd_stm32_fsdev.c#L373

hathach: you are right, that is originally to fix ch32v issue/race that is correctly fixed by your PR

HiFiPhile (approved):
looks good !

PR #3720, opened 2026-06-21
audio: accept class request wIndex addressed to an AS interface

fedurca:
audiod_verify_entity_exists() only matched the function's standard Audio Control (AC) interface number against the request wIndex. Some hosts (notably Linux/ALSA) address class-specific entity requests (e.g. CLOCK_SOURCE GET_CUR SAM_FREQ / CLK_VALID) using one of the function's Audio Streaming (AS) interface numbers instead of the AC interface.

With the previous check those requests failed entity verification and the control transfer was rejected, which on RP2350 manifested as the host timing out on the clock-validity query (err -110) and never starting the isochronous stream.

Accept the request if wIndex refers to the AC interface (original behaviour) OR to any interface descriptor within the function's own descriptor block. The scan is bounded by desc_length and mirrors the existing audiod_verify_itf_exists() pattern, so it cannot read out of bounds and never drops a request the original code accepted.

### DIFF
--- a/src/class/audio/audio_device.c
+++ b/src/class/audio/audio_device.c
@@ -1731,20 +1731,44 @@ bool tud_audio_buffer_and_schedule_control_xfer(uint8_t rhport, tusb_control_req
 static bool audiod_verify_entity_exists(uint8_t itf, uint8_t entityID, uint8_t *func_id) {
   uint8_t i;
   for (i = 0; i < CFG_TUD_AUDIO; i++) {
-    // Look for the correct driver by checking if the unique standard AC interface number fits
-    if (_audiod_fct[i].p_desc && ((tusb_desc_interface_t const *) _audiod_fct[i].p_desc)->bInterfaceNumber == itf) {
-      // Get pointers after class specific AC descriptors and end of AC descriptors - entities are defined in between
-      uint8_t const *p_desc = tu_desc_next(_audiod_fct[i].p_desc);// Points to CS AC descriptor
-      p_desc = tu_desc_next(p_desc);// Get past CS AC descriptor
+    if (_audiod_fct[i].p_desc == NULL) {
+      continue;
+    }
 
-      while (_audiod_fct[i].p_desc_as - p_desc > 0) {
-        // Entity IDs are always at offset 3
-        if (p_desc[3] == entityID) {
-          *func_id = i;
-          return true;
+    // The class-specific request's wIndex (itf) normally carries this function's
+    // standard Audio Control (AC) interface number. However, some hosts (notably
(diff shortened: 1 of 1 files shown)


HiFiPhile (changes requested):
> audiod_verify_entity_exists() only matched the function's standard Audio Control (AC) interface number against the request wIndex. Some hosts (notably Linux/ALSA) address class-specific entity requests (e.g. CLOCK_SOURCE GET_CUR SAM_FREQ / CLK_VALID) using one of the function's Audio Streaming (AS) interface numbers instead of the AC interface.

That's definitely not the case, UAC class has been verified against linux 4.15-7.0, you have likely messed up your configuration descriptor.

That's also against UAC specification.

Running uac2_speaker example on linux 7.0:
Image

fedurca:
Thank you @HiFiPhile for the review and the USB capture — you were  absolutely right.

TLDR - I was wrong and your response helped me greatly to find my error in my project.

**What I got wrong**
I claimed that Linux/ALSA addresses UAC2 entity control requests (e.g.
`CLOCK_SOURCE GET_CUR SAM_FREQ` / `CLK_VALID`) using an Audio Streaming (AS)
interface number in the low byte of `wIndex`, and that TinyUSB's AC-only check
in `audiod_verify_entity_exists()` was therefore rejecting valid host traffic.
That was incorrect. Per the UAC2 spec, entity requests use
`wIndex = (EntityID << 8) | AC_interface_number`, and your capture on Linux 7.0
shows exactly that (`wIndex = 0x0400` → entity 4, AC interface 0). Linux does
not use the AS interface for these requests.

**What actually broke my device**
The `-110` (ETIMEDOUT) on clock-validity was not entity verification failing
on a wrong `wIndex` — the device stopped responding on EP0 due to unrelated
issues (RP2350 DCD abort spin, EP0 busy/claimed state after `SET_INTERFACE`,
deferred ISO IN timing, and `TU_VERIFY` UB under `-O3` in our local tree).
We also had a bug in our **configuration descriptor**: `bNumInterfaces` was 4
while only three interfaces exist (AC=0, AS=1, CDC=2). Fixing that to 3 was
the correct descriptor-side fix; loosening entity verification was not.

**Verification on my side**
After correcting `bNumInterfaces` and removing Fix 13 from my local patches,
6-channel UAC2 capture @ 48 kHz still works reliably on Linux with the
remaining RP2350 / EP0 / deferred-ISO fixes kept locally.
I'll close this PR — the stack behaviour is correct per spec, and the problem
was on our side. Thanks again for catching this and for maintaining UAC2
compliance in TinyUSB.

PR #3721, opened 2026-06-21
fix: release hardware spinlock in tud_deinit/tuh_deinit

renjieah:
On `OPT_OS_PICO`, `tuh_rhport_init()` calls `osal_spin_init(&_usbh_spin)`, which maps to `critical_section_init()` and claims a hardware spinlock. There is no `osal_spin_deinit()`, so `tuh_deinit()` never releases it — every `tuh_init()` / `tuh_deinit()` cycle leaks one spinlock.

This is easy to miss on most targets but bites hard on RP2350, where the spinlock pool is small and mostly taken by the SDK. After only a few host re-inits the pool runs out and `hw_claim_unused_from_range()` panics. I ran into it tearing down and rebuilding a Pico-PIO-USB host at runtime — it panicked reliably on the 4th rebuild. Backtrace at the panic:

    #0  panic
    #1  hw_claim_unused_from_range
    #2  spin_lock_claim_unused
    #3  critical_section_init
    #4  tuh_rhport_init

The fix adds `osal_spin_deinit()` to the OSAL backends and calls it from `tuh_deinit()`, pairing it with the `osal_spin_init()` in `tuh_rhport_init()`:

- `osal_pico.h`: `critical_section_deinit()`
- `osal_none.h` / `osal_freertos.h`: no-op (nothing to free there)

With this, the same teardown/rebuild loop runs 20+ times with the spinlock count staying flat (was dying on the 4th).


HiFiPhile:
Thank you, I've also added deinit to tud_deinit.

HiFiPhile (approved):

PR #3722, opened 2026-06-21
device: clamp EP0 OUT data copy to prevent buffer overrun

sourabhligade:
## Summary

Fixes a memory corruption issue in the USB device control transfer path (`usbd_control_xfer_cb`).

`usbd_control_xfer_cb()` copied `xferred_bytes` from the EP0 bounce buffer into the requester's buffer with no bound. A non-compliant or hostile host that sends an OUT data packet larger than the control transfer's `data_len` (`min(len, wLength)`, the buffer capacity) would overflow that buffer and over-count `total_xferred`.

This lives in the shared control handler, so it affects every DCD.

## Fix

Clamp `xferred_bytes` to the remaining buffer space (`data_len - total_xferred`) before the `memcpy` and accounting. No-op for conforming hosts (they never send more than `wLength`).

## Test plan

- [x] `ceedling test:all` — 61/61 pass
- [x] New regression test `test_usbd_control_out_overrun_clamp` verifies that when the DCD reports a full EP0 packet on an 8-byte vendor OUT transfer, usbd clamps, completes the data stage, and queues the IN status stage (not another OUT data arm)

### DIFF
--- a/src/device/usbd.c
+++ b/src/device/usbd.c
@@ -914,6 +914,8 @@ static bool usbd_control_xfer_cb(uint8_t rhport, uint8_t ep_addr, xfer_result_t 
   // Data stage progress
   if (ctrl_xfer->request.bmRequestType_bit.direction == TUSB_DIR_OUT) {
     TU_VERIFY(ctrl_xfer->buffer);
+    // Clamp host overrun to remaining capacity (data_len) so memcpy can't overflow the caller buffer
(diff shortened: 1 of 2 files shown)


HiFiPhile:
Why are you coping #3705 ??

sourabhligade:
@HiFiPhile I am really sorry it was an llm agent mess-up! Apologies

PR #3731, opened 2026-06-25
fix(dwc2_hcd): correct HFIR for Low-Speed devices via internal FS PHY

zjzhang-cn:
When a Low-Speed device is connected through the internal FS PHY, the effective PHY clock is 6MHz (HCFG_FSLS_PHYCLK_SEL_6MHZ), but phy_clock was incorrectly left at 48. This caused HFIR to be calculated as 47999 (~8ms SOF interval) instead of the correct 5999 (1ms SOF interval), breaking periodic endpoint scheduling.

Fixes: LS mouse only receiving first HID report on OTG_HS + FS PHY


HiFiPhile (changes requested):
> When a Low-Speed device is connected through the internal FS PHY, the effective PHY clock is 6MHz (HCFG_FSLS_PHYCLK_SEL_6MHZ), but phy_clock was incorrectly left at 48. 

~~No the reasoning is wrong, in LS mode PHY clock is divided by 8 with `HCFG.FSLSPclkSel`.~~

If you don't have the databook please refer to Linux driver:
https://github.com/torvalds/linux/blob/ab9de95c9cf952332ab79453b4b5d1bfca8e514f/drivers/usb/dwc2/hcd.c#L326
https://github.com/torvalds/linux/blob/ab9de95c9cf952332ab79453b4b5d1bfca8e514f/drivers/usb/dwc2/hcd_intr.c#L276

zjzhang-cn:
I am using the STM32F405's OTG_HS built-in FS PHY as a USB Host, and encountered this issue when a low-speed mouse was plugged in. After making this modification, the problem disappeared.

zjzhang-cn:
STM32F4xx_HAL_Driver/Src/stm32f4xx_hal_hcd.c
```c
if (hhcd->Init.phy_itface == USB_OTG_EMBEDDED_PHY)
{
    if ((hprt0 & USB_OTG_HPRT_PSPD) == (HPRT0_PRTSPD_LOW_SPEED << 17))
    {
        (void)USB_InitFSLSPClkSel(hhcd->Instance, HCFG_6_MHZ); 
    }
    else
    {
        (void)USB_InitFSLSPClkSel(hhcd->Instance, HCFG_48_MHZ);  
    }
}
```
stm32f4xx_ll_usb.c ->USB_InitFSLSPClkSel
```c

/**
  * @brief  USB_InitFSLSPClkSel : Initializes the FSLSPClkSel field of the
  *         HCFG register on the PHY type and set the right frame interval
  * @param  USBx  Selected device
  * @param  freq  clock frequency
  *          This parameter can be one of these values:
  *           HCFG_48_MHZ : Full Speed 48 MHz Clock
  *           HCFG_6_MHZ : Low Speed 6 MHz Clock
  * @retval HAL status
  */
HAL_StatusTypeDef USB_InitFSLSPClkSel(const USB_OTG_GlobalTypeDef *USBx, uint8_t freq)
{
  uint32_t USBx_BASE = (uint32_t)USBx;

  USBx_HOST->HCFG &= ~(USB_OTG_HCFG_FSLSPCS);
  USBx_HOST->HCFG |= (uint32_t)freq & USB_OTG_HCFG_FSLSPCS;

  if (freq == HCFG_48_MHZ)
  {
    USBx_HOST->HFIR = HFIR_48_MHZ;
  }
  else if (freq == HCFG_6_MHZ)
  {
    USBx_HOST->HFIR = HFIR_6_MHZ;
  }
  else
  {
    return HAL_ERROR;
  }

  return HAL_OK;
}
```

Maybe it should be modified here
```c
  uint32_t hfir = dwc2->hfir & ~HFIR_FRIVL_Msk;
  if (speed == TUSB_SPEED_HIGH) {
    hfir |= 125*phy_clock - 1; // The "- 1" is the correct value. The Synopsys databook was corrected in 3.30a
  } else {
    hfir |= 1000*phy_clock - 1;
  }
```

HiFiPhile (approved):
Thank you, I just rechecked and found out in dedicated FS phy case, PHY clock needs to set to 6MHz running at LS:

Image


Also verified on STM32F407 (GenID 2.81A) and ESP32-S3 (GenID 4.00A) with a keyboard of `bInterval=10`:
Before:
image

After:
image

PR #3737, opened 2026-06-26
Add ST board NUCLEO_H533RE

sgrunza-at-BadgerMeter:
Add ST board NUCLEO_H533RE.  It is very similar to the NUCLEO_H503 board.


HiFiPhile (approved):
Thank you !

PR #3738, opened 2026-06-27
feat(midi2): derive function blocks from the GTB descriptor

sauloverissimo:
The device reports its function blocks straight from the Group Terminal Block
descriptor, so what it sends in UMP Stream FB Info (direction, group span) always
matches what it declares over USB. The built-in responder answers Endpoint and
Function Block Discovery from that descriptor, so a device with more than one
block works out of the box.

Two optional weak callbacks for apps that need more:
- tud_midi2_fb_name_cb: per-block names.
- tud_midi2_stream_msg_cb: see a UMP Stream message before the built-in responder
  (PASS / HANDLED / NEGOTIATED_*), for endpoints with dynamic topology.

This started as the function block direction fix and grew to cover the related
pieces, so I'm closing #3739 and folding its intent into the stream callback here.

Tested on STM32F411 and RP2350 against Windows MIDI Services and ALSA: each block
shows the right name, direction and group on Windows and Linux. Device-only; the
MIDI 1.0 driver is untouched.

Ref #3571

### DIFF
--- a/examples/device/midi2_device/src/main.c
+++ b/examples/device/midi2_device/src/main.c
@@ -546,6 +546,25 @@ void tud_midi2_set_itf_cb(uint8_t itf, uint8_t alt) {
   printf("[ALT] Host selected alt=%u\r\n", (unsigned)alt);
 }
 
+// PROTOTYPE: two Function Blocks with independent directions.
+//   FB0: output only, group 0   FB1: input only, group 1
+static const uint8_t gtb_two_blocks[] = {
+  TUD_MIDI2_GTB_HEADER(2),
+  TUD_MIDI2_GTB_BLOCK(1, MIDI2_GTB_OUTPUT_ONLY, 0, 1, 0),  // block 1: group 0
+  TUD_MIDI2_GTB_BLOCK(2, MIDI2_GTB_INPUT_ONLY,  1, 1, 0),  // block 2: group 1
+};
+
+const uint8_t* tud_midi2_gtb_desc_cb(uint8_t itf, uint16_t* len) {
+  (void)itf;
+  *len = sizeof(gtb_two_blocks);
+  return gtb_two_blocks;
+}
+
+const char* tud_midi2_fb_name_cb(uint8_t itf, uint8_t fb_idx) {
+  (void)itf;
+  return (fb_idx == 0) ? "Synth Out" : "Keys In";
+}
+
 //--------------------------------------------------------------------+
 // Initial Setup - Program Change, CC, Per-Note Management
 //--------------------------------------------------------------------+
--- a/examples/device/midi2_device/src/tusb_config.h
+++ b/examples/device/midi2_device/src/tusb_config.h
@@ -81,6 +81,10 @@ extern "C" {
 //------------- CLASS -------------//
 #define CFG_TUD_MIDI2           1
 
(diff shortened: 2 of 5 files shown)


sauloverissimo:
Addressed the readability note: replaced the `0x33` literal with the explicit bitfield composition (UI hint sender/receiver + bidirectional direction). Same value, now self-documenting.

HiFiPhile:
Do you think it's useful to add a weak callback to define direction per instance ?

sauloverissimo:
Hi @HiFiPhile! Makes sense, and it lines up with the existing callbacks (`num_groups`, `num_function_blocks`, `ep_name`, `product_id`). Direction is the one FB attribute that isn't overridable yet. We could add a `tud_midi2_..._cb(itf)` returning the direction, with bidirectional as the default to keep current behavior, and have the driver apply it to both the GTB `bGrpTrmBlkType` and the FB Info. I can include it in this PR. Were you picturing it per instance (`itf`), like the other callbacks, or per function block? Either way, I'll prototype it here and bring something to show.

sauloverissimo:
Hi @HiFiPhile! Prototyped it per instance: `tud_midi2_direction_cb(itf)`, following the existing callbacks, returning bidirectional (default) / input-only / output-only, driving both the GTB `bGrpTrmBlkType` and the FB Info direction + UI hint.

Validated on hardware with an RP2350 and an STM32F411 (dwc2). On Linux I confirmed the GTB on the wire (the `bGrpTrmBlkType` changes with the callback), and on Windows MIDI Services the ports line up: output-only shows up as MIDI In only, bidirectional as In + Out.

Kept it per instance for now. Per function block would be the natural generalization, but that also pulls in giving each FB its own group span and multiple GTBs, which is a bigger change. Would you rather stay per-instance or aim for per-FB? 2350_fb

HiFiPhile:
> Would you rather stay per-instance or aim for per-FB?

Since you know MIDI2 better you decide it :)

sauloverissimo:
I've been studying M2-104-UM more closely and testing on hardware over the past few days, and this part of the stack makes a lot more sense to me now. The more I worked on it the more the pieces overlapped, so I decided to bring everything together here and close #3739.

It started as the function block direction fix, but it grew a little. The direction and group span for each function block now come straight from the Group Terminal Block descriptor, so what the device reports in FB Info always matches what it actually declares. The built-in UMP Stream responder answers Endpoint and Function Block Discovery from that same descriptor too, so a device with more than one block just works without extra code.

For apps that need more control there are two optional weak callbacks: tud_midi2_fb_name_cb to name the blocks, and tud_midi2_stream_msg_cb to look at a Stream message before the built-in responder does, returning PASS to let it handle the message, HANDLED to answer it yourself, or NEGOTIATED_* to set the protocol.

That last one is really why #3739 isn't needed anymore. It does what the CFG_TUD_MIDI2_USER_RESPONDER flag did, covering the dynamic case like a bridge whose blocks follow whatever is plugged into it, but per message and without a compile-time switch, which felt cleaner than handing the whole responder to the app.

I tested it on an STM32F411 first, then an RP2350. Both Windows MIDI Services and ALSA now enumerate the device fully, each function block showing its name, direction (input, output or bidirectional) and group correctly, on Windows and Linux.

image

image

HiFiPhile (approved):
Thank you

PR #3739, opened 2026-06-27
feat(midi2): opt-in app responder for UMP Stream

sauloverissimo:
## Summary
Small, opt-in change, quick to review. Touches only the MIDI 2.0 device class
(legacy MIDI 1.0 driver untouched) and is byte-identical to current behavior
when the flag is off.

Adds CFG_TUD_MIDI2_USER_RESPONDER (default off). When set, the driver leaves
UMP Stream messages in the RX FIFO so the application can answer Endpoint /
Function Block discovery and Stream Configuration itself, for endpoints whose
topology is dynamic (e.g. a bridge whose function blocks track the devices
plugged into it). Built-in responder is unchanged when off.

## Test plan
- [x] Builds with the flag off (default) and on (RP2350 midi2_device example).
- [x] Used on hardware by an ESP32-P4 USB MIDI bridge to expose dynamic FB names/groups.

Ref #3571

### DIFF
--- a/src/class/midi/midi2_device.c
+++ b/src/class/midi/midi2_device.c
@@ -270,6 +270,7 @@ static uint32_t _tx_ump_write(midi2d_interface_t* p_midi, const uint32_t* words,
 //--------------------------------------------------------------------+
 // Protocol Negotiation
 //--------------------------------------------------------------------+
+#if !CFG_TUD_MIDI2_USER_RESPONDER
 static void _nego_send_ump(midi2d_interface_t* p_midi, const uint32_t* words, uint8_t count) {
   if (!_tx_opened(p_midi)) return;
   if (tu_fifo_remaining(&p_midi->ep_stream.tx.ff) < (uint32_t) count * 4) return;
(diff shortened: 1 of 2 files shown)


sauloverissimo:
Updated to address the automated review: added `tud_midi2_n_set_protocol()` so the application can keep `tud_midi2_n_protocol` / `tud_midi2_n_negotiated` accurate when it owns UMP Stream negotiation, plus header/.c notes on draining the RX FIFO. Builds with the flag off (default) and on.

src/class/midi/midi2_device.c
```diff
@@ -402,6 +411,7 @@ static void _nego_process_rx(midi2d_interface_t* p_midi) {
     tu_edpt_stream_read(ep_rx, buf, pkt_bytes);
```
HiFiPhile, line 411: I'm thinking about adding a weak callback here, with the return value we can differentiate between `Handled`, `Negotiated to V1` , `Negotiated to V2` , etc.

Maybe the user only want to override a one specific response instead of implement full negotiation in app.

sauloverissimo: Hi @HiFiPhile! Are you thinking of a per-message hook at the stream dispatch, where the return value decides fallthrough vs handled? So the app could intercept just FB Discovery, say, and let the built-in responder take the rest.

HiFiPhile: Yes exactely.

sauloverissimo:
Closing this in favor of #3738. While I was working through the function block changes there, I ended up adding a per-message callback (tud_midi2_stream_msg_cb, returning PASS / HANDLED / NEGOTIATED_*) that does what the CFG_TUD_MIDI2_USER_RESPONDER flag here does, but per message and without a compile-time switch. That made this one redundant.

PR #3741, opened 2026-06-27
bound ndp16 wLength against received ntb in recv_validate_datagram

dxbjavid:
the ncm driver validates each incoming ntb in recv_validate_datagram before passing it to the glue logic. the first ndp's wLength is the size of the datagram pointer block and it comes straight from the host on the out endpoint, but the function only checks it against a minimum and never an upper bound. max_ndx is then derived from wLength and used to index ndp16_datagram[max_ndx - 1], and to walk the array, inside ntb->data which is only CFG_TUD_NCM_OUT_NTB_MAX_SIZE bytes (3200 by default). a short transfer carrying a wLength near 0xffff yields a max_ndx of roughly 16000, so the validator reads tens of kilobytes past the receive buffer while it is still deciding whether the packet is valid. i came across it while reading the bounds logic around wNdpIndex, which is capped against len whereas wLength is not. rejecting the ntb when wNdpIndex + wLength runs past the received length keeps the array within the buffer and lines up with the checks already done just above.


dxbjavid:
for anyone picking this up, here is how i triggered it. i built a small standalone harness that mirrors the recv_validate_datagram arithmetic over a heap-allocated recv_ntb_t (data[3200]) and compiled it under -fsanitize=address.

the crafted NTB is a short 64-byte transfer. a valid nth16 (NTH16 signature, wHeaderLength = sizeof(nth16_t), wBlockLength = 64, wNdpIndex = 12), then an ndp16 at data+12 carrying the NCM0 signature, wNextNdpIndex = 0 and wLength = 0xfff0. the checks above only bound wLength from below, so max_ndx works out to (0xfff0 - 8) / 4 = 16380 and the ndp16_datagram[max_ndx - 1] access lands roughly 65 kB into the 3200-byte buffer.

before the patch ASan flags a heap-buffer-overflow read of size 2 at that access. with the added wNdpIndex + wLength > len check the NTB is rejected cleanly (12 + 65520 > 64) and there is no out-of-bounds read.

happy to fold this into the net fuzz harness as a seed if that is the preferred home for the regression.

HiFiPhile:
> happy to fold this into the net fuzz harness as a seed if that is the preferred home for the regression.

Thank you it would be nice.

dxbjavid:
done. added it under test/fuzz/device/net_ncm as a small self-contained harness. it feeds a raw ntb straight into recv_validate_datagram (the driver is pulled in so the static function is reachable, with the handful of usbd symbols stubbed out) and net_ncm_seed_corpus.zip carries the crafted 64-byte ntb.

built against the unpatched driver the seed trips the asan heap-buffer-overflow in recv_validate_datagram; with the wLength bound in place it is rejected cleanly. there is a `make regression` target that replays the seed under asan if you want a quick check without the fuzzing engine. i kept it as its own target rather than enabling ncm in the existing net harness so the ecm/rndis coverage there is untouched, but happy to reshape it if you would rather it live elsewhere.

HiFiPhile (approved):
Thank you.

dxbjavid:
TY!

PR #3742, opened 2026-06-28
Fix RT-Thread MSC inquiry string bounds

Old-Ding:
## Summary
- Bound the RT-Thread MSC inquiry string copies to the fixed SCSI inquiry field sizes.
- Match the existing TinyUSB MSC examples that copy vendor/product/revision strings into 8/16/4 byte fields.

## Root cause
`tud_msc_inquiry_cb()` receives fixed-size output buffers: `vendor_id[8]`, `product_id[16]`, and `product_rev[4]`. The RT-Thread port copied `strlen()` bytes from configurable package strings into those buffers, so longer configuration values could write past the inquiry fields.

## Validation
- `git diff --check`
- Verified the old `memcpy(... strlen(...))` pattern is gone in `lib/rt-thread/port/msc_device_port.c`
- Verified the touched file still uses CRLF line endings

Local RT-Thread package build was not run because this checkout does not have the RT-Thread package build environment/toolchain configured.


HiFiPhile (approved):
Thank you

PR #3748, opened 2026-06-30
max3421: prevent USB host hang at startup

mikeysklar:
The MAX3421E can fail to report its oscillator ready after a soft reset, so `hcd_init()` waits on `USBIRQ_OSCOK` forever and the host hangs at startup. This bounds the wait so it proceeds instead (no-op when OSCOK arrives normally).

Needed on the Adafruit Feather ESP32 V2 + USB Host FeatherWing. Relates to adafruit/circuitpython#10053; complements adafruit/circuitpython#11081.

### DIFF
--- a/src/portable/analog/max3421/hcd_max3421.c
+++ b/src/portable/analog/max3421/hcd_max3421.c
@@ -504,8 +504,15 @@ bool hcd_init(uint8_t rhport, const tusb_rhport_init_t* rh_init) {
   // reset
   reg_write(rhport, USBCTL_ADDR, USBCTL_CHIPRES, false);
   reg_write(rhport, USBCTL_ADDR, 0, false);
+  // Bound the oscillator-OK wait. On boards where the MAX3421E is only
+  // soft-reset (CHIPRES) rather than hardware-reset via nRST, OSCOK may never
+  // latch, which would otherwise hang the host forever (adafruit/circuitpython#10053).
+  // The oscillator is generally running regardless, so proceed after a timeout.
+  uint32_t oscok_retries = 200000;
   while( !(reg_read(rhport, USBIRQ_ADDR, false) & USBIRQ_OSCOK_IRQ) ) {
-    // wait for oscillator to stabilize
+    if (--oscok_retries == 0) {
+      break;
+    }
   }
 
   // Mode: Host and DP/DM pull down
--- a/src/portable/analog/max3421/hcd_max3421.c
+++ b/src/portable/analog/max3421/hcd_max3421.c
@@ -82,6 +82,11 @@ enum {
 };
 
 enum {
+  TIME_TO_EXIT_SUSPEND_MS = 4u, // datasheet: PWRDOWN = 1 to 0 to OSCOKIRQ = 3 ms + 1 ms margin
+  TIME_CHIPRES_DELAY_MS   = 2u  // CHIPRES hold: 2 guarantees >= 1 ms actual despite ms-tick quantization
+};
+
+enum {
   CPUCTL_IE        = 1u << 0,
   CPUCTL_PULSEWID0 = 1u << 6,
   CPUCTL_PULSEWID1 = 1u << 7,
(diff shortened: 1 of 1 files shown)


hathach:
@claude review

hathach (approved):
thank you, I change the timeout to time based that machtes the datasheet 3ms (+1 in case we are in the last us). tusb_time_millis_api() must be implement, which is actually requirement for host stack on non-rtos.

Image

src/portable/analog/max3421/hcd_max3421.c
```diff
@@ -501,12 +504,21 @@ bool hcd_init(uint8_t rhport, const tusb_rhport_init_t* rh_init) {
...
   reg_write(rhport, USBCTL_ADDR, 0, false);
-  while( !(reg_read(rhport, USBIRQ_ADDR, false) & USBIRQ_OSCOK_IRQ) ) {
-    // wait for oscillator to stabilize
-  }
+
+  // Wait for OSCOK (12 MHz oscillator + 48 MHz PLL relock). Bounded as a safety net, so the host never
+  // hangs if the edge is still missed on some board - the clock is running regardless.
+  const uint32_t oscok_start_ms = tusb_time_millis_api();
+  while (!(reg_read(rhport, USBIRQ_ADDR, false) & USBIRQ_OSCOK_IRQ) &&
+         (tusb_time_millis_api() - oscok_start_ms < TIME_TO_EXIT_SUSPEND_MS)) {}
```
hathach, line 521: @mikeysklar after checking the datasheet/programming guide, I think the issue is probably we are writing the CHIPRES too fast within ~us that reset signal isn't long enough for the osc is fully stopped to assert the bit. Would you mind trying to comment out the `(tusb_time_millis_api() - oscok_start_ms < TIME_TO_EXIT_SUSPEND_MS)` the delay in this loop to see if that the above delay work. We will add this anyway, since it prevent actual host hang, any other race condition

mikeysklar: Tested your changes on a Feather V2 + FW MAX3421E. No regression. I see this as safe to merge.

Note: I am using a modded MAX3421E. Changed the [oversized crystal caps from 22pF -> 15pF](https://github.com/adafruit/Adafruit-USB-Host-FeatherWing-PCB/issues/2) which was needed to stabilize detection of lower current SanDisk adapters. 

```
Adafruit CircuitPython 10.3.0-alpha.3-7-gdaa9df856e-dirty on 2026-07-02; Adafruit Feather ESP32 V2 with ESP32
Board ID:adafruit_feather_esp32_v2
UID:8EF9D613E945
MAC:E8:9F:6D:31:9E:54
code.py output:
spi.frequency=20000000
Finding devices:
05e3:0751: USB Storage USB Storage
<switch device>
Finding devices:
0781:cfcb: SanDisk SDDR-B531
<switch device>
Finding devices:
154b:005b: PNY Technologies USB 2.0 FD
<switch device>
Finding devices:
0dd8:0562: Netac MobileDataStar Netac MobileDataStar
```

IMG_0560

hathach: @mikeysklar just for clarification, did you try with only 2ms delay between CHIPRES 1->0 and comment out the `(tusb_time_millis_api() - oscok_start_ms < TIME_TO_EXIT_SUSPEND_MS)` or test the PR as it is. Since I am suspsecting the CHIPRES without delay is too fast and could not generate the lactch.

PS: we will add the timeout regardless, I am just curious if that is real root cause :)

mikeysklar: What I had tested was:

- Full PR - CHIPRES hold + timeout 15pF - all devices found.

After your last reply I also tried:

- Full PR, 22 pF (reversed back to larger caps) - all 4 devices found
- Requested PR changes CHIPRES delay, OSCOK timeout commented out - all 4 devices found

My re-working of the caps seems to have made my crystal more stable. However, I believe all three suggested changes have value:

- Dan's [SPI fixed @ 20 MHz](https://github.com/adafruit/circuitpython/pull/11081)
- [15pF cap sizing for crystal](https://github.com/adafruit/Adafruit-USB-Host-FeatherWing-PCB/issues/2)
- CHIPRES / OSCOK

All are improvements that should help avoid these marginal conditions.

I can still get a failure running a stock CircuitPython 10.3.0-alpha.3 no device detected.

hathach: thank you for the clarification, the chipres delay may or may not help, but I guess it improve the situation and there is reason for it. Since esp32 is a very quick mcu, the time between the write is too fast to latch the event (osc dependent). Thank you very much

PR #3754, opened 2026-07-05
Fix USBTMC status byte interrupt buffer

### LINKED ISSUE #2928: Dangling pointer may result in stack corruption
### Operating System

Others

### Board

Any build using src/class/usbtmc/usbtmc_device.c class driver

### Firmware

CodeQL reported:

> Local variable address stored in non-local memory

While this alert is only a warning, further investigation suggests this
may be a latent stack corruption bug, and whether the corruption occurs
may be affected by compiler optimization settings (or prevented when
compiled with other optimization settings).   Thus, this appears to be
a subtle bug that is extremely difficult to catch, and which may disappear
when using debug builds. 

Therefore, please review carefully the description provided, which walks
the call tree with explanation.

### Relevant call stack

In `usbtmcd_control_xfer_cb()`, while handling [`USB488_bREQUEST_READ_STATUS_BYTE`](https://github.com/henrygab/tinyusb/blob/4fa02c6109817cfa51caf4624afede72008e277a/src/class/usbtmc/usbtmc_device.c#L889-L898):

* `intMsg` is declared as a local (stack) variable.
* `&intMsg` is passed as the `buffer` parameter to `usbd_edpt_xfer()`

Within [`usbd_edpt_xfer()`](https://github.com/henrygab/tinyusb/blob/4fa02c6109817cfa51caf4624afede72008e277a/src/device/usbd.c#L1343), the `buffer` parameter is passed as the `buffer` parameter to `dcd_edpt_xfer()`.

### Requirements for the `buffer` parameter

The contract for `dcd_edpt_xfer()` is generally described in the [porting](https://github.com/henrygab/tinyusb/blob/4fa02c6109817cfa51caf4624afede72008e277a/docs/contributing/porting.rst?plain=1#L197-L198) document:
> Once the transaction is going, the interrupt handler will notify TinyUSB of transfer completion.
> During transmission, the IN data buffer is guaranteed to remain unchanged in memory until the
> `dcd_xfer_complete` function is called.

Said another way, it is a **requirement** that the `buffer` parameter to `dcd_edpt_xfer()` must point
to memory that remains valid **_at least_** until the `dcd_xfer_complete()` function is called.  Therefore,
because `usbd_edpt_xfer()` does not ensure `dcd_xfer_complete()` is called, the same requirement
also applies to the `buffer` parameter to `usbd_edpt_xfer()`.

### Dangling Pointer Bug

It appears that the call at usbtmc_device.c@898 is incorrect, as `intMsg` is a local (stack)
variable, and its stack space may be reclaimed as soon as the variable's scope (the curly braces)
is exited (aka, immediately after the call to `usbd_edpt_xfer()`.

Therefore, because `&intMsg` is the buffer pointer, and there is no waiting for the transfer
to complete before exiting the scope of that local (stack) variable, the buffer pointer will
potentially still be in use when that stack location is reused for other purposes.

### Not mitigated by hardware

It's possible that some specific hardware happens to ensure the transfer is synchronous,
which can prevent the effect of the bug from occurring.
Similarly, if the the asynchronous transfer happens to complete before the stack location
is re-used, the effect of the bug can remain hidden.

However, `dcd_edpt_xfer()` appears to be a hardware-specific routine, which makes reliance
on this being a synchronous request unwise.  Moreover, each project using this source may
compile the code with different optimization settings ... making this a true Heisenbug.

### How to reproduce ?

Review of the source, using the above description to follow the problem.

No debug log is required.  This is a heisenbug, which disappears with timing changes / compilation optimization changes.

### Screenshots

_No response_

Old-Ding:
## Summary
- copy the USB488 READ_STATUS_BYTE interrupt payload into the persistent USBTMC endpoint notification buffer
- avoid passing a stack-local buffer to usbd_edpt_xfer() for an interrupt IN transfer that completes asynchronously

Fixes #2928

## Testing
- git diff --check origin/master..HEAD
- Not run: local cmake/ninja/make/gcc/arm-none-eabi-gcc toolchains are not available in this environment.

### DIFF
--- a/src/class/usbtmc/usbtmc_device.c
+++ b/src/class/usbtmc/usbtmc_device.c
@@ -826,7 +826,7 @@ bool usbtmcd_control_xfer_cb(uint8_t rhport, uint8_t stage, tusb_control_request
                   },
                   .StatusByte = tud_usbtmc_get_stb_cb(&(rsp.USBTMC_status))};
           // Must be queued before control request response sent (USB488v1.0 4.3.1.2)
-          usbd_edpt_xfer(rhport, usbtmc_state.ep_int_in, (void *) &intMsg, sizeof(intMsg), false);
+          (void) tud_usbtmc_transmit_notification_data(&intMsg, sizeof(intMsg));
         }
       } else {
         rsp.statusByte = tud_usbtmc_get_stb_cb(&(rsp.USBTMC_status));


Old-Ding:
The failed HIL entry is 
aspberry_pi_pico / host_info_to_device_cdc, which is outside the USBTMC path touched by this PR. I tried to rerun the failed workflow, but GitHub requires repository admin rights for that run.

HiFiPhile:
I think you can simply use `tud_usbtmc_transmit_notification_data`

Old-Ding:
Updated to queue the READ_STATUS_BYTE interrupt notification through 	ud_usbtmc_transmit_notification_data(). I kept the existing busy check before 	ud_usbtmc_get_stb_cb() so a busy interrupt endpoint still returns USB488_STATUS_INTERRUPT_IN_BUSY without consuming/clearing the status byte.

HiFiPhile (approved):
Thank you

PR #3755, opened 2026-07-05
Fix USBTMC notification endpoint claim

### LINKED ISSUE #2735: USBTMC check for EP not busy should return true in tud_usbtmc_transmit_notification_data
### Operating System

Linux

### Board

RPI Pico

### Firmware

tinyusb/src/class/usbtmc/usbtmc_device.c
tinyusb/examples/device/usbtmc/src/usbtmc_app.c

### What happened ?

PR #2494 introduced `tud_usbtmc_transmit_notification_data()` function. The check if EP is not busy bails out of the function if the EP is NOT busy:

https://github.com/hathach/tinyusb/blob/cfbdc44a8d099240ad5ef208bd639487c2f28153/src/class/usbtmc/usbtmc_device.c#L256

I believe it should be the opposite; bail if the EP IS busy and continue if not busy like so:

```
TU_VERIFY(usbd_edpt_busy(usbtmc_state.rhport, usbtmc_state.ep_int_in) == false);
```

### How to reproduce ?

Original code:

1. Add the following to `usbtmc_app.c`:
```
bool tud_usbtmc_notification_complete_cb(void) {
  usbtmc_srq_interrupt_488_t intMsg =
  {
    .bNotify1 = USB488_bNOTIFY1_SRQ,
    .StatusByte = 0xa5
  };
  bool ret = tud_usbtmc_transmit_notification_data((void*)&intMsg, sizeof(intMsg));
  (void)ret;

  return true;
}
```

2. Use the following python script to poll for interrupt condition: [poll-interrupt-in.txt](https://github.com/user-attachments/files/16349994/poll-interrupt-in.txt)

3. Observe `No data available (timeout)` being printed out by python script.

4. Change the https://github.com/hathach/tinyusb/blob/cfbdc44a8d099240ad5ef208bd639487c2f28153/src/class/usbtmc/usbtmc_device.c#L256
 to 

```
TU_VERIFY(usbd_edpt_busy(usbtmc_state.rhport, usbtmc_state.ep_int_in) == false);
```

5. Using the same script will now continuously print :
```
Data received: array('B', [129, 165])
```

I could not see any useful in the debug logs, I had to use the pico probe debugger to see that the interrupt EP is in fact not busy and the the code is wrong.

### Screenshots

_No response_

Old-Ding:
## Summary
- Use `usbd_edpt_claim()` before queuing USBTMC interrupt notification data.
- Keep the notification payload in the persistent endpoint buffer and release the claim if the copy fails before the transfer is queued.

Fixes #2735.

## Testing
- `git diff --check HEAD~1..HEAD`
- `git ls-files --eol src/class/usbtmc/usbtmc_device.c`

Not run: local build, because this Windows environment does not have CMake, Make, GCC, or Clang available.

### DIFF
--- a/src/class/usbtmc/usbtmc_device.c
+++ b/src/class/usbtmc/usbtmc_device.c
@@ -252,9 +252,12 @@ bool tud_usbtmc_transmit_notification_data(const void *data, size_t len) {
   TU_ASSERT(len > 0);
   TU_ASSERT(usbtmc_state.ep_int_in != 0);
 #endif
-  TU_VERIFY(usbd_edpt_busy(usbtmc_state.rhport, usbtmc_state.ep_int_in));
+  TU_VERIFY(usbd_edpt_claim(usbtmc_state.rhport, usbtmc_state.ep_int_in));
 
-  TU_VERIFY(tu_memcpy_s(usbtmc_epbuf.epnotif, CFG_TUD_USBTMC_INT_EP_SIZE, data, len) == 0);
+  if (tu_memcpy_s(usbtmc_epbuf.epnotif, CFG_TUD_USBTMC_INT_EP_SIZE, data, len) != 0) {
+    usbd_edpt_release(usbtmc_state.rhport, usbtmc_state.ep_int_in);
+    return false;
+  }
   TU_VERIFY(usbd_edpt_xfer(usbtmc_state.rhport, usbtmc_state.ep_int_in, usbtmc_epbuf.epnotif, (uint16_t) len, false));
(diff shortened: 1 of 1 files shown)


HiFiPhile:
Thank you, it's better to use `usbd_edpt_claim()`

Old-Ding:
Updated to use usbd_edpt_claim() for the interrupt notification endpoint, with an explicit release if the payload copy fails before queuing the transfer.

HiFiPhile (approved):
LGTM

PR #3756, opened 2026-07-05
bound rndis data packet offset in handle_incoming_packet

dxbjavid:
**overflow in the rndis data packet window check**

the DataOffset and DataLength fields in handle_incoming_packet come straight from the host on the bulk-out endpoint, and the window check adds them to the header offset in uint32 before comparing against the received length. on a 32-bit target that sum wraps, so a REMOTE_NDIS_PACKET_MSG with a large DataOffset/DataLength slips past the check and leaves pnt pointing well past the small rx buffer while size is taken from DataLength, which tud_network_recv_cb then reads out of bounds. rewrote the check to subtract from the received length instead so it cannot overflow, which keeps well-formed packets unchanged.

verified with a host build under ASAN: a packet with DataOffset=0x1000, DataLength=0xFFFFF000 and len=44 reports a heap read 2502 bytes past the 1602-byte buffer before the change, and is rejected after it.


HiFiPhile (approved):
Thank you.

PR #3757, opened 2026-07-06
Larger at32f403a PMA area

rhgndf:
Without CAN1/CAN2 enabled this can go up to 1280, but 768 is the max safe size if both are enabled.


HiFiPhile (approved):
Thank you

PR #3438, opened 2025-12-28
Fix NCM state on restart

HiFiPhile:
- Fixed notification state handling so that when the interface is deactivated and reactivated, the link state update notification is correctly sent. Fix #3436

- Refactored both `recv_ready_ntb` and `xmit_ready_ntb` buffers in `ncm_interface_t` to use circular buffers, adding head, tail, and count fields for each, replacing the previous linear search and memmove approach.

Issue #3436, opened 2025-12-27
NCM interface stop transmitting packets after ifconfig down+up on Linux side (STM32H7 DWC2)


ceedriic:
### Operating System

Linux

### Commit SHA

3af1bec

### Board

Custom stm32h747 board

### Firmware

Custom stm32h747 board, but the problematic code is very similar to net_lwip_webserver with NCM config.
Alpine Linux 3.23 on the host side.

### What happened ?

When the system run for the first time, everything is fine.

Configuration:
```
#define CFG_TUD_NCM_IN_NTB_N		2
#define CFG_TUD_NCM_IN_NTB_MAX_SIZE	2048
#define CFG_TUD_NCM_OUT_NTB_N		2
#define CFG_TUD_NCM_OUT_NTB_MAX_SIZE	2048
```

1. Start the net_lwip_webserver-similar TinyUSB device.
2. On linux, after eth1 comes up, do `ifconfig eth1 up` and `dhcpcd eth1`
3. At this point, eth1 gets an IP address (given by TinyUSB dhcpd server), and the interface can be pinged, all is fine.
4. Run `ifconfig eth1 down` on the Linux side.
5. At this point, Linux stop listening to the INTERRUPT and BULK endpoints.
6. Run `ifconfig eth1 up` on the Linux side.
7. At this point, Linux is again listening to the INTERRUPT and BULK endpoints.
8. Observe that eth1 does not get an IP address anymore, because the TinyUSB device stop transmitting anything.

If I plug the debugger at this point, I notice that both `xmit_free_ntb` are null, but `xmit_tinyusb_ntb` is non null and `xmit_ready_ntb[0]` is non null (see image below), therefore `tud_network_can_xmit()` return false.

<img width="533" height="531" alt="Image" src="https://github.com/user-attachments/assets/d9cbf015-559e-4dd8-9c72-662a276baf00" />

I've tried with and without DMA it makes no difference.

Anyway, I would love to have some help/idea/suggestion at this point, I'm stuck.

### How to reproduce ?

See above.

### Debug Log as txt file (LOG/CFG_TUSB_DEBUG=2)

The following log starts at point 3) above, i.e. when everything is fine:

[tinyusb.log](https://github.com/user-attachments/files/24356930/tinyusb.log)

We can see at the end of the log, after the last `ifconfig eth1 up` that Linux sends `DHCP Discover` packets continuously, they get received by TinyUSB but generates no answer.

### I have checked existing issues, discussion and documentation

[...]

PR #3761, opened 2026-07-08
ncm: retry link-state notification, fix carrier lost on collision

### LINKED ISSUE #3760: NCM: link-state NETWORK_CONNECTION notification can be silently dropped, leaving the host stuck at NO-CARRIER
### Operating System

Linux

### Commit SHA

0.19

### Board

Company Internal Development

### Firmware

Internal development, using CDC-NCM and CDC-ACM.
0.19 based

### Summary
`tud_network_link_state()` delivers the CDC-NCM `NETWORK_CONNECTION` notification in an edge-triggered, fire-once manner. If a previous notification is still in flight when the link state changes, the new notification is silently discarded and never retried, so the host keeps the previously reported carrier state. In practice a link up transition can be lost, leaving the host network interface permanently at NO-CARRIER (all TX dropped, DHCP fails) even though the device believes the link is up.

`src/class/net/ncm_device.c`
Functions: `tud_network_link_state()`, `notification_xmit()`, `netd_xfer_cb()`
Reproduced on current master. 
Environment: tinyUSB <master>, FreeRTOS, Linux (RPi) host

### Problem description
With a multi-threaded application that calls `tud_network_link_state()` from a task other than the one running `tud_task()`, an occasional link up is not seen by the host:

Expected: after `tud_network_link_state(rhport, true)`, the host receives `NETWORK_CONNECTION` with `wValue=1` and brings the netdev carrier up.
Actual: intermittently the host never receives `wValue=1`; the interface stays <NO-CARRIER … state DOWN>. Because the kernel drops all TX while NO-CARRIER, DHCP (and any other TX) fails. The condition is permanent for that link session.
The rate depends on how often the link toggles and whether a USB bus reset / re-enumeration happens around the transition.

### Root-cause analysis

```
void tud_network_link_state(uint8_t rhport, bool is_up) {
  if (ncm_interface.link_is_up == is_up) {
    return;                                     // (A) dedupe
  }
  ncm_interface.link_is_up = is_up;             // (B) desired state committed BEFORE the send
  if (ncm_interface.itf_data_alt != 1) {
    return;
  }
  ncm_interface.notification_xmit_state = NOTIFICATION_SPEED;
  notification_xmit(rhport, false);             // (C) single, edge-triggered attempt
}

static void notification_xmit(uint8_t rhport, bool force_next) {
  if (!force_next && ncm_interface.notification_xmit_is_running) {
    return;                                     // (D) silent drop if a notification is in flight
  }
  ...
  .wValue = ncm_interface.link_is_up ? 1 : 0,   /* Dynamic link state */
  ...
}
```
Three issues combine:

Silent drop on collision (D). If a notification is still in flight (`notification_xmit_is_running == true`) when `tud_network_link_state()` runs, `notification_xmit(false)` returns immediately and the `NETWORK_CONNECTION` for the new link state is never queued.

No retry + dedupe lock-out (A)+(B). `link_is_up` is updated before the (dropped) send, so any subsequent call with the same value short-circuits at the dedupe. There is no pending flag and nothing re-attempts the send, so the lost notification is never re-delivered. The host's last received `NETWORK_CONNECTION` still carries the previous `wValue` (e.g. 0 from a preceding link-down) → permanent NO-CARRIER.

Cross-context data race. The shared notification state (`notification_xmit_state`, `notification_xmit_is_running`, `link_is_up`) is mutated from both `tud_network_link_state()` (caller/task context) and `netd_xfer_cb()` (USB-stack context) with no synchronization. On RTOS ports where those run in different tasks, this race non-deterministically produces the in-flight collision in (1) and can corrupt the state machine so the completion path (`notification_xmit(rhport, true)` in `netd_xfer_cb`) does not self-heal.

When an in-flight notification is present at the transition a preceding link-down `NETWORK_CONNECTION` still completing when a link-up is signalled.
The `SET_INTERFACE(alt=1)` speed-change → network-connection chain (after enumeration / bus reset / re-enumeration) still in flight when a link-state change arrives.

#### Why it is often latent
Applications that call `tud_network_link_state()` from the same context as `tud_task()` (super-loop) and don't toggle the link rapidly almost never have a notification in flight at a transition. Multi-threaded RTOS apps driving link state from a dedicated task — especially with frequent transitions or bus re-enumeration around them — hit it reproducibly.

### Suggested fix
Make link-state delivery level-triggered and race-safe:

Add a `link_notify_pending` flag: set it whenever `link_is_up` changes; clear it only when a `NETWORK_CONNECTION` carrying the current link_is_up is actually submitted.
After every notification-endpoint completion in `netd_xfer_cb()`, if pending and the channel is free, (re)send — so a collided send is retried on the next completion.
Serialize the notification-state mutations onto the USB-stack context (e.g. `usbd_defer_func()` from `tud_network_link_state()`), so a caller on another task cannot race the completion callback.

Net effect: the carrier notification for the current link state is guaranteed to reach the host — immediately if the channel is free, otherwise on the next completion — regardless of caller context or in-flight collisions.

### Conclusions
A latent correctness bug in the upstream NCM driver, exposed by driving `tud_network_link_state()` from a task separate from `tud_task()` — a normal RTOS usage pattern.

The failure is permanent NO-CARRIER for the affected link session (not transient), because the dropped notification is never retried and the dedupe prevents a resend.

We implemented the approach above and validated it over 250 link up/down cycles with zero recurrences (previously reproducible).

Orthogonal to 6d3d33d… ("ncm: add weak callback for initial link state"), which concerns the initial link state, not delivery of subsequent changes.

### How to reproduce ?

A latent correctness bug in the upstream NCM driver, exposed by driving `tud_network_link_state()` from a task separate from `tud_task()`

NA

### Screenshots

NA

morse-cedricvandenbergh:
# ncm: retry link-state notification, fix carrier lost on collision

Closes #3760

## Problem

`tud_network_link_state()` delivers the CDC-NCM `NETWORK_CONNECTION` notification in an **edge-triggered, fire-once** way. If a previous notification is still in flight when the link state changes, the new notification is silently dropped and never retried, so the host keeps the previously reported carrier state. In practice a link *up* can be lost, leaving the host interface permanently at `NO-CARRIER` (kernel drops all TX → DHCP fails) even though the device thinks the link is up.

Current code:

```c
void tud_network_link_state(uint8_t rhport, bool is_up) {
  if (ncm_interface.link_is_up == is_up) return;      // (A) dedup
  ncm_interface.link_is_up = is_up;                   // (B) commit BEFORE send
  if (ncm_interface.itf_data_alt != 1) return;
  ncm_interface.notification_xmit_state = NOTIFICATION_SPEED;
  notification_xmit(rhport, false);                   // (C) single attempt
}

static void notification_xmit(uint8_t rhport, bool force_next) {
  if (!force_next && ncm_interface.notification_xmit_is_running) return;  // (D) silent drop
  ...
}
```

Three issues combine:

1. **Silent drop (D):** if a notification is in flight (`notification_xmit_is_running`), `notification_xmit(false)` returns and the `NETWORK_CONNECTION` for the new state is never queued.
2. **No retry + dedup lock-out (A)+(B):** `link_is_up` is committed before the dropped send, so a later call with the same value short-circuits at (A). Nothing re-attempts it → the host's last `NETWORK_CONNECTION` keeps the old `wValue` (e.g. `0` from a prior link-down) → permanent `NO-CARRIER`.
3. **Cross-context race:** `notification_xmit_state` / `_is_running` / `link_is_up` are mutated from both `tud_network_link_state()` (caller context) and `netd_xfer_cb()` (usbd context) with no serialisation. On RTOS ports where `tud_network_link_state()` runs in a task other than `tud_task()`, the two race — which is what non-deterministically produces the in-flight collision above.

An in-flight notification is present at a transition when, e.g., a preceding link-down notification is still completing, or the `SET_INTERFACE(alt=1)` speed→connection chain (after a bus reset / re-enumeration) is still in flight when a link-state change arrives. Rapid link toggling and/or a bus reset around the transition make this likely.

It is largely latent for single-threaded/super-loop apps that don't toggle the link rapidly, but reproducible for multi-threaded RTOS apps that drive link state from a dedicated task.

## Fix

Make link-state delivery **level-triggered** and race-safe:

- Add `link_notify_pending`: set when `link_is_up` changes, cleared only when a `NETWORK_CONNECTION` carrying the current state is actually submitted.
- Re-fire from the notify-endpoint completion in `netd_xfer_cb()`: if pending and the channel is free, (re)send — so a collided send is retried on the next completion instead of dropped.
- Defer the send onto the usbd task via `usbd_defer_func()`, so a caller on another task cannot race the completion callback.

Net: the `NETWORK_CONNECTION` for the current link state is guaranteed to reach the host — immediately if the channel is free, otherwise on the next completion — regardless of caller context or in-flight collisions.

## Behavioural note for reviewers

On a link *change* this now re-sends only `NETWORK_CONNECTION`, whereas the previous code reset to `NOTIFICATION_SPEED` and re-sent `CONNECTION_SPEED_CHANGE` as well. A link toggle doesn't change the link speed, and `CONNECTION_SPEED_CHANGE` is still sent at `SET_INTERFACE`, so this is intentional (and is the variant that was validated). If strict behaviour-preservation is preferred, `ncm_flush_link_notification()` can reset to `NOTIFICATION_SPEED` instead — a one-line change.

## Testing

- Reproduced the original failure on an RTOS application that drives `tud_network_link_state()` from a separate task with frequent link down/up transitions plus USB re-enumeration around them: intermittent permanent `NO-CARRIER` after a link-up, DHCP failing on the host.
- With this change, validated over **250 link up/down cycles** (two long soak runs) with **zero** recurrences; every link-up delivered `NETWORK_CONNECTION(wValue=1)` and the host acquired DHCP. Instrumentation counters confirmed the previously-dropped case is now retried and delivered.

## Notes

- Affects only `src/class/net/ncm_device.c`.
- The defect is pre-existing (present well before the recent initial-link-state work in `6d3d33d…`); this is not a regression fix for that change.
- Distinct from #3436 / #3438 (TX-buffer accounting and interface deactivate/reactivate) and #3709 (initial link state at enumeration) — this concerns delivery of a subsequent link-state *change*.

### DIFF
--- a/src/class/net/ncm_device.c
+++ b/src/class/net/ncm_device.c
@@ -800,29 +800,54 @@ static void tud_network_recv_renew_r(uint8_t rhport) {
 } // tud_network_recv_renew
 
 /**
- * Set the link state and send notification to host
+ * usbd-task trampoline for tud_network_link_state(), packing rhport and is_up
+ * into a single pointer-sized argument.
+ *
+ * Runs entirely in the usbd task context, so it cannot race the notify
+ * xfer-completion callback over the notification state machine. Re-arming
+ * notification_xmit_state and kicking notification_xmit() (rather than
+ * sending NETWORK_CONNECTION directly) means a state change that collides
+ * with an in-flight notification is picked up by the existing completion
+ * callback instead of being silently dropped - which would otherwise leave
+ * the host stuck at NO-CARRIER after a link-state change.
  */
-void tud_network_link_state(uint8_t rhport, bool is_up) {
-  TU_LOG_DRV("tud_network_link_state(%d, %d)\n", rhport, is_up);
+static void ncm_link_state_task(void *param) {
+  uintptr_t const arg = (uintptr_t) param;
+  uint8_t const rhport = (uint8_t) (arg >> 1);
+  bool const is_up = (arg & 1u) != 0;
 
   if (ncm_interface.link_is_up == is_up) {
-    // No change in link state
-    return;
+    return; // no change in link state
   }
 
   ncm_interface.link_is_up = is_up;
 
-  // Only send notification if we have an active data interface
   if (ncm_interface.itf_data_alt != 1) {
-    TU_LOG_DRV("  link state notification skipped (interface not active)\n");
-    return;
+    TU_LOG_DRV("  link state notification deferred (interface not active)\n");
+    return; // data interface not active yet; SET_INTERFACE(alt=1) will notify
   }
 
-  // Reset notification state to send speed change notification first, then link state notification
+  // A link toggle does not change the link speed, so strictly only the
+  // NETWORK_CONNECTION notification would need (re)sending. Re-running the
+  // speed-then-connection sequence keeps this on the same state machine the
+  // completion callback already drives, at the cost of a redundant speed
+  // notification on every toggle.
   ncm_interface.notification_xmit_state = NOTIFICATION_SPEED;
-
-  // Trigger notification transmission
   notification_xmit(rhport, false);
+}
+
+/**
+ * Set the link state and notify the host.
+ *
+ * Defers onto the usbd task so a caller running in a different task than
+ * tud_task() cannot race the notification state machine against the notify
+ * xfer-completion callback.
+ */
+void tud_network_link_state(uint8_t rhport, bool is_up) {
+  TU_LOG_DRV("tud_network_link_state(%d, %d)\n", rhport, is_up);
+
+  uintptr_t const arg = ((uintptr_t) rhport << 1) | (is_up ? 1u : 0u);
+  usbd_defer_func(ncm_link_state_task, (void *) arg, false);
 }
 
 //-----------------------------------------------------------------------------
--- a/test/fuzz/device/net_ncm/fuzz.c
+++ b/test/fuzz/device/net_ncm/fuzz.c
@@ -47,6 +47,9 @@ bool usbd_open_edpt_pair(uint8_t rhport, uint8_t const *p_desc, uint8_t ep_count
   (void) rhport; (void) p_desc; (void) ep_count; (void) xfer_type; (void) ep_out; (void) ep_in;
   return true;
 }
+void usbd_defer_func(osal_task_func_t func, void *param, bool in_isr) {
+  (void) func; (void) param; (void) in_isr;
+}
 bool tud_control_xfer(uint8_t rhport, tusb_control_request_t const *request, void *buffer, uint16_t len) {
   (void) rhport; (void) request; (void) buffer; (void) len;
   return true;


HiFiPhile:
Hi, I think the issue can be fixed in a easier way by deferring `tud_network_link_state` completely, no need to change `netd_xfer_cb()` or link_notify_pending:
```c
static void ncm_link_state_task(void *param) {
  uintptr_t const arg = (uintptr_t) param;
  uint8_t const rhport = (uint8_t) (arg >> 1);
  bool const is_up = (arg & 1u) != 0;

  if (ncm_interface.link_is_up == is_up) {
    return;
  }

  ncm_interface.link_is_up = is_up;

  if (ncm_interface.itf_data_alt != 1) {
    return;
  }

  // Preserve existing behavior: speed notification, then connection state.
  ncm_interface.notification_xmit_state = NOTIFICATION_SPEED;
  notification_xmit(rhport, false);
}

void tud_network_link_state(uint8_t rhport, bool is_up) {
  TU_LOG_DRV("tud_network_link_state(%d, %d)\n", rhport, is_up);

  uintptr_t const arg = ((uintptr_t) rhport << 1) | (is_up ? 1u : 0u);
  usbd_defer_func(ncm_link_state_task, (void *) arg, false);
}
```
- All notification state is mutated by the USB task.
- If the channel is idle, transmission starts immediately.
- If it is busy, the changed notification_xmit_state remains pending and is handled by the existing completion callback.
- Multiple queued link changes are processed in order; the latest state ultimately reaches the host.
- Existing speed-then-connection behavior is preserved.

morse-cedricvandenbergh:
Thanks,

I have applied your suggested fix and checked the behaviour on our hardware, and that works nicely.
Will update the PR shortly.

morse-cedricvandenbergh:
Sorry for the delay.
Validated the simplified mechanism on our hardware. Finally updated the PR to reflect the simplied (tested) solution

HiFiPhile (approved):
Thank you

PR #3768, opened 2026-07-10
hil: use stlink for stm32f746disco

HiFiPhile:
Reflashed it back to STLink, hope it's more stable.


hathach (approved):
I re-org the hil pool, changing to other pcie card controller to go around the 32 slot limit. But the moschip controlelr (pure ehci) is buggy, and cause lots of hung/stuck device. The renesas xhci controller is much more stable. I guess I need to do the pcie splitter instead.

PR #3772, opened 2026-07-15
fix(dwc2): defer EP0 buffer advance until DMA completion

tore-espressif:
edpt_schedule_packets() advanced `xfer->buffer` when scheduling EP0 IN DMA, and EP0 OUT completion invalidated cache at that already-advanced pointer. Short control OUT transfers (e.g. 7 bytes) then skipped the start of the buffer and broke class drivers.

Track scheduled EP0 packet size in `ep0_xact_bytes[]`. Invalidate and advance the buffer only on transfer complete, then schedule the next EP0 chunk. EP0 IN advances by `ep0_xact_bytes` before rescheduling.

Fixes https://github.com/hathach/tinyusb/pull/3295

### DIFF
--- a/src/portable/synopsys/dwc2/dcd_dwc2.c
+++ b/src/portable/synopsys/dwc2/dcd_dwc2.c
@@ -43,6 +43,7 @@ static xfer_ctl_t xfer_status[DWC2_EP_MAX][2];
 typedef struct {
   // EP0 transfers are limited to 1 packet - larger sizes has to be split
   uint16_t ep0_pending[2];  // Index determines direction as tusb_dir_t type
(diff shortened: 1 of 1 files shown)


hathach:
Yeah I fixed this exact bug while doing usbtest, which I have been trying to get it merge for the last few days.

hathach:
it is detected when doing usbtest suite https://github.com/hathach/tinyusb/pull/3758/changes/59f02a1c4c18d7e43a1bd6aaad4b50e71931c9ff

hathach:
this should be fixed with https://github.com/hathach/tinyusb/pull/3758, please check if current master work for you.

Issue #3296, opened 2025-10-10
Unplug from hub on RP2040 using PIO-USB fails with a broken unplug event


harbaum:
### Operating System

Linux

### Commit SHA

8f2e3ed4418a08cc13aee4527b7bdd1b8bb1bf55, current master

### Board

Pi Pico 1 / RP2040

### Firmware

Custom firmware for FPGA Companion https://github.com/MiSTle-Dev/FPGA-Companion

Running inside FreeRTOS, using PIO-PICO-USB, also latest master 

This is actually extensively used for the MiSTle retro gaming project, where TinyUSB is used on a rp2040 to control keyboard, mouse and joysticks in a FPGA retro gaming setup.

### What happened ?

Sometimes (with debugging enabled) or very often (without debugging), unplug events on a CH334 hub are missed and the stack goes nuts as it still  tries to communicate with the now missing devices.

### How to reproduce ?

Use the FPGA Companion on a MiSTle setup and unplug a device.

### Debug Log as txt file (LOG/CFG_TUSB_DEBUG=2)

The HUB xfer callback for the unplug event fails:

```
[2] Claimed EP 0x81
  Queue EP 81 with 64 bytes ... 
OK
[:5] on EP 81 with 0 bytes: FAILED
  HUB xfer callback
[5] Claimed EP 0x81
  Queue EP 81 with 1 bytes ... 
OK
```

Afterward, the stack goes nuts as it didn't get the unplug event and still tries to communicate with the now missing devices.

If I ignore the test for a good result in https://github.com/hathach/tinyusb/blob/8f2e3ed4418a08cc13aee4527b7bdd1b8bb1bf55/src/host/hub.c#L364 like so

```
  /*  if (result == XFER_RESULT_SUCCESS) */ {
```

Then, the broken message still contains enough information to actually handle the unplug event correctly. The status change byte is especially correct:

```
[:5] on EP 81 with 0 bytes: FAILED
  HUB xfer callback
  Processing failed hub cb, anyways
  Hub Status Change = 0x08
HUB Get Port Status: addr = 5 port = 3
[1:5] Class Request: A3 00 00 00 03 00 04 00 
[:5] on EP 00 with 8 bytes: OK
[:5] on EP 80 with 4 bytes: OK
[1:5] Control data:
  0000:  00 01 01 00                                      |....|
[:5] on EP 00 with 0 bytes: OK
```

Everything works fine, now. This is ok for now, but this doesn't sound like a good solution.

[...]

PR #2994, opened 2025-02-19
Enhance hub driver

hathach:
**Describe the PR**
Improve host stack compatibility with pio-usb which has often out-of-frame usb transaction error.
- Enhance hub driver, handles most of case where transfer result is not success and requeue interrupt endpoint
- usbh retry up to 3 time if control transferred failed
- update bsp for rp2040, allow to added new board that is not part of sdk e.g FruitJam
- added tuh_connected() API
- Temporarily change pico-pio-usb to forked PR at https://github.com/sekigon-gonnoc/Pico-PIO-USB/pull/159 to retry at transaction level with pio-usb

@tannewt @ladyada

Issue #2971, opened 2025-02-03
RP2040 PIO USB Host does not detect device unplug


rppicomidi:
### Operating System

Linux

### Board

Feather RP2040 with USB A Host

### Firmware

https://github.com/rppicomidi/usb_midi_host/tree/main/examples/C-code/usb_midi_host_pio_example

I think the version in GitHub may not have specific support for the Feather board yet, but it will shortly.

### What happened ?

I have connected a powered USB 2.0 hub to the USB A port of the Feather board. If I plug a MIDI device to the hub, it enumerates correctly and starts working. If I unplug the hub from the Feather board, the software correctly detects the MIDI device is unplugged. If I plug the hub back in with the MIDI device still connected, it enumerates again and starts working again. However, if I leave the hub plugged in and unplug the device, the software does not detect the MIDI device was unplugged.

If I build firmware that uses the native RP2040 USB hardware for the USB host, I do not see this issue.

### How to reproduce ?

1. Build the firmware above for the Feather board
2. Connect a USB C cable to the board and the build computer and load the firmware to the Feather board.
3. Disconnect the Feather board from the computer.
4. Connect a UART to USB serial adapter to the Feather board's ground pin and the Feather board's UART RX and TX pins
5. Plug the serial adapter to the PC and launch a terminal program such as minicom.
6. Plug the Feather board's USB C cable to a 5V power source.
7. Observe the `Pico MIDI Host Example` message on the serial terminal
8. Plug a powered hub with power applied to the Feather's USB A port.
9. Plug a MIDI device to the USB hub. Observe the connection message on the serial terminal.
10. Move a control, press a button, etc. on the MIDI device and confirm the serial terminal registers the MIDI events
11. Unplug the powered hub with the MIDI device still plugged in.
12. Observe the disconnect message on the serial terminal.
13. Plug the hub and device back in to the Feather board's USB A port.
14. Observe the connection message on the serial terminal.
15. Move a control, press a button, etc. on the MIDI device and confirm the serial terminal registers the MIDI events
16. Leave the hub connected to the Feather board's USB A port and unplug the MIDI device from the hub. 
17. Observe no disconnect message. Also observe plugging the device back in does not work.

### Debug Log as txt file (LOG/CFG_TUSB_DEBUG=2)

[...]

PR #3776, opened 2026-07-18
hub: recover a stuck status-change interrupt poll (Fruit Jam CH334F USB host)

mikeysklar:
When a device is unplugged from a hub port, the hub reports it on the status-change interrupt endpoint. On some hosts that interrupt transfer starts failing (`XFER_RESULT_FAILED`) repeatedly instead of delivering the change, and `hub_xfer_cb` just re-arms it, so the removal is never seen and `tuh_umount_cb` is never called. The device stays mounted and every transfer to it fails.

After `HUB_INT_FAIL_RECOVER` consecutive failed polls, poll a port status directly (one port per recovery so every port is covered) so the latched change is still found and handled through the normal path. State is per hub.

Reproduced with CircuitPython on an Adafruit Fruit Jam (RP2350, PIO USB host with an onboard CH334F hub). Run the loop below and unplug the device from a hub port a few times. Before this change, after some unplugs it prints the `USBError` line forever; after it, every unplug prints `no device`. Used a macropad as USB client for unplug/replug tests.

```python
import time, usb.core
while True:
    d = list(usb.core.find(find_all=True))
    if not d:
        print("no device")
    else:
        try:
            print("device:", d[0].product)
        except usb.core.USBError:
            print("USBError, device wedged, not removed")
    time.sleep(0.5)
```

### DIFF
--- a/src/host/hub.c
+++ b/src/host/hub.c
@@ -31,6 +31,9 @@ typedef struct {
   // uint16_t wHubCharacteristics;
   bool mtt;
   hub_port_status_response_t port_status;
+
+  uint8_t int_fail_count; // consecutive failed status-change interrupt polls
+  uint8_t recover_port;   // next port to poll when recovering from a stuck interrupt
 } hub_interface_t;
 
 typedef struct {
@@ -359,14 +362,19 @@ enum {
 static void process_new_status(tuh_xfer_t* xfer);
 
 // callback as response of interrupt endpoint polling
+// Recover the status-change interrupt poll if it keeps failing after this many
+// consecutive failures.
+#define HUB_INT_FAIL_RECOVER 8
+
 bool hub_xfer_cb(uint8_t daddr, uint8_t ep_addr, xfer_result_t result, uint32_t xferred_bytes) {
   (void) xferred_bytes;
   (void) ep_addr;
 
   bool processed = false; // true if new status is processed
+  hub_interface_t* p_hub = get_hub_itf(daddr);
 
   if (result == XFER_RESULT_SUCCESS) {
-    hub_interface_t* p_hub = get_hub_itf(daddr);
+    p_hub->int_fail_count = 0;
     hub_epbuf_t *p_epbuf = get_hub_epbuf(daddr);
     const uint8_t status_change = p_epbuf->status_change[0];
     TU_LOG_DRV("  Hub Status Change = 0x%02X\r\n", status_change);
@@ -387,6 +395,18 @@ bool hub_xfer_cb(uint8_t daddr, uint8_t ep_addr, xfer_result_t result, uint32_t 
         }
       }
     }
+  } else if (p_hub->bNbrPorts && ++p_hub->int_fail_count >= HUB_INT_FAIL_RECOVER) {
+    // The status-change interrupt poll keeps failing (seen on some hubs after a
+    // downstream device is unplugged). The interrupt endpoint only signals that
+    // something changed; if it will not deliver, poll a port status directly so a
+    // latched change (e.g. the disconnect) is still found and handled through the
+    // normal path. Advance one port per recovery so every port is eventually swept.
+    p_hub->int_fail_count = 0;
+    if (p_hub->recover_port < 1 || p_hub->recover_port > p_hub->bNbrPorts) {
+      p_hub->recover_port = 1;
+    }
+    uint8_t const port = p_hub->recover_port++;
+    processed = hub_port_get_status(daddr, port, NULL, process_new_status, STATE_CLEAR_CHANGE);
   }
 
   // If new status event is processed: next status pool is queued by usbh.c after handled this request


HiFiPhile:
Interesting finding, ~~looks like CH334 is less reliable than I thought (has some glitch laterly on my HIL rig with CH334X).~~

Does the port revive when a new device is connected ? Linux handles the failure more aggressively, after 10 errors it reset the hub and disconnect/re-enumerate every child.

mikeysklar:
Good question. I tested this today on the Fruit Jam with a MacroPad.

Running a poll loop that logs every state change.

- On stock master the answer is no. Once the interrupt poll gets stuck, plugging in a new device does not revive it. I left it unplugged for 30 seconds, replugged, and also tried several quick unplug/replug cycles. It stayed stuck the whole time and the new device was never seen. Only reboot recovers.

With this PR: 
- 17 unplug/replug cycles on both hub ports USB 1 and USB 2, including slow and sloppy unplugs. Every unplug was detected within about a second, and every replug enumerated and worked normally.

On the Linux nuclear approach: I could add something similar  (~30 lines of code to reset the hub. It is not necessary for this wedged scenario, but maybe there are more severe situations. Every device plugged into the Fruit Jam would unmount and re-enumerate since this is a root-port bus reset.

hathach:
@mikeysklar is the CircuitPython the tip of main, I will try to hook it with usb analyzer to see what is wrong. Mabe pio-usb has an off timing that cause the hub fail to response.

mikeysklar:
@hathach - 

Yes. The stock repro was CircuitPython main at 30d7addb4e (tip of main on Jul 17) with the submodules exactly as main pins them: tinyusb at 5453e…).

mikeysklar:
The Fruit Jam has JP1 which is a two pin header that exposes the upstream :
- USBH_D+ (RP2350 GPIO1)
- USBH_D- (RP2350 GPIO2)

My logic analyzer is slow for this, but the Cynthion with USB breakout would work. I'll see if I can wire something up.

Screenshot 2026-07-20 at 6 57 26 AM

mikeysklar:
Wired up the Cynthion USB sniffer to the Fruit Jams CH334F upstream on the JP1 header. It worked!

The CH334F is innocent. Just a hub doing its job and still responding to interrupt polls. The host (RP2350) does not ACK once wedged. Probably a deadlock will result in a upstream PR to sekigon-gonnoc/Pico-PIO-USB.

usb-sniff-fruit-jam

hathach:
@mikeysklar superb! that is great to hear, rp2350 has annoying bug with errata19 gpio reading, I fixed it using walkaround in pio-usb, maybe there is still a race. I will work on this soon (still wrap up other long pending issue). Cynthion is nice, I wish I grab one from adafruit when it is first available, more tool is always a plus :).

mikeysklar:
The Cynthion is awesome with excellent open source support and up to 12Mbps. I love using Packetry. 

Still ended up having to hookup a logic analyzer. Pico-PIO-USB seems to be struggling iwth final bits that are 5 bit or more.  

IMG_0655

hathach:
doesn't it support USB HS or the HS is clsoed source ?

HiFiPhile:
My rig is actually using TERMINUS FE2.1, don't know why I'm thinking about CH334 🤦

mikeysklar:
@hathach - It actually does HighSpeed (480Mbps), but I've only used up to Full (12Mbps) for Adafruit USB host controller work. The popular ones don't have the HS PHY yet, but it looks like that will be the norm going forward. 

Screenshot 2026-07-21 at 8 45 42 AM

mikeysklar:
What I've learned about the Fruit Jam USB wedge.

- The USB HUB CH334F is still alive and responding
- The RP2350 is not responding and it is not a silicon issue (E9/E19 ruled out)
- Full CircuitPython build remains problematic
- Bare Metal TinyUSB works (no circuitpython)
- Minimal CircuitPython works (disabled PSRAM heap traffic, DVI/HSTX DMA, audio DMA)

Now doing automated Macropad resets on Fruit Jam USB 1 on different CircuitPython builds to determine which setting(s) are the issue.

mikeysklar:
This took 20 CircuitPython build iterations to prove out and will sound crazy. Disabling `CIRCUITPY_VECTORIO` (not compiled in) will prevent the USB Error wedge condition on the Fruit Jam.

Minimal repro for anyone with a Fruit Jam:

1. Flash stock CircuitPython (I tested at 30d7addb4e, headless, no HDMI needed).
2. Plug any USB device into a USB-A hub port. I used a MacroPad.
3. Run this and unplug the device:

```python
import time, usb.core
while True:
    d = list(usb.core.find(find_all=True))
    if not d:
        print("no device")
    else:
        try:
            print("device:", d[0].product)
        except usb.core.USBError:
            print("USBError, wedged")
    time.sleep(0.5)
```

First or second unplug wedges it: USBError forever, replug is never seen, only reboot recovers.

Separate and worthy of its own issue/pr in CircuitPython repo is `CIRCUITPY_USB_AUDIO` causes random board reboots. Both vectorio and usb_audio being disabled led to very stable USB behavior. 

Still looking into if this is flash or cache as no vectorio code even executes in this scenario. 

## CircuitPython Build Log to narrow down the wedge

| Build | Config | Result |
|---|---|---|
| A-C | minimal, +PSRAM, +display | clean |
| D | +audio | reboots |
| D2 | audio -usb_audio | clean |
| E, F | +misc clusters | wedged |
| G, H | +SD/misc, +ulab | clean |
| I, J | +imaging, +bitmap trio | wedged |
| K | +vectorio only | wedged |
| L | stock -vectorio -usb_audio | clean |
| M | stock -vectorio | reboots, no wedge |
| PAD | +12KB inert flash | clean |
| RAM | K, host path in RAM | wedged |
| QSTR | +vectorio qstrs | clean |
| MOD | +module entry | clean |
| STUB | vectorio silhouette | clean |
| MARK | K + GPIO markers + LA | in progress |

HiFiPhile:
For info I'm using USB Packet Viewer as sniffer, it only costs 700CNY with lua class decoder, but needs a Windows host.
image

mikeysklar:
Follow up on the "flash or cache" question. I instrumented core1's receive path with GPIO markers (no code changes that shift timing, single instruction pin writes) and captured the wedge on a logic analyzer.

Screenshot 2026-07-22 at 9 57 15 AM

Every failed poll shows the same signature: core1 detects end of packet, drains all 5 bytes from the PIO FIFO, but the bytes fail the CRC check so the ACK is never sent. 

wedge-trace

The wire side (Cynthion capture between the RP2350 and the CH334F) shows the hub's bytes are perfect. The hub answers every interrupt poll with a valid CRC16 DATA1 carrying 02, the port 1 status change bitmap. The host never sends the ACK, so the hub keeps the change pending and retransmits the identical byte forever, every 12 ms, for the entire capture. Replug is invisible because the status change interrupt can never be delivered.

```
frame    time      packet       payload
11       0.00801   IN
12       0.00801   DATA1        02
(no ACK, host moves on)
26       0.02001   IN
27       0.02001   DATA1        02
(no ACK)
42       0.03201   IN
43       0.03201   DATA1        02
(no ACK)
```

The corruption correlates with what core0 fetches, not where any USB code lives.

I'll chase the last layer on the CircuitPython side and file issues there. For this PR: the failure class predates RP2350 (#2971 was the same wedge on RP2040, #2994 the previous hub.c recovery). The GetPortStatus fallback keeps hubs usable on affected builds since control transfers bypass the broken interrupt delivery. 

From my side this PR is ready to merge.

Related hardening for a neighboring receive path bug found during this work: sekigon-gonnoc/Pico-PIO-USB#210

mikeysklar:
Wanted to close the loop here. The root cause was a one bit framing offset: core1 locks onto the incoming packet one bit late, every byte arrives shifted by one, and the CRC check fails on wire data that was valid all along. Expected 80 4b 02 c1 7e, received 01 97 04 82 fd, same shift every retry. The flash layout only matters because it changes when core1 arms the receiver relative to the reply.

Fixes that came out of this:

- sekigon-gonnoc/Pico-PIO-USB#211 realigns the misframed packet
- sekigon-gonnoc/Pico-PIO-USB#210 bounds the receive loop
- adafruit/circuitpython#11135 pulls the fix into CircuitPython (draft for now)

Thanks @hathach and @HiFiPhile for the questions and sniffer suggestions along the way, they kept pointing this at the right layer. This PR still stands as written since control transfer recovery helps for any interrupt delivery failure.

hathach:
> Wanted to close the loop here. The root cause was a one bit framing offset: core1 locks onto the incoming packet one bit late, every byte arrives shifted by one, and the CRC check fails on wire data that was valid all along. Expected 80 4b 02 c1 7e, received 01 97 04 82 fd, same shift every retry. The flash layout only matters because it changes when core1 arms the receiver relative to the reply.
> 
> Fixes that came out of this:
> 
> * [Recover packets received with a one bit framing offset sekigon-gonnoc/Pico-PIO-USB#211](https://github.com/sekigon-gonnoc/Pico-PIO-USB/pull/211) realigns the misframed packet
> * [Bound the receive loop and never ACK truncated packets sekigon-gonnoc/Pico-PIO-USB#210](https://github.com/sekigon-gonnoc/Pico-PIO-USB/pull/210) bounds the receive loop
> * [ports/raspberrypi: pull Pico-PIO-USB deterministic SYNC realignment (Fruit Jam USB host wedge) adafruit/circuitpython#11135](https://github.com/adafruit/circuitpython/pull/11135) pulls the fix into CircuitPython (draft for now)
> 
> Thanks @hathach and @HiFiPhile for the questions and sniffer suggestions along the way, they kept pointing this at the right layer. This PR still stands as written since control transfer recovery helps for any interrupt delivery failure.

superb, @mikeysklar great work on the finding. I will jump to pio-usb issue/pr next week to see if there is anything I could help with the fix. Each time I go back to pio-usb, I need to relearn usb timing constraint :smile:

mikeysklar:
@hathach - Thank you for offering assistance on the pio-usb side. I'd like to see these two issues ( sekigon-gonnoc/Pico-PIO-USB#210 and sekigon-gonnoc/Pico-PIO-USB#211 ) go through in some form, if not as is.

This issue must feel familiar as you and tannewt have run up against it multiple times.

Timeline in one place since the pieces were spread across two repos over three years:

| Who | Where | When | Finding |
|---|---|---|---|
| tannewt | sekigon-gonnoc/Pico-PIO-USB#97 | Aug 2023 | Captured the shifted bytes (01 a5 for 80 d2), called it a one bit rotation, suspected RX PIO initial state. Saw it on the first capture so it looked like a cold start quirk |
| hathach | sekigon-gonnoc/Pico-PIO-USB#186 | Jul 2025 | Optimized the same inter packet window from the TX side (reduce delay before sending handshake) |
| harbaum / hathach | hathach/tinyusb#3296 | Oct 2025 | Diagnosed pio timing/race, CPU not responding within USB timing specs, after harbaum found it only fails below 156 MHz |
| mikeysklar  | #3776 + sekigon-gonnoc/Pico-PIO-USB#211 | Jul 2026 | Same bug: the receiver arms a few bits late and decodes valid packets shifted, so the CRC rejects good data on every retry. The shift is derivable from SYNC, fix in #211 |

hathach:
@mikeysklar sorry for the wait, I am struggling with my hil pool getting wedged/D-state with usbtest making it not reliable for dev work. Give me a bit of time, I will get to this soon :(

mikeysklar:
@hathach - Thanks for the update. Let me know when you are ready to pursue the PIO Sync fix.

I am sorry to hear about HIL trouble. I handed in a E-Ink guide yesterday so I can resume HIL work. I have the parts and the $50 hub is working well with uhubctl for PPS. I used an existing FreeCAD Skadis rail mount design and made remixes for secure Metro / Feather mounting. I even made some custom PCBs to reduce the wiring for the Pi debug probes. Should be hanging on my wall complete this weekend.

PR #3785, opened 2026-07-23
Cast BMaxPower in TUD_CONFIG_DESCRIPTOR to uint8_t

TenGui:
## Issue
- Encountered this issue while using C++, `(_power_ma)/2` expands as an int due to integer promotion. When this macro is used to initialize a uint8_t descriptor array in C++, this is a narrowing conversion and fails under `-Wc++11-narrowing`, which is often enabled as an error (as it was for my project).
- Previously, any `_power_ma / 2` that was greater than 255 was truncated (mod 256) when used with a uint8_t array. Every example in tinyusb uses a uint8_t array with `TUD_CONFIG_DESCRIPTOR`.
- **bMaxPower is allocated a single byte in the Standard Configuration Descriptor (USB Spec Rev 2.0 Section Table 9-10), so any value larger should not be valid.**
## Fix
- This does not alter the old truncation behavior, only explicitly makes expression type a uint8_t (assuming uint8_t array)
- Even if not using uint8_t array, per the USB spec, `TUD_CONFIG_DESCRIPTOR` should not produce a `bMaxPower` field larger than uint8_t.
- Fixed `TUSB_DESC_CONFIG_POWER_MA` at same time to prevent similar issues

### DIFF
--- a/src/common/tusb_types.h
+++ b/src/common/tusb_types.h
@@ -13,49 +13,43 @@
 #include "tusb_compiler.h"
 
 #ifdef __cplusplus
-extern "C" {
+ extern "C" {
 #endif
 
 //------------- Device DCache declaration -------------//
-#define TUD_EPBUF_DCACHE_SIZE(_size)                                                                             \
-  (CFG_TUD_MEM_DCACHE_ENABLE ? (TU_DIV_CEIL(_size, CFG_TUD_MEM_DCACHE_LINE_SIZE) * CFG_TUD_MEM_DCACHE_LINE_SIZE) \
-                             : (_size))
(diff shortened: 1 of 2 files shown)


peter-szczeszynski (approved):

HiFiPhile:
Thank you, please also add `TU_MIN` macro to limit the value to 255 before casting to uint8_t to fix the truncate issue.

TenGui:
@HiFiPhile I have added TU_MIN. I kept it fully expanded in `TUD_CONFIG_DESCRIPTOR` like the way it was before, however if it would be preferred to use the `TUSB_DESC_CONFIG_POWER_MA` please let me know, now that it is getting long and hard to read.

HiFiPhile (approved):
Perfect !

PR #3789, opened 2026-07-27
test/hil: replace PCI reset with root-port VBUS cycle for D-state recovery

hathach:
## Problem

`usb_recover.sh pci-reset` wrote `/sys/bus/pci/devices/<addr>/reset`, and the skill documented this as an FLR that fails harmlessly with `ENOTTY` on the Renesas uPD720201. **Neither half is true**, and the combination destroyed the ci HIL rig's controller twice (2026-07-23, 2026-07-27) — both times reported as a plain test hang.

No USB controller on either rig supports FLR:

| controller | `reset_method` |
|---|---|
| Renesas uPD720201 (ci `01:00.0`, `03:00.0`; hifiphile `01:00.0`) | `bus` |
| AMD Matisse (ci `02:00.0`) | `pm bus` |
| Intel ICH9 UHCI/EHCI (all 8) | *(none)* |

The sysfs `reset` attribute exists whenever *any* reset method is available, so the old `[ -e .../reset ]` guard passed and the kernel issued a **PCIe secondary bus reset** on a live, driver-bound xHCI. The HC returns to power-on defaults (`USBCMD=0`, `HCHalted`, zeroed CRCR/DCBAA) while xhci keeps driving rings it no longer owns — card permanently halted, every fixture on it lost until the PVE host is power-cycled. The write **succeeds**, so `usbtest.py`'s `returncode != 0` check never fired.

## Change

Drop `pci-reset` entirely rather than gating it on FLR — no controller here has FLR, so gating would leave a permanently-unreachable branch.

Replace it with **`root-cycle <busport>`**, cutting VBUS at the xHCI root port where per-port power is real. It talks only to the root hub, so it never takes the per-device lock the stuck ioctl holds, and it skips the leaf hubs that advertise ganged switching but never actually cut power.

**`-S` is load-bearing.** By default uhubctl drives port power through `/sys/.../usb<bus>-port<n>/disable` (strace: two `O_WRONLY` opens per cycle). The kernel's `disable_store()` takes the *root hub's* lock and synchronously `usb_disconnect()`s the child **before** `usb_hub_set_port_power()`. Against the wedge this exists for, that disconnect blocks on the very lock being rescued: power never drops and uhubctl itself wedges holding the root hub's lock, taking down the whole bus rather than one port. `-S` selects the libusb path, sending the power-off control transfer straight to the root hub.

`usbtest.py` auto-recovery calls `root-cycle` with the busport it already has (so `pci_addr_of_bus()` is gone), prints the helper's stderr, bounds the call at 60 s, assumes unrecovered until proven otherwise (`sudo()` can `sys.exit()` mid-call and would otherwise unwind past the guard into the forbidden `remove_id`/`unbind` cleanup), and checks `wedged_pids()` on **both** outcomes — matched by device node rather than pid, since `run_case()` may wrap testusb in `sudo` and the Popen pid is then the wrapper.

Skill scripts are now repo-only. The hand-copied `/usr/local/sbin` snapshots were not symlinks, never tracked the repo, and had drifted (`usb_recover.sh` Jul 15 on ci / Jul 6 on hifiphile; `usb_dyndbg.sh` differing outright). Deleted from both rigs; the skills name the script path instead.

## Testing

Verified on the ci rig against an empty root port (bus 11 port 4, no downstream fixtures): port status `0507 power/connect` → `0000 off` → `0507 power/connect`, devnum increments every run, exit 0 in ~8 s, zero opens of the `disable` attribute with `-S`, rig left with no D-state processes. `hub-cycle` re-verified the same way after its liveness check changed.

- Guard logic exercised against real sysfs with the destructive write stubbed, so no controller was reset.
- Port derivation checked for `11-3.7`, `11-3`, `13-1.6.2`, `9-4.7` and path-traversal input; a stale busport is now refused before any power action.
- `/proc/<pid>/stat` parsing verified against a live process, a zombie, a missing pid, and a `comm` containing both `)` and a space.
- Six subagent tabletop scenarios against the skill (nonexistent BDF throughout): before, 3/3 fired `pci-reset` believing it harmless; after, 3/3 choose `root-cycle`, one refusing a hand-written `echo 1 > .../reset` under time pressure.
- `pre-commit run --all-files` passes.

**Not verified:** `root-cycle` has not faced a live D-state wedge. The happy path is measured; the behaviour of the libusb power-on pass against a held root-hub lock is not. The first real hang should be watched rather than trusted.

## Known gaps, not addressed here

- No board-lock coordination before an action that bounces every fixture under the root port. `board_lock.py` is `LOCK_EX | LOCK_NB` and cannot wait, so with `USBTEST_PARALLEL=4` there is no clean way to acquire them mid-run; SKILL.md documents this as a deliberate tradeoff.
- `find_device(..., first=True)` picks arbitrarily among same-serial dual-port parts, and that choice now selects which root port loses VBUS.

https://claude.ai/code/session_012SXzYLicAHtdx7ZTzKK75D

## Disposition of earlier bot findings (head was `fd98eb4f`, now `cb12b2210`)

- **Track testusb rather than its sudo wrapper** — accepted, fixed in `0fc74560`. Detection matches D-state holders by device node across `/proc` rather than by the Popen pid, so a wrapper or its child is caught equally. `run_case()`'s `p.kill()` still targets the wrapper; that is pre-existing and left out of scope.
- **Verify that the device actually re-enumerated** — accepted, fixed in `0fc74560`. `root-cycle` records `devnum` before cycling and requires it to change. `hub-cycle` had the identical flaw and got the same fix.
- **Match only the root hub location (`-e`)** — not applied. In uhubctl 2.6.0 `-e/--exact` is documented as "exact location (no USB3 duality handling)", i.e. it governs USB3 pairing rather than prefix matching, and `uhubctl -l 11` empirically selects only the root hub (the leaf hubs are `ganged` and not controllable without `-f`, which is not passed). The rationale first given for declining this was itself wrong — it claimed omitting `-e` cycles the paired USB3 port, but that pairing does not fire on Linux root hubs. The conclusion rests on the empirical result, not that reasoning.

The recovery path has changed materially since that review: uhubctl is now invoked with `-S`, because strace showed the default backend writes the sysfs `disable` attribute, whose `disable_store()` handler disconnects the child before cutting power and would block on exactly the lock being rescued. That is the part most worth fresh scrutiny.

## Disposition of second review round (head `cb12b2210`, fixed in `5e29ebdb2`)

- **devnum reuse can report a false failure** (Codex, and independently Copilot) — accepted. Linux reuses USB addresses once the per-bus map wraps, so a device that really did power-cycle can return on the same number and be reported as unrecovered, escalating to an unnecessary host power cycle. Not theoretical here: a single `root-cycle` on the idle rig moved devnum `59 → 123`, so the 127-entry map wraps within minutes of normal churn. `root-cycle` and `hub-cycle` now accept **either** an observed disappearance or a changed devnum — the disappearance being the direct evidence that the power cut landed, which devnum inequality was only ever a proxy for.
- **Existence check does not verify identity** — accepted. `[ -e ... ]` only proves *something* occupies that sysfs path; after a reboot renumbers the bus a stale busport can name a different device, whose entire subtree would then lose power. The comment claiming stale busports are refused overstated this. It now prints the target's VID:PID, serial and product before acting so a wrong target is visible.
- **`wedged_pids()` blind under restricted procfs** — accepted, though it cannot fire on these rigs (no `hidepid`; root-owned cmdlines are readable). It swallowed `PermissionError` and returned an empty list, which the caller read as an all-clear and followed with the `remove_id`/`unbind` that must never run while a device lock is held. It now returns `(pids, complete)` and treats incomplete visibility as unrecovered, distinguishing an entry that raced away (`ENOENT`, genuinely gone) from one that is hidden (`PermissionError`) — conflating those would make every scan permanently pessimistic.

## Third review round (head `5e29ebdb2`, fixed in `490dee9c8` + `7c2c98aee`)

Codex was right that the previous round's fix did not resolve its own finding. Watching for the sysfs node to disappear cannot work: `uhubctl -a cycle` holds the entire power-off window inside itself, so the poll only begins once power is already back. Measured on the rig, the disappearance was never observed on any run and success came from the devnum fallback — the very check under question. That left a guard that looked like one and contributed nothing.

**Recovery is now proven by the device's sysfs directory inode.** A real disconnect destroys the `usb_device` and its kobject; reconnecting creates a new one and kernfs allocates a fresh id, so the inode changes. It survives the gap, so no observation window is needed, and it is immune to address reuse. If the disconnect is blocked on the wedged device's lock the kobject is never recreated and the inode is unchanged — which is exactly the case that must fail.

Why it cannot produce a false success: `kernfs_id_ino()` exposes the full 64-bit `(id_highbits<<32 | lowbits)` as `st_ino` on 64-bit `ino_t`, so a repeat needs ~2^64 node creations; and `authorized`-toggle, `set_configuration` and suspend/resume all leave the parent device kobject intact, so none of them can move the marker. The trailing slash in `sysfs_gen()` is load-bearing — `/sys/bus/usb/devices/<busport>` is a symlink with its own inode — and is now called out in the code so it does not get tidied away.

The devnum wrap is not hypothetical: two separate cycles on an idle rig moved devices `123 → 113` and `121 → 61`.

Also in this round: `root-cycle` takes an optional expected serial and refuses on mismatch (the previous round only *printed* the identity, which this PR body previously overstated as a fix); `usbtest.py` drives the helper with `Popen` rather than `subprocess.run`, whose timeout path kills then waits unbounded and so never returns against a D-state child; the helper's stdout is no longer discarded; and `uhubctl` is wrapped in `|| die`.

Verified: end to end on an idle empty root port (`gen 7447801 → 7769803`, exit 0, ~8 s); with `uhubctl` stubbed so no power is cut, the generation is unchanged and it correctly fails; a mismatched serial is refused before any power action. `pre-commit run --all-files` passes.

## Fourth review round (head `7c2c98aee`, fixed in `5b1ff800a`)

- **Processes hidden from `/proc` enumeration** (Codex, P1) — accepted, and a real hole in the previous round's fix. That fix assumed an uninspectable process surfaces as an entry whose `cmdline` read raises `PermissionError`. Under `hidepid=2` or systemd's `ProtectProc=invisible` it does not appear in `iterdir()` at all — no entry, no exception — and `testusb` runs under `sudo` whenever the device node is not writable. The scan returned `([], True)` while hiding the very holder it exists to find, clearing `unrecovered_hang` and running the `remove_id`/`unbind` that must never execute while a device lock is held. `pid 1` is always root-owned, so failing to read it means enumeration is restricted; the scan is now marked incomplete in that case while still reporting whatever is visible.
- **Empty serial silently disables the stale-busport guard** (Copilot) — accepted. `find_device()` strips the sysfs value and it can be empty; an empty third argument reads as "no expectation", so the guard did nothing while the surrounding comment claimed it was active. Only passed when non-empty.
- **SKILL.md overclaimed what is read** (Copilot) — accepted. `root-cycle` reads `idVendor`/`idProduct`/`serial`/`product` to report and check the target, and `devnum` for the success message, not only the directory inode. Reworded to say it never *writes* the wedged device's sysfs and takes no device lock, which is the property that matters.
- **Verify target identity before cycling** (Codex, P2) — not applied; third re-raise of a resolved finding. The `expect=${3:-}` guard and its `die` are present and the comment describes the optional contract accurately. The two findings did interact, though: the empty-serial bug above was the path by which this concern could still bite, so fixing that closes the practical gap.

## Fifth review round (head `5b1ff800a`, fixed in `0f9db2a2e`)

Codex reported no issues on this head. Copilot raised two:

- **`subprocess.Popen()` unwrapped** — accepted. It raises `OSError` when the helper is missing or not executable, or sudo is unavailable (verified: `FileNotFoundError`). The safety property already held, since `unrecovered_hang` is set to `True` before the call and the `finally` block therefore still skipped the unsafe cleanup — but the run died with a traceback instead of a message naming what to fix, and the `--json` payload was lost. Now caught, reported, and the battery ends cleanly.
- **`set -e` aborting on a failed `cat`** — declined as stated. A failed command substitution in an *argument to `echo`* does not trigger errexit (verified: exit 0, script continues); only *assignments* do, and the assignments in that block (`serial=`, `gen=`) already carry `|| echo -` fallbacks, as does `sysfs_gen()` internally. `idVendor`/`idProduct` were given the same fallback anyway so an unreadable attribute prints `-` rather than an empty field, but that is cosmetic, not a crash fix.

### DIFF
--- a/.claude/skills/usb-kernel-debug/SKILL.md
+++ b/.claude/skills/usb-kernel-debug/SKILL.md
@@ -12,11 +12,11 @@ sits in the link — the rig PC when it is the host, or a Linux gadget peer
 (dwc2/UDC + gadget modules) when TinyUSB is the host. It cannot see inside
 the TinyUSB MCU — that is the `target-debug` skill.
 
-Run this skill's `scripts/usb_dyndbg.sh` with `sudo` (abbreviated to
-`usb_dyndbg.sh` in the examples below). It flips the dynamic-debug print flag
-for an allowlisted set of USB modules only:
+Run this skill's `scripts/usb_dyndbg.sh` with `sudo`. It flips the dynamic-debug
+print flag for an allowlisted set of USB modules only:
 
 ```bash
+# all examples below abbreviate:  sudo .claude/skills/usb-kernel-debug/scripts/usb_dyndbg.sh
 sudo usb_dyndbg.sh on  usbcore xhci_hcd   # enable +p; pick modules from `lsusb -t` Driver=
 sudo usb_dyndbg.sh status [module]        # list enabled print sites
 sudo usb_dyndbg.sh off usbcore xhci_hcd   # ALWAYS turn off when done — very noisy
--- a/.claude/skills/usb-kernel-recover/SKILL.md
+++ b/.claude/skills/usb-kernel-recover/SKILL.md
@@ -15,8 +15,9 @@ sudo usb_recover.sh authorized <busport>       # deauthorize+reauthorize: re-enu
 sudo usb_recover.sh rebind     <busport>       # usb driver unbind+bind: re-probe
 sudo usb_recover.sh hub-cycle  <busport>       # uhubctl VBUS cycle of the feeding port, walking parent hub
                                                # -> root port until the device re-enumerates
-sudo usb_recover.sh root-cycle <busport>       # uhubctl VBUS cut straight at the ROOT port (real ppps), no
-                                               # leaf walk, no device-lock touch: the D-state cure
+sudo usb_recover.sh root-cycle <busport> [serial]  # uhubctl VBUS cut straight at the ROOT port (real ppps), no
+                                               # leaf walk, no device-lock touch: the D-state cure.
+                                               # [serial] is checked and a mismatch refused.
 sudo usb_recover.sh pci-rebind <pciaddr>       # whole HCD controller unbind+bind, e.g. 0000:02:00.0
 sudo usb_recover.sh pci-bind   <pciaddr> [drv] # re-bind a DRIVERLESS controller (auto-tries xHCI drivers)
 ```
@@ -43,10 +44,14 @@ sudo usb_recover.sh root-cycle <busport>     # e.g. 11-3.7 -> cycles bus 11 root
 
 This drops power to the wedged device, so its in-flight URB fails and the ioctl
 returns. It targets the *root hub* — a different USB device from the wedged one —
-and only ever *reads* an attribute of the wedged device (`devnum`, which takes no
-lock), so it does not join the convoy the way `authorized`/`rebind`/`pci-rebind`
-do. It exits non-zero if the device does not come back; a **zero exit only means
-it re-enumerated**, so still confirm the D-state process actually let go.
+and only ever *reads* the wedged device's sysfs (its directory inode, which takes
+no lock), so it does not join the convoy the way `authorized`/`rebind`/`pci-rebind`
+do. Recovery is proven by that inode changing — a real disconnect destroys the
+kobject and reconnecting creates a new one, whereas a disconnect blocked on the
+device lock leaves it untouched. It exits non-zero if the device does not come
+back; a **zero exit only means it re-enumerated**, so still confirm the D-state
+process actually let go. Pass the expected serial as a third argument and it
+refuses a busport that now names a different device.
 
 It bounces **every fixture under that root port** — on ci that is up to 25
 devices. Hold the affected boards' locks first if you can, but note
--- a/.claude/skills/usb-kernel-recover/SKILL.md
+++ b/.claude/skills/usb-kernel-recover/SKILL.md
@@ -43,12 +43,24 @@ sudo usb_recover.sh root-cycle <busport>     # e.g. 11-3.7 -> cycles bus 11 root
 
 This drops power to the wedged device, so its in-flight URB fails and the ioctl
 returns. It targets the *root hub* — a different USB device from the wedged one —
-and never touches the wedged device's sysfs, so it does not join the convoy the
-way `authorized`/`rebind`/`pci-rebind` do. It bounces every fixture under that
-root port, so hold their board locks first. It exits non-zero if the device does
-not come back; a **zero exit only means it re-enumerated**, so still confirm the
-D-state process actually let go. (Reasoned from the lock structure and the rigs'
-verified root-port ppps; not yet confirmed against a live D-state wedge. If
+and only ever *reads* an attribute of the wedged device (`devnum`, which takes no
+lock), so it does not join the convoy the way `authorized`/`rebind`/`pci-rebind`
+do. It exits non-zero if the device does not come back; a **zero exit only means
+it re-enumerated**, so still confirm the D-state process actually let go.
+
+It bounces **every fixture under that root port** — on ci that is up to 25
+devices. Hold the affected boards' locks first if you can, but note
+`board_lock.py` uses `LOCK_EX | LOCK_NB` and so fails immediately when CI already
+holds them; there is no wait-for-lock. When CI is mid-run you are choosing
+between bouncing its fixtures and leaving the bus wedged for everything. The
+automated path in `usbtest.py` takes no locks at all and accepts that collateral
+deliberately: by the time a D-state wedge exists the convoy will take the bus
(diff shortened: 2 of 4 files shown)


hathach:
@codex review

hathach:
@codex review

hathach:
@codex review

hathach:
@codex review

hathach:
@codex review

PR #3790, opened 2026-07-27
Fix HFP HIL reliability issue

HiFiPhile:
Fixed and pushed on `fix/lpc43-hfp-reliability`.

Root cause: LPC43 flash remained configured for the 96 MHz boot clock while running at 204 MHz, causing instruction-fetch corruption and HardFaults. `hw/bsp/lpc43/family.c` now applies the documented safe flash timing before switching clocks.

Also included:

- IAR peripheral/DMA reset handling.
- Bounded nonblocking MIDI reads.
- SBUSCFG burst mode `7` for LPC18/43 and i.MX RT, in device and host modes.
- ST-Link configuration for the STM32L412 HFP board.
- No unsupported J-Link reset-type workaround.

Validation:

- No-preflash LPC stress: 8/8 usbtest batteries (`240/240` cases; 8,000 case-9 iterations) and 8/8 throughput runs passed without retries.
- Full no-preflash manual HFP run: 56/56 passed with flash/usbtest parallelism `1/1` and retries disabled.
- [Post-revert IAR HFP](https://github.com/hathach/tinyusb/actions/runs/30377216622/job/90335897774): 56/56 passed.
- [Post-revert strict GCC HFP](https://github.com/hathach/tinyusb/actions/runs/30377216622/job/90336836446): 56/56 passed with `--retry 1`.
- LPC43 GCC/CMake, i.MX RT CMake, pre-commit, and static-analysis checks passed.


hathach:
@HiFiPhile got into issue with the hil pool, controller again. Now all usbtest refuse to run. I will try to figure out why, give me a bit of time. I think the recent no pci-reset just let the usbtest process state in D-state, and blocking all further attempts or so, still better than HC dead, but need more way to out or prevent it

hathach (approved):
thank you, I add an generic SBUSCFG to ci hs register, to make it easier to work with.

PR #3791, opened 2026-07-27
bsp, hil: flash WCH boards with the unified OpenOCD fork

hathach:
Flash with one OpenOCD build instead of a vendor fork per family.

[hathach/openocd](https://github.com/hathach/openocd) (branch `tinyusb`) is mainline OpenOCD plus every config this repo's boards need — the RPi RP2350 configs, all the ADI max32/max78 configs, the WCH configs from the MounRiver Studio toolchain, and the WCH `wlinke` adapter driving mainline's `riscv` target. It is a drop-in **superset** of the vendor forks: vendor configs run unchanged (`sdi` transport, `wch_riscv` target type, probe-mediated resets), so a single `openocd` covers rp2040/rp2350, CH32/CH5xx, MAX32 and everything mainline already supported.

Which build `openocd` is stays the user's choice — the fork, a vendor build, or msdk's. CMake honours `-DOPENOCD=...`, and the msdk openocd is still selected automatically when `MAXIM_PATH` is set; make and the HIL scripts use whatever `openocd` is on PATH, like the existing `flash-openocd` target always has.

### Changes

- `family_support.cmake`: document the fork on `family_flash_openocd`; drop `family_flash_openocd_wch` — the WCH families use the generic helper now. `family_flash_openocd_adi` merges its duplicated `MAXIM_PATH` cmake-variable and environment branches and drops the `~/app/openocd_adi` fallback.
- `family_rules.mk`: `flash-openocd-wch` drops the `OPENOCD_WCH`/`OPENOCD_WCH_OPTION` pair and uses plain `openocd` with the family's `OPENOCD_OPTION`, like every other flash target. It keeps its own command shape (no `verify` — WCH code flash is not readable back over the debug bus).
- `ch583/wch-riscv.cfg`: same work area as the ch32v families (`10000` rather than all 32K of CH58x SRAM — it is only openocd scratch space, and the WCH driver programs through the probe). All four family configs are now byte-identical, matching MounRiver's own config.
- `rp2040/family.cmake`: `rp2350-riscv.cfg` needs the raspberrypi/openocd fork — mainline's riscv target does not accept the `-dap`/`-ap-num` that config uses.
- `test/hil`: WCH boards name the flasher's own config in their args (`-f target/wch-riscv.cfg`) like every other openocd board, instead of `hil_test.py` generating a per-probe copy in the working directory. The `hw/bsp/FAMILY/wch-riscv.cfg` copies stay for the cmake/make flash targets, since the vendor forks do not install configs. `max32666fthr` moves from the `openocd_adi` flasher to plain `openocd`.

### Test

On the ci rig, through the unified build:

| Board | Result |
| --- | --- |
| nanoch32v203 (fsdev + usbfs) | full suite, usbtest 30/30 each |
| ch32v103r_r1_1v0 | full suite, usbtest 30/30 |
| ch32v307v_r1_1v0 (usbfs + usbhs) | full suite, usbtest 30/30 each |
| ch582m_evt | full suite, usbtest 30/30 |
| max32666fthr | full suite |
| raspberry_pi_pico / pico_w / pico2 / adafruit_fruit_jam | flash + verify, HIL |
| stm32h743nucleo / stm32g0b1nucleo | HIL |

Both config styles were exercised on every WCH board: the fork's `target/wch-riscv.cfg` (what HIL uses) and the in-tree `hw/bsp/FAMILY/wch-riscv.cfg` (what the cmake/make targets use). `nanoch32v203` was also flashed through the generic cmake target and the make target with the default `openocd`, and `openocd -f target/wch-riscv.cfg` serves gdb on 3333 for the CLion debug config.

### DIFF
--- a/.idea/debugServers/wch_riscv.xml
+++ b/.idea/debugServers/wch_riscv.xml
@@ -4,7 +4,7 @@
       <debugger kind="GDB" isBundled="true" />
       <env />
     </debugger>
-    <gdbserver dir="$CMakeProjectDir$" exe="$USER_HOME$/app/riscv-openocd-wch/src/openocd" args="-f hw/bsp/ch32v20x/wch-riscv.cfg">
+    <gdbserver dir="$CMakeProjectDir$" exe="openocd" args="-f target/wch-riscv.cfg">
       <env />
     </gdbserver>
     <console enabled="true" port="4444" />
--- a/hw/bsp/ch32v10x/family.cmake
+++ b/hw/bsp/ch32v10x/family.cmake
@@ -93,6 +93,6 @@ function(family_configure_example TARGET RTOS)
 
   # Flashing
   family_add_bin_hex(${TARGET})
-  family_flash_openocd_wch(${TARGET})
+  family_flash_openocd(${TARGET})
   #family_flash_uf2(${TARGET} ${UF2_FAMILY_ID})
 endfunction()
--- a/hw/bsp/ch32v10x/family.mk
+++ b/hw/bsp/ch32v10x/family.mk
@@ -49,5 +49,5 @@ INC += \
 
 FREERTOS_PORTABLE_SRC = $(FREERTOS_PORTABLE_PATH)/RISC-V
 
-OPENOCD_WCH_OPTION=-f $(TOP)/$(FAMILY_PATH)/wch-riscv.cfg
+OPENOCD_OPTION=-f $(TOP)/$(FAMILY_PATH)/wch-riscv.cfg
 flash: flash-openocd-wch
--- a/hw/bsp/ch32v20x/family.cmake
+++ b/hw/bsp/ch32v20x/family.cmake
@@ -125,7 +125,7 @@ function(family_configure_example TARGET RTOS)
 
   # Flashing
   family_add_bin_hex(${TARGET})
-  family_flash_openocd_wch(${TARGET})
+  family_flash_openocd(${TARGET})
   family_flash_wlink_rs(${TARGET})
   #family_flash_uf2(${TARGET} ${UF2_FAMILY_ID})
 endfunction()
--- a/hw/bsp/ch32v20x/family.mk
+++ b/hw/bsp/ch32v20x/family.mk
@@ -63,6 +63,6 @@ INC += \
 
 FREERTOS_PORTABLE_SRC = $(FREERTOS_PORTABLE_PATH)/RISC-V
 
-OPENOCD_WCH_OPTION=-f $(TOP)/$(FAMILY_PATH)/wch-riscv.cfg
+OPENOCD_OPTION=-f $(TOP)/$(FAMILY_PATH)/wch-riscv.cfg
 flash: flash-wlink-rs
 #flash: flash-openocd-wch
--- a/hw/bsp/ch32v30x/family.cmake
+++ b/hw/bsp/ch32v30x/family.cmake
@@ -115,6 +115,6 @@ function(family_configure_example TARGET RTOS)
 
   # Flashing
   family_add_bin_hex(${TARGET})
-  family_flash_openocd_wch(${TARGET})
+  family_flash_openocd(${TARGET})
   family_flash_wlink_rs(${TARGET})
 endfunction()
--- a/hw/bsp/ch32v30x/family.mk
+++ b/hw/bsp/ch32v30x/family.mk
@@ -62,5 +62,5 @@ LD_FILE ?= $(FAMILY_PATH)/linker/ch32v30x.ld
 # For freeRTOS port source
 FREERTOS_PORTABLE_SRC = $(FREERTOS_PORTABLE_PATH)/RISC-V
 
-OPENOCD_WCH_OPTION=-f $(TOP)/$(FAMILY_PATH)/wch-riscv.cfg
+OPENOCD_OPTION=-f $(TOP)/$(FAMILY_PATH)/wch-riscv.cfg
 flash: flash-openocd-wch
--- a/hw/bsp/ch583/family.cmake
+++ b/hw/bsp/ch583/family.cmake
@@ -104,5 +104,5 @@ function(family_configure_example TARGET RTOS)
 
   # Flashing
   family_add_bin_hex(${TARGET})
-  family_flash_openocd_wch(${TARGET})
+  family_flash_openocd(${TARGET})
 endfunction()
--- a/hw/bsp/ch583/family.mk
+++ b/hw/bsp/ch583/family.mk
@@ -52,7 +52,7 @@ INC += \
 
 LD_FILE ?= $(FAMILY_PATH)/linker/ch582.ld
 
-OPENOCD_WCH_OPTION=-f $(TOP)/$(FAMILY_PATH)/wch-riscv.cfg
+OPENOCD_OPTION=-f $(TOP)/$(FAMILY_PATH)/wch-riscv.cfg
 flash: flash-openocd-wch
 
 # For freeRTOS port source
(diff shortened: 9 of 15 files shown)


hathach:
Both Codex P1s were real — thanks. Fixed in `f92b3661d`:

**1. Remote HIL bundle** — `hil_ci.sh` now ships the board openocd configs, and the family lookup that needed `hw/bsp/*/boards/` is gone entirely: WCH boards name their config in the flasher args like every other openocd board (`-f hw/bsp/ch32v20x/wch-riscv.cfg`), resolved with `-s <repo root>` so the cwd doesn't matter.

**2. Hard-coded `openocd_wch`** — the binary now defaults to plain `openocd` and is overridable with `OPENOCD_WCH`, mirroring the make variable of the same name, so a clean unified install works out of the box and a vendor build stays selectable.

Verified on the ci rig after the change: nanoch32v203 (both variants) and ch582m_evt pass with the default `openocd`; ch582m_evt also passes with `OPENOCD_WCH=openocd_wch` and with the script invoked from `/tmp` (cwd-independence).

hathach:
@codex review

hathach:
Round 2 — both findings were real, fixed in `df176e1ab`:

**P1 (rsync parent dir)** — confirmed exactly as described; reproduced locally:
```
rsync: [Receiver] mkdir "…/dst/hw/bsp" failed: No such file or directory (2)
```
Fixed by creating `hw/bsp` in the remote setup step (`mkdir -p -- "$1/test/hil" "$1/examples" "$1/hw/bsp"`) rather than `--mkpath`, which would add an rsync 3.2.3+ requirement on both ends.

**P2 (OPENOCD_WCH not forwarded)** — correct, the ssh wrapper passed only REMOTE_DIR and the CLI args. It is now passed as a positional and exported before python starts.

Both verified by simulating the remote steps locally with the real commands: setup + rsync now lands all four `hw/bsp/*/wch-riscv.cfg`, and the wrapper exports `OPENOCD_WCH` when set while leaving it unset otherwise, with the test args surviving the shifts.

@codex review

hathach:
Ran a deeper local review pass as well; fixed everything that verified, in `794e4c99c`:

**Critical — my previous fix was broken (thanks for catching the class of bug).** Passing `"${OPENOCD_WCH-}"` as a positional was wrong: ssh joins argv into one string, so the empty value collapsed and shifted every following argument. Reproduced over real ssh — the remote saw `$#=4` instead of 5, exported `OPENOCD_WCH=-b`, and hil_test.py got mangled args. It now goes through an `env` prefix that is omitted entirely when unset; both cases verified over real ssh.

Also fixed:
- `flasher["args"]` is optional in the config schema (stm32u083nucleo, hfp.json's stm32f746disco have no `args`) — now `.get("args", "")`, so a KeyError can't escape the flash retry loop.
- `OPENOCD_WCH=` set-but-empty selected `''` — now `or 'openocd'`.
- `-s {TINYUSB_ROOT}` was unquoted under `shell=True` — now `shlex.quote`.
- `reset_openocd_wch` used `program reset exit`, where openocd binds `reset` as the image *filename* — now `init; reset run; exit`, matching `reset_openocd`.
- `family_flash_openocd_adi`: my dedup applied `TO_CMAKE_PATH` to the `-DMAXIM_PATH` form too, which splits a Windows-style value into a CMake list; it is now applied only to the environment value. `OPENOCD` is also guarded with `NOT DEFINED` so an explicit `-DOPENOCD=` wins, as the new comment advertises. Verified: `-DMAXIM_PATH=C:/MaximSDK` → `C:/MaximSDK/Tools/OpenOCD/openocd`, `-DOPENOCD=` → that binary, neither set → `openocd`.
- **rp2350-riscv**: the claim in my rp2040 comment was wrong. The config attaches cores via `-dap`/`-ap-num`, which only raspberrypi/openocd's riscv target accepts — mainline's rejects it (`Unknown param: -dap`). The comment now says the RPi fork is required, and the fork annotates that file and its README instead of implying it works.

Two findings I checked and did **not** act on:
- *"the `*_port disabled` options are no-ops because the command list ends in shutdown"* — not true for these invocations: `gdb_target_add_all()` runs inside `handle_init_command`, and these commands call `init` explicitly. Matches the evidence that prompted the change (135 `Address already in use` errors before adding them, zero after).
- *"`flash_openocd_adi` is now dead code"* — `stflash`, `wlink_rs` and `uniflash` are likewise supported but unused by in-repo configs; out-of-tree configs (e.g. the htpc `local.json`) can still select them.

Re-verified on the rig after all of the above: nanoch32v203 (both variants) and ch582m_evt flash and pass.

hathach:
Updated to `58fde0de6`. Since the last round the HIL side got simpler rather than more clever:

- The unified fork now ships the WCH configs itself (taken from the MounRiver Studio toolchain: `wch-riscv.cfg`, `wch-arm.cfg`, `wch-dual-core.cfg`, plus the `ch32f1x.cfg` that `wch-arm` sources). So the boards just pass `-f target/wch-riscv.cfg` like every other openocd board, and **`hil_ci.sh` is back to being byte-identical to master** — the config shipping, the remote `hw/bsp` mkdir and the `OPENOCD_WCH` ssh forwarding all became unnecessary, which retires both P1s from the earlier rounds rather than patching them.
- The `hw/bsp/FAMILY/wch-riscv.cfg` copies stay for the cmake/make flash targets, since the vendor forks don't install configs.
- `OPENOCD_WCH` is gone entirely (script and make): the binary is plain `openocd`, and which build that is stays the user's choice via PATH. Re: the P2 asking to restore a make override — the neighbouring `flash-openocd` rule hardcodes `openocd` the same way, so this is consistent rather than a regression; the comment no longer implies a selectable binary.
- `ch583/wch-riscv.cfg` now uses the same work area as the ch32v families (10000 rather than all 32K of CH58x SRAM — it is only openocd scratch space, and the WCH driver programs through the probe). All four family configs are byte-identical, and that value matches MounRiver's own config.
- Fixed a regression of my own in `family_flash_openocd_adi`: the msdk scripts dir was gated on `OPENOCD` being unset, so `-DOPENOCD=...` silently dropped `-s .../Tools/OpenOCD/scripts`. Verified override+msdk, msdk-only and default.

Re-verified on the ci rig with the fork-installed config: all four WCH boards flash and pass (nanoch32v203 fsdev+usbfs, ch32v103r, ch32v307v, ch582m_evt — ch582m full suite with usbtest 30/30), max32666fthr passes through the generic openocd flasher, and `openocd -f target/wch-riscv.cfg` serves gdb on 3333 for the CLion debug config.

hathach:
@codex review

hathach:
Both addressed in `cac210a73`.

**Codex (P2, Make executable override)** — taken, in a form that keeps the maintainer's "just `openocd`" decision: rather than resurrecting the WCH-specific `OPENOCD_WCH`, `family_rules.mk` now has a single `OPENOCD ?= openocd` used by **both** `flash-openocd` and `flash-openocd-wch`. Default behaviour is unchanged (plain `openocd` from PATH), an override exists for a vendor build outside PATH, and the variable name now matches the cmake side's `-DOPENOCD=`. Verified: `make BOARD=nanoch32v203 flash-openocd-wch` expands to `openocd -f .../wch-riscv.cfg -c init -c halt -c "flash write_image ..." -c reset -c exit` by default and to `/opt/vendor/openocd ...` with `OPENOCD=/opt/vendor/openocd`; the default command is the one already flash-verified on the rig.

**Copilot (HIL args vs description)** — the description was stale, not the code. It has been rewritten: HIL deliberately uses the flasher's own `target/wch-riscv.cfg` (the fork ships the WCH configs, taken from the MounRiver toolchain), while the in-tree `hw/bsp/FAMILY/wch-riscv.cfg` copies remain for the cmake/make flash targets, since the vendor forks do not install configs. That split is what removed the need for `hil_ci.sh` to ship configs to the remote host — it is now byte-identical to master again.

PR #3799, opened 2026-07-29
device/dfu: fix transfer buffer overwrite issue

### LINKED ISSUE #3798: DFU zeroes out first couple of bytes in block
### Operating System

Others

### Commit SHA

e88fc44 (Current Main)

### Board

STM32 Nucleo-64 Board (Chip STM32C071)

### Firmware

Custom Firmware:

I am running threadx on the nucleo-64 board and put TinyUSB in a threadx task (I am running TinyUSB with the NO OS option) running as a usb device. The task loops and calls tud_task_ext( 1, false ); The Interrupt handler calls tusb_int_handler( 0, true ); I have registered the tud_dfu_download_cb to process data. I have the following descriptors defined:

// Config number, interface count, string index, total length, attribute, power in mA

 // Interface number, string index, protocol, report descriptor length, endpoint in address, size, polling interval
TUD_HID_DESCRIPTOR( ITF_NUM_HID, 0, HID_ITF_PROTOCOL_NONE, sizeof( desc_hid_report ), EPNUM_HID, CFG_TUD_HID_EP_BUFSIZE, 5 ),

// Interface number, string index, notification endpoint, and size, data out endpoint, data in endpoint, and size
TUD_CDC_DESCRIPTOR(ITF_NUM_CDC_0, 4, EPNUM_CDC_0_NOTIF, 16, EPNUM_CDC_0_OUT, EPNUM_CDC_0_IN, 64),

// Interface number, Alternate count, starting string index, attributes, detach timeout, transfer size
TUD_DFU_DESCRIPTOR(ITF_NUM_DFU_RT, ALT_COUNT, 5, FUNC_ATTRS, 1000, CFG_TUD_DFU_XFER_BUFSIZE),

### What happened ?

When TinyUSB calls tud_dfu_download_cb, it seems to be zeroing out the first couple of bytes in the data. I have rolled back all the way to November of last year (hash 3af1bec, Novemeber 19, 2025) and the bug is no longer present (all data received is intact). I am receiving CDC data just fine (tud_cdc_n_read, tud_cdc_n_write, and tud_cdc_n_write_flush are working as expected allowing me to implement a CLI). It might have been introduced in the fsdev refactor.

I know the c071 variant has its own header file, maybe there is some behavior specific to this variant causing the issue?

### How to reproduce ?

1. Run TinyUSB with the following configuration (No OS, tud_task_ext called repeatedly from a task, tusb_int_handler called from the interrupt handler, register tud_dfu_download_cb, setup a DFU descriptor with an alternate, set endpoint0 and dfu xfer )
2. Perform a dfu xfer from a connected linux host to the device to the specified alternate (I was using dfu-util).

N/A Due to how the code is setup, I cannot get the TinyUSB logs to output over uart (there is some C/C++ interleaving going on).

### Screenshots

Screenshot of a hexdump of the first two invocations of the data of tud_dfu_download_cb:

Image

For block = 1, the first 4 bytes should have been 0x4d 0x46 0x57 0x00, but instead, tinyusb is supplying 0x00 0x00 0x00 0x00

HiFiPhile:
Always use internal buffer `_transfer_buf`, using shard control transfer buffer can lead data overwritten issue.

fix #3798

### DIFF
--- a/src/class/dfu/dfu_device.c
+++ b/src/class/dfu/dfu_device.c
@@ -42,23 +42,13 @@ typedef struct {
 
 static dfu_state_ctx_t _dfu_ctx;
 
-#if CFG_TUD_DFU_XFER_BUFSIZE > CFG_TUD_ENDPOINT0_BUFSIZE
-TU_ATTR_ALIGNED(4) uint8_t _transfer_buf[CFG_TUD_DFU_XFER_BUFSIZE];
-#endif
+// Download data must remain valid across the following GETSTATUS control transfer
+TU_ATTR_ALIGNED(4) static uint8_t _transfer_buf[CFG_TUD_DFU_XFER_BUFSIZE];
 
 static void reset_state(void) {
   _dfu_ctx.state = DFU_IDLE;
   _dfu_ctx.status = DFU_STATUS_OK;
   _dfu_ctx.flashing_in_progress = false;
-}
-
-static inline uint8_t* get_xfer_buffer(void) {
-  // Use EP0 buffer if it is large enough, otherwise use dedicated buffer
-  #if CFG_TUD_DFU_XFER_BUFSIZE > CFG_TUD_ENDPOINT0_BUFSIZE
-  return _transfer_buf;
-  #else
-  return usbd_get_ctrl_buf();
-  #endif
 }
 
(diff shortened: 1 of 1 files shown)


AnthonyV-modretro:
@hathach Wondering if this will make it into the next tinyusb release? We are stuck using an old version of tinyusb due to dfu not functioning correctly

me-no-dev (approved):

PR #3802, opened 2026-07-30
clamp committed video payload size to streaming ep buffer

dxbjavid:
The UVC device driver takes dwMaxPayloadTransferSize straight from the probe/commit payload the host sends, and that value ends up sizing each bulk payload it builds. On the fully negotiated path _update_streaming_parameters caps it to the streaming endpoint buffer, but when the host commits before everything is settled (a zero bFormatIndex on a multi-format device, a zero bFrameIndex where the format has several frames, or a frame interval it never pins down) the function returns early and the raw host value is left untouched. handle_video_stm_cs_req then stores it into max_payload_transfer_size without a bound, and for a bulk stream nothing else limits it since only the isochronous open path checks the size against the endpoint packet size. When the application later hands a frame to _prepare_in_payload the memcpy uses that size and can write past the fixed CFG_TUD_VIDEO_STREAMING_EP_BUFSIZE endpoint buffer. I came across it while comparing the committed path against the two get paths, which both already clamp to the same limit. The fix applies that clamp at commit so a partial or hostile commit cannot push the payload beyond the buffer.


HiFiPhile (approved):
Thank you

PR #3813, opened 2026-08-09
device: reset endpoint toggles on SET_INTERFACE fallback

### LINKED ISSUE #2255: resetting DATA PID toggling to DATA0 after SET_INTERFACE
### Operating System

Linux

### Board

raspberry pico (rp2040)

### Firmware

any tinyusb firmware as long as it uses the usbd.c default handling of TUSB_REQ_SET_INTERFACE and doesn't override that in a driver specific control endpoint handler call-back.

### What happened ?

It is my understanding that after any configuration event (to which SET INTERFACE belongs), the endpoints belonging to that interface re-start from DATA0.

See USB Spec 9.1.1.5: _configuing a device or changing an alternate setting causes all of the status and configuration values associated with endpoints in the affected interfaecs to be set to their default values. This includes esetting the data toggle of any endpoint using data toggles to the value of DATA0_

However, I don't see any code in tinyusb for resetting data toggling to DATA0 after a SET_INTERFACE request.  tinyusb simply acknowledges any SET_INTERFACE in its generic code, but data toggles remain unaffected

If I'm creating such a situation (See below how to reproduce), I am seeing BULK out URBs never completing anymore.

If I add USB_INTS_ERROR_DATA_SEQ_BITS to the INTE (interrupts enabled) mask in dcd_rp2040.c I also notice that in this scenario, the related interrupt is raised, and the SIE_STATUS register has the USB_SIE_STATUS_DATA_SEQ_ERROR_BITS set.

I don't really think this is RP2040 specific, but in general I cannot find any code in tinyusb to set the DATA toggle back to DATA0 in case of configuration events. 

### How to reproduce ?

* send an uneven number of BULK OUT URBs to a device
* send a SET_INTERFACE request to the interface
* observe that any further BULK OUT requests are not succesful anymore
* perform the same without the SET_INTERFACE, and it succeeds

none

### Screenshots

_No response_

jaysonndungu:
Fixes #2255

### Summary

* reset non-isochronous endpoints for the requested interface when the generic SET_INTERFACE fallback accepts alt 0
* keep class-specific altsetting handlers in control of classes that implement SET_INTERFACE themselves

### Testing

* `git diff --check`

### DIFF
--- a/src/device/usbd.c
+++ b/src/device/usbd.c
@@ -1060,6 +1060,37 @@ static bool process_get_status(uint8_t rhport, tusb_control_request_t const * re
   return tud_control_xfer(rhport, request, &status, 2);
 }
 
+static bool process_reset_interface_endpoints(uint8_t rhport, uint8_t itf) {
+  TU_VERIFY(_usbd_dev.cfg_num > 0);
+
+  const tusb_desc_configuration_t *desc_cfg =
+    (const tusb_desc_configuration_t *)tud_descriptor_configuration_cb(_usbd_dev.cfg_num - 1);
+  TU_ASSERT(desc_cfg != NULL && desc_cfg->bDescriptorType == TUSB_DESC_CONFIGURATION);
+
+  const uint8_t *p_desc = ((const uint8_t *)desc_cfg) + sizeof(tusb_desc_configuration_t);
+  const uint8_t *desc_end = ((const uint8_t *)desc_cfg) + tu_le16toh(desc_cfg->wTotalLength);
+  bool in_interface = false;
+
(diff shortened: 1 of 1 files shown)


HiFiPhile:
Hi, thank you for the PR, have you actually met the issue ?

jaysonndungu:
Thanks for the review. I have not reproduced the issue on hardware, and the current approach can abort class-owned transfers without coordinating re-arm/restart behavior. I'm closing this PR rather than leaving a risky fix open; a proper fix should use a non-aborting toggle reset path or explicit class coordination with regression coverage.

PR #3823, opened 2026-08-14
midi2 device: complete the UMP stream discovery responder

sauloverissimo:
Small, contained change to `midi2_device.{c,h}`, continuing the #3738 responder. Testing against Linux, Windows MIDI Services and the MIDI 2.0 Workbench surfaced three gaps:

- Device Identity (M2-104 §7.1.3) is never answered. Hosts show an all-zero manufacturer/family/model.
- FB Info always sends MIDI-CI version and SysEx8 streams as 0 (§7.1.8), which on the wire means "not supported". Windows shows "8-bit SysEx not supported" for a device that emits SysEx8.
- The Endpoint Discovery filter bitmap is ignored (§7.1.1), and the full reply burst (~144 bytes) overflows the default 64-byte FIFO: the Workbench received a truncated Product Instance Id.

Changes: an optional identity callback (weak, off by default), two config macros for the FB Info fields (default 0, current behavior), the filter honored, and pending replies drained on TX complete instead of dropped. No new buffer, same flow mechanics the driver already uses.

With this a TinyUSB device shows up fully populated in Windows MIDI Services, which is on its way to shipping in-box in Windows: identity, MIDI-CI and SysEx8 correct in the Settings app and console. Validated on hardware (RP2040 full-speed) against all three hosts; example updated with the identity callback.

@HifiPhile the only behavior change is honoring the filter: a host sending a zero filter no longer gets the three unsolicited replies. Every host I tested sets the bits it wants.

### DIFF
--- a/examples/device/midi2_device/src/main.c
+++ b/examples/device/midi2_device/src/main.c
@@ -565,6 +565,19 @@ const char* tud_midi2_fb_name_cb(uint8_t itf, uint8_t fb_idx) {
   return (fb_idx == 0) ? "Synth Out" : "Keys In";
 }
 
+// Sent when the host asks for a Device Identity Notification (Endpoint
+// Discovery 'd' filter bit). Same four fields as the MIDI 1.0 Device Inquiry
+// reply; 0x7D is the prototyping SysEx ID, placed in the first of the three
+// manufacturer bytes.
+bool tud_midi2_device_identity_cb(uint8_t itf, tud_midi2_device_identity_t* identity) {
+  (void)itf;
+  identity->manufacturer = 0x7D0000;
+  identity->family       = 0x0001;
+  identity->model        = 0x0001;
+  identity->sw_revision  = 0x00010000;
+  return true;
+}
+
 //--------------------------------------------------------------------+
 // Initial Setup - Program Change, CC, Per-Note Management
 //--------------------------------------------------------------------+
--- a/src/class/midi/midi2_device.c
+++ b/src/class/midi/midi2_device.c
@@ -36,6 +36,9 @@ TU_ATTR_WEAK const char* tud_midi2_fb_name_cb(uint8_t itf, uint8_t fb_idx) {
 TU_ATTR_WEAK tud_midi2_stream_result_t tud_midi2_stream_msg_cb(uint8_t itf, const uint32_t* ump_words) {
   (void) itf; (void) ump_words; return MIDI2_STREAM_PASS;
 }
+TU_ATTR_WEAK bool tud_midi2_device_identity_cb(uint8_t itf, tud_midi2_device_identity_t* identity) {
+  (void) itf; (void) identity; return false;
+}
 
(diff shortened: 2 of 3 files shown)


sauloverissimo:
@HifiPhile Codex caught two cases here, and the first one goes back to the raw Tx FIFO we did in #3571: when the Endpoint Name or a Function Block name encodes to more than the FIFO depth, `_nego_send_stream_text` kept emitting and the guard in `_nego_send_ump` dropped the tail packets, leaving the host with a Start/Continue sequence and no End.

Fixed by keeping packet level progress: the text resumes where it stopped on TX complete instead of losing the tail, so the name arrives whole and the sequence always closes. The second one was mine, a new FB Discovery overwrote the pending one; they merge now.

Tested with the device configured with a 90 character Endpoint Name, 7 packets against a 64 byte FIFO: before, 5 packets arrived with no End; now all 7 arrive and close properly. Also ran a 12 case adversarial battery (full filter, discovery burst, zero filter, repeats) and a 3 minute stress run: 45,493 UMPs, zero loss, zero duplicates, zero reordering, with all 22 gestures of the reference sequence recognized.

Confirmed on Linux (snd_ump), Windows MIDI Services and the MIDI 2.0 Workbench, the latter with 0 errors and 0 warnings in the log.

HiFiPhile:
Thank you, Codex detected 2 issues which are unlikely to happen but valid, let me know if the changes are good to you.

Overlapping discovery requests corrupt text responses. A single nego_text_offset is shared by endpoint name, product ID, and every function-block name. If an endpoint discovery arrives while a function-block name is partially queued, endpoint processing reuses that offset, producing a Continue/End packet without its Start or skipping initial characters. See [_nego_send_pending()](https://github.com/hathach/tinyusb/blob/0504faf29825deb130bfeb88dba46bb1c1bdec75/src/class/midi/midi2_device.c#L451-L499). The offset needs to be tied to the specific pending response or processing must not switch responses mid-text.

Alternate-setting changes retain stale discovery state. SET_INTERFACE clears RX/TX FIFOs but leaves all nego_pending_* fields and nego_text_offset intact. Switching away during a partial response and later returning to Alt 1 can resume an old response from the middle. See [midi2d_control_xfer_cb()](https://github.com/hathach/tinyusb/blob/0504faf29825deb130bfeb88dba46bb1c1bdec75/src/class/midi/midi2_device.c#L849-L868). Clear negotiation reply state whenever the alternate setting changes.

sauloverissimo:
Looks good. Tested on hardware: RP2350 running the midi2_device example, with the endpoint name and both block names instrumented to 82-84 characters so the text overflows the TX FIFO. Also ran the adversarial battery and repeated overlap runs, all clean.

Good to merge from my side. Much appreciated!

image

HiFiPhile (approved):

sauloverissimo:
❤️❤️🫶 @HiFiPhile

Psychlist1972:
Nice :)

Pete
Microsoft

PR #3828, opened 2026-08-17
bound endpoint number in tu_bind_driver_to_ep_itf

dxbjavid:
During host enumeration the driver-binding walk writes the driver id into ep2drv indexed by the endpoint number taken from each endpoint descriptor, but that index is never bounded, even though the interface number on the sibling branch a few lines down is checked against itf_max. dev->ep2drv has CFG_TUH_ENDPOINT_MAX rows and that value is commonly reduced below 16 (the bare_api host example ships it at 8), so a rogue device whose bEndpointAddress low nibble is at or above the limit makes the write land past the array in adjacent device state during enumeration. This adds an ep_max argument and the matching bound so a malformed endpoint is rejected the same way an out-of-range interface already is, with the host and device callers passing their respective endpoint maximums.


HiFiPhile (approved):
Thank you !

dxbjavid:
@HiFiPhile Thank you, much appreciated

PR #3832, opened 2026-08-19
hil: convoy-safe openocd_seq recovery flasher for seven J-Link boards

hathach:
## Summary

Follow-up to #3803. That PR landed the wedge-containment machinery (`recover_flasher()`, the `convoy_safe()` gate, `hil_test` substituting the recovery flasher into `--recover-board`) but left it inert: no roster entry defined `flasher_recover`, and no flasher backend actually worked over openocd's jlink transport. This PR completes it:

- **`openocd_seq` flasher backend** (`test/hil/hil_flash.py`): explicit `init` / `reset halt` / `flash write_image erase` / `verify_image` / `reset run` / `shutdown` sequence. The existing `flash_openocd`'s `program <fw> verify reset exit` form fails over the jlink transport (`Examination failed` → `auto_probe failed`, measured on stm32f4x and stm32f0x); the explicit sequence succeeded on all seven candidate boards.
- **Seven `flasher_recover` roster entries** (`test/hil/tinyusb.json`): stm32f407disco, stm32f072disco, stm32f723disco, stm32l476disco, feather_nrf52840_express, metro_m4_express, frdm_k64f — each the same J-Link probe as its primary flasher, driven through openocd's jlink driver instead of the SEGGER tools.
- **Lint/tests** (`test/hil/test/test_ci_select.py`): roster lint for the recovery entries (name, jlink args, required `adapter speed`, uid must match the primary probe, convoy-safe), dispatch/suffix lints extended to `flasher_recover`, command-form regression test pinning "no `program` over jlink".
- The false "no vid_pid pin → probe discovery will open every usbfs node" warning is silenced for `interface/jlink.cfg` args: libjaylink's discovery gates on `idVendor == 0x1366` before ever calling `libusb_open`, so a jlink-args entry never touches foreign usbfs nodes — which is exactly what makes it convoy-safe without a pin. The entries deliberately carry **no `vid_pid`**: openocd's jlink driver never reads `adapter usb vid_pid`.

Why this matters: when a DUT wedges its usbfs node mid-usbtest, the primary SEGGER-tool flashers (JLinkExe) block on libusb enumeration of the poisoned node — the convoy effect. The recovery flasher reaches the same probe without opening any non-SEGGER usbfs node, so the post-hang reflash/reset can run while the bus is still poisoned.

## Bench validation (ci.lan, 2026-08-19, all seven boards, locks held ~8 min)

Each board was flashed and reset through its `flasher_recover` entry exactly as `usbtest.py` receives it (`board['flasher']` replaced by the recovery dict), timeout 90 s / 30 s:

| Board                    | Flash   | Reset  | Verified by                                        |
|--------------------------|---------|--------|----------------------------------------------------|
| stm32f407disco           | 1.81 s  | 0.13 s | fresh enumeration, SN matches roster uid           |
| stm32f072disco           | 1.76 s  | 0.13 s | fresh enumeration, SN matches roster uid           |
| stm32f723disco           | 1.53 s  | 0.13 s | fresh enumeration; transient CM7 errata-3092511 probe retry, non-fatal |
| stm32l476disco           | 2.80 s  | 0.15 s | fresh enumeration, SN matches roster uid           |
| feather_nrf52840_express | 1.26 s  | 0.12 s | clean disconnect → reconnect                       |
| metro_m4_express         | 0.91 s  | 0.13 s | see caveat below                                   |
| frdm_k64f (host-only)    | 20.98 s | 0.14 s | UART boot banner; both attached fixtures enumerated per roster |

All flashes finish far inside `RECOVER_FLASH_TIMEOUT` (90 s). openocd on the rig: `0.12.0+dev-02620-g0ce743125`.

## Known caveats (documented, deliberately not coded around)

- **metro_m4_express**: the SAMD51 UF2 double-tap bootloader stays resident if two reset-run events land within its detection window (~1 s apart reproduced it; isolated flash-only and reset-only calls each boot the app fine). The real `usbtest.py` recovery separates its reset and reflash steps by a ≥5 s settle plus a wedged-pids re-check plus flash time, so the chained case needs unusually fast fall-through; the residual worst case is *visible* (subsequent cases fail against a healthy, enumerable `239a:0021` bootloader — no bus poison), and the next primary flash restores the app. Accepted as a known limitation.
- **frdm_k64f**: openocd's on-target CRC-verify algorithm fails on this MK64FN1M0/k60.cfg/jlink combination and falls back to byte-wise read-back verify (~20 s of the 21 s total). Still comfortably inside budget; verify stays enabled — trading documented slowness for silent-bad-flash risk would be the wrong direction.
- `target/nrf52.cfg` is deprecated in favor of `target/nordic/nrf52.cfg`; the committed args are the bench-validated ones and the alias works on the rig's build.

## Out of scope

- **mimxrt1064_evk** — the board that wedges most still has no automated recovery: the rig's openocd build has no i.MX RT target config. Sourcing/writing one is its own investigation.
- `flash_openocd` keeps the `program` form — it's what the nine pinned CMSIS-DAP boards run in CI daily and was not re-validated here.
- `rescue_openocd` (the deeper vid_pid-pinned rescue) remains exact-`openocd`-gated; `openocd_seq` entries don't take that path.
- The convoy-unsafe operator hint in `usbtest.py` still suggests only "pin vid_pid on an openocd flasher"; mentioning the jlink-args `flasher_recover` alternative there is a possible follow-up touch.

## Testing

- `test/hil/test/test_ci_select.py` 224/224, `test/hil/test/test_hil_bounded.py` 108/108 (both offline, post-rebase on master's test_hil_select→test_ci_select move).
- `pre-commit run --all-files` fully green (incl. unit-test, hil-test, ci-select-test hooks).
- Hardware evidence above; boards and rig left idle/clean, locks released.

https://claude.ai/code/session_01NWTa5C6ahUcWo7Z4kDjiju


hathach:
Superseded by #3953, which carries the J-Link recovery flashers as `flasher_recover` entries on openocd `program`, with the STM32 reset-init fix and the reset-only boards.

PR #3839, opened 2026-08-21
hcd(ch32_usbfs): pace isochronous transfers with SOF

zjzhang-cn:
## Summary

The USBFS SIE executes one transaction the moment `HOST_EP_PID` is
written, with no frame-boundary alignment, so a re-submitted
isochronous endpoint is polled at the task-loop rate instead of once
per 1 ms USB frame. This reworks the CH32V20x USBFS host driver so
isochronous transfers are paced by the SOF interrupt.

## What changed

- **SOF-paced isochronous scheduling** — each ISO endpoint is allowed
  at most one transaction per USB frame, armed from the SOF tick and
  from the ISO completion handler, with a one-slot pending queue per
  endpoint so an early same-frame re-submission cannot starve the
  opposite direction.
- **NAK retry with progressive backoff** — NAK'd control/bulk
  transfers are re-armed once per frame from the SOF interrupt
  (1 → 2 → … → 64 frames) instead of a tight task-loop retry that
  floods the device's control endpoint and starves both sides.
- **Direct DMA to the application buffer** — the 64-byte internal
  TX/RX buffers would be overflowed by larger packets (e.g. FS
  isochronous); they are now used only for setup, status-stage and
  zero-length transfers.
- **Port-reset DETECT edge suppression** — DETECT(attach=0) edges
  asserted while the bus is in reset (SE0) are ignored, otherwise the
  reset is reported as a device removal and kills enumeration.
- **Transfer-flag discipline** — `UIE_TRANSFER` stays enabled for the
  driver lifetime; the ISR STOPS (`HOST_EP_PID = 0`) before clearing
  `UIF_TRANSFER` to avoid re-arming the SIE with a stale PID;
  completions are routed by an `xfer_type` snapshot so a control
  transfer can never complete through the ISO path with a stale
  `RX_LEN`.
- **API completeness** — implement `hcd_edpt_close()` and
  `hcd_edpt_abort_xfer()`; `hcd_frame_number()` returns the SOF frame
  counter.

## Motivation

Host enumeration on CH32V20x USBFS was unreliable after the 0.20.0 →
0.21.0 window: a HID device that mounted fine with 0.20.0 failed to
mount with 0.21.0 on the same hardware.

## Test / build evidence

- `./tools/build.py -b nanoch32v203 -s make`: 34 OK, 0 failed
- Verified on hardware (nanoch32v203): `host_info_to_device_cdc`,
  `host_hid_to_device_cdc`, `device_info`, `msc_file_explorer`
- clang-format and codespell clean; fresh `host/device_info` build
  with `riscv32-wch-elf-gcc` 15 passes

## Summary by CodeRabbit

* **Improvements**
  * Improved USB host transfer timing and reliability, including smoother isochronous audio/video data handling.
  * Added more reliable recovery and retry behavior for interrupted or temporarily unavailable transfers.
  * Improved handling of USB resets, endpoint shutdown, transfer cancellation, and frame timing.
  * Improved synchronization between USB activity and system interrupts for more consistent device communication.


zjzhang-cn:
Verified audio_host and msc_file_explorer on BSP nanoch32v203

PR #3845, opened 2026-08-24
ci: an empty selection must build nothing, plus selector follow-ups

hathach:
A PR whose build axis legitimately selects **nothing** was rebuilding **everything**.
Both merged-since PRs show it:

| PR    | changed                                | selector said         | actually ran |
| ----- | -------------------------------------- | --------------------- | -----------: |
| #3842 | a skill, `README.rst`, `.gitignore`    | no families           |   74 cmake legs |
| #3840 | `test/hil/**` + docs                   | no families (rule 2)  |   74 cmake legs |

The classifier was right both times. From #3842's own `set-matrix` log:

```
ci_select[build]: .gitignore: non-code, no build contribution
ci_select[build]: README.rst: non-code, no build contribution
ci_select[build]: .claude/skills/update-sponsor/*: non-code, no build contribution
```

## The bug

`.build.families` is read twice in the build-extras block: as a `|`-joined regex for the
metrics artifact pattern, and implicitly as "is anything selected". An **empty** list and
one **rejected** by the charset guard both leave the regex empty and mean opposite
things — but the branch tested only `-z "$FAMILY_REGEX"`:

```bash
FAMILY_REGEX=$(jq -r '.build.families | join("|")')   #  []  ->  ""
case "$FAMILY_REGEX" in *[!-A-Za-z0-9_|]*) FAMILY_REGEX='' ;; esac
if [ -z "$FAMILY_REGEX" ]; then                       #  fires for BOTH
  MATRIX_JSON=$(python .github/scripts/ci_set_matrix.py)   # <- discards the correct
fi                                                         #    all-empty matrix
```

`MATRIX_JSON` was already right — `ci_set_matrix --select-file` had returned the all-empty
matrix. The fall-open branch replaced it with the unscoped one, so every toolchain got its
full family list.

Two things made this hard to see from the outside: the HIL axis never passes through this
block, so it stayed correct (rig jobs skipped on #3842, full rig on #3840 per rule 2); and
no `ci_set_matrix: UNSCOPED` marker appears anywhere, because `ci_set_matrix` behaved
perfectly and the override happens in the shell afterwards. It looked like a
classification failure when it was not.

Introduced in the #3841 follow-up that made `FAMILY_REGEX`/`EXAMPLE_MAP`/`BUILD_FILTERED`
drop together — correct for the rejected case, wrong for the empty one.

Branching on which case it is:

| selection                            | matrix legs | `build_filtered` | regex               |
| ------------------------------------ | ----------: | ---------------- | ------------------- |
| `families: []` — nothing selected    |       **0** | `false`          | `''`                |
| `families: [stm32f4, rp2040]`        |           4 | `true`           | `stm32f4\|rp2040`   |
| `families: [stm32f4.*]` — rejected   |         129 | `false`          | `''`                |

An empty list keeps the all-empty matrix so every leg skips, and sets
`build_filtered=false` because nothing was built and there is no baseline to compare
against. A rejected list falls open exactly as before. `FAM_*` is renamed to `FAMILY_*`
along the way: `FAM` meant `hw/bsp` board family, and the abbreviation was ambiguous in a
file that also says "family" for the per-family example map and for `get_deps`' tokens.

This block had **no test at all**, which is how the bug shipped through two merges and a
max-effort review. `test_ci_metrics.py` now extracts it from `build.yml` and executes it
for real against all three shapes. Verified against master it fails with `129 != 0` on the
empty case and `'' != 'stm32f4|rp2040'` on the scoped one. Writing it surfaced two traps
worth recording: the block prints `::warning::` to **stdout** (so the probe needs a
sentinel), and `|` cannot be the field separator because the regex contains one.

## Also in here

Follow-ups to the same machinery, each a self-contained commit.

**`ci_select`: fail-open on an odd checkout path.** Five `glob.glob` sites interpolated
the repo root raw, so a checkout under a path containing `[`, `*` or `?` matched nothing
and the selector failed *closed* — silently selecting less rather than more. An `_rg()`
helper applies `glob.escape` to the root at all five.

**The `ci-full` label is removed.** It never worked: the label is read in a job that has
already computed the matrix from the selection, so setting it changed nothing. Rather than
repair an escape hatch nobody has needed, drop it — a PR that wants the full matrix can
touch any rule-17 path, and the selector falls open on every exception already.

**Dead skip tokens.** 13 `mcu:MKL25ZXX` / `mcu:SAME5X` tokens across 12 `skip.txt` files
name MCUs no board reports, so they excluded nothing. They read as deliberate coverage
gaps and are not.

**The rule table now lives in the module docstring** as a carbon copy of the spec's 23
rows, guarded by a test that fails when the two drift. Contributors editing `ci_select.py`
see the binding table without leaving the file.

**Six selection behaviours had no test** — a mutation pass broke each one and the whole
suite stayed green. Every one is a small expression whose removal silently *narrows* the
selection, which is the direction that merges a regression rather than wasting a runner.

## Performance

Four hot spots, measured before and after. A 260-path differential over every rule shape
reports **0 divergences**, and the whole run is **1.8× faster** (26.0s → 14.6s).

| fix                                                            |     before |  after |
| -------------------------------------------------------------- | ---------: | -----: |
| `_prune_buildable` asks `skip_example` per example, short-circuiting | 6,883 calls | 1,889 |
| `reasons` out of the JSON — stderr only, no consumer reads them back |     453 KB |  12 KB |
| `tusb_config.h` text cached (diff touching all of `src/class`) | 4,240 reads |     46 |
| `lib_examples` globs the two filenames, not `**/*`             | 489 entries |    107 |

`_prune_buildable` alone was 0.63s of 0.64s: it materialised every family's full buildable
list to answer two yes/no questions, probing 46 examples × every board whatever the diff
had selected.

Its one subtlety is preserved and now pinned. A family whose selection is already
everything it can build carries no `-e` list, because `build.py` applies the same
`skip_example` the pruner just did. Dropping that check compiled identical firmware and
only inflated the payload — which is exactly why 216 tests stayed green through it while I
had it wrong.

## Membrowse upload is no longer scoped by the PR filter

The upload step passed `$EX_ARGS`, which never scoped its targets —
`examples-membrowse-upload` is not `all`, so `resolve_example_target_groups` passes it
through as the aggregate — but it *did* move the board, because `--one-first` picks one
that can build the `-e` set.

Dropping it keeps the size history keyed on each family's preferred board whatever a PR
touches. The accepted cost: where that board is not the one the Build step compiled, there
is no elf and every example uploads `--identical`, so the leg contributes no real data.
Eleven families are in that position (`imxrt lpc11 lpc18 lpc54 mcx rp2040 rx samd11
stm32l0 stm32l4 tm4c`); the rest are unaffected. A test derives that set from `build.py`
and pins it, so the cost stays measured — it already caught an undercount of four.

## Validation

`pre-commit run --all-files` green. Suites: `test_ci_select` 218, `test_ci_metrics` 26,
`test_hil_bounded` 118. End-to-end, both PR shapes above now yield `legs=0`; a
`dcd_dwc2.c` diff yields 22 families / 8 boards / 37 legs.


test/hil/test/test_ci_metrics.py
```diff
@@ -447,6 +447,70 @@ def test_the_build_extras_drop_when_the_matrix_falls_open(self):
...
+                _json.dump(sel, fh)
+            matrix = subprocess.run(
+                [sys.executable, os.path.join(repo, '.github/scripts/ci_set_matrix.py'),
+                 '--select-file', selp], capture_output=True, text=True, cwd=repo).stdout.strip()
+            self.assertTrue(matrix, 'ci_set_matrix produced nothing')
+            sh = os.path.join(d, 'probe.sh')
+            with open(sh, 'w') as fh:
+                # shlex.quote, not hand-rolled quoting: a TMPDIR with a space in it
+                # made this fail for a reason that had nothing to do with the block
+                fh.write('BUILD_SELECT_FILE=' + shlex.quote(selp) + '\n')
```
hathach, line 471: Fixed in b7c2f8060. Reproduced first: with `TMPDIR="/tmp/has space"`, `test_a_real_family_list_stays_scoped` fails before the change and the whole suite is green after. Both assignments now go through `shlex.quote`.

`MATRIX_JSON` was the less likely of the two (it is JSON, so no single quotes to break out of), but quoting both is the right habit rather than reasoning about which values happen to be safe today.

PR #3849, opened 2026-08-26
test/hil: run blocking device IO in a child, not on the worker

hathach:
Two call sites in the HIL harness could hang a pool worker indefinitely. Both now run in a child process bounded by `run_alongside`, which already carries the own-session / killpg / bounded-reap contract.

### Why each one blocks

**`usblp_open()` never consults `O_NONBLOCK`.** It takes the driver-global `usblp_mutex`, then calls `usb_autopm_get_interface()` — a runtime-PM resume that does I/O (v6.12.96 `usblp.c`). One wedged printer blocks opens of *every* usblp node.

`test_device_printer_to_cdc` opened `/dev/usb/lp*` on the worker itself for its write half — while its read half, twenty lines below, already forked for exactly this reason:

> a blocked thread cannot be abandoned without keeping that fd, and usblp allows a SINGLE opener, so the next open returns `-EBUSY` for the life of the worker. A killed process takes its fd with it.

**`hid.enumerate()` reads `manufacturer` and `product` for every HID device it lists** (hidapi's hidraw backend), and both are `usb_string_attr` — served under the device lock a wedged usbfs ioctl holds (`sysfs.c:141-143`). So a wedged HID device that isn't even ours stalls the walk, unbounded, in-process. The enumeration retry moves into the child so a poll doesn't re-pay interpreter start-up.

### Not converted, deliberately

`test_device_midi_test` opens an ALSA rawmidi node and looks identical — but rawmidi **honours `O_NONBLOCK` on open** (v6.12.96 `rawmidi.c:489`, `-EBUSY` rather than queueing on `open_wait`), unlike usblp. Same-looking call, opposite behaviour.

### Tests

A scripted `hid` stub covers the wedged enumerate, the wedged open, the absent device and both bad-echo shapes. The printer writer runs against a **FIFO with a payload past the pipe buffer**, so its `select()` bound and partial-write resume actually execute — a regular-file target would run neither, since `select()` is always writable there and `os.write` never returns short. A stalled reader proves the timeout message, and a reader-less FIFO proves a blocked open leaves no readiness marker.

### One subtlety worth flagging

The parent and child share **one** readiness deadline. Two different ones let a slow open finish after the parent stopped reading — the writer then sent with nobody listening, and the byte-compare reported *firmware data corruption* for what was only a slow interpreter start. The wedged-node verdict keys on the readiness marker never appearing, not on the exit code, so a slow drain and a wedged node stay distinguishable.

All six suites green, `pre-commit run --all-files` clean.

### DIFF
--- a/test/hil/hil_test.py
+++ b/test/hil/hil_test.py
@@ -44,7 +44,6 @@ import itertools
 import os
 import random
 import re
-import select
 import signal
 import shlex
 import sys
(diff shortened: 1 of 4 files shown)


hathach:
Closing this — the printer half solves a problem master does not have, and the HID half is built on a misused API.

`hil_test.py:1002` already calls `hil_util.bounded_open(lp_dev, O_WRONLY | O_NONBLOCK, 5)`, whose docstring cites the same `usb_autopm_get_interface` / driver-global `usblp_mutex` mechanism this PR quotes as its justification. The hang is already bounded, and the abandoned-fd case that motivated forking (usblp is single-opener) is covered too — `bounded_open`'s `handoff` lock closes the fd if the caller gave up. The conversion came from a branch where `bounded_open` was being deleted; without that, it is a rewrite with no defect behind it, trading a 5s in-process bound for a fork, a payload file, a marker file and a shared-deadline dance.

The HID gap is real — `hid.enumerate()` reads `manufacturer` and `product` for every HID device it lists, both `usb_string_attr` served under the device lock, with no bound anywhere. But `run_alongside` is the wrong runner for it: that helper exists for the read-while-we-write shape, and this passes `lambda: None` because there is no concurrent work. Reopening as a narrower change that bounds the enumerate itself, reusing the mechanism already in the tree.

PR #3850, opened 2026-08-26
test/hil: bound the two hidapi calls that can stall a worker

hathach:
`test_device_hid_generic_inout` holds the last unbounded blocking IO in `hil_test.py`.

**Why it can stall.** hidapi's hidraw backend reads `manufacturer` and `product` for **every** HID device it lists (`linux/hid.c`), and both are `usb_string_attr` — served under the device lock a wedged usbfs ioctl holds (v6.12.96 `sysfs.c:141-143`). So one wedged HID device, which need not even be the board under test, stalls `hid.enumerate()` forever on the worker with nothing to abandon it. `hid_open` walks the same descriptors to match vid/pid/serial, so it carries the same exposure.

Every sibling in this file is already bounded — `run_cmd` for the subprocess paths, `bounded_open` for the printer's lp node, pyserial's own timeouts, the MTP session in its own process. These two were the gap.

**`bounded_call` is the generic form of `bounded_open`**, for a library call that hands back no fd we could close: same daemon thread, same wall-clock join, same `SYSFS_UNKNOWN` sentinel, and the same strand accounting — so a poll loop doesn't re-pay the bound on a call already known stuck, and the blindness cap still governs how many threads can accumulate.

`bounded_open` is **not** folded into it, deliberately: their resource semantics differ. `bounded_open` has a `handoff` lock that closes an fd arriving *after* the deadline — usblp is single-opener, so a leaked fd poisons the node — and `bounded_call` cannot know what `fn()` allocated. That is also why `bounded_open`'s memo clears on an inode change while `bounded_call`'s entry is permanent.

### The dedup that came out of it

What the two *did* share was the bookkeeping, and adding `bounded_call` made that three copies — `read_sysfs` and `bounded_open` carried an identical eight-line inode block. `_record_strand` now holds it once.

This is the piece where being subtly wrong is easy, and there is precedent: the credit must be spent per **key**, not per caller, or four `hil_pool_check` poll threads on *one* wedged device spend the whole blindness budget between them and latch on the single wedge the tool was run to find. And the memo must key on the kernfs **inode**, which a re-enumeration changes, or a board that recovered on the same busport stays invisible for the life of the process.

Two contract details that matter:

- **A give-up answers `SYSFS_UNKNOWN`, never `None` or `[]`.** Reading "did not answer" as "no such device" is the absence/unknown conflation `read_sysfs` exists to prevent; here it would report an unreadable bus as a missing board and send the operator after hardware that is fine.
- **The callee's own exception is re-raised**, not folded into the sentinel — a hidapi error is a fact about the device, a timeout is not.

### Tests

Cover the bound, the sentinel staying falsy for `if not listed:` callers, the memo not re-paying, and the raise passing through.

Mutation-checked: removing the bound **hangs** the suite; returning `None` instead of the sentinel fails two tests; swallowing the callee's raise fails one.

All six suites green, `pre-commit run --all-files` clean.

### DIFF
--- a/test/hil/helper/hil_util.py
+++ b/test/hil/helper/hil_util.py
@@ -253,19 +253,38 @@ def read_sysfs(path: str, grace: float = SYSFS_READ_GRACE) -> str | None | _Sysf
         # of them recorded it -- spending the whole blindness budget on a single device,
         # which is what this memo exists to prevent. note_sysfs_strand takes the lock
         # itself, so call it after releasing.
-        with _sysfs_stuck_lock:
-            first = path not in _sysfs_stranded
-            if first:
-                try:
-                    # stat, never the thread's own open(): stat does not call ->show(), so
-                    # it cannot block on the device lock the reader is stuck behind
-                    _sysfs_stranded[path] = os.stat(path).st_ino
-                except OSError:
-                    _sysfs_stranded[path] = None   # unstattable, but still known-stranded
-        if first:
-            note_sysfs_strand()
+        _record_strand(path, stat_path=path)
         return SYSFS_UNKNOWN
     return out.get('v')
+
+
+def _record_strand(key: str, stat_path: str | None = None) -> None:
+    """Memoise `key` as stranded and spend ONE blindness credit for it.
+
+    The credit is per KEY, not per caller: hil_pool_check runs a ThreadPoolExecutor of
+    SYSFS_STUCK_MAX workers over one bus, so counting each reader let four threads on ONE
+    wedged device spend the whole budget between them -- latching blind on the single
+    wedge the tool was run to find.
+
+    `stat_path` memoises by kernfs INODE, which a re-enumeration changes: that is the
+    all-clear for a device that came back on the same busport, and without it a recovered
+    board would stay invisible for the life of the process. os.stat is safe on a wedged
+    device -- it does not call ->show(), so it cannot block on the lock the reader is
+    stuck behind. Omit it when there is no node to key on (see bounded_call); the entry is
+    then permanent, which is correct for something with no way to signal recovery.
+    """
+    with _sysfs_stuck_lock:
+        first = key not in _sysfs_stranded
+        if first:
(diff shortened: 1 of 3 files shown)


hathach:
Closing: the `h.open()` bound in this PR cannot work, and the mechanism is wrong for hidapi generally.

cython-hidapi wraps `hid_enumerate` in `with nogil` but calls `hid_open` bare (hid.pyx). A daemon thread cannot bound a GIL-holding C call — `done.wait()` parks off-GIL but must reacquire the GIL to return, and the worker thread never yields it. Measured against the real module: a GIL-releasing callee honoured a 1.0s bound exactly; a GIL-holding one ran the full 5.0s and returned the callee's value instead of the sentinel. So the assert this PR adds is dead code for a genuine wedge.

Per-call GIL matrix (hid.pyx 0.15.0):

| call | releases GIL | bounded here |
|---|---|---|
| enumerate | yes | yes — works |
| open | no | yes — inert |
| write | yes | no — would have worked |
| read | yes (+ hidapi's own 2000ms) | no — already safe |
| close | no | no |

So this bounds the one call that was already boundable, adds a bound that can never fire, and leaves `write` and `close` untouched — `close` runs in a `finally` on every failure path and takes `down_write(&minors_rwsem)`, the same rwsem `hidraw_open` needs, so a stalled close blocks every hidraw node.

The tests could not have caught it: all three give-up tests use `time.sleep`, which releases the GIL, so they validate only the half that already worked.

One correction to this PR's own justification while it is on the record: `hid.enumerate(0xCafe)` is a FILTERED call — hidapi skips on VID mismatch before the locked manufacturer/product reads — so "every HID device on the bus" is wrong. The bound is still warranted because every DUT on this rig is VID 0xCafe and a wedged sibling does pass the filter, but the hazard is narrower than stated.

Redoing out of process, which is what the rest of this harness already does for MTP and the printer, and what bounded_open's own docstring prescribes: 'A PROCESS, not a thread'.

PR #3859, opened 2026-08-27
feat(dcd/dwc2): Add Periodic Transfer Interrupt feature

tore-espressif:
## DWC2 DCD: Periodic Transfer Interrupt feature

Enable `DCTL.IgnrFrmNum` in Buffer DMA so ISOC endpoints can be armed without even/odd frame matching.
This is required to schedule ISOC OUT on alternating 7/8 ms intervals (Bluetooth HCI 7.5 ms) while keeping bInterval = 1.

Gated by CFG_TUD_DWC2_PTI_ENABLE (default off, DMA only). When on, mask IncompISOCIN and report dropped ISOC OUT (PktDrpSts) as a failed transfer.

### DIFF
--- a/src/portable/synopsys/dwc2/dcd_dwc2.c
+++ b/src/portable/synopsys/dwc2/dcd_dwc2.c
@@ -16,6 +16,10 @@
 #error DWC2 require either CFG_TUD_DWC2_SLAVE_ENABLE or CFG_TUD_DWC2_DMA_ENABLE to be enabled
 #endif
 
+#if CFG_TUD_DWC2_PTI_ENABLE && !CFG_TUD_DWC2_DMA_ENABLE
+#error CFG_TUD_DWC2_PTI_ENABLE requires CFG_TUD_DWC2_DMA_ENABLE (Buffer DMA)
+#endif
+
 // Debug level for DWC2
 #define DWC2_DEBUG    2
 
@@ -113,6 +117,32 @@ TU_ATTR_ALWAYS_INLINE static inline bool dma_device_enabled(const dwc2_regs_t* d
   // Internal DMA only
   const dwc2_ghwcfg2_t ghwcfg2 = {.value = dwc2->ghwcfg2};
   return CFG_TUD_DWC2_DMA_ENABLE && ghwcfg2.arch == GHWCFG2_ARCH_INTERNAL_DMA;
+}
+
+// Periodic Transfer Interrupt: DCTL.IgnrFrmNum in Buffer DMA only (not Slave, not thresholding)
+TU_ATTR_ALWAYS_INLINE static inline bool pti_device_enabled(const dwc2_regs_t* dwc2) {
+#if CFG_TUD_DWC2_PTI_ENABLE
+  return dma_device_enabled(dwc2);
+#else
+  (void) dwc2;
+  return false;
+#endif
+}
+
+static void pti_configure(dwc2_regs_t* dwc2) {
+#if CFG_TUD_DWC2_PTI_ENABLE
+  if (!dma_device_enabled(dwc2)) {
+    return;
+  }
+  dwc2->dctl |= DCTL_IGNRFRMNUM;
+  const dwc2_ghwcfg4_t ghwcfg4 = {.value = dwc2->ghwcfg4};
+  // Explicitly disable AHB thresholding (defensive programming only)
+  if (ghwcfg4.dedicated_fifos) {
+    dwc2->dthrctl &= (uint32_t) ~(DTHRCTL_NONISOTHREN | DTHRCTL_ISOTHREN | DTHRCTL_RXTHREN);
+  }
+#else
+  (void) dwc2;
+#endif
 }
 
 static void dma_setup_prepare(uint8_t rhport) {
@@ -455,6 +485,8 @@ bool dcd_init(uint8_t rhport, const tusb_rhport_init_t* rh_init) {
   dcfg |= DCFG_NZLSOHSK; // send STALL back and discard if host send non-zlp during control status
   dwc2->dcfg = dcfg;
 
+  pti_configure(dwc2); // No operation if !CFG_TUD_DWC2_PTI_ENABLE
+
   dcd_disconnect(rhport);
 
   // Force device mode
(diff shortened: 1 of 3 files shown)


tore-espressif:
cc @peter-marcisovsky @hathach @HiFiPhile

HiFiPhile:
Hi @tore-espressif , didn't know HCI does this weird stuff, I got some review:
P1 — Incorrect PTI OUT transfer sizing.
[edpt_schedule_packets()](https://github.com/hathach/tinyusb/blob/1d73cd43ce76e65e5e044ae81fb0cf221f3c6acf/src/portable/synopsys/dwc2/dcd_dwc2.c#L377-L402) programs XferSize = total_bytes. DWC2 Programming Guide §9.3.5, PDF pp. 274–278, requires PTI OUT XferSize to be packet_count × MPS, with each packet rounded to a DWORD boundary. For Bluetooth altsetting 6 with MPS 63, one packet therefore occupies 64 DMA bytes, not 63. The implementation needs separate logical-length and padded-DMA-span accounting, plus guaranteed padded buffer capacity. See [DWC2 Programming Guide](D:\\doc\\doc_classified\\dwc2\\DWC_otg_programming_changes.pdf).

P1 — Packet-drop detection is incomplete.
[handle_epout_dma()](https://github.com/hathach/tinyusb/blob/1d73cd43ce76e65e5e044ae81fb0cf221f3c6acf/src/portable/synopsys/dwc2/dcd_dwc2.c#L1067-L1084) checks only PktDrpSts. Programming Guide p. 275 warns that consecutive drops can produce XferComplete without PktDrpSts; remaining PktCnt/XferSize must also be inspected and the transfer reprogrammed. The current code can report that case as success.

tore-espressif:
@HiFiPhile Thank you for the review!

Ad point 1: If I understand it correctly, this problem is not PTI specific, it is valid also for mode currently implemented in DCD?

Refactoring the way how transfer sizes are calculated seems too risky to me... I'd suggest users to always use MPS = 64 to workaround this problem.

In case this is not acceptable to you, feel free to close this PR

HiFiPhile:
> If I understand it correctly, this problem is not PTI specific, it is valid also for mode currently implemented in DCD?

You are right I've overlooked it.

@tore-espressif 

I read the HCI transport [spec](https://www.bluetooth.com/wp-content/uploads/Files/Specification/HTML/Core-61/out/en/host-controller-interface/usb-transport-layer.html) and I think we don't need PTI. The 7.5 ms value is the cadence of non-empty mSBC HCI packets, not the USB endpoint polling interval:
- For ISO OUT just arm the transfer every 1ms, rearm if no data received.
- For ISO IN, queue payload when available and a ZLP otherwise.


From Codex:
Linux keeps the USB endpoint on a 1 ms schedule:
- Linux builds the alternating 7/8-frame pattern here: [`[__fill_isoc_descriptor_msbc()](https://github.com/torvalds/linux/blob/786262be6048deab760f68c8acc2c85607165894/drivers/bluetooth/btusb.c#L1762-L1802)`](https://github.com/torvalds/linux/blob/786262be6048deab760f68c8acc2c85607165894/drivers/bluetooth/btusb.c#L1762-L1802). It inserts 6 or 7 empty ISO descriptors, followed by the payload descriptor.

- The URB uses the endpoint descriptor’s `bInterval`: [`[alloc_isoc_urb()](https://github.com/torvalds/linux/blob/786262be6048deab760f68c8acc2c85607165894/drivers/bluetooth/btusb.c#L2198-L2224)`](https://github.com/torvalds/linux/blob/786262be6048deab760f68c8acc2c85607165894/drivers/bluetooth/btusb.c#L2198-L2224).

- xHCI treats each zero-length descriptor as an actual packet opportunity: [`[xhci-ring.c](https://github.com/torvalds/linux/blob/786262be6048deab760f68c8acc2c85607165894/drivers/usb/host/xhci-ring.c#L4128-L4141)`](https://github.com/torvalds/linux/blob/786262be6048deab760f68c8acc2c85607165894/drivers/usb/host/xhci-ring.c#L4128-L4141).

- The Bluetooth specification defines altsetting 6 as **1 ms USB polling / 7.5 ms HCI packet interval**: [[Bluetooth Core Specification, USB Transport Layer, Table 2.1](https://www.bluetooth.com/wp-content/uploads/Files/Specification/HTML/Core-61/out/en/host-controller-interface/usb-transport-layer.html)](https://www.bluetooth.com/wp-content/uploads/Files/Specification/HTML/Core-61/out/en/host-controller-interface/usb-transport-layer.html). Search that page for `1 ms/7.5 ms`.

Thus Linux maintains 1 ms USB slots while placing non-empty HCI packets every 7 or 8 slots.

gm-jiang:
> Thus Linux maintains 1 ms USB slots while placing non-empty HCI packets every 7 or 8 slots.

@HiFiPhile Yes, for linux platform, linux usb host core stack schedule the ISO packet every 1ms, like this sequence
```
Frame 0: 63 bytes
Frame 1: 0 bytes
Frame 2: 0 bytes
Frame 3: 0 bytes
Frame 4: 0 bytes
Frame 5: 0 bytes
Frame 6: 0 bytes
Frame 7: 63 bytes
Frame 8: 0 bytes
Frame 9: 0 bytes
Frame 10: 0 bytes
Frame 11: 0 bytes
Frame 12: 0 bytes
Frame 13 0 bytes
Frame 14 0 bytes
Frame 15: 63 bytes
Frame 16: 0 bytes
Frame 17: 0 bytes
Frame 18: 0 bytes
Frame 19: 0 bytes
Frame 20: 0 bytes
Frame 21: 0 bytes
Frame 22: 63 bytes
Frame 23: 0 bytes
Frame 24: 0 bytes
Frame 25: 0 bytes
Frame 26: 0 bytes
Frame 27: 0 bytes
Frame 28: 0 bytes
Frame 29: 0 bytes
Frame 30: 63 bytes
```
But for windows platform, windows usb host usb stack schedule like this
```
Frame 0 : 63 bytes
+7
Frame 7 : 63 bytes
+8
Frame 15: 63 bytes
+7
Frame 22: 63 bytes
+8
Frame 30: 63 bytes
+7
Frame 37: 63 bytes
+8
Frame 45: 63 bytes
+7
```
Using current tinyusb stack, we always lose one fixed ISO frame packet. This is because the data does not alternate changes as odd/even frames, but instead of `odd/even/even/odd/odd/even/even/odd/odd/.../`

HiFiPhile:
> But for windows platform, windows usb host usb stack schedule like this
Frame 0 : 63 bytes
+7
Frame 7 : 63 bytes
+8

Interesting to see how Windows abuse EHCI

HiFiPhile (approved):

PR #3815, opened 2026-08-11
hcd/dwc2: fix FIFO allocation and periodic transfer

HiFiPhile:
## Problem

DWC2 periodic transfers could remain active indefinitely during repeated audio
stream start/stop cycles. EP0 traffic continued, but the affected endpoint
stopped receiving tokens and never reported success or failure.

A reproduced failure left the channel in this state:

- `HCCHAR=0xe044881c` (`CHENA|CHDIS`)
- one packet still pending in `HCTSIZ`
- `HCINT=0` and `HAINT=0`
- software result still `XFER_RESULT_INVALID`

## Root causes

Hardware traces identified several independent ways to lose the terminal
channel transition:

- Slave mode acknowledged a simultaneous `ChHltd` before the channel handler
  consumed it.
- Periodic channels selected a frame before request-queue waits or interrupt
  preemption had completed.
- Slave OUT enabled the channel but deferred the FIFO write that creates the
  request-queue entry.
- Buffer-DMA code requested `CHDIS` for non-split periodic channels, which the
  DWC2 Programming Guide explicitly prohibits.
- Disconnect and abort paths could release software ownership before channel
  activity was safely retired.

## Changes

- Correct the host RX, periodic TX, and non-periodic TX FIFO layout.
- Preserve periodic endpoint cadence and defer early submissions through SOF.
- Select `ODDFRM` and enable the channel from one protected `HFNUM` sample.
- Write the initial slave OUT packet in the same bounded critical section as
  channel enablement. Reject late periodic submissions using `HFNUM.FrRem`.
- Preserve a simultaneous slave `ChHltd` for the following handler pass.
- Report missed isochronous intervals as failed instead of fabricating success.
- Keep single-transaction isochronous transfers on `DATA0`.
- Let periodic DMA channels halt naturally and retain endpoint ownership while
  an abort is waiting for that halt.
- Clean up host channels on disconnect without reinitializing the PHY from ISR
  context.
- Propagate enumeration submission failures and release address-zero and
  provisional device state correctly.

## Validation

The negative control reproduced the stuck channel in its first uninstrumented
capture with no analyzer loss.

The hardware-tested candidate then completed ten-minute segmented runs on an
STM32U5A5 in all four configurations:

- `-O0`, slave mode: 600 seconds
- `-O2`, slave mode: 600 seconds
- `-O0`, DMA mode: 600 seconds
- `-O2`, DMA mode: 600 seconds

Total: 2,400 accepted seconds across 40 clean segments, with zero analyzer loss
[...]

PR #3864, opened 2026-08-28
host: recover from hub and roothub failures, and fix low-speed devices behind a hub on ESP32-S2/S3

P-R-O-C-H-Y:
Depends on #3815, now merged. That PR keeps the hub status transfer alive, so the hub watchdog commit is gone.

Three leftover fixes from a full-speed DWC2 host dying when an LS mouse and LS keyboard share one FS hub.

### host: close dev0 when the whole roothub port goes away

A remove for `hub_addr = 0, hub_port = 0` covers everything below the root port, but `process_remove_event()` only closed dev0 when the event named the exact bus address it was enumerating on. Mid-enumeration behind a downstream hub then left `enumerating_daddr` at 0 and deferred every later attach.

### host: retry enumeration of the roothub port after a failed attach

`enum_full_complete()` already closes the failed device (#3815). Nothing retries the root port: the host sits idle until reboot. Re-post the attach up to `USBH_ENUM_ROOT_RETRY_MAX` times while the port still reports a connection. Budget is per port and given back when the attachment ends, including when an attach event tears down an in-flight enumeration. `tuh_umount_cb()` only runs for a device that reached `tuh_mounted()`.

### dwc2: space low-speed transactions one per frame on esp32s2/s3

ESP32-S2/S3 cannot run two preamble transactions in the same 1 ms frame. A second one clears `HPRT.PENA` and the bus stays down. Not an errata and not a `GHWCFG` bit — ESP-IDF hits the same limit (IDF-12986 / espressif/esp-idf#15683). Gated at runtime by `ls_needs_preamble()` (FS port + LS device, no split), compiled in only for those MCUs.

Reservation hooks through `channel_enable()` from #3815, after the request-queue wait and outside the interrupt-masked window. Periodic INs from SOF defer a frame rather than spin. A port the core disables on its own now posts attach or remove; that used to hit a commented-out `TU_ASSERT(false, )` and stay dead.

## Testing

| Device | VID:PID | Speed |
| --- | --- | --- |
| Keyboard | `03f0:034a` | LS |
| Mouse | `413c:301a` | LS |
| Flash drive | `058f:6387` | FS |

ESP32-S3 and ESP32-P4, all three on one hub. Cold boot and unplug/replug of each, including the drive, on #3815 plus these three commits. P4 compiles the frame gate out and exercises the two generic commits on HS.

### DIFF
--- a/src/portable/synopsys/dwc2/hcd_dwc2.c
+++ b/src/portable/synopsys/dwc2/hcd_dwc2.c
@@ -1858,15 +1858,10 @@ static void handle_hprt_irq(uint8_t rhport, bool in_isr) {
       // Port enable
       const tusb_speed_t speed = hprt_speed_get(dwc2);
       port0_enable(dwc2, speed);
-    } else {
-      // Port disabled by the core (port error e.g babble). SOF generation stops, so no transfer
-      // on this bus can ever complete again. Tear down the tree; if something is still connected
-      // re-attach so the port is reset and devices are enumerated again.
-      if (hprt_bm.conn_status == 1u) {
-        hcd_event_device_attach(rhport, in_isr);
-      } else {
-        hcd_event_device_remove(rhport, in_isr);
-      }
+    } else if (hprt_bm.conn_status == 1u && hprt_bm.conn_detected == 0u) {
+      // The core disabled a still-connected port. Reuse attach handling
+      // to tear down the stale device tree and restart enumeration.
+      hcd_event_device_attach(rhport, in_isr);
     }
   }
 
--- a/src/portable/synopsys/dwc2/hcd_dwc2.c
+++ b/src/portable/synopsys/dwc2/hcd_dwc2.c
@@ -176,6 +176,83 @@ TU_ATTR_ALWAYS_INLINE static inline bool channel_is_periodic(uint32_t hcchar) {
   return hcchar_bm.ep_type == HCCHAR_EPTYPE_INTERRUPT || hcchar_bm.ep_type == HCCHAR_EPTYPE_ISOCHRONOUS;
 }
 
+#if TU_CHECK_MCU(OPT_MCU_ESP32S2, OPT_MCU_ESP32S3)
+// The FS core cannot run two low-speed transactions in the same frame: a low-speed device behind a
+// full-speed hub is addressed with a preamble packet, and a second one in the same frame makes the
+// core clear HPRT.PENA, which kills the whole root port. ESP-IDF works around this in hcd_dwc.c with
+// a 1 ms delay per low-speed control stage (espressif/esp-idf#15683). Reserve one frame per
+// low-speed transaction instead, which also covers the periodic IN of a second low-speed device.
+// Directly attached low-speed devices use no preamble and are not affected.
+#define DWC2_LS_ONE_XACT_PER_FRAME 1
+#else
+#define DWC2_LS_ONE_XACT_PER_FRAME 0
+#endif
+
+#if DWC2_LS_ONE_XACT_PER_FRAME
+#define DWC2_LS_FRAME_MASK     0x3FFFu // FrNum counts 0..0x3FFF
+#define DWC2_LS_FRAME_SPIN_MAX 200000u // covers a frame, but never spins forever if SOFs have stopped
+
+static uint32_t _ls_frame_end = 0; // last frame claimed by a low-speed transaction
+
+TU_ATTR_ALWAYS_INLINE static inline uint32_t ls_frame_now(const dwc2_regs_t* dwc2) {
+  return dwc2->hfnum & DWC2_LS_FRAME_MASK;
+}
+
+TU_ATTR_ALWAYS_INLINE static inline int32_t ls_frame_diff(uint32_t a, uint32_t b) {
+  const uint32_t d = (a - b) & DWC2_LS_FRAME_MASK;
+  return (int32_t)(d << 18) >> 18; // sign-extend 14-bit so the FrNum wrap compares correctly
+}
+
+// A transaction is free to run once the frames claimed by the previous one have passed. offset is
+// where this channel would start: periodic channels are armed for the next frame (odd_frame),
+// non-periodic ones go out in the current frame.
+TU_ATTR_ALWAYS_INLINE static inline bool ls_frame_free(const dwc2_regs_t* dwc2, uint32_t offset) {
+  return ls_frame_diff(ls_frame_now(dwc2) + offset, _ls_frame_end) > 0;
+}
+
+TU_ATTR_ALWAYS_INLINE static inline bool ls_needs_preamble(const dwc2_regs_t* dwc2, uint32_t hcchar, uint32_t hcsplt) {
+  const dwc2_channel_char_t char_bm = {.value = hcchar};
+  const dwc2_channel_split_t splt_bm = {.value = hcsplt};
+  const dwc2_hprt_t hprt = {.value = dwc2->hprt};
+  // high/full speed device, split transaction or low-speed root port: no preamble, nothing to space out
+  return char_bm.low_speed_dev && !splt_bm.split_en && hprt.speed == HPRT_SPEED_FULL;
+}
+
+// Claim the frames this channel's transaction will occupy, waiting out the ones a previous low-speed
+// transaction still holds. Must be called on every path that enables a low-speed channel, including
+// the NAK retries that re-issue an IN token without going through channel_xfer_start(), and always
+// before a caller masks interrupts: this can wait out a whole frame.
+static void ls_frame_reserve(const dwc2_regs_t* dwc2, dwc2_channel_t* channel) {
+  if (!ls_needs_preamble(dwc2, channel->hcchar, channel->hcsplt)) {
+    return;
+  }
+
+  const bool is_period = channel_is_periodic(channel->hcchar);
+  const dwc2_channel_tsize_t hctsiz = {.value = channel->hctsiz};
+  const uint32_t offset = is_period ? 1u : 0u;
+  // the core runs the packets of a multi-packet transfer back-to-back, which at low speed can spill
+  // past the end of the frame it started in
+  const uint32_t span = (!is_period && hctsiz.packet_count > 1) ? 1u : 0u;
+
+  uint32_t spin = DWC2_LS_FRAME_SPIN_MAX;
+  while (spin-- && !ls_frame_free(dwc2, offset)) {}
+
+  const uint32_t now = ls_frame_now(dwc2);
+  if (is_period) {
+    dwc2_channel_char_t rearm = {.value = channel->hcchar};
+    rearm.odd_frame = 1 - (now & 1); // waiting may have crossed a frame boundary
+    channel->hcchar = rearm.value;   // channel_enable() re-picks the frame when it selects one itself
+  }
+  _ls_frame_end = (now + offset + span) & DWC2_LS_FRAME_MASK;
+}
+
+// Periodic low-speed IN armed from the SOF interrupt: defer to the next frame rather than spin, the
+// endpoint already has a retry path for that.
+TU_ATTR_ALWAYS_INLINE static inline bool ls_edpt_deferred(const dwc2_regs_t* dwc2, const hcd_endpoint_t* edpt) {
+  return ls_needs_preamble(dwc2, edpt->hcchar, edpt->hcsplt) && !ls_frame_free(dwc2, 1);
+}
+#endif
+
 TU_ATTR_ALWAYS_INLINE static inline uint8_t req_queue_avail(const dwc2_regs_t* dwc2, bool is_period) {
   if (is_period) {
     const dwc2_hptxsts_t hptxsts = {.value = dwc2->hptxsts};
@@ -292,6 +369,11 @@ TU_ATTR_ALWAYS_INLINE static inline uint16_t channel_send_in_token(dwc2_regs_t* 
   while (0 == req_queue_avail(dwc2, channel_is_periodic(channel->hcchar))) {
     // blocking wait for request queue available
   }
+#if DWC2_LS_ONE_XACT_PER_FRAME
+  // after the queue wait, not before: a wait that spans a frame boundary would leave the
+  // reservation naming a frame this channel no longer starts in
+  ls_frame_reserve(dwc2, channel);
+#endif
   return channel_enable(dwc2, channel, next_periodic_frame);
 }
 
@@ -760,6 +842,9 @@ static bool channel_xfer_start(dwc2_regs_t* dwc2, uint8_t ch_id, bool defer_peri
       periodic_frame = channel_send_in_token(dwc2, channel, is_period);
     } else {
       hcd_dcache_clean(edpt->buffer, edpt->buflen);
+#if DWC2_LS_ONE_XACT_PER_FRAME
+      ls_frame_reserve(dwc2, channel); // IN is covered inside channel_send_in_token()
+#endif
       periodic_frame = channel_enable(dwc2, channel, is_period);
     }
   }
@@ -789,6 +874,13 @@ static bool channel_xfer_start(dwc2_regs_t* dwc2, uint8_t ch_id, bool defer_peri
     if (hcchar_bm->ep_dir == TUSB_DIR_IN) {
       periodic_frame = channel_send_in_token(dwc2, channel, is_period);
     } else {
+#if DWC2_LS_ONE_XACT_PER_FRAME
+      // the path taken whenever host DMA is off, which is the default: every low-speed OUT and SETUP
+      // goes out here, including the control stages of a device enumerating next to a live low-speed
+      // interrupt IN. Before the masked region below, since waiting out a frame with the controller's
+      // interrupt disabled would hold off every other channel too.
+      ls_frame_reserve(dwc2, channel);
+#endif
       // The final FIFO word creates the OUT request. Keep CHENA and that write
       // atomic with respect to this controller's ISR.
       // This region never waits for FIFO or queue space.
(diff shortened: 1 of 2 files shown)


P-R-O-C-H-Y:
@me-no-dev

src/portable/synopsys/dwc2/hcd_dwc2.c
```diff
@@ -200,8 +212,75 @@ TU_ATTR_ALWAYS_INLINE static inline bool channel_disable(const dwc2_regs_t* dwc2
...
+static uint32_t _ls_frame_end = 0; // last frame claimed by a low-speed transaction
+
+TU_ATTR_ALWAYS_INLINE static inline uint32_t ls_frame_now(const dwc2_regs_t* dwc2) {
+  return dwc2->hfnum & DWC2_LS_FRAME_MASK;
+}
+
+TU_ATTR_ALWAYS_INLINE static inline int32_t ls_frame_diff(uint32_t a, uint32_t b) {
+  const uint32_t d = (a - b) & DWC2_LS_FRAME_MASK;
+  return (int32_t)(d << 18) >> 18; // sign-extend 14-bit so the FrNum wrap compares correctly
+}
```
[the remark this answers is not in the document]
P-R-O-C-H-Y: `HFNUM_FRNUM_Msk` is `0xFFFF`, the register field width, but the frame counter wraps at `0x3FFF`. Using the register mask would break the comparison across a wrap, so the 14-bit mask and sign-extension are fine.

HiFiPhile:
Hi @P-R-O-C-H-Y,

ba8d20f3ce917248dd40e9968ffe6ba1e6bfc0b3 and ba8d20f3ce917248dd40e9968ffe6ba1e6bfc0b3(with post fix) looks good.

For 29fd95c250081845989d69c82a742d5e3ea0570e I'd like to know the know the cause instead of adding a polling loop:
- Is the transfer jammed without success/fail return, could be fixed by #3815
- Or somehow the re-arm is missed

For ce9b3a4e48538674cfc1a34d7860766cfa133fc1 is this limitation from hardware config (GHWCFGn) or an errata ? So we can check it instead of hard coding device family.

P-R-O-C-H-Y:
Hi @HiFiPhile, thanks for the review.

> For [29fd95c](https://github.com/hathach/tinyusb/commit/29fd95c250081845989d69c82a742d5e3ea0570e) I'd like to know the know the cause instead of adding a polling loop:

it's the first case, the transfer jams with no
success or fail return.

I know because the watchdog on its own did not fix it. While the endpoint
stays claimed, the re-arm can't even be attempted. Wedged transfer and a
healthy pending one look the same to `usbh_edpt_claim()`. It only started
recovering once the removal path aborted the pending transfer first. If the
re-arm had simply been missed, the endpoint would have been free and the
watchdog would have worked on its first pass.

It always dies in the same place: the channel teardown in
`remove_device_tree()`, when a device below the hub is removed.

So the polling is covering for a lost transfer rather than fixing it. 

> * Is the transfer jammed without success/fail return, could be fixed by [hcd/dwc2: fix FIFO allocation and periodic transfer #3815](https://github.com/hathach/tinyusb/pull/3815)
> * Or somehow the re-arm is missed

#3815 touches abort handling and channel disable, so it may be the same bug. I'll
rebase onto it and re-run the removal/replug test that reproduces this. If
the jam is gone, this commit shrinks to just the abort on removal, or drops
out entirely.

> For [ce9b3a4](https://github.com/hathach/tinyusb/commit/ce9b3a4e48538674cfc1a34d7860766cfa133fc1) is this limitation from hardware config (GHWCFGn) or an errata ? So we can check it instead of hard coding device family.

neither, as far as I can tell. It isn't in the S2/S3
errata, and no GHWCFG field describes preamble scheduling.

ESP-IDF hits the same limit and works around it with a 1 ms delay in the ISR,
commented "The HW can't handle two transactions with preamble in one frame"
(IDF-12986). They gate it at runtime on FS port + LS device, not on the chip.

So the condition is detectable even though the defect isn't, and it's exactly
what `ls_needs_preamble()` already checks. I'm happy to drop TU_CHECK_MCU and
apply the spacing wherever a preamble is needed. The cost for other FS cores
is one low-speed transaction per frame, which for interrupt-only LS devices
should be unmeasurable. Or a `CFG_TUH_DWC2_LS_ONE_XACT_PER_FRAME` macro
defaulting on for esp32s2/s3, if you'd rather not impose it.


Let me know what seems to be better and in meantime I will test the #3815

P-R-O-C-H-Y:
@HiFiPhile Good news, tested against #3815. The hub commit isn't needed with it.

Build: #3815 changes onto master + the other three commits of this PR on top (the hub status abort + watchdog removed). ESP32-S3, FS hub with an LS keyboard (03f0:034a), LS mouse (413c:301a) and an FS flash drive (058f:6387) behind it.

15 connect/disconnect events across the three devices, including removals of the flash drive. The teardown that used to leave the hub deaf and one removal after three minutes of an idle bus, so nothing was keeping the host task awake. Every edge was reported and every device kept running.

So the status transfer was jammed rather than never re-armed, and the channel teardown rework in #3815 keeps it alive. Happy to drop hub: recover a hub whose status endpoint stopped being re-armed and rebase the rest onto #3815 changes. Is there any plan to merge that soon so it is part of the master?

HiFiPhile:
> Good news, tested against https://github.com/hathach/tinyusb/pull/3815. The hub commit isn't needed with it.

That's nice, I'm still fighting with a unplug/replug race. Maybe Thach is off for Vietnam's National Day.

HiFiPhile:
@P-R-O-C-H-Y I'm done with #3815

P-R-O-C-H-Y:
@HiFiPhile Do you want me to retest with final version of your PR? Once merged, I will rebase my PR and remove the commits no longer needed.

HiFiPhile:
> @HiFiPhile Do you want me to retest with final version of your PR? Once merged, I will rebase my PR and remove the commits no longer needed.

Yes please

P-R-O-C-H-Y:
> > @HiFiPhile Do you want me to retest with final version of your PR? Once merged, I will rebase my PR and remove the commits no longer needed.
> 
> Yes please

@HiFiPhile Retested your latest #3815 (e461c6f9c) plus the other three commits of this PR, hub watchdog still out. All good, everything works as expected.

So I’m dropping hub: recover a hub whose status endpoint stopped being re-armed. I’ll rebase the rest onto #3815 once it lands. The dwc2 commit hooks through your `channel_enable()`— reservation sits outside the interrupt-masked window.

src/portable/synopsys/dwc2/hcd_dwc2.c
```diff
@@ -1473,7 +1567,14 @@ static void handle_hprt_irq(uint8_t rhport, bool in_isr) {
...
       port0_enable(dwc2, speed);
     } else {
-      // TU_ASSERT(false, );
+      // Port disabled by the core (port error e.g babble). SOF generation stops, so no transfer
+      // on this bus can ever complete again. Tear down the tree; if something is still connected
+      // re-attach so the port is reset and devices are enumerated again.
+      if (hprt_bm.conn_status == 1u) {
+        hcd_event_device_attach(rhport, in_isr);
+      } else {
+        hcd_event_device_remove(rhport, in_isr);
```
HiFiPhile: An attach event already tears down any existing root-port tree before re-enumerating, so `attach` is enough.

```diff
-    } else {
-      // Port disabled by the core (port error e.g babble). SOF generation stops, so no transfer
-      // on this bus can ever complete again. Tear down the tree; if something is still connected
-      // re-attach so the port is reset and devices are enumerated again.
-      if (hprt_bm.conn_status == 1u) {
-        hcd_event_device_attach(rhport, in_isr);
-      } else {
-        hcd_event_device_remove(rhport, in_isr);
-      }
+    } else if (hprt_bm.conn_status == 1u && hprt_bm.conn_detected == 0u) {
+      // The core disabled a still-connected port. Reuse attach handling
+      // to tear down the stale device tree and restart enumeration.
+      hcd_event_device_attach(rhport, in_isr);
```

HiFiPhile:
@P-R-O-C-H-Y if your schedule is tight @me-no-dev or @tore-espressif can review and approve #3815.

src/portable/synopsys/dwc2/hcd_dwc2.c
P-R-O-C-H-Y: Applied, thanks. Attach-only -> skipped when `conn_detected` is already set so we don’t post attach twice.

P-R-O-C-H-Y:
PR was rebased, removed no longer needed commits as #3815 is merged + updated the PR description.

P-R-O-C-H-Y:
Retested on S3 and P4 with the same setup, everything works as expected.

HiFiPhile:
@P-R-O-C-H-Y 
I've tested the LS workaround on STM32U5 (don't have ESP32 board ATM) but still observed > 2 preambles in one frame, due to sniffer limitation I can't inspect LS traffic but looks like an additional transaction attempt in the frame.
{CAD01569-4A81-44B8-94CF-7169EB3F6FED}

I've tried another approach in https://github.com/hathach/tinyusb/tree/pr3864-alt d0999293426ccbc8904474eaf5d182dd21a66cce:
- schedule one 8 bytes packet per SOF, prevent long transfer from creating multiple PREs.
- avoid 1ms wait in ISR context

b09f21f328e12c7296c554dbaad9e8d607cc17f3, 47a432a3683a0dbc77de7e4ae3ddbbb9dffa98fd, 6f2824e97bbc7226e06eddd1429fc4fa64f6b2a5 are newly identified fixes.

P-R-O-C-H-Y:
> @P-R-O-C-H-Y I've tested the LS workaround on STM32U5 (don't have ESP32 board ATM) but still observed > 2 preambles in one frame, due to sniffer limitation I can't inspect LS traffic but looks like an additional transaction attempt in the frame. {CAD01569-4A81-44B8-94CF-7169EB3F6FED}
> 
> I've tried another approach in https://github.com/hathach/tinyusb/tree/pr3864-alt [d099929](https://github.com/hathach/tinyusb/commit/d0999293426ccbc8904474eaf5d182dd21a66cce):
> 
> * schedule one 8 bytes packet per SOF, prevent long transfer from creating multiple PREs.
> * avoid 1ms wait in ISR context
> 
> [b09f21f](https://github.com/hathach/tinyusb/commit/b09f21f328e12c7296c554dbaad9e8d607cc17f3), [47a432a](https://github.com/hathach/tinyusb/commit/47a432a3683a0dbc77de7e4ae3ddbbb9dffa98fd), [6f2824e](https://github.com/hathach/tinyusb/commit/6f2824e97bbc7226e06eddd1429fc4fa64f6b2a5) are newly identified fixes.

@HiFiPhile So you have turned the gate for STM? Currently in the PR it's compiled only for ESP32-S2 and S3. So it shouldn't affect any other chips.

I have tested your alt branch with all the commits on ESP32-S3 (FS hub, LS keyboard, LS mouse, FS MSC).

Quiet plug/unplug of each device is fine.

The stress case is not. If I unplug the MSC, plug it back, and immediately use the mouse and keyboard, the MSC never mounts / never prints anything. In one run, plugging the MSC while typing and clicking left the mouse dead after a few seconds.

So the SOF one-packet scheduler looks OK on an idle bus, but concurrent MSC enum/bulk plus live LS HID still loses the drive, and can silence the mouse. Happy to retest if you have a new revision.

P-R-O-C-H-Y:
@HiFiPhile To be sure I have retested the same stress test on my PR changes, and there its fine.

P-R-O-C-H-Y:
For tomorrow I have a local-only follow-up on top of this PR, not pushed yet:

- keep `ls_frame_reserve` (still one LS channel start per frame, no parking)
- program only one packet  (`min(remain, ep_size)`) on each preamble activation, so a long control stage cannot emit a second PRE in the same frame
- periodic NAK that would otherwise spin in the ISR is deferred to the next SOF

Plan is to capture on S3 for idle kbd+mouse, typing/moving, and the same during MSC enum/bulk. Goal is ≤1 PRE per SOF. If I can confirm on the analyzer, I will push it; if not I will drop it and stay on 9db4010ed.

HiFiPhile:
> For tomorrow I have a local-only follow-up on top of this PR, not pushed yet:
Plan is to capture on S3 for idle kbd+mouse, typing/moving, and the same during MSC enum/bulk. Goal is ≤1 PRE per SOF. If I can confirm on the analyzer, I will push it; if not I will drop it and stay on

That's nice.

> So you have turned the gate for STM? Currently in the PR it's compiled only for ESP32-S2 and S3. So it shouldn't affect any other chips.

It's only a temporary test to see how it works.

> @HiFiPhile To be sure I have retested the same stress test on my PR changes, and there its fine.

So >= 2 PRE doesn't always crash the controller ?

P-R-O-C-H-Y:
> So >= 2 PRE doesn't always crash the controller ?

Analyzer follow-up:

Two PRE in the same 1 ms frame, ~42–45 µs apart, is what we see when `HPRT.PENA` drops. It is not every pair: the same 2-PRE spacing also appears during LS control (Get Report Descriptor / Set Configuration / Set Address) and the port often survives. 
So ≥2 PRE does not always crash the controller. It is the condition that *can* kill the FS core, and it is timing-dependent.

The PR gate (S2/S3 only, one LS channel start per frame) is enough for the stress case I retested. I also tried a local-only follow-up to force ≤1 PRE (one packet per activation, then defer every new LS start to SOF). That did get long control stages down to one PRE per SOF, but it did not eliminate the remaining 2-PRE frames (those are two independent starts in the same frame, e.g. HID poll next to a control stage), and the SOF-side defer deadlocked the host (hub SETUP timeout, HID/MSC stuck). I am dropping that follow-up and staying on this PR as-is.

The `#if` stays ESP32-S2/S3 only. Fine if you keep a temporary STM build for a look; I would not ship the gate on HS cores (splits, no preamble).

P-R-O-C-H-Y:
@HiFiPhile @hathach Anything I can do to move with this PR? Thanks

HiFiPhile:
> @HiFiPhile @hathach Anything I can do to move with this PR? Thanks

> For tomorrow I have a local-only follow-up on top of this PR, not pushed yet:
> - keep ls_frame_reserve (still one LS channel start per frame, no parking)
> - program only one packet (min(remain, ep_size)) on each preamble activation, so a long control stage cannot emit a second PRE in the same frame
> - periodic NAK that would otherwise spin in the ISR is deferred to the next SOF

Ah I thought you still have works pending.

Please test https://github.com/hathach/tinyusb/commit/b09f21f328e12c7296c554dbaad9e8d607cc17f3, https://github.com/hathach/tinyusb/commit/47a432a3683a0dbc77de7e4ae3ddbbb9dffa98fd, https://github.com/hathach/tinyusb/commit/6f2824e97bbc7226e06eddd1429fc4fa64f6b2a5 on current tree, https://github.com/hathach/tinyusb/commit/b09f21f328e12c7296c554dbaad9e8d607cc17f3 is for this PR's USBH enumeration handling, others 2s are post fixup for #3815

P-R-O-C-H-Y:
> > @P-R-O-C-H-Y I've tested the LS workaround on STM32U5 (don't have ESP32 board ATM) but still observed > 2 preambles in one frame, due to sniffer limitation I can't inspect LS traffic but looks like an additional transaction attempt in the frame. {CAD01569-4A81-44B8-94CF-7169EB3F6FED}
> > I've tried another approach in https://github.com/hathach/tinyusb/tree/pr3864-alt [d099929](https://github.com/hathach/tinyusb/commit/d0999293426ccbc8904474eaf5d182dd21a66cce):
> > 
> > * schedule one 8 bytes packet per SOF, prevent long transfer from creating multiple PREs.
> > * avoid 1ms wait in ISR context
> > 
> > [b09f21f](https://github.com/hathach/tinyusb/commit/b09f21f328e12c7296c554dbaad9e8d607cc17f3), [47a432a](https://github.com/hathach/tinyusb/commit/47a432a3683a0dbc77de7e4ae3ddbbb9dffa98fd), [6f2824e](https://github.com/hathach/tinyusb/commit/6f2824e97bbc7226e06eddd1429fc4fa64f6b2a5) are newly identified fixes.
> 
> @HiFiPhile So you have turned the gate for STM? Currently in the PR it's compiled only for ESP32-S2 and S3. So it shouldn't affect any other chips.
> 
> I have tested your alt branch with all the commits on ESP32-S3 (FS hub, LS keyboard, LS mouse, FS MSC).
> 
> Quiet plug/unplug of each device is fine.
> 
> The stress case is not. If I unplug the MSC, plug it back, and immediately use the mouse and keyboard, the MSC never mounts / never prints anything. In one run, plugging the MSC while typing and clicking left the mouse dead after a few seconds.
> 
> So the SOF one-packet scheduler looks OK on an idle bus, but concurrent MSC enum/bulk plus live LS HID still loses the drive, and can silence the mouse. Happy to retest if you have a new revision.

@HiFiPhile I have tested your alt branch already, didn't pass the stress test. Is there anything new in that branch? So it's worth retesting again?

I don't have anything to add, I have commented that all my new approaches I have tried didn't worked out, there were always multiple PREs and it was working worse than commits in this PR. The PR as is works perfectly for me and ESP32S3.

HiFiPhile:
@P-R-O-C-H-Y As I said earlier (also detected by Chatgpt) `ls_frame_reserve()` can be run from ISR context, thus the maximum 1ms busy waiting inside is unacceptable.

The call chain is:
```
hcd_int_handler()
  → handle_channel_irq()
    → handle_channel_in_slave() / handle_channel_out_slave()
      → retry or continuation
        → ls_frame_reserve()
```

A simple trigger is a control IN NAK in the frame already reserved by that transfer. The immediate retry checks the same frame, finds it occupied, and spins inside the ISR until the frame becomes free or the spin limit expires.

PR #3869, opened 2026-08-31
dcd/samx7x: fix ZLP being dropped on DMA-capable IN endpoints

chrishamm:
First of all, thank you for this project!

On SAM E70/S70/V70 the terminating zero-length packet of a bulk IN transfer is silently discarded, so a transfer whose length is an exact multiple of the endpoint size is never terminated on the bus. With a Linux host the data sits in a partially filled cdc-acm URB and is not handed to userspace until something else arrives, which usually looks like a hung or truncated connection.

## Why

`dcd_edpt_xfer()` with `total_bytes == 0` falls through to `dcd_transmit_packet()`, which clears FIFOCON right away to hand the bank to the controller. That is fine when the previous packet completed via TXINI, because TXINI means a bank is free. It is not fine when the previous transfer went through DMA: the END_BUFF interrupt that completes a DMA transfer only says DPRAM was written, and the last bank is normally still being transmitted. Clearing FIFOCON on a busy bank does nothing, and the next TXINI then takes the "TX complete" path and reports the ZLP as sent although it never went out.

The stack hits this on its own, no application code required. `cdcd_xfer_cb()` calls `tu_edpt_stream_write_zlp_if_needed()` from the completion of the preceding data transfer, which is exactly the DMA completion described above. The same applies to the other class drivers that terminate with a ZLP (MIDI, printer, NCM, ECM/RNDIS, MTP).

## Proposed fix

A zero-length IN transfer on a non-control endpoint no longer touches FIFOCON at submit time. It sets `zlp_pending` and enables TXINES, and the TXINI handler validates the empty bank when the bank is guaranteed to be free. The transfer then completes on the following TXINI as before. EP0 is unaffected, it does not use FIFOCON and keeps the existing path.

## Verification

The fix was developed against an older master snapshot in our fork, where the affected code is the same, and verified on a SAME70Q20B at high speed with usbmon on the host. Before the change, CDC writes of an odd multiple of 512 bytes deterministically stalled the link within seconds. After it, 6/6 transfers of exactly 2560 bytes were delivered intact, and normal traffic was unaffected over longer runs.

The port to current master in this PR is compile-tested only, since the surrounding file has been reformatted and reworked since that snapshot.

### DIFF
--- a/src/portable/microchip/samx7x/dcd_samx7x.c
+++ b/src/portable/microchip/samx7x/dcd_samx7x.c
@@ -531,10 +531,11 @@ bool dcd_edpt_xfer(uint8_t rhport, uint8_t ep_addr, uint8_t *buffer, uint16_t to
   } else {
     if (dir == TUSB_DIR_OUT) {
       USB_REG->DEVEPTIER[epnum] = DEVEPTIER_RXOUTES;
-    } else if (total_bytes == 0 && epnum != 0) {
-      // A ZLP is typically submitted from the completion of a DMA transfer, whose END_BUFF interrupt only
-      // means DPRAM was written: the last bank is usually still being transmitted. Clearing FIFOCON on a busy
-      // bank is a no-op that would swallow the ZLP, so let the TXINI handler validate it once the bank is free
+    } else if (total_bytes == 0 && EP_DMA_SUPPORT(epnum)) {
+      // Such a ZLP is submitted from the completion of a DMA transfer, whose END_BUFF interrupt only means
+      // DPRAM was written: the last bank is usually still being transmitted. Clearing FIFOCON on a busy bank
+      // is a no-op that would swallow the ZLP, so let the TXINI handler validate it once the bank is free.
+      // Endpoints without DMA complete from TXINI with the bank already free and can send the ZLP right away
       xfer->zlp_pending         = true;
       USB_REG->DEVEPTIER[epnum] = DEVEPTIER_TXINES;
     } else {
--- a/src/portable/microchip/samx7x/dcd_samx7x.c
+++ b/src/portable/microchip/samx7x/dcd_samx7x.c
@@ -38,6 +38,7 @@ typedef struct {
   uint16_t   queued_len;
   uint16_t   max_packet_size;
   uint8_t    interval;
+  bool       zlp_pending;
   tu_fifo_t *fifo;
 } xfer_ctl_t;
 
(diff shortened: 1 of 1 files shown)


chrishamm:
The first commit here didn't fix the problem reliably. `TXINI` is sticky: hardware sets it whenever the current bank becomes free and only `TXINIC` clears it, and DMA transfers run with the interrupt masked. So when the class driver submits the terminating ZLP, the flag is practically always still set from the data phase (331 of 331 submissions in an instrumented build), and enabling `TXINES` fires the handler immediately, regardless of the actual bank state. In my tests the bank was still busy in about 6% of the submissions and those ZLPs were swallowed exactly like before the patch.

The latest commit acknowledges the stale `TXINI` first and then decides based on `NBUSYBK`, which is a live counter and not a latched flag: if no bank is busy the ZLP is validated right away, otherwise it is sent from the next genuine `TXINI`. That logic holds for FIFO endpoints as well, and for a ZLP submitted right after another ZLP, so I dropped the `EP_DMA_SUPPORT` restriction from the intermediate commit again.

Verified on a Duet 3 MB6HC (SAME70, high-speed, CDC to a Linux host): three runs of roughly 230 bulk IN transfers at sizes 3072, 3584, 4096, 4608 and 5120 bytes plus two non-multiple controls, zero lost ZLPs. The same test loses 7 to 13 transfers per run with each of the previous versions.

HiFiPhile:
Thank you, an issue is raised in my side which looks valid:


- **[P2] Re-acknowledge `TXINI` before the immediate ZLP send** — [dcd_samx7x.c:539]
  The final busy bank can become free between clearing stale `TXINI` and reading `NBUSYBK`. Hardware then asserts a fresh `TXINI`, while `NBUSYBK` reads zero. The immediate branch validates the ZLP and enables `TXINE` without clearing that fresh flag, so the handler may immediately report the ZLP complete before its bus transmission. A subsequent FIFO transfer can consequently be submitted against a busy bank.

  Clear `TXINI` again in the zero-busy branch immediately before `dcd_transmit_packet()`. Once `NBUSYBK == 0`, no previous busy bank remains to race that second clear.

chrishamm:
Thanks, that was a real race. I went one step further than clearing TXINI again: the immediate path is gone. The ZLP is now always validated from the TXINI handler, where the bank is guaranteed free. After acknowledging the stale TXINI the driver checks RWALL, and if a bank is writable already it raises TXINI by software (DEVEPTIFR TXINIS), which is what the ASF USBHS driver does after a DMA transfer. That leaves one path, the completion always follows the transmission, and the NBUSYBK check is no longer needed. NBUSYBK could not distinguish a free current bank from a busy one on dual-bank endpoints anyway, which the full-speed configuration uses.

Tested on a Duet 3 MB6HC (SAME70, high speed, dual CDC to a Linux host running our SBC protocol over the second interface): 3 hours at roughly 450 request/response transfers per second with bulk IN sizes up to 8 KiB, zero lost or truncated transfers.

HiFiPhile (approved):
LGTM

PR #3871, opened 2026-09-01
tools: support RTT capture lifecycle on Windows

HiFiPhile:
Use platform-specific process group creation and termination so J-Link, OpenOCD, and dump helpers are retired reliably on Windows as well as POSIX. Resolve Windows executable names while preserving environment overrides and report missing tools cleanly.

Add an optional stop file for ending indefinite captures from automation without signals. Cover executable selection, process cleanup, and stop-file handling in the RTT HIL tests.

### DIFF
--- a/test/hil/test/test_hil_rtt.py
+++ b/test/hil/test/test_hil_rtt.py
@@ -12,6 +12,7 @@ import time
 import unittest
 from contextlib import suppress as contextlib_suppress
 from pathlib import Path
+from unittest import mock
 
 # the module under test lives in the parent dir's helper/ package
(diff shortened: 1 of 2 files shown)


hathach:
Superseded by #3925, which folded this into `tools/rtt.py`: the Windows process-group lifecycle, the `RTT_JLINK_EXE`/`RTT_OPENOCD_EXE` overrides and `--stop-file` cancellation. It also validates arguments before honoring a stop file, and sends SIGINT/SIGBREAK through the same teardown as SIGTERM. `rtt.py` itself now lives in agentrc, so the extended lifecycle tests belong there. Closing.

PR #3885, opened 2026-09-06
Fix IAR warnings and nRF54H20 Make build

hathach:
Fix three CI build failures:

- Snapshot GINTMSK before the live GINTSTS read to sequence volatile accesses and remove IAR Pa082.
- Return audioh_open failures through one cleanup helper so control never bypasses initialized locals, removing IAR Pe546 without goto.
- Use nRF54H20's native USB host in Make builds instead of the stale MAX3421 configuration inherited from nRF5340.

Verified with pre-commit, an IAR 9.70.1 full stm32h743eval build, and all Make examples for nrf54h20dk.

### DIFF
--- a/hw/bsp/nrf/boards/nrf54h20dk/board.mk
+++ b/hw/bsp/nrf/boards/nrf54h20dk/board.mk
@@ -6,9 +6,6 @@ CFLAGS += -DNRF54H20_XXAA
 # RAM00 (512 KB) and drop this.
 CFLAGS += -DCFG_EXAMPLE_VIDEO_READONLY
 
-# enable max3421 host driver for this board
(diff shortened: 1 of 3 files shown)


hathach:
All eight CircleCI IAR shards pass after the DWC2 and audio-host fixes. The full stm32h743eval example set also passes locally with IAR 9.70.1.

The nrf54h20dk Make failure is fixed by removing its stale MAX3421 setting. The full board Make matrix passes locally with 34 OK and 0 failed, and the refreshed CircleCI nRF Make shard passes.

STM32F723 HIL host/device_info still fails identically from base commit 42138394 on the same rig.

PR #3914, opened 2026-09-13
fix(ncm): wNdbInDivisor/wNdbOutDivisor rejected as 1 by Windows KB5124008

### LINKED ISSUE #3913: NCM device driver's hardcoded wNdbIn/OutDivisor=1 fails Windows' updated usbncm.sys validation (Code 10)
### Operating System

Windows 11

### Commit SHA

ad54f72e10859596044918b76ed5a73402ca545b

### Board

Waveshare RP2040-Zero

### Firmware

Reproducible with the stock, completely unmodified `examples/device/net_lwip_webserver`
example — this is not specific to any custom firmware. The bug is in the default
`ntb_parameters` struct in `src/class/net/ncm_device.c` itself (`wNdbInDivisor`/
`wNdbOutDivisor` hardcoded to `1`), which every CFG_TUD_NCM=1 project inherits
identically since there's no config override for it.

### What happened ?

Windows Update KB5124008 (Sept 2026) enabled a feature-flag-gated NTB parameter
validation in Microsoft's inbox usbncm.sys driver. It requires wNdbOutDivisor to be
>=4, a power of 2, and <= dwNtbOutMaxSize. TinyUSB hardcodes wNdbInDivisor/
wNdbOutDivisor to 1 (spec-legal, but fails the new ">=4" check immediately).

This causes the whole device to fail enumeration with STATUS_DEVICE_HARDWARE_ERROR,
shown to the end user as Device Manager's "Code 10 — This device cannot start
(fatal device hardware error)" — on any Windows machine that's received this update,
regardless of project configuration.

Root-caused via WinDbg kernel debugging (KDNET), breaking on
UsbNcm!UsbNcmHostDevice::SelectConfiguration and its callees, and confirmed against
Microsoft's own published NCM host driver reference source
(https://github.com/microsoft/NCM-Driver-for-Windows). See attached log for the
disassembly/memory evidence. Confirmed the bug is still present on current `master`.

Suggested fix: bump both to 4 (still power-of-2, safely under any reasonable
dwNtbIn/OutMaxSize), or better, add CFG_TUD_NCM_IN/OUT_DIVISOR overrides matching
the existing CFG_TUD_NCM_IN/OUT_NTB_MAX_SIZE pattern.

### How to reproduce ?

1. Flash any TinyUSB NCM device firmware (e.g. stock net_lwip_webserver example,
   unmodified) onto a supported board.
2. Plug it into a Windows PC that has installed KB5124008 (or a later cumulative
   update carrying the same usbncm.sys feature rollout) — this is host-side and
   PC-specific, so it may not reproduce on every Windows install depending on
   Windows' own feature rollout ring.
3. Check Device Manager: the resulting network adapter shows "Code 10 — This
   device cannot start (fatal device hardware error)".

[.tinyusb-issue-debug-log.txt](https://github.com/user-attachments/files/32145987/default.tinyusb-issue-debug-log.txt)

### Screenshots

_No response_

Warumski:
Fixes #3913
A Windows Update (KB5124008, Sept 2026) enabled a feature-flag-gated NTB
parameter validation in usbncm.sys requiring wNdbOutDivisor (and by
symmetry wNdbInDivisor) to be >=4 and a power of 2. TinyUSB's NCM class
driver hardcodes both to 1 — spec-legal per NCM1.0, but now rejected,
causing STATUS_DEVICE_HARDWARE_ERROR (Code 10) in Device Manager before
the device gets past SelectConfiguration().

Root-caused via WinDbg kernel debugging (KDNET) cross-referenced against
Microsoft's own published NCM host driver reference source
(microsoft/NCM-Driver-for-Windows). Confirmed fixed on real hardware
(RP2040, Windows 11 post-KB5124008; also verified no regression on
Android).

4 is spec-legal on the device side and is the value most other NCM
devices already ship with, so this shouldn't affect any other host.

## Bug Fixes

- Improved USB network device compatibility with Windows systems.
- Updated network device configuration values to satisfy Windows validation requirements.
- Resolved an issue that could prevent device enumeration and cause a Code 10 hardware error.


HiFiPhile:
Thank you, please fix trailing white space pre-commit issue.

Warumski:
I hope I fixed it now ;)

HiFiPhile (approved):
Thank you !

Issue #3908, opened 2026-09-09
[audio examples] uac2_headset / uac2_speaker_fb: feature-unit handlers index mute[]/volume[] with an unvalidated host-controlled channel number (wValue low byte, 0..255) — OOB read/write


shinoryo1216:
### Operating System

Linux

### Commit SHA

master 8348d43f7179667a5083e7fdb3030e05bf488eb0 (fetched 2026-09-09; all line numbers in this report match this snapshot). The handler pattern has been in the tree for years — it predates the 0.19 → 0.21 example restructuring and is carried by both the UAC1 (audio10_*) and UAC2 (audio20_*) paths.

### Board

ESP32-S3-DevKitC-1 (my hardware; any board applies — the defect is in the example's request handlers, host-protocol level, not port-specific)

### Firmware

Stock examples, unmodified:

- `examples/device/uac2_headset` — primary; all line numbers below refer to its `src/main.c` (656 lines) unless stated otherwise. This example returns a UAC1 configuration at Full-Speed and UAC2 at High-Speed (per its README), so both handler families are reachable depending on the board's link speed.
- `examples/device/uac2_speaker_fb` — same four handlers, own copy (`src/main.c`, 676 lines; arrays at lines 70-71).

Not affected:

- The stack itself, `src/class/audio/audio_device.c` — see "Where the responsibility lies" under "What happened ?" (verified: it forwards `bChannelNumber` to the application callbacks and never indexes a per-channel array itself).
- `examples/host/audio_host` — host-side example, no per-channel state arrays indexed by a device-supplied channel number.

> **Correction:** the first version of this section also listed `examples/device/audio_test` as not affected ("computes `channelNum` but never uses it as an index"). That was my verification error — I had only checked its EP/ITF handlers, which really do cast `(void) channelNum`. Its **entity** handlers are affected: `tud_audio_set_req_entity_cb()` / `tud_audio_get_req_entity_cb()` index `mute[channelNum]` / `volume[channelNum]` (arrays sized `CFG_TUD_AUDIO_FUNC_1_N_CHANNELS_TX + 1`) inside the `entityID == 2` feature-unit branches, exactly like the examples above.
>
> Re-checking the whole tree on the same master snapshot, **all eight device audio examples** carry the same pattern, not just the two listed above:
>
> | example | vulnerable FU handlers | array bound |
> |---|---|---|
[...]

PR #3915, opened 2026-09-13
audio: fix cdc_uac2 speaker descriptor channel mismatch

shinoryo1216:
## Description

The cdc_uac2 example has a pre-existing mismatch between its speaker-side
descriptors and its configuration/state:

- CFG_TUD_AUDIO_FUNC_1_N_CHANNELS_RX is 1 (tusb_config.h), mute[]/volume[]
hold N+1 = 2 entries (master + ch1, src/uac2_app.c), and the AS interface
declares bNrChannels = 1.
- But the speaker Input Terminal declares 2 logical channels, and the Feature
Unit is sized TUD_AUDIO20_DESC_FEATURE_UNIT_LEN(2) and advertises RW
mute+volume for master/ch1/ch2 (src/usb_descriptors.h).

The mismatch is present in the 0.21.0 tag, so it predates the recent
channel-number guards (#3909).

Two consequences:

1. In the stock example, a conforming host addressing the advertised ch2
reads/writes one entry past the 2-entry mute[]/volume[] arrays. The arrays
are globals, so this is an OOB access into .data/.bss — exactly the class
of bug the guards in #3909 prevent.
2. With those guards, that request now STALLs instead, i.e. the guard changes
behaviour for a conforming host in this one example (noted in #3909).

## Change

Align the speaker descriptors with the actual 1-channel configuration
(mono):

- Speaker Input Terminal: bNrChannels 2 -> 1
- Speaker Feature Unit: master + ch1 only (drop _ctrlch2; the bLength
follows automatically via TUD_AUDIO20_DESC_FEATURE_UNIT_LEN(1))
- CS AC wTotalLength and the example's total-length macro track the LEN
macro, so both adjust accordingly

The alternative — stereo (RX = 2) — would also change the advertised stream
format and the EP size, so mono is the surgical fix: the streaming path,
sample format, EP sizing and alternate settings are untouched. After this
change the descriptor-verified model matches the runtime state everywhere:
IT logical = AS physical = RX = array entries - 1 = FU channels, and the
guards reject only channels that are neither advertised nor backed by state.

## Notes

- Only src/usb_descriptors.h is touched (4 lines). The mic path is already
consistent (TX = 1, IT logical = 1, no mic feature unit).
- Verified: the 2-argument (master + ch1) TUD_AUDIO20_DESC_FEATURE_UNIT
invocation compiles and sizes to 14 bytes (same pattern as the built-in
mono templates in usbd.h and audio_test_multi_rate), and the example's
TUD_AUDIO_HEADSET_STEREO_DESC_LEN still equals the emitted byte count.

Related: #3909 (channel-number guards), #3908 (issue). Caught by the
automated review on #3909.

## Summary by CodeRabbit

- **Bug Fixes**
  - Updated the CDC UAC2 USB audio device example to use a mono, single-channel speaker configuration.
  - Corrected audio descriptor information so the device reports its channel count and available controls accurately.


HiFiPhile (approved):
LGTM

PR #3923, opened 2026-09-15
rp2040: fix E15 deferred ZLP resume missing SEL reset

bnielsen1965:
# [rp2040] Fix E15 deferred ZLP resume missing buffer-select reset

## Summary

Under light CDC-NCM traffic the device's **Bulk-IN path permanently wedges** after ~5–10 min. A required trailing **Zero-Length Packet** (issued when an NTB's `wBlockLength` is a multiple of the 64-byte MPS) is never transmitted when that ZLP falls inside the E15 "critical window" (last ~200 µs of a full-speed frame). The SIE then NAKs the endpoint forever: `usbd_edpt_busy(0x81)` stays true, the NCM glue's `xmit_start_if_possible()` early-returns every time, the TX ring fills and all subsequent frames are dropped. **RX is unaffected** — the host keeps sending while the device stops responding — until a bus reset.

## Root Cause

Two bugs in the SOF-handler E15 resume block in `src/portable/raspberrypi/rp2040/dcd_rp2040.c` (the *only* path that re-arms transfers the errata deferred):

**A — SEL not reset for a fresh deferred transfer.**
A fresh deferred transfer (e.g. a ZLP, `remaining_len == 0`) was gated out of the `rp2usb_buffer_start()` path (which resets the internal buffer select to `buf0`). It was armed with a bare `bufctrl_write16` instead. After an even number of double-buffered 64-byte packets the SIE's internal select sits on `buf1`, so the armed `buf0` is never polled → NAK forever.

**B — Marked ACTIVE before arming.**
`ep->state = EPSTATE_ACTIVE` was set at the top of the branch *before* any buffer was confirmed idle. If neither buffer was idle at SOF time, nothing was armed but the state was already ACTIVE, so no later SOF retried — the transfer was orphaned.

## Trigger Conditions

A wedge requires all three:
1. An IN NTB whose `wBlockLength % 64 == 0` (a trailing ZLP is required)
2. `wBlockLength % 128 == 0` (even 64-byte packet count → the double-buffered select ends on `buf1`)
3. That ZLP issued inside the last ~200 µs of a frame (~20% probability per qualifying NTB)

~16 KB device→host responses yield 1280-byte NTBs satisfying (1) and (2); (3) is a ~20% roll, so the hit lands after several minutes of light traffic.

## Fix

Both changes are in the SOF-handler `EPSTATE_PENDING` resume block:

1. Drop the `remaining_len > 0` gate so a ZLP reaches the `rp2usb_buffer_start()` path (which sets SEL). "Both idle" means nothing is in flight, so resetting the select to `buf0` and re-arming in order is always correct for both a fresh and a mid-transfer resume.

2. Use `xferred_len == 0` to distinguish a fresh deferred transfer from a mid-transfer resume. A fresh transfer must go through the SEL-resetting path; if a buffer is still in flight, stay `EPSTATE_PENDING` and retry on the next SOF. Only set `EPSTATE_ACTIVE` after a buffer is actually armed.

## Verification

- **Before (unpatched):** hard wedge at ~7 min of light traffic. Serial log shows the trigger — a 1280-byte NTB whose trailing ZLP is never polled:
  ```
  DBG in  xfer len=1280
  DBG in  done bytes=1280 res=0
  # expected "done bytes=0" (ZLP) never arrives
  DBG st  gd=0 nr=0 ntb=0 busy=1 can=1   # IN EP stuck
  net: WEDGE - NCM cannot transmit
  ```
  531 `WEDGE` lines; TX frozen until manual USB reset.

- **After (patched):** same light-traffic profile, 10+ min (well past the original ~7-min failure):
  - Every qualifying 1280-byte NTB's trailing ZLP completes 1:1 (20/20)
  - **0** `WEDGE` lines
  - Large multi-NTB (16 KB, ping-pong) transfers intact
  - `dcd_rp2040.c` compiles cleanly for `raspberry_pi_pico` (both `cdc_msc` and `net_lwip_webserver` examples)

## Environment

- tinyusb `0.21.0`, RP2040 (Pico), full-speed, CDC-NCM
- Bulk-IN EP `0x81`, double-buffered, 64 B MPS
- `CFG_TUSB_RP2_ERRATA_E15=1` (via Pico SDK `PICO_RP2040_USB_DEVICE_UFRAME_FIX=1`)
- Commit `087ceb741` (keep_sof_alive `>=` fix) already present

## Reproduction

Repro firmware (CDC-NCM + lwIP) is available at https://github.com/bnielsen1965/pico-usbnet. Build with the TinyUSB at `dae3f9a36` (pre-fix) to reproduce, or with this patch applied to confirm the fix.

```sh
# Reproduce (unfixed):
git clone https://github.com/bnielsen1965/pico-usbnet && cd pico-usbnet/example
./setup.sh
cmake -DPICO_USBNET_TINYUSB_PATCHES=OFF .. && make
# Flash, connect, then:
ping -i 0.5 192.168.7.1    # wedges in ~5-10 min

# Confirm fix (patched — default):
cmake .. && make           # patch applied by default
# Same traffic: 0 wedges after 10+ min
```


bnielsen1965:
FYI for anyone looking at this patch. It resolves a severe and reproducible TX stall issue, but I have ran into another TX stall that is difficult to reproduce and I don't know yet if it is related to this patch or not. It can take an hour or more for this new TX stall to show up so it is going to take awhile to characterize. If it turns out to be related and affects this patch then I'll amend the patch if the PR is still open.

This new issue was unrelated, it appear to be related to using FreeRTOS and my network task was roaming cores while the USB IRQ was pinned to one core. Pinning my network task to the same core as the IRQ now. No impact on this PR.

PR #3940, opened 2026-09-20
dcd: dwc2: clear TX FIFO empty mask on bus reset

nicklasb:
The slave-mode transmit path sets endpoint bits in `DIEPEMPMSK` when additional data remains to be written. If a bus reset interrupts an active IN transfer, this mask can remain set while `xfer_status` and endpoint state are reset. That leaves stale TX FIFO-empty interrupt servicing enabled after the reset.

## Fix

Reset `DIEPEMPMSK` before clearing the software transfer state in `handle_bus_reset()`.

## Verification

Reproduced on ESP32-P4 using the DWC2 high-speed device controller:

- Interrupted an active CDC IN transfer by resetting the USB host.
- Observed stale `DIEPEMPMSK = 0x00000004` for endpoint 2.
- Confirmed the reset handler cleared the stale mask.
- Confirmed the device subsequently re-enumerated and resumed CDC transfers.


HiFiPhile:
Thank you, I think it's safer to also clear it in `dcd_edpt_close_all()`

nicklasb:
> Thank you, I think it's safer to also clear it in `dcd_edpt_close_all()`

Feel free to modify the PR if you want to, I have not tested that combination.

nicklasb:
May be worth noting that that this issue causes a crash on host reset on the platform if warning messages are enabled, as that isn't perhaps evident of the original wording.

